PR #10, opened 2021-11-04
Fix buf in bytestream_server when NotFound was returned

allada:
Fixes bug found where the backend store could return NotFound but
would deadlock the request because the spawn that received the
error was never awaited properly.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/rust_cas/10)


allada (commented):
> Reviewed 9 of 10 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/rust_cas/10)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #19, opened 2021-11-11
`.has()` now returns size on success instead of boolean

allada:
We already have the size in each of our stores for the objects,
so we now return the size of the entry or None. This will make it
easier to implement certain features in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/rust_cas/19)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/rust_cas/19)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #28, opened 2021-11-16
Refactor code to use streams instead of AsyncRead/AsyncWrite

allada:
This will greatly reduce the amount of copies needed when paired
with using `bytes` package. This is because `bytes` package will
not make copies of the data much of the time. Instead we will ship
these Bytes structs around and only when needed make a copy. These
structs also don't make copies most of the time when passed from
module to module and instead just point to the same memory.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/rust_cas/28)


allada (commented):
> Reviewed 21 of 25 files at r1, 9 of 13 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/rust_cas/28)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #48, opened 2022-04-11
[RE] Various changes to worker_api.proto

allada:
* ConnectWorker is no longer stream<->stream, instead has function
  calls and returns a stream for server->client updates.
* Added `reserved` keyword representing each `NextId`.
* Added ConnectionResult so worker itself knows it's own ID.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/48)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/48)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #51, opened 2022-04-13
Restructure config format in prep for WorkerApiServer

allada:
Restructure the format of how scheduler and platform properties
are in the config spec. This will allow more intuitive sharing
of scheduler and platform properties to different services.

* `schedulers` is now a root-level map and services reference the
  name of that scheduler instead of defined inline.
* Platform properties are now owned by the scheduler.
* Updated example config files.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/51)

### DIFF
--- a/cas/BUILD
+++ b/cas/BUILD
@@ -12,6 +12,7 @@ rust_binary(
         "//cas/grpc_service:capabilities_server",
         "//cas/grpc_service:cas_server",
         "//cas/grpc_service:execution_server",
+        "//cas/scheduler",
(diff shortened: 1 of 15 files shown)


allada (commented):
> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/51)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #57, opened 2022-04-16
Services in main() now run in their own spawn

allada:
This is a minor performance improvement that allows different
services to run in their own spawn, since they don't depend on
each other anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/57)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/57)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #60, opened 2022-04-18
Adds ability in proto for worker to return internal errors

allada:
Updates the worker_api.proto so it can return error codes for
internal errors in the event that an internal error occurred.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/60)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/60)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #64, opened 2022-04-23
Fix bug in memory store when receiving a zero byte object

allada:
An edge case was discovered in another unit test where a zero
byte object (empty file) would error out during receive.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/64)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/64)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #75, opened 2022-04-29
Support deprecated symlink fields & fix bug for workers use CWD

allada:
The documentation in the bazel remote execution proto is confusing.
It states that the output_paths will be relative to working directory
which was interpreted as CWD. This PR fixes this miss-understanding
and now files are relative to input_root.

Also, it turns out that bazel does not yet support 2.1, so we
are now populating the deprecated fields as well as the new fields.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/75)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,6 +7,9 @@ build:debug -c dbg
 build:self_test --remote_instance_name=main
 build:self_test --remote_cache=grpc://127.0.0.1:50051
 
+build:self_execute --remote_executor=grpc://127.0.0.1:50051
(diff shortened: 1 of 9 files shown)


allada (commented):
> Reviewed 9 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/75)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #77, opened 2022-05-05
Filesystem store now delays before deleting temp file

allada:
There was a rare condition that could happen where the store would
atomically swap the temp file and the final location then delete
the temp file immediately after. Normally this would be fine, however
during a hardlink, the interal kernel command first queries the inode
then creates a hardlink for the queried inode. If the file was not
swapped yet and the inode was queried, it'll get the inode of the
file about to be deleted, then if the file is deleted then the
hardlink will use an inode that doesn't exist any more.

This is solved by adding a 100ms delay. In testing 1ms would be
enough on my 16 core testbench, but to give plenty of leeway we add
a substantial delay.

This is also incredibly hard to write a test for, since it was a
flake condition and I don't have control over the innards of the
`hardlink` command, so this is one of the rare conditions there is
not a unit test for a bug, but it is documented now.

---
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/77)

### DIFF
--- a/cas/store/filesystem_store.rs
+++ b/cas/store/filesystem_store.rs
@@ -5,7 +5,7 @@ use std::path::Path;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::SystemTime;
+use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
 use bytes::BytesMut;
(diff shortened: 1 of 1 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/77)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #100, opened 2022-10-17
Fix bug if no instance_name/resource_name is given upload does not work

allada:
Fixes a bug that if an upload is attempted but no instance_name is configured it will error out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/100)

### DIFF
--- a/util/BUILD
+++ b/util/BUILD
@@ -144,6 +144,16 @@ rust_library(
 )
 
 rust_test(
+    name = "resource_info_test",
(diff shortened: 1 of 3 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/100)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #104, opened 2023-06-25
Move from fast-async-mutex to async-lock

chrisstaite-menlo:
The async-lock crate implements the same functionality as fast-async-mutex, however it is maintained and actually implements a Mutex rather than a spin-lock that pretends to be a Mutex as documented in the fast-async-mutex Issues on GitHub.

In addition, there is a bug in the spinlock implementation of fast-async-mutex that causes the Send trait to be unsafe for the Mutex, however that trait is used within turbo-cache and causes turbo-cache to lock up when under heavy load.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/104)


allada (approved):
> Reviewed 50 of 50 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #105, opened 2023-06-25
Remove use of renameat2 by adding RW locks for all managed files.

chrisstaite-menlo:
I've seen a number of race conditions and temporary files not being cleaned up and the use of renameat2 adds a dependency on Linux.  Adding RWLocks does not appear to affect performance in any meaningful way during operation, and I've not seen any warnings about double-deleting temporary files or left over temporary files since this change in our production system.


chrisstaite-menlo:
Owned RWLocks don't exist in a release of async-locks yet, they are in the unreleased branch.  I was using that originally, but have realised that tokio has a sync feature, so decided that would be a better choice.  Now I'm wondering if I should remove the async-locks dependency entirely and use the tokio Mutex instead since I'm bringing it in with this PR anyway.

chrisstaite-menlo:
After reading the documentation for tokio Mutex (https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html) I realised that I should be using a sync lock for the HashMap, so I swapped out the implementation.  I also removed the necessity to keep clone the Arc again, so this seems a much better solution.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/105#-NYnYBtkDQUF12NlQ9sG:-NYnYBtkDQUF12NlQ9sH:bnhqpov) (no related file):*
Can we update the documentation in `backends.rs` where it talks about `RENAME_EXCHANGE`?

___
___
*[`cas/store/filesystem_store.rs` line 103 at r1](https://reviewable.io/reviews/allada/turbo-cache/105#-NYnfpxE3-MiyjN8oge-:-NYnfpxE3-MiyjN8oge0:bxtj90c) ([raw file](https://github.com/allada/turbo-cache/blob/0458c16c4deae561aa1290505510b60a29f1ea2d/cas/store/filesystem_store.rs#L103)):*
> ```XML
>         let content_path = to_full_path_from_digest(&self.content_path, &self.digest);
>         log::info!("\x1b[0;31mFilesystem Store\x1b[0m: Deleting: {}", &content_path);
>         if let Err(err) = fs::remove_file_locked(&content_path).await {
> ```

We can't actually do this in `unref()`.  `unref()` simply marks the item as no longer in the `EvictingMap`, but it could still be in use in another section of code. The `FileEntry` struct is only a "placeholder" to say/guarantee, "this file exists and here's the meta details" it cannot hold a file descriptor open, because there's a `FileEntry` for every CAS object.

This presents a problem with deleting the file, because there needs to be a way for other parts of the code to check if file exists and if the file does exist perform the operation in an atomic way. Since we cannot hold a file-descriptor we need to guarantee that if a `FileEntry` is ever handed out, the underlying file will never change and will always exist while struct is alive.

Master current gets around this by using `unref()` to move the file to a temp file, then when all references to the struct are dropped does it actually delete the temp file.

The comment in `unref()` should make a lot more sense since in those 4 lines it calls `rename()` on the filesystem then changes the struct to point to the new file and if someone tried to open the file between those two operations it could error.

ref: https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L81

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 112 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYno95e8RWU42BPDIZM:-NYno95e8RWU42BPDIZN:bfj1uwo) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/cas/store/filesystem_store.rs#L112)):*
> ```XML
> }
> 
> fn temp_file_name() -> u64 {
> ```

nit: I wounder if a `uuid1()` might be better here? For debugging reasons it might be nice to know when the temp file was created and `uuid1()` encodes this info.
___
*[`util/fs.rs` line 103 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYntvKn44OHidl1chZ5:-NYntvKn44OHidl1chZ6:b-wdq73) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/util/fs.rs#L103)):*
> ```XML
> static OPEN_FILE_LOCKS: Mutex<Option<HashMap<String, Arc<RwLock<()>>>>> = parking_lot::const_mutex(None);
> 
> fn get_lock(path: impl AsRef<Path>) -> Arc<RwLock<()>> {
> ```

In thinking about this, I think there's a couple questions/ideas I want to entertain. I'm in no way saying we should or should not do this, I just want to think about this, since I think we are fighting something that we may not be needed.

1. Would it be better to simply use `flock`?
Offloading this onto the kernel to manage might be much easier and a lot less chance of introducing bugs. The big down side is how `rename()` would need to be processed, since we'd need to open a file descriptor, and under-the-hood the rename operation will also open a file descriptor. Since `rename()` is pretty rare (during create and delete operations only), I think it might be reasonable. (also `hardlink()` would have similar concerns), however in the future we could easily add a `if-nix` condition and call the raw calls on the inodes if performance was a concern).

2. Do we actually need to lock at all?
Currently the only place that uses this code is the `FilesystemStore`, which we discussed above. My concern is that currently, when the `FileEntry` exists, the file must also exist and must be immutable. Keeping the guarantees on the layer above might be better met in the `FileEntry` object instead of here in the functional calls. Also, any file name on disk that has a `FileEntry` should only have exactly one `FileEntry` pointing to that file at any given time (ie: Arc pointer).


After digging into this code for the last couple hours, I'm pretty confident that `renameat2` was a bit of an overkill for guarantees and with maybe 1-2 locks could be removed completely and converted to a simple move/replace call.

My big question for you is if these locks fixed a bug that you were experiencing, or are these here to re-create the promises that `renameat2` was giving?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @allada)

___
*[`cas/store/filesystem_store.rs` line 103 at r1](https://reviewable.io/reviews/allada/turbo-cache/105#-NYnfpxE3-MiyjN8oge-:-NYnv7iOBCQDEJEukTtP:b-ck4nn4) ([raw file](https://github.com/allada/turbo-cache/blob/0458c16c4deae561aa1290505510b60a29f1ea2d/cas/store/filesystem_store.rs#L103)):*

We can't actually do this in `unref()`.  `unref()` simply marks the item as no longer in the `EvictingMap`, but it could still be in use in another section of code. The `FileEntry` struct is only a "placeholder" to say/guarantee, "this file exists and here's the meta details" it cannot hold a file descriptor open, because there's a `FileEntry` for every CAS object.

This presents a problem with deleting the file, because there needs to be a way for other parts of the code to check if file exists and if the file does exist perform the operation in an atomic way. Since we cannot hold a file-descriptor we need to guarantee that if a `FileEntry` is ever handed out, the underlying file will never change and will always exist while struct is alive.

Master current gets around this by using `unref()` to move the file to a temp file, then when all references to the struct are dropped does it actually delete the temp file.

The comment in `unref()` should make a lot more sense since in those 4 lines it calls `rename()` on the filesystem then changes the struct to point to the new file and if someone tried to open the file between those two operations it could error.

ref: https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L81
</blockquote></details>

That makes sense.  I'll revert to deleting on Drop.  I did think that any other thread would hold a read lock, however they may have a FileEntry and have not yet done anything with it...
___
*[`util/fs.rs` line 103 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYntvKn44OHidl1chZ5:-NYnyjW37SP0oc5W2f9d:bya1zse) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/util/fs.rs#L103)):*

In thinking about this, I think there's a couple questions/ideas I want to entertain. I'm in no way saying we should or should not do this, I just want to think about this, since I think we are fighting something that we may not be needed.

1. Would it be better to simply use `flock`?
Offloading this onto the kernel to manage might be much easier and a lot less chance of introducing bugs. The big down side is how `rename()` would need to be processed, since we'd need to open a file descriptor, and under-the-hood the rename operation will also open a file descriptor. Since `rename()` is pretty rare (during create and delete operations only), I think it might be reasonable. (also `hardlink()` would have similar concerns), however in the future we could easily add a `if-nix` condition and call the raw calls on the inodes if performance was a concern).

2. Do we actually need to lock at all?
Currently the only place that uses this code is the `FilesystemStore`, which we discussed above. My concern is that currently, when the `FileEntry` exists, the file must also exist and must be immutable. Keeping the guarantees on the layer above might be better met in the `FileEntry` object instead of here in the functional calls. Also, any file name on disk that has a `FileEntry` should only have exactly one `FileEntry` pointing to that file at any given time (ie: Arc pointer).


After digging into this code for the last couple hours, I'm pretty confident that `renameat2` was a bit of an overkill for guarantees and with maybe 1-2 locks could be removed completely and converted to a simple move/replace call.

My big question for you is if these locks fixed a bug that you were experiencing, or are these here to re-create the promises that `renameat2` was giving?
</blockquote></details>

The locks fixed the issue of dangling temp files and log messages stating that temp files didn't exist when they should.  That was the reason I swapped to an incrementing number rather than a random number because I attributed it to random collisions (however unlikely that would be).

In the implementation of this I also changed every `fs::` function to take a `locked` parameter, but in doing so I realised that the use cases were actually pretty small, as such I reverted to implementing specialised functions for the situations required. It also allowed me to remove a lot of complication with owned locks for directory reading that was never used.

I think you're right though, I think this can be fixed through FileEntry instead.  Not least because, as you say, the FileEntry should be immutable until it is Dropped, however, the file may change in size (through update_file) and there are then FileEntry structs laying around with a pointer to a file that is not the size that it says it is.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @allada)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/105#-NYnYBtkDQUF12NlQ9sG:-NYo-NlpApvCtZH4LfZ7:b-896fix) (no related file):*

Can we update the documentation in `backends.rs` where it talks about `RENAME_EXCHANGE`?
</blockquote></details>

Done.

___

allada (commented):
Thanks BTW!

Code review in open source projects are often difficult because it's some stranger scrutinizing your code and design choices rather than someone that you have met personally or have a reputation with. This ends up causing the reviewee to start to feel that the reviewer is an "emotionless-bot that just tells me what to do", because there has never been any other interactions. The best way I've seen around this is to present messages like this one and to encourage others to not feel attacked when questions are raised.

> the 7%-38%-55% Rule, 7% is what you say, 38% is your tone and 55% is your facial expression
It's hard to question/critique others when I loose ~93% of the normal tools we use to communicate empathy.

https://en.wikipedia.org/wiki/Albert_Mehrabian#Attitudes_and_congruence

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105#-:-NYnyCw59ySvK2V8j8Zr:b-t6jy84)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @chrisstaite-menlo)

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/fs.rs` line 103 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYntvKn44OHidl1chZ5:-NYo28pX6OCsG_qFtM1X:bnliu5r) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/util/fs.rs#L103)):*
> however, the file may change in size (through update\_file) and there are then FileEntry structs laying around with a pointer to a file that is not the size that it says it is.

I don't think the file should change in size. If it does, it is likely a bug. Do you have a case in mind that you are aware of where the file can change sizes (or contents)?

allada (commented):
> Reviewed 1 of 12 files at r1, 2 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 3 of 13 files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 395 at r3](https://reviewable.io/reviews/allada/turbo-cache/105#-NYo4ME244ICiUKf_U94:-NYo4ME244ICiUKf_U95:b-7lajpf) ([raw file](https://github.com/allada/turbo-cache/blob/69954cb59ec357f070cc0cec11c555ea6b4c42a0/cas/store/filesystem_store.rs#L395)):*
> ```XML
>             let pending_delete_file_name = old_item.pending_delete_file_name.load(Ordering::Relaxed);
>             if pending_delete_file_name == 0 {
>                 if let Some(callback) = old_item.file_evicted_callback {
> ```

nit: This happens automatically in the `drop()` call now, so you can just call `self.evicting_map.insert()` without needing to do anything with the old item.

However, given the comment about `hardlink` it might be a good idea to delay the delete operation. I am pretty sure I personally experienced this. The delay is probably best to be done inside the `drop()` though.
___
*[`cas/store/tests/filesystem_store_test.rs` line 235 at r3](https://reviewable.io/reviews/allada/turbo-cache/105#-NYo7V_yBGyQa_2M220y:-NYo7V_yBGyQa_2M220z:boynaw0) ([raw file](https://github.com/allada/turbo-cache/blob/69954cb59ec357f070cc0cec11c555ea6b4c42a0/cas/store/tests/filesystem_store_test.rs#L235)):*
> ```XML
>         tokio::task::yield_now().await;
> 
>         {
> ```

nit: This entire section is actually bad to have in a test, it's testing implementation details that are irreverent to callers.

allada (commented):
> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 4 of 13 files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

allada (commented):
> Reviewed 4 of 12 files at r1, 2 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 10 of 13 files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 10 of 13 files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 411 at r3](https://reviewable.io/reviews/allada/turbo-cache/105#-NYoBkB_AmAX26rOrEaz:-NYoBkB_AmAX26rOrEb-:b-bizzsz) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/cas/store/filesystem_store.rs#L411)):*
> ```XML
> 
>         if let Some(old_item) = self.evicting_map.insert(digest, entry).await {
>             // Even though the move is atomic above, it is possible that during a hardlink operation
> ```

BTW, I found why this bug even exists. I'll fix it in a follow up patch:
https://github.com/allada/turbo-cache/issues/106

chrisstaite-menlo:
*[`util/fs.rs` line 103 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYntvKn44OHidl1chZ5:-NYoG1En4bY4ZfNlHwfm:bn5ejki) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/util/fs.rs#L103)):*

> however, the file may change in size (through update\_file) and there are then FileEntry structs laying around with a pointer to a file that is not the size that it says it is.

I don't think the file should change in size. If it does, it is likely a bug. Do you have a case in mind that you are aware of where the file can change sizes (or contents)?
</blockquote></details>

This happens a lot in an Action Cache since the Digest isn't over the file itself.  I actually have some other fixes pending to sort that out...

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/105#-NYoF-nz1eL__wj2MnbF:-NYoF-o-C15FBchEERob:bdxtncq) (no related file):*
After going deep down the rabbit hole of trying to figure out why there might be dangling files and double deletes, I think I found the issue.

It has to do with the complex interaction with `unref()`, `drop()` and `EvictingCache::insert()`. I think when I originally wrote this I was going to have a `FileEntry` represent a digest, but later I changed paradimes where it'd create a new `FileEntry` for each file change. This resulted in `EvictingCache::insert()` to not call `unref()` and the unref logic ended up in `FileSystemStore`'s update call.

I can create a follow up on this or I can get it in now. The concern I have with "fixing" this is that it is difficult to say if it will fix the issue in production. I don't want to end up saying it's fixed, but find out it is not. My question to you is if you are able to run the proposed fixes in an environment that could reproduce the issues to see if they were resolved?

___

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/fs.rs` line 103 at r2](https://reviewable.io/reviews/allada/turbo-cache/105#-NYntvKn44OHidl1chZ5:-NYoIYPC4Th75UJvqZxj:beuszci) ([raw file](https://github.com/allada/turbo-cache/blob/10fa6632ab7f2bfd814f18ab6496b46f9d67edc3/util/fs.rs#L103)):*

This happens a lot in an Action Cache since the Digest isn't over the file itself.  I actually have some other fixes pending to sort that out...
</blockquote></details>

Yeah, the file's contents can change of the `Digest`, but the `FileEntry`'s contents should never change once the struct is created.

CAS `Digest`'s are not allowed to change ever.
AC `Digest`'s are allowed to change.

In the case of the `FileEntry` we just want to ensure any actions performed on it are consistent throughout it's lifetime. But the same `Digest` may result in the same `FileEntry` or a different one throughout the `Digest`'s lifetime.

chrisstaite-menlo:
I'm a fan of code reviews, I love a collaborative process.  It means that I think we can throw away this whole PR in favour of a much better solution from your comments about the purpose of FileEntry.  https://github.com/allada/turbo-cache/pull/107

By the way, hi, and thanks for creating turbo-cache because every other Bazel system was way too complicated (so much so that it was worth me putting a few weeks of developer effort into fixing the issues I was having over the months I had battling other systems).

chrisstaite-menlo:

After going deep down the rabbit hole of trying to figure out why there might be dangling files and double deletes, I think I found the issue.

It has to do with the complex interaction with `unref()`, `drop()` and `EvictingCache::insert()`. I think when I originally wrote this I was going to have a `FileEntry` represent a digest, but later I changed paradimes where it'd create a new `FileEntry` for each file change. This resulted in `EvictingCache::insert()` to not call `unref()` and the unref logic ended up in `FileSystemStore`'s update call.

I can create a follow up on this or I can get it in now. The concern I have with "fixing" this is that it is difficult to say if it will fix the issue in production. I don't want to end up saying it's fixed, but find out it is not. My question to you is if you are able to run the proposed fixes in an environment that could reproduce the issues to see if they were resolved?
</blockquote></details>

I had noticed this with the duplicate call to file_evicted_callback being required.  I think this is fixed in both of my implementations.

chrisstaite-menlo:

I had noticed this with the duplicate call to file_evicted_callback being required.  I think this is fixed in both of my implementations.
</blockquote></details>

I'll spin up our build system with the changes for a week and review the logs.

allada (commented):
> every other Bazel system was way too complicated (so much so that it was worth me putting a few weeks of developer effort into fixing the issues I was having over the months I had battling other systems).

Fun fact, I wrote `turbo-cache` because I implemented Bazel Remote Execution at a previous employer using `buildbarn`. It was processing around 9 petabytes of blobs per month and around 150k hours of single-core CPU hours per day and another 3k hours of GPU hours per day. To get the system working under such high load required a ton of hacks and fixes. When I submitted the fixes to upstream they were met with a "your use case is too obscure for such complicated edge case fixes". The worst part was the testing. It was so painful to write regression tests because ... well... `golang`. So, `turbo-cache` was born. I used my previous employer's use case as the base case for all design. Sadly, I left my previous employer then got `turbo-cache` more production-worthy, so wasn't able to test it on so much expensive hardware.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/105#-:-NYoJhAP6fqdTcgy_d4d:b-lwykvt)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

PR #108, opened 2023-06-26
Overhaul filesystem store to no longer use renameat2

allada:
This PR introduces a massive overhaul to remove the dependency on renameat2 as well as fix bugs in the process. The bugs found are:
* unref() was not being called when insert() replaced an item
* Various extremely unlikely corner cases around files being created, replaced, deleted
* Edge case where the contents of a FileEntry could use the wrong data on data replacement (although very unlikely).

The paradime of "FileEntry" entries and it's corresponding file are now full immutable, meaning if the file gets replaced it will generate a new FileEntry and shuffle the files around so both may be read uninterrupted at any point.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)

### DIFF
--- a/cas/store/BUILD
+++ b/cas/store/BUILD
@@ -216,6 +216,7 @@ rust_library(
     srcs = ["filesystem_store.rs"],
     deps = [
         "//config",
+        "//third_party:async_lock",
         "//third_party:bytes",
         "//third_party:filetime",
         "//third_party:futures",
@@ -395,6 +396,8 @@ rust_test(
     deps = [
         "//config",
         "//third_party:filetime",
+        "//third_party:futures",
+        "//third_party:lazy_static",
         "//third_party:pretty_assertions",
         "//third_party:rand",
         "//third_party:tokio",
--- a/cas/store/filesystem_store.rs
+++ b/cas/store/filesystem_store.rs
@@ -13,21 +13,19 @@
 // limitations under the License.
 
 use std::fmt::{Debug, Formatter};
-use std::path::Path;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::{Duration, SystemTime};
+use std::time::SystemTime;
 
+use async_lock::RwLock;
 use async_trait::async_trait;
 use bytes::BytesMut;
 use filetime::{set_file_atime, FileTime};
 use futures::stream::{StreamExt, TryStreamExt};
-use nix::fcntl::{renameat2, RenameFlags};
-use rand::{thread_rng, Rng};
+use futures::Future;
 use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt, SeekFrom, Take};
 use tokio::task::spawn_blocking;
-use tokio::time::sleep;
 use tokio_stream::wrappers::ReadDirStream;
 
 use buf_channel::{DropCloserReadHalf, DropCloserWriteHalf};
@@ -40,45 +38,141 @@ use traits::{StoreTrait, UploadSizeInfo};
 // Default size to allocate memory of the buffer when reading files.
 const DEFAULT_BUFF_SIZE: usize = 32 * 1024;
 
-struct FileEntry {
-    digest: DigestInfo,
+#[derive(Debug)]
+struct FolderPaths {
+    temp_path: String,
+    content_path: String,
+}
+
+#[derive(Eq, PartialEq, Debug)]
+enum PathType {
+    Content,
+    Temp,
+}
+
+type FileNameDigest = DigestInfo;
+struct EncodedFilePath {
+    folder_paths: Arc<FolderPaths>,
+    path_type: PathType,
+    digest: FileNameDigest,
+}
+
+impl EncodedFilePath {
+    #[inline]
+    fn get_file_path(&self) -> String {
+        let folder = match self.path_type {
+            PathType::Content => &self.folder_paths.content_path,
+            PathType::Temp => &self.folder_paths.temp_path,
+        };
+        to_full_path_from_digest(folder, &self.digest)
+    }
+}
+
+#[inline]
+fn to_full_path_from_digest(folder: &str, digest: &DigestInfo) -> String {
+    format!("{}/{}-{}", folder, digest.str(), digest.size_bytes)
+}
+
+// Note: We don't store the full path of the file here because it would cause
+// a lot of needless memeory bloat. There's a high chance we'll end up with a
+// lot of small files, so to prevent storing duplicate data, we store an Arc
+// to the path of the directory where the file is stored and the packed digest.
+// Resulting in usize + sizeof(DigestInfo).
+pub struct FileEntry {
     file_size: u64,
-    temp_path: Arc<String>,
-    content_path: Arc<String>,
-    // Will be the name of the file in the temp_path if it is flagged for deletion.
-    pending_delete_file_name: AtomicU64,
+
+    // Encoded this way to save memory.
+    encoded_file_path: RwLock<EncodedFilePath>,
+
+    // TODO(allada) Figure out a way to only setup these fields for tests, they
+    // are not used in production code.
     file_evicted_callback: Option<&'static (dyn Fn() + Sync)>,
+    file_unrefed_callback: Option<&'static (dyn Fn(&FileEntry) + Sync)>,
 }
 
 impl FileEntry {
-    async fn read_file_part(&self, offset: u64, length: u64) -> Result<Take<fs::FileSlot<'_>>, Error> {
-        let full_content_path = to_full_path_from_digest(&self.content_path, &self.digest);
-        let mut file = fs::open_file(&full_content_path)
-            .await
-            .err_tip(|| format!("Failed to open file in filesystem store {}", full_content_path))?;
+    pub async fn read_file_part(&self, offset: u64, length: u64) -> Result<Take<fs::FileSlot<'_>>, Error> {
+        let (mut file, full_content_path_for_debug_only) = self
+            .get_file_path_locked(|full_content_path| async move {
+                let file = fs::open_file(&full_content_path)
+                    .await
+                    .err_tip(|| format!("Failed to open file in filesystem store {}", full_content_path))?;
+                Ok((file, full_content_path))
+            })
+            .await?;
 
         file.seek(SeekFrom::Start(offset))
             .await
-            .err_tip(|| format!("Failed to seek file: {}", full_content_path))?;
+            .err_tip(|| format!("Failed to seek file: {}", full_content_path_for_debug_only))?;
         Ok(file.take(length))
     }
 
-    #[inline]
-    fn flag_moved_to_temp_file(&self, rand_file_name: u64) {
-        self.pending_delete_file_name.store(rand_file_name, Ordering::Relaxed);
+    /// This function is a safe way to extract the file name of the underlying file. To protect users from
+    /// accidentally creating undefined behavior we encourage users to do the logic they need to do with
+    /// the filename inside this function instead of extracting the filename and doing the logic outside.
+    /// This is because the filename is not guaranteed to exist after this function returns, however inside
+    /// the callback the file is always guaranteed to exist and immutable.
+    /// DO NOT USE THIS FUNCTION TO EXTRACT THE FILENAME AND STORE IT FOR LATER USE.
+    pub async fn get_file_path_locked<T, Fut: Future<Output = Result<T, Error>>, F: FnOnce(String) -> Fut>(
+        &self,
+        handler: F,
+    ) -> Result<T, Error> {
+        let encoded_file_path = self.encoded_file_path.read().await;
+        handler(encoded_file_path.get_file_path()).await
     }
 }
 
 impl Debug for FileEntry {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
         f.debug_struct("FileEntry")
-            .field("digest", &self.digest)
             .field("file_size", &self.file_size)
-            .field("temp_path", &self.temp_path)
-            .field("content_path", &self.content_path)
-            .field("pending_delete_file_name", &self.pending_delete_file_name)
+            .field("encoded_file_path", &"<behind mutex>")
             .finish()
     }
+}
+
+/// This encapsolates the logic for the edge case of if the file fails to create
+/// the cleanup of the file is handled without creating a FileEntry, which would
+/// try to cleanup the file as well during drop().
+async fn make_file_entry(
+    encoded_file_path: EncodedFilePath,
+    file_evicted_callback: Option<&'static (dyn Fn() + Sync)>,
+    file_unrefed_callback: Option<&'static (dyn Fn(&FileEntry) + Sync)>,
+) -> Result<(FileEntry, fs::FileSlot<'static>, String), Error> {
+    let temp_full_path = encoded_file_path.get_file_path();
+    let temp_file_result = fs::create_file(&temp_full_path)
+        .await
+        .err_tip(|| format!("Failed to create {} in filesystem store", temp_full_path));
+
+    match temp_file_result {
+        Ok(file) => {
+            Ok((
+                FileEntry {
+                    file_size: 0, // Unknown yet, we will fill it in later.
+                    encoded_file_path: RwLock::new(encoded_file_path),
+                    file_evicted_callback,
+                    file_unrefed_callback,
+                },
+                file,
+                temp_full_path,
+            ))
+        }
+        Err(mut err) => {
+            let remove_result = fs::remove_file(&temp_full_path)
+                .await
+                .err_tip(|| format!("Failed to remove file {} in filesystem store", temp_full_path));
+            if let Err(remove_err) = remove_result {
+                err = err.merge(remove_err);
+            }
+            log::warn!("\x1b[0;31mFilesystem Store\x1b[0m: {:?}", err);
+            Err(err)
+        }
+    }
+}
+
+fn make_temp_digest(digest: &mut DigestInfo) {
+    static DELETE_FILE_COUNTER: AtomicU64 = AtomicU64::new(0);
+    digest.packed_hash[24..].clone_from_slice(&DELETE_FILE_COUNTER.fetch_add(1, Ordering::Relaxed).to_le_bytes());
 }
 
 #[async_trait]
@@ -90,61 +184,84 @@ impl LenEntry for FileEntry {
 
     #[inline]
     async fn touch(&self) {
-        let full_content_path = to_full_path_from_digest(&self.content_path, &self.digest);
-        let set_atime_fut = spawn_blocking(move || {
-            set_file_atime(&full_content_path, FileTime::now())
-                .err_tip(|| format!("Failed to touch file in filesystem store {}", full_content_path))
-        });
-        let res = match set_atime_fut.await {
-            Ok(res) => res,
-            Err(_) => Err(make_err!(
-                Code::Internal,
-                "Failed to change atime of file due to spawn failing"
-            )),
-        };
-        if let Err(err) = res {
-            log::error!("{:?}", err);
+        let result = self
+            .get_file_path_locked(move |full_content_path| async move {
+                Ok(spawn_blocking(move || {
+                    set_file_atime(&full_content_path, FileTime::now())
+                        .err_tip(|| format!("Failed to touch file in filesystem store {}", full_content_path))
+                })
+                .await
+                .map_err(|e| {
+                    make_err!(
+                        Code::Internal,
+                        "Failed to change atime of file due to spawn failing {:?}",
+                        e
+                    )
+                })??)
+            })
+            .await;
+        if let Err(e) = result {
+            log::error!("{}", e);
         }
     }
 
+    // unref() only triggers when an item is removed from the eviction_map. It is possible
+    // that another place in code has a reference to `FileEntry` and may later read the
+    // file. To support this edge case, we first move the file to a temp file and point
+    // target file location to the new temp file. `unref()` should only ever be called once.
     #[inline]
     async fn unref(&self) {
-        let rand_file_name = thread_rng().gen::<u64>();
-
-        let from_path = to_full_path_from_digest(&self.content_path, &self.digest);
-        let to_path = to_full_path(&self.temp_path, &format!("{}", rand_file_name));
-        log::info!("\x1b[0;31mFilesystem Store\x1b[0m: Deleting: {}", &from_path);
-        // It is possible (although extremely unlikely) that another thread is reading
-        // this file while we want to delete it here. To prevent errors in either case
-        // we rename the file (since that other thread would have an open file handle)
-        // to the temp folder then delete it when the Arc reference is dropped.
-        if let Err(err) = fs::rename(&from_path, &to_path).await {
-            log::warn!("Failed to rename file from {} to {} : {:?}", from_path, to_path, err);
+        if let Some(callback) = self.file_unrefed_callback {
+            (callback)(&self);
         }
-        self.flag_moved_to_temp_file(rand_file_name);
+        {
+            let mut encoded_file_path = self.encoded_file_path.write().await;
+            if encoded_file_path.path_type == PathType::Temp {
+                // We are already a temp file that is now marked for deletion on drop.
+                // This is very rare, but most likely the rename into the content path failed.
+                return;
+            }
+            let from_path = encoded_file_path.get_file_path();
+            let mut new_digest = encoded_file_path.digest.clone();
+            make_temp_digest(&mut new_digest);
+
+            let to_path = to_full_path_from_digest(&encoded_file_path.folder_paths.temp_path, &new_digest);
+
+            log::info!(
+                "\x1b[0;31mFilesystem Store\x1b[0m: Unref {}, moving file {} to {}",
+                encoded_file_path.digest.str(),
+                &from_path,
+                &to_path
+            );
+            if let Err(err) = fs::rename(&from_path, &to_path).await {
+                log::warn!("Failed to rename file from {} to {} : {:?}", from_path, to_path, err);
+            } else {
+                encoded_file_path.path_type = PathType::Temp;
+                encoded_file_path.digest = new_digest;
+            }
+        }
     }
 }
 
+// TODO(allada) When the `file_evicted_callback` is removed, move `Drop` to EncodedFilePath impl.
 impl Drop for FileEntry {
     fn drop(&mut self) {
-        // If the file was flagged to be deleted (ie: only if unref() was called) delete
-        // the file, but in another spawn. This will ensure we don't delete the files
-        // on safe shutdown as well as not block this thread while we wait on an OS
-        // blocking call.
-        let pending_delete_file_name = self.pending_delete_file_name.load(Ordering::Relaxed);
-        if pending_delete_file_name == 0 {
+        let encoded_file_path = self.encoded_file_path.get_mut();
+        // `drop()` can be called during shutdown, so we use `path_type` flag to know if the
+        // file actually needs to be deleted.
+        if encoded_file_path.path_type == PathType::Content {
             return;
         }
+
+        let file_path = encoded_file_path.get_file_path();
         let file_evicted_callback = self.file_evicted_callback.take();
-        let full_temp_path = to_full_path(&self.temp_path, &pending_delete_file_name.to_string());
         tokio::spawn(async move {
-            log::info!("\x1b[0;31mFilesystem Store\x1b[0m: Store deleting: {}", &full_temp_path);
-            if let Err(err) = fs::remove_file(&full_temp_path).await {
-                log::warn!(
-                    "\x1b[0;31mFilesystem Store\x1b[0m: Failed to remove file {} {:?}",
-                    full_temp_path,
-                    err
-                );
+            log::info!("\x1b[0;31mFilesystem Store\x1b[0m: Deleting: {}", &file_path);
+            let result = fs::remove_file(&file_path)
+                .await
+                .err_tip(|| format!("Failed to remove file {}", file_path));
+            if let Err(err) = result {
+                log::warn!("\x1b[0;31mFilesystem Store\x1b[0m: {:?}", err);
             }
             if let Some(callback) = file_evicted_callback {
                 (callback)();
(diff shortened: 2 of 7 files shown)


allada:
@chrisstaite-menlo, if you want to review this and verify. I am pretty sure this will fix your issues. I'm still currently running manual cloud-based tests to ensure it did not break remote execution, but the logic seems like it will solve the issues you were having.

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 76 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYt_X_K1baV3mG409mW:-NYt_X_K1baV3mG409mX:bnij53e) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L76)):*
> ```XML
>     // represents which item in `folder_paths` to use. The second item is the file
>     // name in the folder.
>     file_path: Mutex<(PathType, FileNameDigest)>,
> ```

I feel like an `tokio::sync::RwLock` would be more appropriate as there can be many concurrent readers for the file (indeed, that is the common case).  It's only on update or delete that a "write" needs to take place.  Given that those operations are rare, an uncontended read lock will be much more efficient here I believe.

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 106 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtaDpL1zhgN_2-y8YP:-NYtaDpL1zhgN_2-y8YQ:bzdm23h) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L106)):*
> ```XML
>     /// This is because the filename is not guaranteed to exist after this function returns, however inside
>     /// the callback the file is always guaranteed to exist and immutable.
>     /// DO NOT USE THIS FUNCTION TO EXTRACT THE FILENAME AND STORE IT FOR LATER USE.
> ```

This is exactly what you do in `read_file_part`, that's fine though because it's for error reporting, but maybe that could be clarified?

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 195 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtb0gwDzMsCCfl4YOx:-NYtb0gwDzMsCCfl4YOy:bzcys4b) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L195)):*
> ```XML
>                 file_path_lock.0 = PathType::Temp;
>                 file_path_lock.1 = new_digest;
>                 self.unrefed.store(true, Ordering::Relaxed);
> ```

I'm wondering if this ought to be `Ordering::Release` and the `Drop` check have `Ordering::Acquire` in case it's immediately dropped on another core.  I'm not very good with Ordering, so I may be being overly cautious here...

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 76 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtbNjmAg5HuNbOPDqo:-NYtbNjn1q2DPKQ2lTb3:b-poe2kp) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L76)):*
> ```XML
>     // represents which item in `folder_paths` to use. The second item is the file
>     // name in the folder.
>     file_path: Mutex<(PathType, FileNameDigest)>,
> ```

There's a lot of references to `file_path.0` and `file_path.1` which isn't very readable.  I wonder if this should be a `struct` of its own and then you could put `to_full_path_for_digest` in its `impl` to tidy it all up a bit.

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 457 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtc1LI9C90RDplx3IJ:-NYtc1LI9C90RDplx3IK:b-m1ntra) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L457)):*
> ```XML
>         //    FileEntry (which has not yet been placed on disk) will not be able to read the file's
>         //    contents until we are done inside `get_file_path_locked()`.
>         entry
> ```

I'm really not convinced by this logic at all.  You need to hold the `Mutex` for `entry.file_path` *before* it's entered into `evicting_map` otherwise another thread could pre-empt you.  See https://github.com/chrisstaite-menlo/turbo-cache/commit/db14b377d2464dd794cd4319eb94c711ec5333af#diff-e057efe67ca2c291064f36fbd348402e9a384f8ea7efe2af36f69a65330f3a19R427 for how I resolved this race.

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 461 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcTsI8fPIpHu1uLvg:-NYtcTsI8fPIpHu1uLvh:bdhrl6k) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L461)):*
> ```XML
>             .get_file_path_locked(|final_path| async move {
>                 let old_item = self.evicting_map.insert(digest, entry).await;
>                 drop(old_item); // Ensure we trigger drop() of old item asap.
> ```

This should not be important at all since it will have been moved in the `unref`.

chrisstaite-menlo:
*[`util/evicting_map.rs` line 266 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcoQiAmaMOZ3QHIl_:-NYtcoQiAmaMOZ3QHIla:b-cua2bn) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L266)):*
> ```XML
>         if let Some(entry) = state.lru.pop(digest) {
>             state.sum_store_size -= entry.data.len() as u64;
>             drop(state);
> ```

I don't think this should have been removed, it releases the `state` lock earlier, which is a good thing, it is not required for the `unref` call.

chrisstaite-menlo (commented):
I really like the method here.  It's a good solution, but I think there's a few places it needs to be touched up before it can be considered correct.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108#-:-NYtczwq1RRLk4Dlu3_K:b-qm40fb)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @allada)

allada (commented):
> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 1 of 6 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 195 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtb0gwDzMsCCfl4YOx:-NYtd40c1vKw5krkpuSo:b7f5aqq) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L195)):*

I'm wondering if this ought to be `Ordering::Release` and the `Drop` check have `Ordering::Acquire` in case it's immediately dropped on another core.  I'm not very good with Ordering, so I may be being overly cautious here...
</blockquote></details>

Ordering is used to tell the optimizer what is allowed to happen, not used at runtime to give promises to other threads. In this case we are not working with anything else that has ordering issues that the optimizer could possibly re-order.

It doesn't matter the ordering in this case, in fact we are already in a mutex, so it would be impossible for `drop()` to be called here. I'm fine changing it, but generally ordering only matters in "hot-paths" where there reading and writing to memory must be preserved in a specific order.

Example:
```
Compiler might change:
fn foo() {
  let b = 1;
  let a = 2;
  if b > 2 {
    println!("bar");
  }
}

to:
fn foo() {
  let a = 2;
  let b = 1;
  if b > 2 {
    println!("bar");
  }
}

so "b" can be in eax register and avoid a potential load instruction. Although this specific example it is not allowed to re-order it this way by spec, atomics work a bit different because they have if-conditions inside the assign operations that does not alter memory layout guarantees, so the optimizer is allowed to be more aggressive unless very unique edge cases are needed to be handled.
```
___
*[`cas/store/filesystem_store.rs` line 461 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcTsI8fPIpHu1uLvg:-NYtgUQg1PzewMP7skc6:b-bteckk) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L461)):*

This should not be important at all since it will have been moved in the `unref`.
</blockquote></details>

It's more about allowing the delete of the file to run in the background place while we are renaming this file. Might be a bit of a needless optimization. I could probably just not assign the `old_item` and I believe rust will drop it on exiting `insert()`.
___
*[`util/evicting_map.rs` line 266 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcoQiAmaMOZ3QHIl_:-NYthJGw9pO1BJRy0fpd:b-rjnltw) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L266)):*

I don't think this should have been removed, it releases the `state` lock earlier, which is a good thing, it is not required for the `unref` call.
</blockquote></details>

This is more about trying to keep consistency of when `unref()` is called. In other locations we will have `state` locked and here is the only place it is allowed to be called without state being locked. I'm worried about in the future the code being changed inside `unref()` where the makes bad assumptions and randomly get race conditions or corner cases because of it.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 1 of 6 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 457 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtc1LI9C90RDplx3IJ:-NYtincZBjoZvdV6dkLQ:b30pi1r) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L457)):*

I'm really not convinced by this logic at all.  You need to hold the `Mutex` for `entry.file_path` *before* it's entered into `evicting_map` otherwise another thread could pre-empt you.  See https://github.com/chrisstaite-menlo/turbo-cache/commit/db14b377d2464dd794cd4319eb94c711ec5333af#diff-e057efe67ca2c291064f36fbd348402e9a384f8ea7efe2af36f69a65330f3a19R427 for how I resolved this race.
</blockquote></details>

I'm not quite sure I follow the issue. Here's where locks are:
* All code in `get_file_path_locked()` callback is protected for the given `FileEntry` (but there may be multiple `FileEntry`'s trying to become the same file at the same time).
  * `EvictingMap.insert()` will hold a lock preventing multiple inserts at the same time.
     * `insert()` may trigger `unref()`, which then will call `get_file_path_locked()` which protects it from being moved while we are in this code.

We can make the guarantee that while `insert()` is executing the newly inserted item cannot be `unrefed()` until `insert()` completes. This prevents a deadlock.

Since you are not allowed to even see where the file is without holding a lock for the entire duration of the location access it protects users from having their file deleted on them.

Now, it is possible for two `FileEntry` objects battle to become the actual file, in which case we should be protected because one of them needs to be replaced/evicted, which triggers `unref()`, which gets blocked until the `rename()` of the first item is done. 

Keep in mind, users may have open file descriptors active to these files that were elsewhere, linux (and I believe all other OSs) do not actually delete the underlying data until the last inode deref and all file descriptors are closed.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 1 of 6 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 76 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYt_X_K1baV3mG409mW:-NYttnBhEsTGX7Rqas9P:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L76)):*

I feel like an `tokio::sync::RwLock` would be more appropriate as there can be many concurrent readers for the file (indeed, that is the common case).  It's only on update or delete that a "write" needs to take place.  Given that those operations are rare, an uncontended read lock will be much more efficient here I believe.
</blockquote></details>

Done.
___
*[`cas/store/filesystem_store.rs` line 76 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtbNjmAg5HuNbOPDqo:-NYttoDf9W0xj-vJkYpx:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L76)):*

There's a lot of references to `file_path.0` and `file_path.1` which isn't very readable.  I wonder if this should be a `struct` of its own and then you could put `to_full_path_for_digest` in its `impl` to tidy it all up a bit.
</blockquote></details>

Done.
___
*[`cas/store/filesystem_store.rs` line 106 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtaDpL1zhgN_2-y8YP:-NYtu-dg4PhAgrgCeSXh:b8bo8k7) ([raw file](https://github.com/allada/turbo-cache/blob/9670a717efa59200ac134afae4d175e540555986/cas/store/filesystem_store.rs#L106)):*

This is exactly what you do in `read_file_part`, that's fine though because it's for error reporting, but maybe that could be clarified?
</blockquote></details>

Done. Added `_for_debug_only` suffix.

chrisstaite-menlo (commented):
Thanks for the discussion on ordering, I guess my confusion comes from the documentation (https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html) that has the line:
"In particular, all previous writes become visible to all threads that perform an Acquire (or stronger) load of this value."

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108#-:-NYvkEku0pb07rQ0yT4d:b-5jncj8)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`cas/store/filesystem_store.rs` line 74 at r4](https://reviewable.io/reviews/allada/turbo-cache/108#-NYvi4hV9swHjcFMliNQ:-NYvi4hV9swHjcFMliNR:b4i4w1r) ([raw file](https://github.com/allada/turbo-cache/blob/62b981f57059061f1ba7f1ee1d8796e6ac3da39e/cas/store/filesystem_store.rs#L74)):*
> ```XML
> pub struct FileEntry {
>     file_size: u64,
>     folder_paths: Arc<FolderPaths>,
> ```

If you move `folder_paths` into `EncodedFilePath` you can actually perform the `Drop` without obtaining the `RwLock` on `encoded_file_path` as it will have all the information you need.  In addition, you can do away with `unrefed` since all `EncodedFilePath` instances with `PathType::Temp` should be deleted on `Drop`.
___
*[`cas/store/filesystem_store.rs` line 176 at r4](https://reviewable.io/reviews/allada/turbo-cache/108#-NYvjYogBMj6HGhfiSYR:-NYvjYogBMj6HGhfiSYS:b-kai46s) ([raw file](https://github.com/allada/turbo-cache/blob/62b981f57059061f1ba7f1ee1d8796e6ac3da39e/cas/store/filesystem_store.rs#L176)):*
> ```XML
>     #[inline]
>     async fn unref(&self) {
>         if let Some(callback) = self.file_unrefed_callback {
> ```

There's some inconsistency with these callbacks (I know they're for testing only, but I thought I'd point it out).  In `Drop` it's called after the file is deleted, but here it's called before the file is renamed, was that intentional?
___
*[`util/evicting_map.rs` line 266 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcoQiAmaMOZ3QHIl_:-NYvgx7b4o5-lC3YOBsT:bwtm8us) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L266)):*

This is more about trying to keep consistency of when `unref()` is called. In other locations we will have `state` locked and here is the only place it is allowed to be called without state being locked. I'm worried about in the future the code being changed inside `unref()` where the makes bad assumptions and randomly get race conditions or corner cases because of it.
</blockquote></details>

I have some stacked commits that are actually trying to do the opposite.  I think the scope of the `Mutex` should be well defined and only operate on the things inside it.  In this case an arbitrary extra constraint is being imposed that would be easy to accidentally ignore.  After spending some time in the last two days reading up about tokio and Mutex's I've got a little clarity.  I have been working on a commit that swaps all of the async mutexes in the code base to sync ones with very tight scopes and swaps the use of a mutex in the scheduler to a single scheduler thread that operates through a channel.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: all files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/evicting_map.rs` line 266 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcoQiAmaMOZ3QHIl_:-NYxO3A_5SeYrEnPPUul:bq2dwgh) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L266)):*

I have some stacked commits that are actually trying to do the opposite.  I think the scope of the `Mutex` should be well defined and only operate on the things inside it.  In this case an arbitrary extra constraint is being imposed that would be easy to accidentally ignore.  After spending some time in the last two days reading up about tokio and Mutex's I've got a little clarity.  I have been working on a commit that swaps all of the async mutexes in the code base to sync ones with very tight scopes and swaps the use of a mutex in the scheduler to a single scheduler thread that operates through a channel.
</blockquote></details>

I am pretty sure we need a semi-global lock while all file moves to happen. In this PR we are using `EvictingMap` to make these guarantees.

If we don't hold a lock here what could happen:
1. `remove()` called
2. `state` locked
3. new `insert()` with same destination attempted
4. `insert()` blocks waiting on `state` lock
3. entry removed from map
6. `state` unlocked (from `remove()`)
7. `insert()` unblocked and adds entry and creates the file.
8. `unref()` called in `remove()` and file is moved.

The issue here is the last step's file is actually the new file, not the old one.

I do notice that this PR is abusing insider-knowledge of how `EvictingMap` works internally. I'll think about a way to do this without adding a new lock, but keeping the magical `EvictingMap` lock knowledge in `FileSystemStore`.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 180 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxSkl7BMD8t7QRd6jS:-NYxe2EO2jb99QsymzaI:b-wllhko) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L180)):*

Is it worth checking that the FileEntry isn't already in the temp path before trying to move it there again?
</blockquote></details>

I added an assert to ensure `unref()` is only called one time and documented the requirement.
___
*[`cas/store/filesystem_store.rs` line 203 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxStDsEMbFvWuA2757:-NYxbedf7tEbWaen_dWG:b3aescq) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L203)):*

There's no need to implement `Drop` for `FileEntry` now, it would be better to implement it for `EncodedFilePath`.
</blockquote></details>

Because of the callback this cant easily be done. I added a TODO, since this should be possible without those callbacks.
___
*[`util/evicting_map.rs` line 266 at r3](https://reviewable.io/reviews/allada/turbo-cache/108#-NYtcoQiAmaMOZ3QHIl_:-NYxeyQUB24cCOb5csKJ:b-277h9n) ([raw file](https://github.com/allada/turbo-cache/blob/e172756613b5398f1ccdaaf258f3f7b80ac4b08e/util/evicting_map.rs#L266)):*

Yes, I think the lock is required until `unref` completes with the current implementation.  If that's a hard requirement for `EvictingMap` (which I think it is) then it may be best to document that in it's API?
</blockquote></details>

Added comments.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 4 of 6 files reviewed, all discussions resolved (waiting on @allada)

___
*[`cas/store/filesystem_store.rs` line 180 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxSkl7BMD8t7QRd6jS:-NYxgE6t4m0YLDKtnyco:bxvl17e) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L180)):*

I added an assert to ensure `unref()` is only called one time and documented the requirement.
</blockquote></details>

It's worth noting that the rename failing when inserting could cause this assert to trigger.  Although, that's probably fine as the disk will be in a weird state.
___
*[`cas/store/filesystem_store.rs` line 203 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxStDsEMbFvWuA2757:-NYxgTJlAGc3K3Q9Ci9l:b-7xvwu8) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L203)):*

Because of the callback this cant easily be done. I added a TODO, since this should be possible without those callbacks.
</blockquote></details>

Can just move the callback into the struct too, but I don't mind that much...

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

chrisstaite-menlo:
*[`cas/store/filesystem_store.rs` line 212 at r7](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxinop-yMUFy9jmp-n:-NYxinoq6TUy7auxKn-s:b-834n6v) ([raw file](https://github.com/allada/turbo-cache/blob/c4209da905ff51c18d724c4d8a260a909ecf3851/cas/store/filesystem_store.rs#L212)):*
> ```XML
>         // `drop()` can be called during shutdown, so we use `path_type` flag to know if the
>         // file actually needs to be deleted.
>         if encoded_file_path.path_type == PathType::Content {
> ```

Oops!

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

chrisstaite-menlo (approved):

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 3 of 7 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 180 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxSkl7BMD8t7QRd6jS:-NYxhRMh8_Ds7bXsZpIv:b-jyh8bk) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L180)):*

It's worth noting that the rename failing when inserting could cause this assert to trigger.  Although, that's probably fine as the disk will be in a weird state.
</blockquote></details>

Oh interesting point, a rename fail on insert should remove the item from the map too. I believe I cleaned up this corner case and added a test for this too.
___
*[`cas/store/filesystem_store.rs` line 203 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxStDsEMbFvWuA2757:-NYyha1LARvWR5mQ3c8A:b-6qdpsk) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L203)):*

Can just move the callback into the struct too, but I don't mind that much...
</blockquote></details>

To keep the two callbacks in the same place, I'd prefer to keep them the way they are. It'll make fixing it later easier.

chrisstaite-menlo (commented):
> Reviewed 3 of 4 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: all files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/store/filesystem_store.rs` line 180 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxSkl7BMD8t7QRd6jS:-NZ-vWXNEuB_WE7_DrQ-:b9te1tl) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L180)):*

Oh interesting point, a rename fail on insert should remove the item from the map too. I believe I cleaned up this corner case and added a test for this too.
</blockquote></details>

Removing the assert and removing the entry from the map if the rename fails causes there to be a case where renaming in the temp directory fails, it is `unref`'d and then it tries to rename from temp to temp.
___
*[`cas/store/filesystem_store.rs` line 483 at r9](https://reviewable.io/reviews/allada/turbo-cache/108#-NZ-xdL98gkmQtQMNvyZ:-NZ-xdL98gkmQtQMNvy_:b-2npsi6) ([raw file](https://github.com/allada/turbo-cache/blob/426131c604413ab82ac098bc0f000a4dbce76679/cas/store/filesystem_store.rs#L483)):*
> ```XML
>             }
>             *encoded_file_path = final_encoded_file_path;
>             return Ok(());
> ```

Could be `Ok(())` as before.  However, I feel like this end may be more readable if both `return`s were removed and was reduced to:
```
    Err(err)
} else {
    *encoded_file_path = final_encoded_file_path;
    Ok(())
}
```
As  the `return` in the `if` is doing heavy lifting due to the `drop(encoded_file_path)`.  I suppose this is the kind of thing that might cause a bug in other languages, but Rust would catch us in the future if we made a mistake.
___
*[`cas/store/filesystem_store.rs` line 512 at r9](https://reviewable.io/reviews/allada/turbo-cache/108#-NZ-yTHqBsug6NO6B0tA:-NZ-yTHqBsug6NO6B0tB:bpj1s9r) ([raw file](https://github.com/allada/turbo-cache/blob/426131c604413ab82ac098bc0f000a4dbce76679/cas/store/filesystem_store.rs#L512)):*
> ```XML
>         // We construct this first, so if the entry is dropped, it will clean up the temp
>         // file during drop().
>         let entry = FileEntry {
> ```

Creating this entry *before* creating the file means that if the file creation fails then trying to delete it in the `Drop` will also fail.  I'd suggest moving the creation of `FileEntry` until after the `fs::create_file` has completed successfully.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/filesystem_store.rs` line 180 at r5](https://reviewable.io/reviews/allada/turbo-cache/108#-NYxSkl7BMD8t7QRd6jS:-NZ1MdwN1bXpvkTaopOX:b-4ta) ([raw file](https://github.com/allada/turbo-cache/blob/f05fdedf11f9cebf1719fa1bf301a2aa1d99a5de/cas/store/filesystem_store.rs#L180)):*

Removing the assert and removing the entry from the map if the rename fails causes there to be a case where renaming in the temp directory fails, it is `unref`'d and then it tries to rename from temp to temp.
</blockquote></details>

I wrote a test for this exact case in `rename_on_insert_fails_due_to_filesystem_error_proper_cleanup_happens`.

I made a short circuit though for this edge case to prevent another rename and possible failure resulting from it.
___
*[`cas/store/filesystem_store.rs` line 483 at r9](https://reviewable.io/reviews/allada/turbo-cache/108#-NZ-xdL98gkmQtQMNvyZ:-NZ1Niuc6_eE5z1bw_cf:b5lkz2v) ([raw file](https://github.com/allada/turbo-cache/blob/426131c604413ab82ac098bc0f000a4dbce76679/cas/store/filesystem_store.rs#L483)):*

Could be `Ok(())` as before.  However, I feel like this end may be more readable if both `return`s were removed and was reduced to:
```
    Err(err)
} else {
    *encoded_file_path = final_encoded_file_path;
    Ok(())
}
```
As  the `return` in the `if` is doing heavy lifting due to the `drop(encoded_file_path)`.  I suppose this is the kind of thing that might cause a bug in other languages, but Rust would catch us in the future if we made a mistake.
</blockquote></details>

I follow the early-return philosophy for code styles. I try to avoid using `else` statements where possible. The rest of the code also avoids `else` statements, to keep things consistent I'd prefer to keep it the way it is here :smile: 
___
*[`cas/store/filesystem_store.rs` line 512 at r9](https://reviewable.io/reviews/allada/turbo-cache/108#-NZ-yTHqBsug6NO6B0tA:-NZ1LUQf8X9RCvawnEVI:bd08dmw) ([raw file](https://github.com/allada/turbo-cache/blob/426131c604413ab82ac098bc0f000a4dbce76679/cas/store/filesystem_store.rs#L512)):*

Creating this entry *before* creating the file means that if the file creation fails then trying to delete it in the `Drop` will also fail.  I'd suggest moving the creation of `FileEntry` until after the `fs::create_file` has completed successfully.
</blockquote></details>

Yeah, I thought about this, but deleting files in the drop is always "best effort", meaning it will only show warnings and not actually cause problems at runtime. Even if the file failed to create, we would want to attempt to delete it because we are unsure what reason it failed to create.... An example is an NFS drive might fail due to a creation timeout, but it really did get created, so we'd want to clean it up if possible.

To help improve this, I moved the logic to it's own function and only create the `FileEntry` when we are sure to have a file.

chrisstaite-menlo (commented):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

allada (commented):
> Reviewed 1 of 6 files at r1, 1 of 1 files at r3, 2 of 4 files at r8, 1 of 1 files at r9, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/108)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #113, opened 2023-06-28
Add timestamps to executor jobs

chrisstaite-menlo:
It kinda does what it says on the tin.  Hopefully this is a nice easy one.

I know you asked for refactors before features, but I thought this might be an easy one over the long discussion I can see us having on things like swapping async Mutex's to sync ones.

### DIFF
--- a/cas/scheduler/tests/scheduler_test.rs
+++ b/cas/scheduler/tests/scheduler_test.rs
@@ -103,6 +103,28 @@ mod scheduler_tests {
 
     const WORKER_TIMEOUT_S: u64 = 100;
 
+    fn check_and_reset_timestamp(msg_for_worker: &mut UpdateForWorker) {
+        match &mut msg_for_worker.update {
+            Some(update) => match update {
+                update_for_worker::Update::StartAction(start_execute) => {
+                    if let Some(timestamp) = &start_execute.queued_timestamp {
+                        assert!(
+                            SystemTime::now()
+                                .duration_since(SystemTime::try_from(timestamp.clone()).expect("Unknown time set"))
+                                .expect("Time went backwards")
+                                .as_millis()
+                                < 500,
+                            "Queued timestamp is out of range"
+                        );
+                    }
+                    start_execute.queued_timestamp = None;
+                }
+                _ => {}
+            },
+            _ => {}
+        }
+    }
+
     #[tokio::test]
     async fn basic_add_action_with_one_worker_test() -> Result<(), Error> {
         const WORKER_ID: WorkerId = WorkerId(0x123456789111);
@@ -124,9 +146,11 @@ mod scheduler_tests {
                         ..Default::default()
                     }),
                     salt: 0,
+                    queued_timestamp: None,
                 })),
             };
-            let msg_for_worker = rx_from_worker.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, expected_msg_for_worker);
         }
         {
@@ -174,11 +198,13 @@ mod scheduler_tests {
                     ..Default::default()
                 }),
                 salt: 0,
+                queued_timestamp: None,
             })),
         };
         {
             // Worker1 should now see execution request.
-            let msg_for_worker = rx_from_worker1.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker1.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, execution_request_for_worker);
         }
 
@@ -211,7 +237,8 @@ mod scheduler_tests {
         }
         {
             // Worker2 should now see execution request.
-            let msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, execution_request_for_worker);
         }
 
@@ -260,9 +287,11 @@ mod scheduler_tests {
                         ..Default::default()
                     }),
                     salt: 0,
+                    queued_timestamp: None,
                 })),
             };
-            let msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, expected_msg_for_worker);
         }
         {
@@ -322,9 +351,11 @@ mod scheduler_tests {
                         ..Default::default()
                     }),
                     salt: 0,
+                    queued_timestamp: None,
                 })),
             };
-            let msg_for_worker = rx_from_worker.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, expected_msg_for_worker);
         }
 
@@ -406,12 +437,14 @@ mod scheduler_tests {
                     ..Default::default()
                 }),
                 salt: 0,
+                queued_timestamp: None,
             })),
         };
 
         {
             // Worker1 should now see execution request.
-            let msg_for_worker = rx_from_worker1.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker1.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, execution_request_for_worker);
         }
 
@@ -448,7 +481,8 @@ mod scheduler_tests {
         }
         {
             // Worker2 should now see execution request.
-            let msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker2.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, execution_request_for_worker);
         }
 
@@ -652,9 +686,11 @@ mod scheduler_tests {
                         ..Default::default()
                     }),
                     salt: 0,
+                    queued_timestamp: None,
                 })),
             };
-            let msg_for_worker = rx_from_worker.recv().await.unwrap();
+            let mut msg_for_worker = rx_from_worker.recv().await.unwrap();
+            check_and_reset_timestamp(&mut msg_for_worker);
             assert_eq!(msg_for_worker, expected_msg_for_worker);
         }
 
--- a/cas/scheduler/worker.rs
+++ b/cas/scheduler/worker.rs
@@ -141,6 +141,7 @@ impl Worker {
         self.send_msg_to_worker(update_for_worker::Update::StartAction(StartExecute {
             execute_request: Some(action_info_clone.into()),
             salt: *action_info.salt(),
+            queued_timestamp: Some(action_info.insert_timestamp.into()),
         }))
     }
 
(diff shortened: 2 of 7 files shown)


allada (changes requested):
> Reviewed 5 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/113)* status: 5 of 7 files reviewed, 8 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/tests/scheduler_test.rs` line 112 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ46NO86twZG7bbNBNW:-NZ46NO86twZG7bbNBNX:bspa0uf) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/scheduler/tests/scheduler_test.rs#L112)):*
> ```XML
>                     if let Some(timestamp) = &start_execute.queued_timestamp {
>                         assert!(
>                             SystemTime::now()
> ```

ditto on `SystemTime` in tests as below.
___
*[`cas/worker/running_actions_manager.rs` line 458 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ43A0_BvRZmqJ4m2FD:-NZ43A0_BvRZmqJ4m2FE:b-jiqm03) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L458)):*
> ```XML
>         {
>             let mut state = self.state.lock().await;
>             state.execution_metadata.worker_start_timestamp = SystemTime::now();
> ```

There is actually a potential for a substantial amount of time to happen between when we received the task and here because we create directories and such earlier. I suggest changing: `create_and_add_action()` to accept an `ExecutionMetadata` then we can easily create one right after:
```rs
Update::StartAction(start_execute) => {
```
and pass it along.
___
*[`cas/worker/running_actions_manager.rs` line 740 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ3tteeCwmgrb8bIdAc:-NZ3tteeCwmgrb8bIdAd:bdmjpao) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L740)):*
> ```XML
>             });
>         }
>         execution_metadata.output_upload_completed_timestamp = SystemTime::now();
> ```

This should be below. At this point we've only created our futures, but the futures may have not completed yet.
___
*[`cas/worker/running_actions_manager.rs` line 754 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ3u0O5Ed-RI31ee2Rm:-NZ3u0O5Ed-RI31ee2Rn:b-z2t9iz) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L754)):*
> ```XML
>             }
>         }
>         drop(output_path_futures);
> ```

After this line is where we are actually done uploading our files.
___
*[`cas/worker/running_actions_manager.rs` line 867 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ3xUAkDPGLHic2Noi-:-NZ3xUAkDPGLHic2Noi0:b-owz8gq) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L867)):*
> ```XML
>             ActionInfo::try_from_action_and_execute_request_with_salt(execute_request, action, start_execute.salt)
>                 .err_tip(|| "Could not create ActionInfo in create_and_add_action()")?;
>         action_info.insert_timestamp = start_timestamp;
> ```

I'm not a super big fan of using this field for when the worker started the job. Can we make a new field for this and use it? It might be worth renaming the `insert_timestamp` to something more like: `scheduled_timestamp` too.
___
*[`cas/worker/running_actions_manager.rs` line 893 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ41CqwBE5_M74NkidU:-NZ41CqwBE5_M74NkidV:b-702kuv) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L893)):*
> ```XML
>     ) -> Result<Arc<RunningActionImpl>, Error> {
>         let queued_timestamp = match &start_execute.queued_timestamp {
>             Some(time) => SystemTime::try_from(time.clone()).unwrap_or(SystemTime::UNIX_EPOCH),
> ```

nit: `time.try_into()` might be easier to read, since it should be able to infer the type elsewhere.
___
*[`cas/worker/tests/running_actions_manager_test.rs` line 87 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ44vbW2zXTTC4bUBDn:-NZ44vbW2zXTTC4bUBDo:btnburi) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/tests/running_actions_manager_test.rs#L87)):*
> ```XML
> 
>     fn check_and_reset_timestamp(name: &str, timestamp: &mut SystemTime) {
>         let now = SystemTime::now();
> ```

It's generally considered bad to use system time in tests. Unit tests are supposed to be allowed to take infinite amount of time between operations and the test should still pass.

The way around this is usually to use fixed time intervals and times that are purely predictable.

Example:
https://github.com/allada/turbo-cache/blob/eebd6bea6ca80c89cfd185f804320e478b5a3524/cas/grpc_service/tests/worker_api_server_test.rs#L166
___
*[`proto/com/github/allada/turbo_cache/remote_execution/worker_api.proto` line 163 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ3qvgi1G4GL_lUdl-r:-NZ3qvgi1G4GL_lUdl-s:b-4nsr0b) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/proto/com/github/allada/turbo_cache/remote_execution/worker_api.proto#L163)):*
> ```proto
> 
>     /// The time at which the command was added to the queue to allow population
>     /// of the ActionResult
> ```

nit: Can you end it in a period? I'm a bit of a stickler for consistency, which is why I prefer to use linters where possible (eg: let the software do the work), but cant easily make a rule for ending comments/sentences in a period.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/113)* status: 5 of 7 files reviewed, 9 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ46UQz4vE4Mu8ypppN:-NZ46UQz4vE4Mu8ypppO:b57127f) (no related file):*
If you'd like I can help out with the tests, it'll require a tiny amount of refactoring... essentially you'll need to make a `now_fn()` that is used instead of `SystemTime::now()` and make it default to`SystemTime::now()` but overrideable with a different constructor.

___

chrisstaite-menlo:
*[`cas/worker/running_actions_manager.rs` line 893 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ41CqwBE5_M74NkidU:-NZ5eHWHBA02NN0mrma3:b-x2t6of) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L893)):*

nit: `time.try_into()` might be easier to read, since it should be able to infer the type elsewhere.
</blockquote></details>

"the trait \`From<&prost\_types::protobuf::Timestamp>\` is not implemented for \`SystemTime\`"

chrisstaite-menlo:
*[`cas/worker/running_actions_manager.rs` line 893 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ41CqwBE5_M74NkidU:-NZ5eTep4I-dLneJgLPi:b-u1mmnc) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L893)):*

"the trait \`From<&prost\_types::protobuf::Timestamp>\` is not implemented for \`SystemTime\`"
</blockquote></details>

Oh, I was missing a `.clone()`!

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/113)* status: 1 of 10 files reviewed, 7 unresolved discussions (waiting on @allada)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ46UQz4vE4Mu8ypppN:-NZ5h0CA9znHUhaYwSqv:b-896fix) (no related file):*

If you'd like I can help out with the tests, it'll require a tiny amount of refactoring... essentially you'll need to make a `now_fn()` that is used instead of `SystemTime::now()` and make it default to`SystemTime::now()` but overrideable with a different constructor.
</blockquote></details>

Done.

___
___
*[`cas/scheduler/tests/scheduler_test.rs` line 112 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ46NO86twZG7bbNBNW:-NZ5h1dOEfrnA0g0fT1X:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/scheduler/tests/scheduler_test.rs#L112)):*

ditto on `SystemTime` in tests as below.
</blockquote></details>

Done.
___
*[`cas/worker/running_actions_manager.rs` line 458 at r1](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ43A0_BvRZmqJ4m2FD:-NZ5o13GCCxnn69v91fi:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/aa2b9484785aa5c7268ebb524ac17b9e9ae12ada/cas/worker/running_actions_manager.rs#L458)):*

There is actually a potential for a substantial amount of time to happen between when we received the task and here because we create directories and such earlier. I suggest changing: `create_and_add_action()` to accept an `ExecutionMetadata` then we can easily create one right after:
```rs
Update::StartAction(start_execute) => {
```
and pass it along.
</blockquote></details>

Done.

allada (approved):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/113)* status: all files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/action_messages.rs` line 83 at r2](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ6rIqPD9FJ-XtSCNMv:-NZ6rIqPD9FJ-XtSCNMw:b-wuz1vg) ([raw file](https://github.com/allada/turbo-cache/blob/1efe437a62e4b0482ee9c6987df87f54ff5db89f/cas/scheduler/action_messages.rs#L83)):*
> ```XML
>     /// The priority of the action. Higher value means it should execute faster.
>     pub priority: i32,
>     /// When this action started to be loaded from the CAS
> ```

super nit: End comments in period.
___
*[`cas/worker/running_actions_manager.rs` line 745 at r2](https://reviewable.io/reviews/allada/turbo-cache/113#-NZ6vQpXEhquAAdiMbUk:-NZ6vQpXEhquAAdiMbUl:b-9q09ql) ([raw file](https://github.com/allada/turbo-cache/blob/1efe437a62e4b0482ee9c6987df87f54ff5db89f/cas/worker/running_actions_manager.rs#L745)):*
> ```XML
>         output_file_symlinks.sort_unstable_by(|a, b| a.name_or_path.cmp(&b.name_or_path));
>         output_directory_symlinks.sort_unstable_by(|a, b| a.name_or_path.cmp(&b.name_or_path));
>         execution_metadata.worker_completed_timestamp = (self.running_actions_manager.now_fn)();
> ```

This is definitely nit picky, but it might be worth putting this below the `self.state.lock().await;` below in order to capture how long it takes even to get the lock... But this is very nit picky.

chrisstaite-menlo:
Oh, it merged before I fixed the nits...

PR #122, opened 2023-07-03
Workers may be running multiple actions, so track them individually.

chrisstaite-menlo:
Currently each worker only tracks the last action that was sent to it,
and they might be running multiple, so if the worker is evicted only
the last action is retried.

Testing this exposes some issues in the do_try_match algorithm, namely
that it continues after performing an immediate_evict_worker which will
invalidate any of its assumptions due to the clone of queued_actions.

This also means that if a single action fails on a worker, we don't abort
all of the actions on that worker, only the action that failed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)

### DIFF
--- a/cas/scheduler/scheduler.rs
+++ b/cas/scheduler/scheduler.rs
@@ -233,72 +233,78 @@ impl SchedulerImpl {
         Ok(rx)
     }
 
+    fn retry_action(&mut self, action_info: &Arc<ActionInfo>, worker_id: &WorkerId) {
+        match self.active_actions.remove(action_info) {
+            Some(running_action) => {
+                let mut awaited_action = running_action.action;
+                let send_result = if awaited_action.attempts >= self.max_job_retries {
+                    Arc::make_mut(&mut awaited_action.current_state).stage = ActionStage::Error((
+                        awaited_action.last_error.unwrap_or_else(|| {
+                            make_err!(
+                                Code::Internal,
+                                "Job cancelled because it attempted to execute too many times and failed"
+                            )
+                        }),
+                        ActionResult {
+                            output_files: Default::default(),
+                            output_folders: Default::default(),
+                            output_directory_symlinks: Default::default(),
+                            output_file_symlinks: Default::default(),
+                            exit_code: INTERNAL_ERROR_EXIT_CODE,
+                            stdout_digest: DigestInfo::empty_digest(),
+                            stderr_digest: DigestInfo::empty_digest(),
+                            execution_metadata: ExecutionMetadata {
+                                worker: format!("{}", worker_id),
+                                queued_timestamp: SystemTime::UNIX_EPOCH,
+                                worker_start_timestamp: SystemTime::UNIX_EPOCH,
+                                worker_completed_timestamp: SystemTime::UNIX_EPOCH,
+                                input_fetch_start_timestamp: SystemTime::UNIX_EPOCH,
+                                input_fetch_completed_timestamp: SystemTime::UNIX_EPOCH,
+                                execution_start_timestamp: SystemTime::UNIX_EPOCH,
+                                execution_completed_timestamp: SystemTime::UNIX_EPOCH,
+                                output_upload_start_timestamp: SystemTime::UNIX_EPOCH,
+                                output_upload_completed_timestamp: SystemTime::UNIX_EPOCH,
+                            },
+                            server_logs: Default::default(),
+                        },
+                    ));
+                    awaited_action.notify_channel.send(awaited_action.current_state.clone())
+                    // Do not put the action back in the queue here, as this action attempted to run too many
+                    // times.
+                } else {
+                    Arc::make_mut(&mut awaited_action.current_state).stage = ActionStage::Queued;
+                    let send_result = awaited_action.notify_channel.send(awaited_action.current_state.clone());
+                    self.queued_actions_set.insert(action_info.clone());
+                    self.queued_actions.insert(action_info.clone(), awaited_action);
+                    send_result
+                };
+
+                if send_result.is_err() {
+                    // Don't remove this task, instead we keep them around for a bit just in case
+                    // the client disconnected and will reconnect and ask for same job to be executed
+                    // again.
+                    log::warn!(
+                        "Action {} has no more listeners during evict_worker()",
+                        action_info.digest().str()
+                    );
+                }
+            }
+            None => {
+                log::error!("Worker stated it was running an action, but it was not in the active_actions : Worker: {:?}, ActionInfo: {:?}", worker_id, action_info);
+            }
+        }
+    }
+
     /// Evicts the worker from the pool and puts items back into the queue if anything was being executed on it.
     /// Note: This will not call .do_try_match().
     fn immediate_evict_worker(&mut self, worker_id: &WorkerId) {
         if let Some(mut worker) = self.workers.remove_worker(&worker_id) {
             // We don't care if we fail to send message to worker, this is only a best attempt.
             let _ = worker.notify_update(WorkerUpdate::Disconnect);
-            if let Some(action_info) = worker.running_action_info {
-                match self.active_actions.remove(&action_info) {
-                    Some(running_action) => {
-                        let mut awaited_action = running_action.action;
-                        let send_result = if awaited_action.attempts >= self.max_job_retries {
-                            Arc::make_mut(&mut awaited_action.current_state).stage = ActionStage::Error((
-                                awaited_action.last_error.unwrap_or_else(|| {
-                                    make_err!(
-                                        Code::Internal,
-                                        "Job cancelled because it attempted to execute too many times and failed"
-                                    )
-                                }),
-                                ActionResult {
-                                    output_files: Default::default(),
-                                    output_folders: Default::default(),
-                                    output_directory_symlinks: Default::default(),
-                                    output_file_symlinks: Default::default(),
-                                    exit_code: INTERNAL_ERROR_EXIT_CODE,
-                                    stdout_digest: DigestInfo::empty_digest(),
-                                    stderr_digest: DigestInfo::empty_digest(),
-                                    execution_metadata: ExecutionMetadata {
-                                        worker: format!("{}", worker_id),
-                                        queued_timestamp: SystemTime::UNIX_EPOCH,
-                                        worker_start_timestamp: SystemTime::UNIX_EPOCH,
-                                        worker_completed_timestamp: SystemTime::UNIX_EPOCH,
-                                        input_fetch_start_timestamp: SystemTime::UNIX_EPOCH,
-                                        input_fetch_completed_timestamp: SystemTime::UNIX_EPOCH,
-                                        execution_start_timestamp: SystemTime::UNIX_EPOCH,
-                                        execution_completed_timestamp: SystemTime::UNIX_EPOCH,
-                                        output_upload_start_timestamp: SystemTime::UNIX_EPOCH,
-                                        output_upload_completed_timestamp: SystemTime::UNIX_EPOCH,
-                                    },
-                                    server_logs: Default::default(),
-                                },
-                            ));
-                            awaited_action.notify_channel.send(awaited_action.current_state.clone())
-                            // Do not put the action back in the queue here, as this action attempted to run too many
-                            // times.
-                        } else {
-                            Arc::make_mut(&mut awaited_action.current_state).stage = ActionStage::Queued;
-                            let send_result = awaited_action.notify_channel.send(awaited_action.current_state.clone());
-                            self.queued_actions_set.insert(action_info.clone());
-                            self.queued_actions.insert(action_info.clone(), awaited_action);
-                            send_result
-                        };
-
-                        if send_result.is_err() {
-                            // Don't remove this task, instead we keep them around for a bit just in case
-                            // the client disconnected and will reconnect and ask for same job to be executed
-                            // again.
-                            log::warn!(
-                                "Action {} has no more listeners during evict_worker()",
-                                action_info.digest().str()
-                            );
-                        }
-                    }
-                    None => {
-                        log::error!("Worker stated it was running an action, but it was not in the active_actions : Worker: {:?}, ActionInfo: {:?}", worker.id, action_info);
-                    }
-                }
+            // We create a temporary Vec to avoid doubt about a possible code
+            // path touching the worker.running_action_infos elsewhere.
+            for action_info in worker.running_action_infos.drain() {
+                self.retry_action(&action_info, &worker_id);
             }
         }
     }
@@ -313,7 +319,6 @@ impl SchedulerImpl {
     // of capabilities of each worker and then try and match the actions to the worker using
     // the map lookup (ie. map reduce).
     fn inner_do_try_match(&mut self) -> ShouldRunAgain {
-        let mut should_run_again = false;
         // TODO(blaise.bruer) This is a bit difficult because of how rust's borrow checker gets in
         // the way. We need to conditionally remove items from the `queued_action`. Rust is working
         // to add `drain_filter`, which would in theory solve this problem, but because we need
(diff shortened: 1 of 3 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)* status: all files reviewed, 8 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 306 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTZFm39XApC5pQBuiE:-NZTZFm39XApC5pQBuiF:b835447) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L306)):*
> ```XML
>             // We create a temporary Vec to avoid doubt about a possible code
>             // path touching the worker.running_action_infos elsewhere.
>             for action_info in worker.running_action_infos.drain().collect::<Vec<Arc<ActionInfo>>>() {
> ```

nit: There's no need to put this into a vec, we have a mutable reference to it, so unless the rustlang compiler is broken or somewhere else is doing some very-unsafe stuff we are the only one with access to all properties on `running_action_infos` (minus any Arc, which has it's own protection).
___
*[`cas/scheduler/scheduler.rs` line 332 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTihuI6PhXJb1HtAgP:-NZTihuI6PhXJb1HtAgQ:b-spnzkf) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L332)):*
> ```XML
>                 Some(awaited_action) => awaited_action,
>                 _ => {
>                     log::warn!(
> ```

nit: This should probably be an `error`, it is pretty dangerous for this to happen.
___
*[`cas/scheduler/scheduler.rs` line 344 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTgsFpD-hQXjEC2EWf:-NZTgsFpD-hQXjEC2EWg:bj36ik6) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L344)):*
> ```XML
>                 // No worker found, check the next action to see if there's a
>                 // matching one for that.
>                 continue;
> ```

This might cause some issues with tasks that have a lot of requirements (lets say `cpu=32`) while most tasks use little resources (lets say `cpu=1`). This might cause the large tasks to never be able to complete. I see a couple ways to solve this, some are easy (configuration) and some are complicated (pre-declare the big-task for the worker).

I'm fine changing it to this, but wanted to make sure you are aware of what might happen with this change. I also think it might be worth a test for this if it's the desired behavior.
___
*[`cas/scheduler/scheduler.rs` line 357 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTiUIV8_If-SJUDhBa:-NZTiUIWBs62F5srr4MJ:bprf42f) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L357)):*
> ```XML
>                 // After evicting the worker, all running actions have been
>                 // placed back in the queue, so start again now.
>                 return true;
> ```

good catch :smile:   
___
*[`cas/scheduler/scheduler.rs` line 404 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTearw2sEJWAJFoE_f:-NZTearw2sEJWAJFoE_g:b1ll2al) ([raw file](https://github.com/allada/turbo-cache/blob/e2e62d20b8badadf20970dde763394310fb24cb7/cas/scheduler/scheduler.rs#L404)):*
> ```XML
>         self.active_actions.insert(action_info, running_action);
> 
>         self.immediate_evict_worker(worker_id);
> ```

My logic of killing the worker completely here is that this kind of error should only happen if there was an error not related to the task, like file not present, kernel errors, wrapper docker container missing, exc...

This stems from a previous deployment of BRE, where we ran everything through a wrapper shell script, that would read a magic docker image variable, then download the container (if needed) to a special shared drive, then run the task inside the container. Then once the task finished, it'd do some health checks on the node to see if the node was "bad". For example, we used to run a lot of GPU tasks, and GPU drivers are notorious for crashing and causing new GPU processes to also fail due to previously ran stuff. To solve this, we would run a light-weight GPU test after every test that used a GPU to see if the results might have failed because the GPU driver got out of wack.

I'm not opposed to putting the node back in the queue, I would just like to have a short discussion about this to see your point of view.
___
*[`cas/scheduler/scheduler.rs` line 412 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTiwo_1WiYn3ISfnui:-NZTiwo_1WiYn3ISfnuj:bcxyiuw) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L412)):*
> ```XML
>             );
>         } else {
>             // Don't set the error on an action that's running somewhere else
> ```

super nit: period.
___
*[`cas/scheduler/scheduler.rs` line 420 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTiycJE3wep2AxXmhH:-NZTiycJE3wep2AxXmhI:bcxyiuw) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L420)):*
> ```XML
>         self.active_actions.insert(action_info.clone(), running_action);
> 
>         // Clear this action from the current worker
> ```

super nit: period.
___
*[`cas/scheduler/scheduler.rs` line 425 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTizifAQoOjPUH4ScW:-NZTizifAQoOjPUH4ScX:bcxyiuw) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L425)):*
> ```XML
>         }
> 
>         // Re-queue the action or fail on max attempts
> ```

super nit: period.
___
*[`cas/scheduler/worker.rs` line 170 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTlveXD2_i4w3_XGK9:-NZTlveXD2_i4w3_XGKA:b-dnd6p4) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/worker.rs#L170)):*
> ```XML
>     }
> 
>     fn restore_platform_properties(&mut self, props: &PlatformProperties) {
> ```

nit: This is not used anywhere else other than `complete_action`, I suggest inlining this function.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)* status: all files reviewed, 8 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/122#-NZToILz1uuKHKuvu1QG:-NZToIM-DQkQj9FxduWg:b54uyxm) (no related file):*
Thanks for this fix!

There was so much design of so many different parts of the code happening when this code was written that some things like this got overlooked. Usually unit tests help find them, but this apparently slipped through. I think this was one of those places where I first was thinking a single node would have multiple "workers" (like buildbarn), then changed my mind and thought it'd be better to use properties to dynamically allocate jobs and each node would only have one "worker" that is multiplexed and so some of the code got caught up in the old design.

___

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)* status: all files reviewed, 4 unresolved discussions (waiting on @allada)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/122#-NZToILz1uuKHKuvu1QG:-NZUpZhH7OfA7ESrQbHQ:b-qqckqq) (no related file):*

Thanks for this fix!

There was so much design of so many different parts of the code happening when this code was written that some things like this got overlooked. Usually unit tests help find them, but this apparently slipped through. I think this was one of those places where I first was thinking a single node would have multiple "workers" (like buildbarn), then changed my mind and thought it'd be better to use properties to dynamically allocate jobs and each node would only have one "worker" that is multiplexed and so some of the code got caught up in the old design.
</blockquote></details>

Yeah, it felt like it was written for a different design.  That makes a lot of sense.

___
___
*[`cas/scheduler/scheduler.rs` line 306 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTZFm39XApC5pQBuiE:-NZUpe973rKrDOai1Kmi:b-tun1pr) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L306)):*

nit: There's no need to put this into a vec, we have a mutable reference to it, so unless the rustlang compiler is broken or somewhere else is doing some very-unsafe stuff we are the only one with access to all properties on `running_action_infos` (minus any Arc, which has it's own protection).
</blockquote></details>

When I wrote it I wasn't confident it was required.  My Rust newbie is showing.
___
*[`cas/scheduler/scheduler.rs` line 332 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTihuI6PhXJb1HtAgP:-NZUpzMC6KLMmt8dvZsS:bsry9xc) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L332)):*

nit: This should probably be an `error`, it is pretty dangerous for this to happen.
</blockquote></details>

I think this is left over from my debugging of the fast-async-mutex not being safe and having races.  I don't think it's actually possible any more and could be put back to an `unwrap()`, but it was happening regularly when I wrote this code (you're seeing the commits in a different order to how they were written).
___
*[`cas/scheduler/scheduler.rs` line 344 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTgsFpD-hQXjEC2EWf:-NZUqMTYDZR3T5CDANGA:bh4unpx) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L344)):*

This might cause some issues with tasks that have a lot of requirements (lets say `cpu=32`) while most tasks use little resources (lets say `cpu=1`). This might cause the large tasks to never be able to complete. I see a couple ways to solve this, some are easy (configuration) and some are complicated (pre-declare the big-task for the worker).

I'm fine changing it to this, but wanted to make sure you are aware of what might happen with this change. I also think it might be worth a test for this if it's the desired behavior.
</blockquote></details>

I don't believe this is actually a change in behaviour over the previous algorithm.  Previously it was breaking out of the inner loop that was looking for a worker that would respond for the job.  The new method does away with the inner loop because it just starts again after evicting a worker.

That's not to say the situation you describe isn't a problem, but it was previously a problem also.
___
*[`cas/scheduler/scheduler.rs` line 357 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTiUIV8_If-SJUDhBa:-NZUquPv7gdKXUWLHf_C:b-t3ozon) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L357)):*

good catch :smile:   
</blockquote></details>

Trust me, this was through a lot of trial and error.
___
*[`cas/scheduler/scheduler.rs` line 404 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTearw2sEJWAJFoE_f:-NZUr4FGE8ovrNPVDUO-:bvipp82) ([raw file](https://github.com/allada/turbo-cache/blob/e2e62d20b8badadf20970dde763394310fb24cb7/cas/scheduler/scheduler.rs#L404)):*

My logic of killing the worker completely here is that this kind of error should only happen if there was an error not related to the task, like file not present, kernel errors, wrapper docker container missing, exc...

This stems from a previous deployment of BRE, where we ran everything through a wrapper shell script, that would read a magic docker image variable, then download the container (if needed) to a special shared drive, then run the task inside the container. Then once the task finished, it'd do some health checks on the node to see if the node was "bad". For example, we used to run a lot of GPU tasks, and GPU drivers are notorious for crashing and causing new GPU processes to also fail due to previously ran stuff. To solve this, we would run a light-weight GPU test after every test that used a GPU to see if the results might have failed because the GPU driver got out of wack.

I'm not opposed to putting the node back in the queue, I would just like to have a short discussion about this to see your point of view.
</blockquote></details>

The reason I changed this (and I actually have some more changes that relate to this later) is because we have some areas of the code base that use a lot of RAM to compile (around 2GB per file).  Usually, each job uses less than 1GB, so we don't have any issues.  However, if we happen to hammer a worker with a lot of RAM hungry jobs the Kernel gets out the OOM killer and takes out one of the compile jobs.

The process then went:
- Job killed
- All jobs re-queued (even the successful ones)
- Worker comes back after being dumped by the scheduler (and all the other non-OOM jobs still running)
- Worker gets allocated all the same tasks
- OOM killed even quicker
- All jobs fail

It was ugly and terrible.  This way, we only tend to loose that one job that ran out of RAM.  With some load-balancing improvements, killing all running jobs on disconnect and backpressure we've removed this problem entirely now.
___
*[`cas/scheduler/worker.rs` line 170 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTlveXD2_i4w3_XGK9:-NZUsQ4Z30T4yvK9Ww76:b-qc6wpi) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/worker.rs#L170)):*

nit: This is not used anywhere else other than `complete_action`, I suggest inlining this function.
</blockquote></details>

The same is true for `reduce_platform_properties`, but I think it makes logical sense to have them as a pair.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)* status: all files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 344 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTgsFpD-hQXjEC2EWf:-NZWK6SSAuBItoUucU-e:b-fl5ppa) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L344)):*

I don't believe this is actually a change in behaviour over the previous algorithm.  Previously it was breaking out of the inner loop that was looking for a worker that would respond for the job.  The new method does away with the inner loop because it just starts again after evicting a worker.

That's not to say the situation you describe isn't a problem, but it was previously a problem also.
</blockquote></details>

I don't think so...

The old flow iterated actions in queued order, if a worker was not found for an action it would essentially return from `inner_do_try_match` The `do_try_match` loop would try again, then break out of the outer loop because no actions got matched. This would make sure we never give a task out of order, since if it did not hand off the task to a worker it would not even evaluate the next task.

The new flow iterates actions in queued order, if a worker was not found for an action it would go to the next action and try and find a worker for it. This means that if the first item in the queue could not find a worker capable of running it, it would continue scheduling more actions and leave the first item in the queue.

This behavior might be semi-desired, but I could see it causing issues if people want to mix jobs with a lot of requirements with jobs that have few requirements on nodes that can handle both.
___
*[`cas/scheduler/scheduler.rs` line 404 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTearw2sEJWAJFoE_f:-NZWMY3Q7EB88MvJTHwS:b-6dml89) ([raw file](https://github.com/allada/turbo-cache/blob/e2e62d20b8badadf20970dde763394310fb24cb7/cas/scheduler/scheduler.rs#L404)):*

The reason I changed this (and I actually have some more changes that relate to this later) is because we have some areas of the code base that use a lot of RAM to compile (around 2GB per file).  Usually, each job uses less than 1GB, so we don't have any issues.  However, if we happen to hammer a worker with a lot of RAM hungry jobs the Kernel gets out the OOM killer and takes out one of the compile jobs.

The process then went:
- Job killed
- All jobs re-queued (even the successful ones)
- Worker comes back after being dumped by the scheduler (and all the other non-OOM jobs still running)
- Worker gets allocated all the same tasks
- OOM killed even quicker
- All jobs fail

It was ugly and terrible.  This way, we only tend to loose that one job that ran out of RAM.  With some load-balancing improvements, killing all running jobs on disconnect and backpressure we've removed this problem entirely now.
</blockquote></details>

Hmmm I see. Here's some ideas:

1. We put the worker in a "draining" state where no more jobs are queued onto it, but the remaining jobs can continue. Once no more jobs are running (or timeout happens), it fully evicts the worker.
2. We send a special command to the worker asking it to evaluate if it should be fully evicted (like a health check).
3. Adjust `ExecuteResult::result::internal_error` to request that the scheduler not evict the instance since it was able to self-repair.

I feel 1 is the easiest to implement.

I do not want to block this PR based on this, but I am interested to see how you guys prefer to deal with spurious errors. Previous job we would prefer to just kill the entire instance on any error, but our stuff was 100% cloud and having a single job give false negatives cost more than re-running everything.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122)* status: all files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/scheduler.rs` line 344 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTgsFpD-hQXjEC2EWf:-NZWQWDY5wKNmICnhD0g:b-xoqaua) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L344)):*

I don't think so...

The old flow iterated actions in queued order, if a worker was not found for an action it would essentially return from `inner_do_try_match` The `do_try_match` loop would try again, then break out of the outer loop because no actions got matched. This would make sure we never give a task out of order, since if it did not hand off the task to a worker it would not even evaluate the next task.

The new flow iterates actions in queued order, if a worker was not found for an action it would go to the next action and try and find a worker for it. This means that if the first item in the queue could not find a worker capable of running it, it would continue scheduling more actions and leave the first item in the queue.

This behavior might be semi-desired, but I could see it causing issues if people want to mix jobs with a lot of requirements with jobs that have few requirements on nodes that can handle both.
</blockquote></details>

This might have been the intention, but it was not the reality:
```
for action_info in action_infos {
    while worker_received_msg == false {
...
                    // No worker found, break out of our loop.
                    break;
...
    }
}
```
___
*[`cas/scheduler/scheduler.rs` line 404 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTearw2sEJWAJFoE_f:-NZWRPZT1IUjbYB_R8-j:b-c0sind) ([raw file](https://github.com/allada/turbo-cache/blob/e2e62d20b8badadf20970dde763394310fb24cb7/cas/scheduler/scheduler.rs#L404)):*

Hmmm I see. Here's some ideas:

1. We put the worker in a "draining" state where no more jobs are queued onto it, but the remaining jobs can continue. Once no more jobs are running (or timeout happens), it fully evicts the worker.
2. We send a special command to the worker asking it to evaluate if it should be fully evicted (like a health check).
3. Adjust `ExecuteResult::result::internal_error` to request that the scheduler not evict the instance since it was able to self-repair.

I feel 1 is the easiest to implement.

I do not want to block this PR based on this, but I am interested to see how you guys prefer to deal with spurious errors. Previous job we would prefer to just kill the entire instance on any error, but our stuff was 100% cloud and having a single job give false negatives cost more than re-running everything.
</blockquote></details>

Option 1 seems the simplest solution that deals with all cases.

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122#-:-NZX-9tdAh2KXxoEAP4a:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 344 at r1](https://reviewable.io/reviews/allada/turbo-cache/122#-NZTgsFpD-hQXjEC2EWf:-NZWzBaI2c_Erqkblccp:bbia9fx) ([raw file](https://github.com/allada/turbo-cache/blob/69ac295826dba9cd70d2c1a22a83cce352b5d6e0/cas/scheduler/scheduler.rs#L344)):*

This might have been the intention, but it was not the reality:
```
for action_info in action_infos {
    while worker_received_msg == false {
...
                    // No worker found, break out of our loop.
                    break;
...
    }
}
```
</blockquote></details>

Oh, interesting. I completely missed that while statement.

allada (approved):
I think you are now able to push the changes yourself.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/122#-:-NZX-Az_-ntL9Cz638Gn:bbwk4yl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #129, opened 2023-07-04
If a stale work directory exists, try deleting it.

chrisstaite-menlo:
There are times when a job is not cleaned up due to improper shutdown and the work directory still existed from before.  This PR tries to clean it up and create a new one if this occurs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/129)

### DIFF
--- a/cas/worker/BUILD
+++ b/cas/worker/BUILD
@@ -135,6 +135,7 @@ rust_test(
         "//util:common",
         "//util:error",
         "@crate_index//:futures",
+        "@crate_index//:hex",
         "@crate_index//:pretty_assertions",
         "@crate_index//:prost",
         "@crate_index//:rand",
--- a/cas/worker/running_actions_manager.rs
+++ b/cas/worker/running_actions_manager.rs
@@ -847,6 +847,11 @@ impl RunningActionsManagerImpl {
     async fn make_work_directory(&self, action_id: &ActionId) -> Result<String, Error> {
         let work_directory = format!("{}/{}", self.root_work_directory, hex::encode(action_id));
         fs::create_dir(&work_directory)
+            .or_else(|original_error| {
+                fs::remove_dir_all(&work_directory)
+                    .and_then(|_| fs::create_dir(&work_directory))
+                    .or_else(|_| async move { Err(original_error) })
+            })
             .await
             .err_tip(|| format!("Error creating work directory {}", work_directory))?;
         Ok(work_directory)
--- a/cas/worker/tests/running_actions_manager_test.rs
+++ b/cas/worker/tests/running_actions_manager_test.rs
@@ -22,6 +22,7 @@ use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use futures::{FutureExt, TryFutureExt};
+use hex;
 use prost::Message;
 use rand::{thread_rng, Rng};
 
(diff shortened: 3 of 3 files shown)


allada (changes requested):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/129)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/129#-NZXozw01mz2FzlCoEV-:-NZXozw01mz2FzlCoEV0:bhnnayl) (no related file):*
I'm looking over the code and I think if this is happening something is very wrong. 

Here we delete everything in the root directory on service startup:
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/local_worker.rs#L230C13-L230C27

Here, the cleanup is supposed to happen:
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/running_actions_manager.rs#L769

Here's where cleanup should be invoked. It should happen always if `create_and_add_action()` succeeds (which is where you are trying to cleanup before action is invoked, because it calls `make_work_directory()`):
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/local_worker.rs#L151C63-L151C63
I double checked that `.then()` works with chaining the way it is dealt with:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e07dfac8760d67d6ee386e0d1cb034ec

I'm curious if this might have been as a side-effect of the previous bug you fixed, where the worker would continue running on the node when the scheduler terminated it, then the scheduler sent the same job to the same node after re-connect and it tried to use the same directory?

___

chrisstaite-menlo:

I'm looking over the code and I think if this is happening something is very wrong. 

Here we delete everything in the root directory on service startup:
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/local_worker.rs#L230C13-L230C27

Here, the cleanup is supposed to happen:
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/running_actions_manager.rs#L769

Here's where cleanup should be invoked. It should happen always if `create_and_add_action()` succeeds (which is where you are trying to cleanup before action is invoked, because it calls `make_work_directory()`):
https://github.com/allada/turbo-cache/blob/d5468dc3965be32ffffa516f4f302269715288bf/cas/worker/local_worker.rs#L151C63-L151C63
I double checked that `.then()` works with chaining the way it is dealt with:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e07dfac8760d67d6ee386e0d1cb034ec

I'm curious if this might have been as a side-effect of the previous bug you fixed, where the worker would continue running on the node when the scheduler terminated it, then the scheduler sent the same job to the same node after re-connect and it tried to use the same directory?

</blockquote></details>

It may well be true that this can't happen any more.

allada:
If you are ok with it, can we hold off on this PR to see if it was fixed by the other PR?

chrisstaite-menlo:
Yeah, makes sense.  I've got to get everything else up to date before I can test all of this because we haven't even started on the Goma PRs yet, but happy to leave this open indefinately.

allada:
I think I found the problem. Here's the scenario:
* Scheduler disconnects
* https://github.com/allada/turbo-cache/blob/b354b2511eafd0f58625904d0bb6c3d45d353080/cas/worker/local_worker.rs#L356 triggers
* `kill_action` is triggered
* `kill_action` sends the kill message on the message channel then returns
* Scheduler reconnects
* Same action is received as before
* First action is still running, but about to be killed and cleaned up
* New action (with same name) is created and fails because folder exists

What should happen is `kill_action` should return only when we are certain the action is 100% finished and no all related objects are dropped. Maybe a `CondVar` or a `OneShot` channel would fix this.

chrisstaite-menlo:
I've not seen this error with all the other changes, so I'm going to close this off.

chrisstaite-menlo:
```
[2023-09-07T08:29:17.678Z WARN  simple_scheduler] Internal error for worker 7b2e3f8d-0afb-4774-82e3-cecef835dbf3: Error { code: AlreadyExists, messages: ["File exists (os error 17) : Error creating work directory /root/.cache/turbo-cache/work/f5596fdea33511e1c7f513873c087faee689c6ee024c9d90d2908c351e5b0e83"] }
```
A worker OOM-d and when it came back it was assigned the same tasks

PR #133, opened 2023-07-05
Move trivial data access Mutex's to sync Mutex.

chrisstaite-menlo:
The Tokio documentation suggests that using an async Mutex for trivial data access (where the data is not required over an `await`) is more expensive and therefore should be avoided.  This is due to the fact that the operations performed while the Mutex is locked are trivial and therefore switching task on the tokio thread is far more inefficient than simply waiting for the Mutex to become available.

See: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html and https://tokio.rs/tokio/tutorial/shared-state

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/133)

PR #134, opened 2023-07-05
Move the scheduler to a thread rather than use a Mutex.

chrisstaite-menlo:
For the same reasoning as #133 the scheduler makes heavy use of a Mutex.  This means that the threads that perform operations on the data in the SchedulerImpl are constantly switching context.  It is much more efficient to have a single thread dealing with all data requests for the internal data of SchedulerImpl and then have channels send requests to perform mutations on that data.

By ensuring all data access to SchedulerImpl is performed on a single thread, all locking is removed and instead channels are used to pass data in and out of the scheduler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/134)

### DIFF
--- a/cas/scheduler/BUILD
+++ b/cas/scheduler/BUILD
@@ -28,7 +28,6 @@ rust_library(
         "//config",
         "//util:common",
         "//util:error",
-        "@crate_index//:async-lock",
         "@crate_index//:lru",
         "@crate_index//:rand",
         "@crate_index//:tokio",
--- a/cas/scheduler/scheduler.rs
+++ b/cas/scheduler/scheduler.rs
@@ -17,7 +17,6 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::Arc;
 use std::time::SystemTime;
 
-use async_lock::Mutex;
 use lru::LruCache;
 use rand::{thread_rng, Rng};
 use tokio::sync::watch;
@@ -426,7 +425,7 @@ impl SchedulerImpl {
         self.retry_action(&action_info, &worker_id);
     }
 
-    async fn update_action(
+    fn update_action(
         &mut self,
         worker_id: &WorkerId,
         action_info_hash_key: &ActionInfoHashKey,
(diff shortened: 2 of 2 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/134)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/134#-NZax2pdAdBMaaE8RB8N:-NZax2peBijaDk8kYo71:b-nsh946) (no related file):*
I'm not sure I agree with this PR. Each call now incurs about a 32-64 byte heap allocation and I think it complicates the code a bit more.

It also looks like none of the actual inner functions need async at all, `update_action()` is marked wrong.

Now the only concern is some of the expensive operations should not be taking up tokio runtime threads, like `do_try_match()`, since it's pretty much an `O(n^2)`. The cool thing about `do_try_match()` is that it can be yieldable. So what we could do is any time `do_try_match()` is invoked, it instead spawns it in a spawn (if there is not one running) and in the spawn it could grab a lock, run a few iterations then release the lock, then run a few more again until it completes.

This would allow us to go to a pure blocking Mutex and it would allow us to even save resources, because `do_try_match()` would queue up multiple requests to run it and do them all together, instead of how it currently is, and every new job causes it to run again... (ie: 100 jobs get added quickly, only 1-2 iterations of `do_try_match()` get invoked).

Thoughts?

___

chrisstaite-menlo:
Yeah, that sounds like a sensible approach.  I had originally switched this to use a sync Mutex, but decided that it would be locking for too long, hence changing the approach.  Your suggestion mitigates that.  Did you want me to have a go at it, or would you like to take the reins on this one?

allada:
> Did you want me to have a go at it, or would you like to take the reins on this one?

It is up to you. I could probably nock it out pretty quick if you'd like, but if you think it'd be fun for you to do, go for it :smile:

chrisstaite-menlo:
Happy for you to take it.

PR #135, opened 2023-07-05
Optionally cache and retrieve ActionResults in an Action Cache.

chrisstaite-menlo:
This PR makes use of the action cache at the scheduler to returned cached results for actions that have previously been executed.  It also cached completed actions back into the action cache.

Previously this was configured as an option for the worker, but was never implemented.  It seems to make sense to bring this forward rather than requiring a worker to perform a store lookup.

This is currently lacking tests, but I thought it was worth starting the review to get comments on the approach before spending a lot of time on writing tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/135)


chrisstaite-menlo:
The integration tests are failing (presumably because I removed the ac_store from the worker config).  I don't know how to fix them, if you have any idea what I've missed I would be grateful.  I can't figure out how to run the test locally.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/135)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/135#-NZbDJrb8xMhRo7SUiKg:-NZbDJrc7GNnjIN-OPJ4:bxh384q) (no related file):*
So Bazel usually takes care of this, but if you are using Goma or another BRE client, it is very possible they don't do this work.

At first I was thinking... maybe this should be offloaded to a CAS server, since we want to keep the Scheduler ultra-light-weight, since it is the single-point-of-failure in this design. Then after thinking about it, I think the scheduler should work similar to how the `Store` interface works and there should be multiple implementations of `Scheduler` and one of them would support cache lookup and cache saving.

The reason I'm thinking about it this way is because it would easily allow us to build a `GrpcScheduler` similar to `GrpcStore` and it could forward these lookups to another instance to help offload this heavy lifting. For example a CAS instance could serve both normal CAS operations and `CacheLookupScheduler` API calls through the same instance (and it'd likely be 100x faster, since lookups can be quite IO intensive and it'd live on or near the CAS store).

Lastly, @aaronmondal and I were talking about possibly making different `Scheduler` implementations here https://github.com/allada/turbo-cache/issues/116 to support k8s and redis/(or similar) backed `Scheduler` so you could spin up unlimited amount of schedulers and put them behind a load balancer allowing any worker or client to access any scheduler at any point without the worry of any scheduler going offline causing issues.

If you'd like, I can abstract out the scheduler into it's own config. Converting this PR to a scheduler implementation should be fairly straight-forward from that point.

___

chrisstaite-menlo:
That's actually kinda perfect as we have a use case for a remote cache that we currently have to forward AC requests over to the real scheduler and I have to fool around with nginx to make it work properly.  That idea would solve that problem too.

chrisstaite-menlo:
Going to implement as two separate PRs, one to populate the AC at the worker and one to create a Scheduler implementation that can lookup in the AC before scheduling an action.

PR #136, opened 2023-07-05
Move scheduler to use blocking mutexes

allada:
This will improve the performance of the scheduler by quite a bit since the scheduler will not need to context switch very much.

The big change here is that `do_try_match()` will now run in a spawn and is notified when it needs to be run. This will add another layer of improvements, since this function is so expensive to run, it allows many changes to tasks<->workers to be queued up and run the matching all at once instead of each time any change happens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)


chrisstaite-menlo (changes requested):
I like the solution, but I'm worried about the edge cases here.

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136#-:-NZf8qmKBVjLeRS1miAr:b-t7rb2v)* status: all files reviewed, 4 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/scheduler.rs` line 353 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf3XyNCkD4_mzJdXCX:-NZf3XyNCkD4_mzJdXCY:b-tnxruu) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L353)):*
> ```XML
>                 // Remove worker, as it is no longer receiving messages and let it try to find another worker.
>                 log::warn!("Worker command failed, removing worker {}", worker_id);
>                 self.immediate_evict_worker(&worker_id);
> ```

I see an issue here, when `immediate_evict_worker` is called from within the scheduler thread it the `notify_one` will be discarded as it only gets taken when it's waiting.  I'm also trying to convince myself that because it holds the mutex when running the loop that it will get to the await before the next time it awaits the `Notify`, but it won't because of the sleep callback.  I think there's a lot of races here with the `Notify`.  I think something a bit more complex is required unfortunately.

I'm wondering about a `Semaphore` but one-way.  Each "notify" adds a permit to the "Semaphore" and then "acquire" takes all of the permits.  This is pretty simple to do with a `AtomicU64`, except you can't `await` that becoming non-zero (I don't think!).  The `tokio::sync::Semaphore` doesn't support taking all of the permits, so I don't think it's possible to implement with that either.

Having looked at the implementations of `tokio::sync::Semaphore` and `async_locks::Semaphore` this should be fairly trivial to implement with an `Event` and an `AtomicUsize`.  Add to the atomic for every `notify` and then wake the `Event` then at the start of each loop swap the atomic for 0 and if the exchanged value is 0 wait on the `Event` otherwise continue.  That should make sure that the inner loop is run at least once for each time it is required.
___
*[`cas/scheduler/scheduler.rs` line 504 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf6cXMCBP4VMEdCpXn:-NZf6cXMCBP4VMEdCpXo:b-7xhqg) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L504)):*
> ```XML
>     inner: Arc<Mutex<SchedulerImpl>>,
>     platform_property_manager: PlatformPropertyManager,
>     task_worker_matching_future: JoinHandle<()>,
> ```

Out of interest, why are you storing the `JoinHandle` I don't think it's required for the thread to execute?
___
*[`cas/scheduler/scheduler.rs` line 558 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf7Nud9JZtjBZdi-kn:-NZf7Nud9JZtjBZdi-ko:b-wsdauf) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L558)):*
> ```XML
>             inner: inner.clone(),
>             platform_property_manager,
>             task_worker_matching_future: tokio::spawn(async move {
> ```

Why not just run this on a standard Rust thread?  That way we don't have to worry about yielding.  Nothing in the loop uses IO, so it makes sense to use a `std::thread::spawn`.  If you're using a Rust thread you'd just have to use a `std::sync::Condvar` instead of a `tokio::sync::Event`.
___
*[`cas/scheduler/scheduler.rs` line 691 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf8_AE8Iu6Cik73SoK:-NZf8_AE8Iu6Cik73SoL:b-xvcir) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L691)):*
> ```XML
> impl Drop for Scheduler {
>     fn drop(&mut self) {
>         self.task_worker_matching_future.abort();
> ```

Should probably do a `notify` to allow the thread to try and upgrade the `Arc` and fail otherwise it will never die.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: all files reviewed, 4 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 353 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf3XyNCkD4_mzJdXCX:-NZfbdq0B0hQfkN7A3_o:bh16tfh) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L353)):*
According to the `Notify` documentation, it will always hold at most 1 permit. This means that if no notify signal is waiting and another place sends a notify signal, it'll hold 1 permit so when the next `.notified()` call is done, it will grab the permit.

from the documentation (https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html):
> A `Notify` can be thought of as a [`Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html "struct tokio::sync::Semaphore") starting with 0 permits. The [`notified().await`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notified "method tokio::sync::Notify::notified") method waits for a permit to become available, and [`notify_one()`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_one "method tokio::sync::Notify::notify_one") sets a permit **if there currently are no available permits**.

...

> If `notify_one()` is called **before** `notified().await`, then the next call to `notified().await` will complete immediately, consuming the permit. Any subsequent calls to `notified().await` will wait for a new permit.

Keep in mind `Notify` acts just like a conditional variable, but not documented that way.

Here's a sample to show `Notify` doesn't need an active listener:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df51a40c100a7eb5fbc5ea6e0b28d80e
___
*[`cas/scheduler/scheduler.rs` line 504 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf6cXMCBP4VMEdCpXn:-NZfcxfL8ODr9m7lzMBJ:b-htaf9h) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L504)):*

Out of interest, why are you storing the `JoinHandle` I don't think it's required for the thread to execute?
</blockquote></details>

I need to store either the `JoinHandle` or `AbortHandle`, so when the `Scheduler` is dropped it terminates the spawn/thread.
___
*[`cas/scheduler/scheduler.rs` line 558 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf7Nud9JZtjBZdi-kn:-NZfdTbM6MiS_r6h5QBA:b-r5oe6n) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L558)):*

Why not just run this on a standard Rust thread?  That way we don't have to worry about yielding.  Nothing in the loop uses IO, so it makes sense to use a `std::thread::spawn`.  If you're using a Rust thread you'd just have to use a `std::sync::Condvar` instead of a `tokio::sync::Event`.
</blockquote></details>

I actually made a bunch of changes to do just that and after getting everything all done and nice and tidy I started working on fixing the tests. This was a nightmare because when things run in other threads, it is extremely difficult for the tests to manage order of operations.

I fixed all the order of operations, but it became really easy to introduce flakes. It also made tests much more difficult to debug, because we would now need to wait for state to equal something rather than reading state and expecting it to equal something. For example, right now you can do things like:
```rs
assert_eq!(client_rx.borrow_and_update().stage, ActionStage::Executing);
```

But it'd need to be changed to something like:
```rs
client_rx.wait_for(|v| v.stage == ActionStage::Executing).await
```

It may look about the same, but in terms of debugging it is horrible, because the `wait_for` will just never finish, where the first one will at least show you where the expectations where missed/wrong.
___
*[`cas/scheduler/scheduler.rs` line 691 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf8_AE8Iu6Cik73SoK:-NZff3Sg29zL8CPWALDM:bnza10p) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L691)):*

Should probably do a `notify` to allow the thread to try and upgrade the `Arc` and fail otherwise it will never die.
</blockquote></details>

Actually, `.abort()` manages this for us. I originally did just that, I was dropping the `inner` then sending a notify, but then after thinking about it, I realized that `.abort()` tells tokio to completely remove the spawn/strand and drop everything about it (including dropping the captured variables). In a nut shell, `.abort()` tells tokio to remove the future (and all decedents) from the runtime at the next `.await`, so it just never continues.

chrisstaite-menlo:
*[`cas/scheduler/scheduler.rs` line 354 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZjQBikExakchuE4DRK:-NZjQBikExakchuE4DRL:b-4ev6mh) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L354)):*
> ```XML
>                 log::warn!("Worker command failed, removing worker {}", worker_id);
>                 self.immediate_evict_worker(&worker_id);
>                 return;
> ```

I think this needs a `notify_one` otherwise it won't re-run the loop immediately as `immediate_evict_worker` won't re-queue.  In fact, we probably don't even want the 1ms wait in this case, right?

chrisstaite-menlo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: all files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/scheduler.rs` line 353 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf3XyNCkD4_mzJdXCX:-NZjPhtl46eHNRxTf4Xr:b-15fqy2) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L353)):*

According to the `Notify` documentation, it will always hold at most 1 permit. This means that if no notify signal is waiting and another place sends a notify signal, it'll hold 1 permit so when the next `.notified()` call is done, it will grab the permit.

from the documentation (https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html):
> A `Notify` can be thought of as a [`Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html "struct tokio::sync::Semaphore") starting with 0 permits. The [`notified().await`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notified "method tokio::sync::Notify::notified") method waits for a permit to become available, and [`notify_one()`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_one "method tokio::sync::Notify::notify_one") sets a permit **if there currently are no available permits**.

...

> If `notify_one()` is called **before** `notified().await`, then the next call to `notified().await` will complete immediately, consuming the permit. Any subsequent calls to `notified().await` will wait for a new permit.

Keep in mind `Notify` acts just like a conditional variable, but not documented that way.

Here's a sample to show `Notify` doesn't need an active listener:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df51a40c100a7eb5fbc5ea6e0b28d80e
</blockquote></details>

You're right! I don't know what happened when I was reading the docs for it to check myself... I must have drifted off or something!
___
*[`cas/scheduler/scheduler.rs` line 504 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf6cXMCBP4VMEdCpXn:-NZjPuIt4aP6biheQ2Ek:b-owbya9) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L504)):*

I need to store either the `JoinHandle` or `AbortHandle`, so when the `Scheduler` is dropped it terminates the spawn/thread.
</blockquote></details>

In which case, I don't believe we need to store a `Weak` and upgrade it every time?  Can simply store the `Arc`, right?
___
*[`cas/scheduler/scheduler.rs` line 558 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf7Nud9JZtjBZdi-kn:-NZjQ1gG40mo8ilKNfWx:bxxx4gf) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L558)):*

I actually made a bunch of changes to do just that and after getting everything all done and nice and tidy I started working on fixing the tests. This was a nightmare because when things run in other threads, it is extremely difficult for the tests to manage order of operations.

I fixed all the order of operations, but it became really easy to introduce flakes. It also made tests much more difficult to debug, because we would now need to wait for state to equal something rather than reading state and expecting it to equal something. For example, right now you can do things like:
```rs
assert_eq!(client_rx.borrow_and_update().stage, ActionStage::Executing);
```

But it'd need to be changed to something like:
```rs
client_rx.wait_for(|v| v.stage == ActionStage::Executing).await
```

It may look about the same, but in terms of debugging it is horrible, because the `wait_for` will just never finish, where the first one will at least show you where the expectations where missed/wrong.
</blockquote></details>

That makes sense.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: all files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 354 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZjQBikExakchuE4DRK:-NZlV2sU2zqgg_4Sjpw3:b-uyte52) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L354)):*

I think this needs a `notify_one` otherwise it won't re-run the loop immediately as `immediate_evict_worker` won't re-queue.  In fact, we probably don't even want the 1ms wait in this case, right?
</blockquote></details>

I actually moved all the notify stuff inside the actions that trigger them. Before this was not done because triggering `do_try_match()` was expensive and it was not easy to figure out if somewhere else triggered it. Since now it runs in a different spawn, we can call `notify_one()` each time the action is triggered. In this case it's inside `immediate_evict_worker()`.
___
*[`cas/scheduler/scheduler.rs` line 504 at r1](https://reviewable.io/reviews/allada/turbo-cache/136#-NZf6cXMCBP4VMEdCpXn:-NZlVWZe58lop9HpKt4V:b-imh4i4) ([raw file](https://github.com/allada/turbo-cache/blob/668ee3aa629a38569b923c94a238882b2e180d64/cas/scheduler/scheduler.rs#L504)):*

In which case, I don't believe we need to store a `Weak` and upgrade it every time?  Can simply store the `Arc`, right?
</blockquote></details>

It is out of good habit to not use a strong reference unless we intend to keep it alive. You are right, but I would say it is just for good book keeping.

chrisstaite-menlo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136#-:-NZpU49K530XLQGtWZck:bnfp4nl)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/scheduler.rs` line 302 at r2](https://reviewable.io/reviews/allada/turbo-cache/136#-NZpU9xFEstEYW0UeiId:-NZpU9xFEstEYW0UeiIe:b3oeum) ([raw file](https://github.com/allada/turbo-cache/blob/3fd9cbffe545d26145eac37fba5c9bf6a09619d2/cas/scheduler/scheduler.rs#L302)):*
> ```XML
> 
>     /// Evicts the worker from the pool and puts items back into the queue if anything was being executed on it.
>     /// Note: This will not call tasks_or_workers_change_notify.notify_one().
> ```

This note is incorrect.
___
*[`cas/scheduler/scheduler.rs` line 597 at r2](https://reviewable.io/reviews/allada/turbo-cache/136#-NZpUKCVAyKz9nv5JOmg:-NZpUKCVAyKz9nv5JOmh:bjl4fwk) ([raw file](https://github.com/allada/turbo-cache/blob/3fd9cbffe545d26145eac37fba5c9bf6a09619d2/cas/scheduler/scheduler.rs#L597)):*
> ```XML
>             inner.immediate_evict_worker(&worker_id);
>         }
>         inner.tasks_or_workers_change_notify.notify_one();
> ```

Only need to do this in the else case, but I suppose it doesn't matter calling it twice.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/scheduler.rs` line 302 at r2](https://reviewable.io/reviews/allada/turbo-cache/136#-NZpU9xFEstEYW0UeiId:-NZqDbZx6Wwm5EqH8-h1:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/3fd9cbffe545d26145eac37fba5c9bf6a09619d2/cas/scheduler/scheduler.rs#L302)):*

This note is incorrect.
</blockquote></details>

Done.
___
*[`cas/scheduler/scheduler.rs` line 597 at r2](https://reviewable.io/reviews/allada/turbo-cache/136#-NZpUKCVAyKz9nv5JOmg:-NZqDdEE97covCwmQX0-:b-l0b401) ([raw file](https://github.com/allada/turbo-cache/blob/3fd9cbffe545d26145eac37fba5c9bf6a09619d2/cas/scheduler/scheduler.rs#L597)):*

Only need to do this in the else case, but I suppose it doesn't matter calling it twice.
</blockquote></details>

Yeah, I decided that it is so cheap to call, it's fine to call it twice.

This way has less chance it'll cause future bugs.

allada (commented):
> Dismissed @chrisstaite-menlo from a discussion.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

allada (commented):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/136)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #143, opened 2023-07-08
Update all dependencies to their latest versions

aaronmondal:
Update Rust to 1.70.0 and bring the Cargo.toml file in sync with the versions currently used in Cargo.Bazel.lock.

Explicitly depend on protobuf to make future transition to bzlmod simpler and to future-proof the repository for upcoming rust prost toolchains.

The fixed-buffer crate is excluded from these updates as changing it is a bit more involved and that code will likely be reworked as part of migration away from rusoto.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/143)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Use latest supported C++ version for protoc.
+build --cxxopt=-std=c++20 --host_cxxopt=-std=c++20 --repo_env=BAZEL_CXXOPTS=-std=c++20
+
 build --test_output=errors
 
 build:debug -c dbg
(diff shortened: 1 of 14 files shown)


aaronmondal:
Failure caused by outdated Ubuntu. Will update that in a separate commit first.

aaronmondal:
Rebased after #144 Let's see whether our C++ toolchain supports enough of C++ 20 now :laughing:

allada (approved):
Verified the lock files are generated from `Cargo.toml`

:lgtm:

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/143#-:-NZnZ_NsDmjQ4ZZvWGj7:b-b4y1mr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

___
*[`.bazelrc` line 35 at r1](https://reviewable.io/reviews/allada/turbo-cache/143#-NZnVRXnCTg9p8pvZSGr:-NZnVRXnCTg9p8pvZSGs:b-caddbf) ([raw file](https://github.com/allada/turbo-cache/blob/5ac2c5551eeb36f41df3e713a7751fdced5cdc09/.bazelrc#L35)):*
> ```
> # "indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since".
> # If cargo raze ever supports something that will make this error go away we can remove this.
> build --ui_event_filters=-DEBUG
> ```

aaah, thank you... this was such a pain and hated that it was done this way.

allada:
I think this PR broke `CC=clang`. I'll wait for `master` to fail, then revert if it does.

allada:
Ok maybe not... interesting... I purged my docker cache and things work in docker, but not on my host machine. False alarm.

PR #144, opened 2023-07-08
Use recent Ubuntu with Clang as build environment

aaronmondal:
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/144)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,13 +10,13 @@ on:
 jobs:
   unit-test:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v3.0.1
+    - uses: actions/checkout@v3.5.3
       with:
         fetch-depth: 0
     - name: Cache bazel
-      uses: actions/cache@v3.0.2
+      uses: actions/cache@v3.3.1
       env:
         cache-name: bazel-cache
       with:
@@ -27,9 +27,9 @@ jobs:
     - name: Bazel test
       run: bazel test //...
   integration-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v3.0.1
+    - uses: actions/checkout@v3.5.3
       with:
         fetch-depth: 0
 
--- a/deployment-examples/docker-compose/Dockerfile
+++ b/deployment-examples/docker-compose/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:20.04 AS builder
+FROM ubuntu:22.04 AS builder
 
 # Install bazel and needed deps.
 RUN apt update && \
@@ -21,20 +21,21 @@ RUN apt update && \
         git \
         pkg-config \
         libssl-dev \
-        gcc \
-        g++ \
+        clang \
         python3 && \
     npm install -g @bazel/bazelisk
 
 WORKDIR /root/turbo-cache
 ADD . .
 
+ARG CC=clang
+
 # Compile `cas` binary.
 RUN bazel build --compilation_mode=opt //cas && \
     cp ./bazel-bin/cas/cas /root/turbo-cache-bin
 
 # Go back to a fresh ubuntu container and copy only the compiled binary.
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 COPY --from=builder /root/turbo-cache-bin /usr/local/bin/turbo-cache
 
 # Install runtime packages.


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/144)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`.github/workflows/main.yml` line 28 at r1](https://reviewable.io/reviews/allada/turbo-cache/144#-NZnDl7B9Gtx83zijoM1:-NZnDl7B9Gtx83zijoM2:b-vykbo1) ([raw file](https://github.com/allada/turbo-cache/blob/93607006cce4978ec9b029f753bc0fc06e4137fa/.github/workflows/main.yml#L28)):*
> ```YAML
> 
>     - name: Bazel test
>       run: bazel test --config=generic_clang //...
> ```

Is this required now? I'd rather do this as a separate test if you want to test this config.

Also, I thought there was a ubuntu `clang` package replaces `gcc`?

aaronmondal:
*[`.github/workflows/main.yml` line 28 at r1](https://reviewable.io/reviews/allada/turbo-cache/144#-NZnDl7B9Gtx83zijoM1:-NZnFrLPEJuEbWfS1DvU:b-3yczln) ([raw file](https://github.com/allada/turbo-cache/blob/93607006cce4978ec9b029f753bc0fc06e4137fa/.github/workflows/main.yml#L28)):*

Is this required now? I'd rather do this as a separate test if you want to test this config.

Also, I thought there was a ubuntu `clang` package replaces `gcc`?
</blockquote></details>

The ubuntu clang package is what is used in the Dockerfile. But that doesn't set the CC environment variable which is required by Bazel's C++ toolchain autodetection. So instead this variable is conditionally set by the `generic_clang` config in `.bazelrc`. This way Clang isn't a hard requirement when building locally (if the CC override was always set we'd require every user to locally install clang).

I agree that it would be nicer to test both Clang and GCC as GCC builds would be untested after this change. We're only using it to build protoc, but it's probably still a good idea to test both.

Maybe I could split the unit-test workflow into `unit-tests-clang` and `unit-tests-gcc`?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/144)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`.github/workflows/main.yml` line 28 at r1](https://reviewable.io/reviews/allada/turbo-cache/144#-NZnDl7B9Gtx83zijoM1:-NZnJ7tV0oEkJhr4vOsL:b-xl0zx7) ([raw file](https://github.com/allada/turbo-cache/blob/93607006cce4978ec9b029f753bc0fc06e4137fa/.github/workflows/main.yml#L28)):*

The ubuntu clang package is what is used in the Dockerfile. But that doesn't set the CC environment variable which is required by Bazel's C++ toolchain autodetection. So instead this variable is conditionally set by the `generic_clang` config in `.bazelrc`. This way Clang isn't a hard requirement when building locally (if the CC override was always set we'd require every user to locally install clang).

I agree that it would be nicer to test both Clang and GCC as GCC builds would be untested after this change. We're only using it to build protoc, but it's probably still a good idea to test both.

Maybe I could split the unit-test workflow into `unit-tests-clang` and `unit-tests-gcc`?
</blockquote></details>

I'm more worried that we are not testing defaults. At least, can we use bazel's auto-detection and just globally set `CC` to clang in the Dockerfile, so users don't see this magic config and have to go digging to figure out what it does?

aaronmondal:
@allada WDYT about something like this:

```yaml
# main.yaml

matrix:
  compiler: [gcc, clang]

unit-test:
  ...
  env:
    CC: ${{ matrix.compiler }}
```

aaronmondal:
*[`.github/workflows/main.yml` line 28 at r1](https://reviewable.io/reviews/allada/turbo-cache/144#-NZnDl7B9Gtx83zijoM1:-NZnOgdk8lv1ZfV6_kLZ:bkgvi00) ([raw file](https://github.com/allada/turbo-cache/blob/93607006cce4978ec9b029f753bc0fc06e4137fa/.github/workflows/main.yml#L28)):*

I'm more worried that we are not testing defaults. At least, can we use bazel's auto-detection and just globally set `CC` to clang in the Dockerfile, so users don't see this magic config and have to go digging to figure out what it does?
</blockquote></details>

I've removed the `generic_clang` config and set it directly in the Dockerfile. This way we're testing just a default build configuration now and `bazel build ...` and `bazel test ...` should keep working as before with autodetection handled by the user's local toolchain config.

allada (approved):
Yeah that looks good :smile: 

:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/144#-:-NZnRT3-EgbuE9ct_ulC:bqv2lwm)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

PR #151, opened 2023-07-09
Remove dependency on `satackey/action-docker-layer-caching`

### LINKED ISSUE #145: Find an alternative to the docker layer caching action
`satackey/action-docker-layer-caching` has been outdated for a long time.

We're working on docker builds and new integration tests, so this will probably resolve itself.

allada:
closes #145

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+target/
+*/*/.terraform-turbo-cache-builder
+*/*/.update_scheduler_ips.zip
+*/*/terraform.tfstate
+*/*/.terraform
+.git
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,12 +33,17 @@ jobs:
       with:
         fetch-depth: 0
 
-    - uses: satackey/action-docker-layer-caching@v0.0.11
-      continue-on-error: true
+    - uses: docker/setup-buildx-action@v2
+    - uses: docker/build-push-action@v4
       with:
-        key: ${{ runner.os }}-{hash}-v1
-        restore-keys: |
-          docker-compose-cache-
+        context: .
+        file: ./deployment-examples/docker-compose/Dockerfile
+        load: true # This brings the build into `docker images` from buildx.
+        tags: allada/turbo-cache:latest
+        cache-from: |
+          type=gha,scope=$GITHUB_REF_NAME
+          type=gha,scope=master
+        cache-to: type=gha,scope=$GITHUB_REF_NAME,mode=max
 
     - name: Run tests
-      run: sudo ./run_integration_tests.sh
+      run: ./run_integration_tests.sh
--- a/deployment-examples/docker-compose/Dockerfile
+++ b/deployment-examples/docker-compose/Dockerfile
@@ -12,39 +12,98 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:22.04 AS builder
+# Override this if you want to run on a different version of ubuntu.
+ARG OS_VERSION=22.04
+# `--compilation_mode` to pass into bazel (eg: opt, dbg, fastbuild).
+ARG OPT_LEVEL=opt
+# Compiler to use.
+ARG CC=clang
 
-# Install bazel and needed deps.
-RUN apt update && \
-    DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends -y \
+
+# Builder that contains the OS dependencies.
+FROM ubuntu:${OS_VERSION} AS builder-deps
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
         npm \
         git \
         pkg-config \
         libssl-dev \
         clang \
         python3 && \
-    npm install -g @bazel/bazelisk
+    npm install -g @bazel/bazelisk@1.12.1
+SHELL [ "/bin/bash", "-euo", "pipefail", "-c" ]
 
+
+# Builder that only contains the downloaded externals.
+FROM builder-deps as builder-with-externals
 WORKDIR /root/turbo-cache
-ADD . .
+ARG CC
+# Only xfer the minimum needed to query external packages..
+COPY [ "WORKSPACE", ".bazel*", "Cargo.*", ".rustfmt.toml", "BUILD", "./" ]
+# Download external dependencies. This is to improve docker layer caching.
+RUN bazel fetch //...
 
-ARG CC=clang
 
-# Compile `cas` binary.
-RUN bazel build --compilation_mode=opt //cas && \
+# Builder that only contains the files related to bazel, but none of the source code.
+# This layer improves the cache hits of the following layer, since it'll bypass the next
+# layer if only the code files changed.
+FROM builder-with-externals as builder-only-bazel-and-build-files
+WORKDIR /root/turbo-cache
+COPY . .
+RUN find . -type f -regextype egrep ! -regex '.*/(BUILD|WORKSPACE|\.bazel.*|Cargo\..*|\.rustfmt\.toml)' -delete && \
+    find . -type d -empty -delete
+
+
+# Builder to get a list of externals and targets rarely changed that take a while to build
+# and push them into a file. This action takes a few seconds, but most of the time the
+# output file will not change between builds, which dramatically improves our layer cache
+# hits.
+FROM builder-with-externals AS builder-external-deps
+WORKDIR /root/turbo-cache
+ARG OPT_LEVEL
+ARG CC
+# Copy only the BUILD files and other bazel related files.
+COPY --from=builder-only-bazel-and-build-files /root/turbo-cache/ .
+# Special case for `//proto/... because we have a bit of generated code here that runs under
+# a different platform (host platform). Since this folder rarely changes we build it as well
+# at this cache layer.
+RUN bazel cquery -c ${OPT_LEVEL} \
+            'filter("^@", deps(attr(testonly, 0, //...), 1))' \
+            --universe_scope=//... \
+            --implicit_deps=false \
+            --tool_deps=false \
+        | cut -d' ' -f1 \
+        | sort \
+        | uniq > /root/external-targets.txt && \
+    echo "//proto/..." >> /root/external-targets.txt
+
+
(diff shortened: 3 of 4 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/151#-NZtP0D46o277I2KCn-2:-NZtP0D46o277I2KCn-3:b-ijskxk) (no related file):*
I spent some time to re-work the caching of the docker layers. It does complicate the docker layer, but should dramatically reduce CI build times assuming we get decent layer caching.

I'm also thinking about moving all tests to run under docker. This will make it much easier to test on other OS versions.

___

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/151#-NZtP0D46o277I2KCn-2:-NZtPKI-77oDTXiL3LUG:bb9iune) (no related file):*

I spent some time to re-work the caching of the docker layers. It does complicate the docker layer, but should dramatically reduce CI build times assuming we get decent layer caching.

I'm also thinking about moving all tests to run under docker. This will make it much easier to test on other OS versions.
</blockquote></details>

@aaronmondal , thoughts?

___

aaronmondal (changes requested):
I like the the distributed layer structure. I'm not a fan that we have to do it in such a rather complicated way, but I don't see another way to improve layer cache hit rates with Dockerfiles.

Just a few nits.  and this log from hadolint:

```
Dockerfile:18 DL3016 warning: Pin versions in npm. Instead of `npm install <package>` use `npm install <package>@<version>`
Dockerfile:18 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
Dockerfile:34 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:35 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:36 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:37 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:38 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:58 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
Dockerfile:75 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:82 DL3020 error: Use COPY instead of ADD for files and folders
Dockerfile:91 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
```

> Reviewed 3 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151#-:-NZvENwJCh4ZzY1y9Mzw:b-veqdjn)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`deployment-examples/docker-compose/Dockerfile` line 20 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvDmX_Bng3OLaYVtSs:-NZvDmX_Bng3OLaYVtSt:b-od408b) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L20)):*
> ```Dockerfile
> RUN apt update && \
>     DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends -y \
>         npm \
> ```

While we're at it, WDYT about changing the bazelisk install from npm to go, or fetching the binary straight from their git repo?
___
*[`deployment-examples/docker-compose/Dockerfile` line 38 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvBy-D9tDRY52poemz:-NZvBy-D9tDRY52poen-:b-7zrmlj) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L38)):*
> ```Dockerfile
> ADD Cargo.* .
> ADD .rustfmt.toml .
> ADD BUILD .
> ```

All these ADDs create separate layers. Since these files all depend fairly heavily on each other and are rather small, I think these can be merged into a single ADD.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @aaronmondal)

___
*[`deployment-examples/docker-compose/Dockerfile` line 20 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvDmX_Bng3OLaYVtSs:-N_-vyjF59qhDxqdoz_4:b-i2u0jd) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L20)):*

While we're at it, WDYT about changing the bazelisk install from npm to go, or fetching the binary straight from their git repo?
</blockquote></details>

I started to use `go install`, but abandoned it because it requires us to install go, setup the environ and then use it. `npm` is just install `npm` then install it.

The git repo is pretty much off limits because it would be super complicated to support arm & x86 and any other architecture easily.
___
*[`deployment-examples/docker-compose/Dockerfile` line 38 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvBy-D9tDRY52poemz:-N_-wMVN4Vy16os1Em8v:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L38)):*

All these ADDs create separate layers. Since these files all depend fairly heavily on each other and are rather small, I think these can be merged into a single ADD.
</blockquote></details>

Done.

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`deployment-examples/docker-compose/Dockerfile` line 20 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvDmX_Bng3OLaYVtSs:-N_0_GhnBF-_MiGf14bl:b-bwcpq5) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L20)):*

I started to use `go install`, but abandoned it because it requires us to install go, setup the environ and then use it. `npm` is just install `npm` then install it.

The git repo is pretty much off limits because it would be super complicated to support arm & x86 and any other architecture easily.
</blockquote></details>

Ah I see. I think just sticking to npm is fine then.

I don't think this affects build time too much since this will be reused from cache in 99% of cases, but for the record, after testing things like this a lot recently:

- The fastest way to install `npm` packages *by far* is actually via `bun`. They use some extremely voodoo black magic to install things that take 10 min via npm within 5 seconds. Only downside is that its still extremely experimental and doesn't always work "out of the box". But if it works, its incredibly fast. (~1500 deps, 4 second install time :scream:)
- For other cases, I've found `yarn` to basically always work, but it still installs packages ~3-10x faster than npm. Would generally be my go-to option but for true reproducibility this would require a yarn lockfile and that's probably underirable in this case.

aaronmondal (approved):
> Reviewed 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

___
*[`deployment-examples/docker-compose/Dockerfile` line 20 at r1](https://reviewable.io/reviews/allada/turbo-cache/151#-NZvDmX_Bng3OLaYVtSs:-N_1k5dQB9AYW2VI8MEK:b-tfkkz6) ([raw file](https://github.com/allada/turbo-cache/blob/4753c04ae85e0363981e41ab3a8cf4d17076e9e4/deployment-examples/docker-compose/Dockerfile#L20)):*

Ah I see. I think just sticking to npm is fine then.

I don't think this affects build time too much since this will be reused from cache in 99% of cases, but for the record, after testing things like this a lot recently:

- The fastest way to install `npm` packages *by far* is actually via `bun`. They use some extremely voodoo black magic to install things that take 10 min via npm within 5 seconds. Only downside is that its still extremely experimental and doesn't always work "out of the box". But if it works, its incredibly fast. (~1500 deps, 4 second install time :scream:)
- For other cases, I've found `yarn` to basically always work, but it still installs packages ~3-10x faster than npm. Would generally be my go-to option but for true reproducibility this would require a yarn lockfile and that's probably underirable in this case.
</blockquote></details>

Yeah, but the suggested way to install `yarn` is through `npm` :frowning:

allada (commented):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r3, 2 of 3 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @allada)

allada (commented):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @allada)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #157, opened 2023-07-09
Add a test for filestore loading from disk.

chrisstaite-menlo:
I had a concern that the loading of filetimes from the disk may have been broken while looking into the ordering of entries in the evicting map.  It turns out that this was unfounded as writing a test proves that the function works as expected.

I then broke the function on purpose by changing the sort order and found this new test was the only one that failed.  Therefore adding it to the test suite seems reasonable since it is now written.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/157)

### DIFF
--- a/cas/store/tests/filesystem_store_test.rs
+++ b/cas/store/tests/filesystem_store_test.rs
@@ -33,8 +33,7 @@ use futures::task::Poll;
 use futures::{poll, Future};
 use lazy_static::lazy_static;
 use rand::{thread_rng, Rng};
-use tokio::io::AsyncReadExt;
-use tokio::io::Take;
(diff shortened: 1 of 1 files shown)


allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/157)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/157#-:-NZxy7aT3KIJSNsFDkzj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #158, opened 2023-07-09
Implement Into traits indirectly via From

aaronmondal:
Also addresses most clippy warnings related to these files as long as they don't affect other files too much.

The impls that could panic should be implemented via TryFrom, but changing this would affect more files. For now, keep it as From and change those in the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/158)

PR #161, opened 2023-07-10
Implementation of a cache lookup scheduler

chrisstaite-menlo:
This scheduler sits in front of another one and looks up actions in
the cache if requested before queueing them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)

### DIFF
--- a/cas/cas_main.rs
+++ b/cas/cas_main.rs
@@ -105,7 +105,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut worker_schedulers = HashMap::new();
     if let Some(schedulers_cfg) = cfg.schedulers {
         for (name, scheduler_cfg) in schedulers_cfg {
-            let (maybe_action_scheduler, maybe_worker_scheduler) = scheduler_factory(&scheduler_cfg)
+            let (maybe_action_scheduler, maybe_worker_scheduler) = scheduler_factory(&scheduler_cfg, &store_manager)
                 .await
                 .err_tip(|| format!("Failed to create scheduler '{}'", name))?;
             if let Some(action_scheduler) = maybe_action_scheduler {
--- a/cas/grpc_service/execution_server.rs
+++ b/cas/grpc_service/execution_server.rs
@@ -188,9 +188,12 @@ impl ExecutionServer {
             .build_action_info(instance_name, digest, &action, priority, execute_req.skip_cache_lookup)
             .await?;
 
+        // Warning: This name is ignored by GrpcScheduler, so don't use it for
+        //          operations mapping here...
+        let name = format!("{:X}", thread_rng().gen::<u128>());
         let rx = instance_info
             .scheduler
-            .add_action(action_info)
+            .add_action(name, action_info)
             .await
             .err_tip(|| "Failed to schedule task")?;
 
--- a/cas/grpc_service/tests/worker_api_server_test.rs
+++ b/cas/grpc_service/tests/worker_api_server_test.rs
@@ -291,7 +291,10 @@ pub mod execution_response_tests {
             },
             skip_cache_lookup: true,
         };
-        let mut client_action_state_receiver = test_context.scheduler.add_action(action_info).await?;
+        let mut client_action_state_receiver = test_context
+            .scheduler
+            .add_action("name".to_string(), action_info)
+            .await?;
 
         let mut server_logs = HashMap::new();
         server_logs.insert(
--- a/cas/scheduler/BUILD
+++ b/cas/scheduler/BUILD
@@ -32,6 +32,40 @@ rust_library(
 )
 
 rust_library(
+    name = "mock_scheduler",
+    testonly = True,
+    srcs = ["tests/utils/mock_scheduler.rs"],
+    visibility = [
+        "//cas:__pkg__",
+        "//cas:__subpackages__",
+    ],
+    proc_macro_deps = ["@crate_index//:async-trait"],
+    deps = [
+        ":action_messages",
+        ":platform_property_manager",
+        ":scheduler",
+        "//util:error",
+        "@crate_index//:tokio",
+    ],
+)
+
+rust_library(
+    name = "scheduler_utils",
+    testonly = True,
+    srcs = ["tests/utils/scheduler_utils.rs"],
+    visibility = [
+        "//cas:__pkg__",
+        "//cas:__subpackages__",
+    ],
+    proc_macro_deps = ["@crate_index//:async-trait"],
+    deps = [
+        ":action_messages",
+        ":platform_property_manager",
+        "//util:common",
+    ],
+)
+
+rust_library(
     name = "simple_scheduler",
     srcs = ["simple_scheduler.rs"],
     visibility = [
@@ -50,8 +84,57 @@ rust_library(
         "@crate_index//:parking_lot",
         "@crate_index//:futures",
         "@crate_index//:lru",
-        "@crate_index//:rand",
         "@crate_index//:tokio",
+    ],
+)
+
+rust_library(
+    name = "cache_lookup_scheduler",
+    srcs = ["cache_lookup_scheduler.rs"],
+    visibility = [
+        "//cas:__pkg__",
+        "//cas:__subpackages__",
+    ],
+    proc_macro_deps = ["@crate_index//:async-trait"],
+    deps = [
+        ":action_messages",
+        ":platform_property_manager",
+        ":scheduler",
+        ":worker",
+        "//cas/store",
+        "//cas/store:ac_utils",
+        "//cas/store:grpc_store",
+        "//config",
+        "//proto",
+        "//util:common",
+        "//util:error",
+        "@crate_index//:futures",
+        "@crate_index//:tonic",
+        "@crate_index//:tokio",
+        "@crate_index//:tokio-stream",
+    ],
+)
+
+rust_test(
+    name = "cache_lookup_scheduler_test",
+    srcs = ["tests/cache_lookup_scheduler_test.rs"],
+    deps = [
+        ":action_messages",
+        ":cache_lookup_scheduler",
+        ":mock_scheduler",
+        ":platform_property_manager",
+        ":scheduler",
+        ":scheduler_utils",
+        "//cas/store",
+        "//cas/store:memory_store",
+        "//config",
+        "//proto",
+        "//util:common",
+        "//util:error",
+        "@crate_index//:pretty_assertions",
+        "@crate_index//:prost",
+        "@crate_index//:tokio",
+        "@crate_index//:tokio-stream",
     ],
 )
 
@@ -85,9 +168,11 @@ rust_library(
         "//cas:__subpackages__",
     ],
     deps = [
+        ":cache_lookup_scheduler",
         ":grpc_scheduler",
         ":scheduler",
         ":simple_scheduler",
+        "//cas/store",
         "//config",
         "//util:error",
         "@crate_index//:futures",
@@ -155,6 +240,7 @@ rust_test(
         ":action_messages",
         ":platform_property_manager",
         ":scheduler",
+        ":scheduler_utils",
         ":simple_scheduler",
         ":worker",
         "//config",
--- a/cas/scheduler/cache_lookup_scheduler.rs
+++ b/cas/scheduler/cache_lookup_scheduler.rs
@@ -1,0 +1,183 @@
+// Copyright 2023 The Turbo Cache Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::pin::Pin;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use futures::stream::{FuturesUnordered, StreamExt};
+use tokio::sync::watch;
+use tokio_stream::wrappers::WatchStream;
+use tonic::Request;
+
+use ac_utils::get_and_decode_digest;
+use action_messages::{ActionInfo, ActionResult, ActionStage, ActionState};
+use common::DigestInfo;
+use error::Error;
+use grpc_store::GrpcStore;
+use platform_property_manager::PlatformPropertyManager;
+use proto::build::bazel::remote::execution::v2::{
+    ActionResult as ProtoActionResult, FindMissingBlobsRequest, GetActionResultRequest,
+};
+use scheduler::ActionScheduler;
+use store::Store;
+
+pub struct CacheLookupScheduler {
+    /// A reference to the CAS which is used to validate all the outputs of a
+    /// cached ActionResult still exist.
+    cas_store: Arc<dyn Store>,
+    /// A reference to the AC to find existing actions in.
+    ac_store: Arc<dyn Store>,
+    /// The "real" scheduler to use to perform actions if they were not found
+    /// in the action cache.
+    action_scheduler: Arc<dyn ActionScheduler>,
+}
+
+async fn get_action_from_store(
+    ac_store: Arc<dyn Store>,
+    action_digest: &DigestInfo,
+    instance_name: String,
+) -> Option<ProtoActionResult> {
+    // If we are a GrpcStore we shortcut here, as this is a special store.
+    let any_store = ac_store.clone().as_any();
+    let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
+    if let Some(grpc_store) = maybe_grpc_store {
+        let action_result_request = GetActionResultRequest {
+            instance_name,
+            action_digest: Some(action_digest.into()),
+            inline_stdout: false,
+            inline_stderr: false,
+            inline_output_files: Vec::new(),
+        };
+        grpc_store
+            .get_action_result(Request::new(action_result_request))
+            .await
+            .map(|response| response.into_inner())
+            .ok()
+    } else {
+        get_and_decode_digest::<ProtoActionResult>(Pin::new(ac_store.as_ref()), action_digest)
+            .await
+            .ok()
+    }
+}
+
+async fn validate_outputs_exist(
+    cas_store: Arc<dyn Store>,
+    action_result: &ProtoActionResult,
+    instance_name: String,
+) -> bool {
+    // Verify that output_files and output_directories are available in the cas.
+    let required_digests = action_result
+        .output_files
+        .iter()
+        .filter_map(|output_file| output_file.digest.clone())
+        .chain(
+            action_result
+                .output_directories
+                .iter()
+                .filter_map(|output_directory| output_directory.tree_digest.clone()),
+        )
+        .collect();
+
+    // If the CAS is a GrpcStore store we can check all the digests in one message.
+    let any_store = cas_store.clone().as_any();
+    let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
+    if let Some(grpc_store) = maybe_grpc_store {
+        grpc_store
+            .find_missing_blobs(Request::new(FindMissingBlobsRequest {
+                instance_name,
+                blob_digests: required_digests,
+            }))
+            .await
+            .is_ok_and(|find_result| find_result.into_inner().missing_blob_digests.is_empty())
+    } else {
+        let cas_pin = Pin::new(cas_store.as_ref());
+        required_digests
+            .into_iter()
+            .map(|digest| async move { cas_pin.has(DigestInfo::try_from(digest)?).await })
+            .collect::<FuturesUnordered<_>>()
+            .all(|result| async { result.is_ok_and(|result| result.is_some()) })
+            .await
+    }
+}
+
+impl CacheLookupScheduler {
+    pub fn new(
+        cas_store: Arc<dyn Store>,
+        ac_store: Arc<dyn Store>,
+        action_scheduler: Arc<dyn ActionScheduler>,
+    ) -> Result<Self, Error> {
+        Ok(Self {
+            cas_store,
+            ac_store,
+            action_scheduler,
+        })
+    }
+}
+
+#[async_trait]
+impl ActionScheduler for CacheLookupScheduler {
+    async fn get_platform_property_manager(&self, instance_name: &str) -> Result<Arc<PlatformPropertyManager>, Error> {
+        self.action_scheduler.get_platform_property_manager(instance_name).await
+    }
+
+    async fn add_action(
+        &self,
+        name: String,
+        action_info: ActionInfo,
+    ) -> Result<watch::Receiver<Arc<ActionState>>, Error> {
+        if action_info.skip_cache_lookup {
+            // Cache lookup skipped, forward to the upstream.
+            return self.action_scheduler.add_action(name, action_info).await;
+        }
+        let action_digest = action_info.digest();
+        let mut current_state = Arc::new(ActionState {
+            name: name.clone(),
+            stage: ActionStage::CacheCheck,
+            action_digest: *action_digest,
+        });
+        let (tx, rx) = watch::channel(current_state.clone());
+        let ac_store = self.ac_store.clone();
+        let cas_store = self.cas_store.clone();
+        let action_scheduler = self.action_scheduler.clone();
+        tokio::spawn(async move {
+            let instance_name = action_info.instance_name.to_string();
+            if let Some(proto_action_result) =
+                get_action_from_store(ac_store, &current_state.action_digest, instance_name.clone()).await
+            {
+                if validate_outputs_exist(cas_store, &proto_action_result, instance_name).await {
+                    // Found in the cache, return the result immediately.
+                    Arc::make_mut(&mut current_state).stage = ActionStage::CompletedFromCache(proto_action_result);
+                    let _ = tx.send(current_state);
+                    return;
+                }
+            }
+            // Not in cache, forward to upstream and proxy state.
+            let mut watch_stream = match action_scheduler.add_action(name, action_info).await {
+                Ok(rx) => WatchStream::new(rx),
+                Err(err) => {
+                    Arc::make_mut(&mut current_state).stage = ActionStage::Error((err, ActionResult::default()));
+                    let _ = tx.send(current_state);
+                    return;
+                }
+            };
+            while let Some(action_state) = watch_stream.next().await {
+                if tx.send(action_state).is_err() {
+                    break;
+                }
+            }
+        });
+        Ok(rx)
+    }
+}
--- a/cas/scheduler/default_scheduler_factory.rs
+++ b/cas/scheduler/default_scheduler_factory.rs
@@ -17,16 +17,19 @@ use std::sync::Arc;
 
 use futures::Future;
 
+use cache_lookup_scheduler::CacheLookupScheduler;
 use config::schedulers::SchedulerConfig;
-use error::Error;
+use error::{Error, ResultExt};
 use grpc_scheduler::GrpcScheduler;
 use scheduler::{ActionScheduler, WorkerScheduler};
 use simple_scheduler::SimpleScheduler;
+use store::StoreManager;
 
 pub type SchedulerFactoryResults = (Option<Arc<dyn ActionScheduler>>, Option<Arc<dyn WorkerScheduler>>);
 
 pub fn scheduler_factory<'a>(
     scheduler_type_cfg: &'a SchedulerConfig,
+    store_manager: &'a StoreManager,
 ) -> Pin<Box<dyn Future<Output = Result<SchedulerFactoryResults, Error>> + 'a>> {
     Box::pin(async move {
         let scheduler: SchedulerFactoryResults = match scheduler_type_cfg {
@@ -34,7 +37,22 @@ pub fn scheduler_factory<'a>(
                 let scheduler = Arc::new(SimpleScheduler::new(&config));
                 (Some(scheduler.clone()), Some(scheduler))
             }
-            SchedulerConfig::grpc(config) => (Some(Arc::new(GrpcScheduler::new(&config).await?)), None),
+            SchedulerConfig::grpc(config) => (Some(Arc::new(GrpcScheduler::new(config).await?)), None),
+            SchedulerConfig::cache_lookup(config) => {
+                let cas_store = store_manager
+                    .get_store(&config.cas_store)
+                    .err_tip(|| format!("'cas_store': '{}' does not exist", config.cas_store))?;
+                let ac_store = store_manager
+                    .get_store(&config.ac_store)
+                    .err_tip(|| format!("'ac_store': '{}' does not exist", config.ac_store))?;
+                let (action_scheduler, worker_scheduler) = scheduler_factory(&config.scheduler, store_manager).await?;
+                let cache_lookup_scheduler = Arc::new(CacheLookupScheduler::new(
+                    cas_store,
+                    ac_store,
+                    action_scheduler.err_tip(|| "Nested scheduler is not an action scheduler")?,
+                )?);
+                (Some(cache_lookup_scheduler), worker_scheduler)
+            }
         };
         Ok(scheduler)
     })
--- a/cas/scheduler/grpc_scheduler.rs
+++ b/cas/scheduler/grpc_scheduler.rs
@@ -82,7 +82,11 @@ impl ActionScheduler for GrpcScheduler {
         Ok(platform_property_manager)
     }
 
-    async fn add_action(&self, action_info: ActionInfo) -> Result<watch::Receiver<Arc<ActionState>>, Error> {
+    async fn add_action(
+        &self,
+        _name: String,
+        action_info: ActionInfo,
+    ) -> Result<watch::Receiver<Arc<ActionState>>, Error> {
         let execution_policy = if action_info.priority == DEFAULT_EXECUTION_PRIORITY {
             None
         } else {
--- a/cas/scheduler/scheduler.rs
+++ b/cas/scheduler/scheduler.rs
@@ -30,7 +30,11 @@ pub trait ActionScheduler: Sync + Send + Unpin {
     async fn get_platform_property_manager(&self, instance_name: &str) -> Result<Arc<PlatformPropertyManager>, Error>;
 
     /// Adds an action to the scheduler for remote execution.
-    async fn add_action(&self, action_info: ActionInfo) -> Result<watch::Receiver<Arc<ActionState>>, Error>;
+    async fn add_action(
+        &self,
+        name: String,
+        action_info: ActionInfo,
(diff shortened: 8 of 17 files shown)


chrisstaite-menlo:
I'm not sure if the cache lookup should validate the results are in the CAS, it's probably a good optimisation as we can probably talk to it a lot faster than the client.

aaronmondal (commented):
This might benefit from #158. I'm not sure how we can make sure that we don't regress clippy warnings when we can't enable it globally just yet. So far I've used something like this:

```python
load("@rules_rust//rust:defs.bzl", ..., "rust_clippy")

rust_clippy(
    name = "clippy",
    deps = [
        ":grpc_lookup_scheduler",
        ":cache_lookup_scheduler",
    ]
)
```
```rust
// grpc_cheduler.rs

#![warn(clippy::all, clippy::pedantic, clippy::nursery)]
```

But not all those warnings should be followed. I think we'd want `clippy::all` and `clippy::pedantic` (apart from some doc-related things that make more sense to follow when things are more stable), but while `clippy::nursery` has some really nice lints, e.g. the `use_self` lint, some of the nursery lints aren't meant to always be followed.

Maybe just run it to some degree and call it a day when "most" things look fine?

Don't let this block you though. I'm totally fine with fixing things later.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161#-:-N_0jyzA2Ozo7VV6csaA:b-fvyf3w)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/scheduler/action_messages.rs` line 499 at r1](https://reviewable.io/reviews/allada/turbo-cache/161#-N_0iXYMA1xi0yX63Pqh:-N_0iXYMA1xi0yX63Pqi:b5pa6ys) ([raw file](https://github.com/allada/turbo-cache/blob/0d0ed4d80939c641a317d6fad786c11dfc682d94/cas/scheduler/action_messages.rs#L499)):*
> ```XML
> 
> /// Exit code sent if there is an internal error.
> pub const INTERNAL_ERROR_EXIT_CODE: i32 = -178;
> ```

I wonder whether there is a way to avoid this magic number, maybe using an enum or adding it to some existing error codes?
___
*[`cas/scheduler/action_messages.rs` line 716 at r1](https://reviewable.io/reviews/allada/turbo-cache/161#-N_0gdPTAMU_V28lki-N:-N_0gdPTAMU_V28lki-O:b8gba1) ([raw file](https://github.com/allada/turbo-cache/blob/0d0ed4d80939c641a317d6fad786c11dfc682d94/cas/scheduler/action_messages.rs#L716)):*
> ```XML
> }
> 
> impl TryFrom<Operation> for ActionState {
> ```

Clippy-sensei will complain about https://rust-lang.github.io/rust-clippy/master/index.html#/use_self here, i.e. occurrences of `ActionState` within this block should be renamed to `Self` :laughing:

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/action_messages.rs` line 499 at r1](https://reviewable.io/reviews/allada/turbo-cache/161#-N_0iXYMA1xi0yX63Pqh:-N_1hIh0CyHd3G1MGTNn:basp90c) ([raw file](https://github.com/allada/turbo-cache/blob/0d0ed4d80939c641a317d6fad786c11dfc682d94/cas/scheduler/action_messages.rs#L499)):*

I wonder whether there is a way to avoid this magic number, maybe using an enum or adding it to some existing error codes?
</blockquote></details>

I added this because I had a deployment where we would have a wrapper shell script all jobs would go through and in that script in the event that we ran out of disk space or if the job failed and our node health checker said the node was bad (like GPU kernel crash), the node would tell the scheduler the node was bad with this magic number. I think this got removed in a recent change though, but I believe it should be added back in to allow nodes to tell the scheduler to disconnect it. But then again, we would need a way for the scheduler to ensure the node doesn't just re-connect to the scheduler :shrug:

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @aaronmondal and @allada)

___
*[`cas/scheduler/action_messages.rs` line 499 at r1](https://reviewable.io/reviews/allada/turbo-cache/161#-N_0iXYMA1xi0yX63Pqh:-N_3qSqe4vyGWd87DeCF:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0d0ed4d80939c641a317d6fad786c11dfc682d94/cas/scheduler/action_messages.rs#L499)):*

I added this because I had a deployment where we would have a wrapper shell script all jobs would go through and in that script in the event that we ran out of disk space or if the job failed and our node health checker said the node was bad (like GPU kernel crash), the node would tell the scheduler the node was bad with this magic number. I think this got removed in a recent change though, but I believe it should be added back in to allow nodes to tell the scheduler to disconnect it. But then again, we would need a way for the scheduler to ensure the node doesn't just re-connect to the scheduler :shrug: 
</blockquote></details>

Done.
___
*[`cas/scheduler/action_messages.rs` line 716 at r1](https://reviewable.io/reviews/allada/turbo-cache/161#-N_0gdPTAMU_V28lki-N:-N_4x0Za5260n-TZDAL6:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0d0ed4d80939c641a317d6fad786c11dfc682d94/cas/scheduler/action_messages.rs#L716)):*

Clippy-sensei will complain about https://rust-lang.github.io/rust-clippy/master/index.html#/use_self here, i.e. occurrences of `ActionState` within this block should be renamed to `Self` :laughing: 
</blockquote></details>

Done.

allada (changes requested):
> Reviewed 7 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 7 of 15 files reviewed, 3 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 60 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5-IXF1Y1dEFtwkCKq:-N_5-IXF1Y1dEFtwkCKr:bc4hzhl) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L60)):*
> ```XML
>             .get_store(&config.ac_store)
>             .ok_or_else(|| make_input_err!("'ac_store': '{}' does not exist", config.ac_store))?;
>         let scheduler = scheduler_manager
> ```

We should not guarantee order schedulers are created here. Instead we should do what `RefStore` does by looking up the store on first invocation. In all honesty, I would not use support `RefScheduler` at all and instead have the scheduler be nested (`VerifyStore` is a good config example of this).

In the future we can add a custom `RefScheduler` if we want to reference the scheduler by name instead of have it owned by another scheduler.

chrisstaite-menlo:
*[`cas/scheduler/cache_lookup_scheduler.rs` line 60 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5-IXF1Y1dEFtwkCKq:-N_5DjF38Y_QMDFu1yXq:b-ou5386) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L60)):*

We should not guarantee order schedulers are created here. Instead we should do what `RefStore` does by looking up the store on first invocation. In all honesty, I would not use support `RefScheduler` at all and instead have the scheduler be nested (`VerifyStore` is a good config example of this).

In the future we can add a custom `RefScheduler` if we want to reference the scheduler by name instead of have it owned by another scheduler.
</blockquote></details>

That raises the question of whether `CacheLookupScheduler` should implement `WorkerScheduler` or not... it should only implement it if the nested one does.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 7 of 15 files reviewed, 5 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 1 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5s4gL3ElUHenxVtYX:-N_5s4gL3ElUHenxVtYY:b-nmx4dh) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L1)):*
> ```XML
> // Copyright 2023 The Turbo Cache Authors. All rights reserved.
> ```

You know, I spent some time going through this file and to make sure the code snippets I was sending you would work, but I realized I kept changing stuff and by the end it was easier to just send the file with all the changes at once...

It might work better to have you modify this file to your desire then I'll re-review it with any additional comments.

Most of it had to do with trying to get the future stream working properly.

https://gist.github.com/allada/583a3c7781196d509c2fb29b94175949
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 54 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_513SfC2coifp910jc:-N_513SfC2coifp910jd:b-y9ethd) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L54)):*
> ```XML
>         scheduler_manager: &HashMap<String, Arc<dyn ActionScheduler>>,
>     ) -> Result<Self, Error> {
>         let cas_store = store_manager
> ```

lets pass these stores in instead of querying them here. It'll make testing a bit more straight forward.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 7 of 15 files reviewed, 5 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 60 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5-IXF1Y1dEFtwkCKq:-N_5sjb5-MWuYO9TH2TQ:b-vnri9s) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L60)):*

That raises the question of whether `CacheLookupScheduler` should implement `WorkerScheduler` or not... it should only implement it if the nested one does.
</blockquote></details>

Yeah, I think that works. Downcast it to on construction and forward if it implements it.

I'd try and avoid `RefScheduler` as long as possible.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 7 of 15 files reviewed, 5 unresolved discussions (waiting on @aaronmondal and @allada)

___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 1 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5s4gL3ElUHenxVtYX:-N_86rrUE5aoqiVM0jFt:bornntr) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L1)):*

You know, I spent some time going through this file and to make sure the code snippets I was sending you would work, but I realized I kept changing stuff and by the end it was easier to just send the file with all the changes at once...

It might work better to have you modify this file to your desire then I'll re-review it with any additional comments.

Most of it had to do with trying to get the future stream working properly.

https://gist.github.com/allada/583a3c7781196d509c2fb29b94175949
</blockquote></details>

Yeah, the streaming API is not at all user friendly I found.  I did a number of iterations to get something to work.

Do you think we need to limit the number of parallel validations with the store, that's really the problem of `fs` to handle, right?  If it's all in-memory then we can validate it immediately without any issues.  Also, I'm not sure that we need to count all of the entries, we can assume that the `Stream` won't throw away results, so we just have to check that all results are valid, which is what I had in my implementation.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 54 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_513SfC2coifp910jc:-N_89PNt8CAIZqCzPLC_:b5uavav) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L54)):*

lets pass these stores in instead of querying them here. It'll make testing a bit more straight forward.
</blockquote></details>

I kept going back and forth with this one, I couldn't decide the best way to do it.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 60 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5-IXF1Y1dEFtwkCKq:-N_8Dkog2lQYs5vAUP_r:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L60)):*

Yeah, I think that works. Downcast it to on construction and forward if it implements it.

I'd try and avoid `RefScheduler` as long as possible.
</blockquote></details>

Done.

allada (changes requested):
> Reviewed 3 of 13 files at r3, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: all files reviewed, 12 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 1 at r3](https://reviewable.io/reviews/allada/turbo-cache/161#-N_5s4gL3ElUHenxVtYX:-N_9VbBm3yrPJcO8dGA-:brojc4) ([raw file](https://github.com/allada/turbo-cache/blob/450f9db2272e7000eebd904b1bba11a16934d3da/cas/scheduler/cache_lookup_scheduler.rs#L1)):*

Yeah, the streaming API is not at all user friendly I found.  I did a number of iterations to get something to work.

Do you think we need to limit the number of parallel validations with the store, that's really the problem of `fs` to handle, right?  If it's all in-memory then we can validate it immediately without any issues.  Also, I'm not sure that we need to count all of the entries, we can assume that the `Stream` won't throw away results, so we just have to check that all results are valid, which is what I had in my implementation.


</blockquote></details>

I'm worried about things like `GrpcStore`, `S3Store`, or any other store that uses a lot of resources internally...

However, I think the places that matter already have internal rate limiting in the store. `S3Store` does. `GrpcStore` does not, but it's a single HTTP2 connection usually.

So yeah, I guess we don't need to rate limit it.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 173 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9Snnq3musuC9U85ee:-N_9Snnq3musuC9U85ef:b5w3tsy) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L173)):*
> ```XML
>         self.worker_scheduler
>             .as_ref()
>             .ok_or(make_input_err!("Adding worker to action only scheduler"))?
> ```

nit: try:
```rs
.err_tip(|| "Adding worker to action only scheduler")?
```
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 184 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9TpsL0QJpIswdIog9:-N_9T1G42lTtp_-mPgr6:b-51tnqe) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L184)):*
> ```XML
>         err: Error,
>     ) {
>         if let Some(worker_scheduler) = self.worker_scheduler.as_ref() {
> ```

nit: Suggest:



_Code snippet:_
```XML
let Some(worker_scheduler) = self.worker_scheduler.as_ref() else {
    log::error!("Inner worker scheduler is None for update_action_with_internal_error");
    return;
};
worker_scheduler
    .update_action_with_internal_error(&worker_id, &action_info_hash_key, err)
    .await;
```
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 199 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9TxVbD8AEy_hZBC4o:-N_9TxVbD8AEy_hZBC4p:b-r3fegs) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L199)):*
> ```XML
>         self.worker_scheduler
>             .as_ref()
>             .ok_or(make_input_err!("Update action for action only scheduler"))?
> ```

nit: try:
```rs
.err_tip(|| "Update action for action only scheduler")?
```
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 207 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9Tzh31-SmA_s2lpYw:-N_9Tzh31-SmA_s2lpYx:b1wyjjx) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L207)):*
> ```XML
>         self.worker_scheduler
>             .as_ref()
>             .ok_or(make_input_err!("Worker keep alive for action only scheduler"))?
> ```

nit: try:
```rs
.err_tip(|| "Worker keep alive for action only scheduler")?
```
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 213 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9U3DwB9us3zuK1Q-M:-N_9U3DwB9us3zuK1Q-N:b74vacn) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L213)):*
> ```XML
> 
>     async fn remove_worker(&self, worker_id: WorkerId) {
>         if let Some(worker_scheduler) = self.worker_scheduler.as_ref() {
> ```

nit:
```rs
let Some(worker_scheduler) = self.worker_scheduler.as_ref() else {
    log::error!("Inner worker scheduler is None for remove_worker");
    return;
};
worker_scheduler.remove_worker(&worker_id).await;
```
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 221 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9UHgbEHICI4XKSXMr:-N_9UHgbEHICI4XKSXMs:broyajc) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/cache_lookup_scheduler.rs#L221)):*
> ```XML
>         self.worker_scheduler
>             .as_ref()
>             .ok_or(make_input_err!("Worker timeout for action only scheduler"))?
> ```

nit: try:
```rs
.err_tip(|| "Worker timeout for action only scheduler")?
```
___
*[`cas/scheduler/default_scheduler_factory.rs` line 44 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9UPIU9C10R8v9OzIt:-N_9UPIU9C10R8v9OzIu:bufbk00) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L44)):*
> ```XML
>                 let cas_store = store_manager
>                     .get_store(&config.cas_store)
>                     .ok_or_else(|| make_input_err!("'cas_store': '{}' does not exist", config.cas_store))?;
> ```

nit:

```rs
.err_tip(|| format!("'cas_store': '{config.cas_store}' does not exist")))?;
```
___
*[`cas/scheduler/default_scheduler_factory.rs` line 47 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9U_3f7sup3w-9ZJiA:-N_9U_3f7sup3w-9ZJiB:b-5v3gaq) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L47)):*
> ```XML
>                 let ac_store = store_manager
>                     .get_store(&config.ac_store)
>                     .ok_or_else(|| make_input_err!("'ac_store': '{}' does not exist", config.ac_store))?;
> ```

nit: ditto.
___
*[`cas/scheduler/default_scheduler_factory.rs` line 53 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9UbBLBQ-qrZN75Gr5:-N_9UbBLBQ-qrZN75Gr6:b-5v3gaq) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L53)):*
> ```XML
>                     cas_store,
>                     ac_store,
>                     action_scheduler.ok_or(make_input_err!("Nested scheduler is not an action scheduler"))?,
> ```

nit: ditto.
___
*[`cas/scheduler/default_scheduler_factory.rs` line 57 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9V3ktBEzES5oKyN_z:-N_9V3ktBEzES5oKyNa-:b-886r6x) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L57)):*
> ```XML
>                 )?);
>                 if has_worker_instance {
>                     (Some(cache_lookup_scheduler.clone()), Some(cache_lookup_scheduler))
> ```

Oh interesting, Can we just pass `worker_scheduler` here and not even implement the `WorkerScheduler` for `CacheLookupScheduler`?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: all files reviewed, 13 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9WdpEEO24XYq17fF-:-N_9WdpEEO24XYq17fF0:b-23xnim) (no related file):*
Can we also write a test just to make sure the basic functionality is working?

___

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 11 of 17 files reviewed, 5 unresolved discussions (waiting on @aaronmondal and @allada)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9WdpEEO24XYq17fF-:-N_DWGmgC9ylU_rVTpPy:b-896fix) (no related file):*

Can we also write a test just to make sure the basic functionality is working?
</blockquote></details>

Done.

___
___
*[`cas/scheduler/default_scheduler_factory.rs` line 44 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9UPIU9C10R8v9OzIt:-N_DDkKY6Os9XQ74n8-7:bfwslkj) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L44)):*

nit:

```rs
.err_tip(|| format!("'cas_store': '{config.cas_store}' does not exist")))?;
```
</blockquote></details>

`error: invalid format string: field access isn't supported`
___
*[`cas/scheduler/default_scheduler_factory.rs` line 57 at r5](https://reviewable.io/reviews/allada/turbo-cache/161#-N_9V3ktBEzES5oKyN_z:-N_DDKUc2u5a-GuutSx9:b-jyvdw5) ([raw file](https://github.com/allada/turbo-cache/blob/1ae1d88ea980b1821dbb6b84a2078c002bc5bcbf/cas/scheduler/default_scheduler_factory.rs#L57)):*

Oh interesting, Can we just pass `worker_scheduler` here and not even implement the `WorkerScheduler` for `CacheLookupScheduler`?
</blockquote></details>

YES!!! :D

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: 6 of 21 files reviewed, 3 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 193 at r7](https://reviewable.io/reviews/allada/turbo-cache/161#-N_F-_cH3SKfEr9s8O1Y:-N_F-_cH3SKfEr9s8O1Z:b-18bbqp) ([raw file](https://github.com/allada/turbo-cache/blob/88f6c7d55594e7d045d8f4b25ded4148b02bae7e/cas/grpc_service/execution_server.rs#L193)):*
> ```XML
>         // Warning: This name is ignored by GrpcScheduler, so don't use it for
>         //          operations mapping here...
>         let name = format!("{:X}", thread_rng().gen::<u128>());
> ```

I'm not sure I agree with the code calling into the scheduler's `add_action()` should own this responsibility.

At first I thought... "oh that makes sense", but then when I started thinking about it, the scheduler technically is responsible for preventing collisions (which we don't do, but technically should not happen). The `name` is only used internally between the scheduler and worker to ensure when the worker talks to the scheduler it knows what task the responses are for and to inform the caller (through `ActionState`) what name was assigned to the job for future lookups.

chrisstaite-menlo:
Now we've got most of the changes into the main branch I managed to test this out this evening on our cluster.  I had a few little hiccups, but generally it's gone well.  It's using way less CPU than it used to on the storage server and everything appears to be incredibly responsive.

allada (changes requested):
> Reviewed 11 of 15 files at r7, 3 of 4 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: all files reviewed, 9 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 193 at r7](https://reviewable.io/reviews/allada/turbo-cache/161#-N_F-_cH3SKfEr9s8O1Y:-N_GhWi80eIbOVzVJ7LX:b-a4m6gp) ([raw file](https://github.com/allada/turbo-cache/blob/88f6c7d55594e7d045d8f4b25ded4148b02bae7e/cas/grpc_service/execution_server.rs#L193)):*

I put that comment in to remind me, but I think the GrpcScheduler should map between upstream `name` and local `name` and then we can have a lookup table within the `execution_server` for operations rather than stumbling around the different instances looking for it.
</blockquote></details>

for now, do you want to change this in a future PR or keep it this way?
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 150 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_Ghjgo-EIqVW6ormV4:-N_Ghjgo-EIqVW6ormV5:bkvj397) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L150)):*
> ```XML
>             name: name.clone(),
>             stage: ActionStage::CacheCheck,
>             action_digest: action_digest.clone(),
> ```

nit: I don't think cloning is needed any more.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 169 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GkxwK2UM54VI9sSUb:-N_GkxwK2UM54VI9sSUc:b-h2qcbc) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L169)):*
> ```XML
>             }
>             // Not in cache, forward to upstream and proxy state.
>             let rx = action_scheduler.add_action(name, action_info).await.map_or_else(
> ```

nit: I think you want `.unwrap_or_else()`
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 178 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GoPPPCvBw07PYbOlz:-N_GmAqi3Fq7A7Fm_P8w:bd344po) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L178)):*
> ```XML
>             WatchStream::new(rx)
>                 .for_each(|action_state| async {
>                     let _ = tx.send(action_state);
> ```

I think we should terminate the `rx` here or we may end up with either some lingering data or worst case memory leak (depending on impl details of `watch` and `WatchStream` (if they hold Arc references).

I would expand `rx` into a normal `while`:

```rs
let Some(action_state) = rx.changed().await {
  tx.send(*rx.borrow_and_update()).err_tip(|| "Client disconnected in CacheLookupScheduler")?;
}
```

This should prevent most of the issues at least.
___
*[`cas/scheduler/simple_scheduler.rs` line 395 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GqBgU4RzxQ6rtOUrN:-N_GqBgU4RzxQ6rtOUrO:bqyx7yk) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/simple_scheduler.rs#L395)):*
> ```XML
>         };
> 
>         let due_to_backpressure = err.code == Code::ResourceExhausted;
> ```

nit: Maybe inline this?
___
*[`cas/scheduler/tests/cache_lookup_scheduler_test.rs` line 89 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GqJmb6VL03h4ax19A:-N_GqJmb6VL03h4ax19B:buyg47o) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/tests/cache_lookup_scheduler_test.rs#L89)):*
> ```XML
>             .await?;
>         let mut watch_stream = WatchStream::new(watch_channel);
>         if watch_stream.next().await.err_tip(|| "")?.stage != ActionStage::CacheCheck {
> ```

nit: You don't need to `err_tip` if there's no content. Try `.into()`
___
*[`cas/scheduler/tests/cache_lookup_scheduler_test.rs` line 92 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GqPEbB8GwKw9F25TH:-N_GqPEbB8GwKw9F25TI:b-5v3gaq) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/tests/cache_lookup_scheduler_test.rs#L92)):*
> ```XML
>             panic!("Not performing a cache check");
>         }
>         let cached_action_state = watch_stream.next().await.err_tip(|| "")?;
> ```

nit: ditto.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/161)* status: all files reviewed, 7 unresolved discussions (waiting on @aaronmondal and @allada)

___
*[`cas/grpc_service/execution_server.rs` line 193 at r7](https://reviewable.io/reviews/allada/turbo-cache/161#-N_F-_cH3SKfEr9s8O1Y:-N_Hyiso1kPLR8uNhTZm:b18nss1) ([raw file](https://github.com/allada/turbo-cache/blob/88f6c7d55594e7d045d8f4b25ded4148b02bae7e/cas/grpc_service/execution_server.rs#L193)):*

for now, do you want to change this in a future PR or keep it this way?
</blockquote></details>

I think it's probably worth considering in the future instead if we want to support operations listing.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 150 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_Ghjgo-EIqVW6ormV4:-N_Hzxj-4HlexZBpX2nU:bbzu59k) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L150)):*

nit: I don't think cloning is needed any more.
</blockquote></details>

This is required as `action_digest` is a `&DigestInfo`.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 169 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GkxwK2UM54VI9sSUb:-N_Hz32v5_Y6zU7Mfhsy:b-1agwp7) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L169)):*

nit: I think you want `.unwrap_or_else()`
</blockquote></details>

Yeah, I kept changing how I was doing this...
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 178 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GoPPPCvBw07PYbOlz:-N_HzLvF4mznuS9V1mag:b-mn3o9k) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/cache_lookup_scheduler.rs#L178)):*

I think we should terminate the `rx` here or we may end up with either some lingering data or worst case memory leak (depending on impl details of `watch` and `WatchStream` (if they hold Arc references).

I would expand `rx` into a normal `while`:

```rs
let Some(action_state) = rx.changed().await {
  tx.send(*rx.borrow_and_update()).err_tip(|| "Client disconnected in CacheLookupScheduler")?;
}
```

This should prevent most of the issues at least.
</blockquote></details>

I couldn't decide how to do it, I did read that `for_each` would take a `Result` but this one didn't.
___
*[`cas/scheduler/simple_scheduler.rs` line 395 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GqBgU4RzxQ6rtOUrN:-N_HzT629-uKGeQYyTJS:b-8k5e4d) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/simple_scheduler.rs#L395)):*

nit: Maybe inline this?
</blockquote></details>

I need it below for the pause state.
___
*[`cas/scheduler/tests/cache_lookup_scheduler_test.rs` line 89 at r9](https://reviewable.io/reviews/allada/turbo-cache/161#-N_GqJmb6VL03h4ax19A:-N_HzWd7Cewhz3M47qV_:b-cfqs0p) ([raw file](https://github.com/allada/turbo-cache/blob/d9974dd8b55d6613b388653a434789bdaa0cc2f5/cas/scheduler/tests/cache_lookup_scheduler_test.rs#L89)):*

nit: You don't need to `err_tip` if there's no content. Try `.into()`
</blockquote></details>

I actually meant to fill in a context...

chrisstaite-menlo:
*[`cas/scheduler/tests/utils/mock_scheduler.rs` line 28 at r10](https://reviewable.io/reviews/allada/turbo-cache/161#-N_JsNq3AHmD1NYMf8fJ:-N_K1bUr7UIsZItg6G58:betkhb2) ([raw file](https://github.com/allada/turbo-cache/blob/52d903506f8b3de5fe87b34038c531edd7d8ee84/cas/scheduler/tests/utils/mock_scheduler.rs#L28)):*

nit: In the future use `Box` on the large struct. (I'll fix in the scheduler clippy PR).
</blockquote></details>

I wasn't sure it was worth it for a mock.

PR #163, opened 2023-07-11
Add is_empty to LenEntry

aaronmondal:
Since this clippy warning requires touching several files we handle it separately. Most remaining warnings should be self-contained enough that we can handle them on a file-by-file basis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)

### DIFF
--- a/cas/store/filesystem_store.rs
+++ b/cas/store/filesystem_store.rs
@@ -254,6 +254,10 @@ impl LenEntry for FileEntryImpl {
         self.file_size as usize
     }
 
+    fn is_empty(&self) -> bool {
+        self.file_size == 0
+    }
+
     #[inline]
     async fn touch(&self) {
         let result = self
--- a/cas/store/memory_store.rs
+++ b/cas/store/memory_store.rs
@@ -41,6 +41,11 @@ impl LenEntry for BytesWrapper {
     fn len(&self) -> usize {
         Bytes::len(&self.0)
     }
+
+    #[inline]
+    fn is_empty(&self) -> bool {
+        Bytes::is_empty(&self.0)
+    }
 }
 
 pub struct MemoryStore {
--- a/cas/store/tests/filesystem_store_test.rs
+++ b/cas/store/tests/filesystem_store_test.rs
@@ -110,6 +110,10 @@ impl<Hooks: FileEntryHooks + 'static + Sync + Send> LenEntry for TestFileEntry<H
         self.inner.as_ref().unwrap().len()
     }
 
+    fn is_empty(&self) -> bool {
+        self.inner.as_ref().unwrap().is_empty()
+    }
+
     async fn touch(&self) {
         self.inner.as_ref().unwrap().touch().await
     }
--- a/util/evicting_map.rs
+++ b/util/evicting_map.rs
@@ -64,6 +64,9 @@ pub trait LenEntry {
     /// Length of referenced data.
     fn len(&self) -> usize;
 
+    /// Returns `true` if `self` has zero length.
+    fn is_empty(&self) -> bool;
+
     /// Called when an entry is touched.
     #[inline]
     async fn touch(&self) {}
@@ -92,13 +95,18 @@ impl<T: LenEntry + Send + Sync> LenEntry for Arc<T> {
     }
 
     #[inline]
+    fn is_empty(&self) -> bool {
+        T::is_empty(self.as_ref())
+    }
+
+    #[inline]
     async fn touch(&self) {
-        self.as_ref().touch().await
+        self.as_ref().touch().await;
     }
 
     #[inline]
     async fn unref(&self) {
-        self.as_ref().unref().await
+        self.as_ref().unref().await;
     }
 }
 
@@ -148,7 +156,7 @@ where
             };
             serialized_lru
                 .data
-                .push((serialized_digest, eviction_item.seconds_since_anchor as i32));
+                .push((serialized_digest, eviction_item.seconds_since_anchor));
         }
         serialized_lru
     }
@@ -187,11 +195,8 @@ where
     }
 
     async fn evict_items(&self, state: &mut State<T>) {
-        let mut peek_entry = if let Some((_, entry)) = state.lru.peek_lru() {
-            entry
-        } else {
-            return;
-        };
+        let Some((_, mut peek_entry)) = state.lru.peek_lru() else { return; };
+
         while self.should_evict(state.lru.len(), peek_entry, state.sum_store_size) {
             let (key, eviction_item) = state.lru.pop_lru().expect("Tried to peek() then pop() but failed");
             state.sum_store_size -= eviction_item.data.len() as u64;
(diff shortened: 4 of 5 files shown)


allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`util/evicting_map.rs` line 68 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_1fCsrCN7jTKM_JIzm:-N_1fCsrCN7jTKM_JIzn:b-gbp8uk) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/evicting_map.rs#L68)):*
> ```XML
> 
>     /// Returns `true` if `self` has zero length.
>     fn is_empty(&self) -> bool;
> ```

nit: Maybe put the default impl of `self.len() == 0` here and remove the others?
___
*[`util/evicting_map.rs` line 198 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_1fTWN0S3F3utzy3qb:-N_1fTWN0S3F3utzy3qc:b-wv5hw1) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/evicting_map.rs#L198)):*
> ```XML
> 
>     async fn evict_items(&self, state: &mut State<T>) {
>         let Some((_, mut peek_entry)) = state.lru.peek_lru() else { return; };
> ```

Wow this is new syntax I didn't even know you could use :confused:

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163#-:-N_1fgGl5nXI6yJMDpYR:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`util/evicting_map.rs` line 68 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_1fCsrCN7jTKM_JIzm:-N_2I6KIBa58RYf5gtEi:b-d611cu) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/evicting_map.rs#L68)):*

nit: Maybe put the default impl of `self.len() == 0` here and remove the others?
</blockquote></details>

I'm not sure that's what we would want. According to https://rust-lang.github.io/rust-clippy/master/index.html#/len_without_is_empty the runtimes of `len` and `is_empty` might be different, with `is_empty` being potentially faster for some structures.

In our implementation that might not always be the case, but I think somewhere in this commit we have a `T::is_empty`. Depending on `T` that might actually make a difference.
___
*[`util/evicting_map.rs` line 198 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_1fTWN0S3F3utzy3qb:-N_2HPtsDDlb-ZJlN0Um:bfjqy0g) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/evicting_map.rs#L198)):*

Wow this is new syntax I didn't even know you could use :confused: 
</blockquote></details>

This beautiful pattern is the is the `let-else` pattern, available since Rust 1.65. So this wasn't possible before the recent dependency update :heart_eyes:  

Example: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html
RFC: https://rust-lang.github.io/rfcs/3137-let-else.html

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`util/evicting_map.rs` line 68 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_1fCsrCN7jTKM_JIzm:-N_2OJE-CTqyPsMwfBpw:bwyb0fw) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/evicting_map.rs#L68)):*

I'm not sure that's what we would want. According to https://rust-lang.github.io/rust-clippy/master/index.html#/len_without_is_empty the runtimes of `len` and `is_empty` might be different, with `is_empty` being potentially faster for some structures.

In our implementation that might not always be the case, but I think somewhere in this commit we have a `T::is_empty`. Depending on `T` that might actually make a difference.
</blockquote></details>

Yeah, that seems reasonable.
___
*[`util/tests/evicting_map_test.rs` line 285 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_2OZ43CAR4dMQCCm8r:-N_2OZ43CAR4dMQCCm8s:b-mbzx1r) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/tests/evicting_map_test.rs#L285)):*
> ```XML
>             fn is_empty(&self) -> bool {
>                 // Note: We are not testing this functionality.
>                 return true;
> ```

nit: Use `unreachable!()` when we are not testing the functionality so it is clear to any future test writer why it is acting weird.

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`util/tests/evicting_map_test.rs` line 285 at r1](https://reviewable.io/reviews/allada/turbo-cache/163#-N_2OZ43CAR4dMQCCm8r:-N_5yy2h2RaQHJQNCp5Y:b-lsdz9b) ([raw file](https://github.com/allada/turbo-cache/blob/2d9d86f8224b454dd130fb469ea487bda80a072d/util/tests/evicting_map_test.rs#L285)):*

Thanks, I didn't think of that! Two sidenotes:

1. Turns out we're actually reaching `len` and `touch`. Adding `unreachable` there breaks the test. Bug? 
2. Super nit: The string in `unreachable` deliberatly doesn't end in a period because otherwise then the error message for it would end in a confusing `.',` pattern :laughing: 
</blockquote></details>

If it really is hitting this code, then just make it return false. It looks like `len()` already returns zero, so it should probably trigger `is_empty()` to be true.

I only said use `unreachable!()` because of the comment, not the code.

As for a period... I don't know why, I just want comments to be consistent. I don't actually care either way. As for error messages... They are designed to be read only when debugging not when reading the code. But :shrug:

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/163)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

PR #164, opened 2023-07-11
Add convenience config to test clippy

aaronmondal:
Since the Bazel Aspect for Clippy fails on warnings, we need to gate this behind a config flag for now. The config gate is temporary until we can enable clippy by default during tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,3 +31,18 @@ build --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
 test --@rules_rust//:rustfmt.toml=//:.rustfmt.toml
 # This will make rustfmt only run on `bazel test`.
 test --output_groups=+rustfmt_checks
+
+# TODO: The repo doesn't build with clippy enabled yet and this aspect will fail
+#       even on warnings. To test a single build target:
+#
+#           bazel build --config=clippy mytarget
+#
+#       To test current conformance of the entire repository, run tests with:
+#
+#           bazel test --config=clippy --keep_going ...
+#
+#       When all warnings are fixed the config gate will be removed and clippy
+#       will be enabled by default during tests.
+build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
+test:clippy --output_groups=+clippy_checks
+build:clippy --output_groups=+clippy_checks


aaronmondal:
@allada @chrisstaite-menlo With this you can run clippy against single targets via Bazel. The "usual" way of invoking it won't work since the repo isn't buildable with cargo.

Running these at the moment will produce a gigantic list of errors. That list is actually *too small* since it doesn't include the `pedantic` warning set which covers many relevant lints like bad 64-32 conversions, "... shouldn't panic" and potential async issues.

For maximum fun I highly encourage you to try adding this line to the top of a `.rs` file and then building it with `--config=clippy` :rofl::

```
#![warn(clippy::all, clippy::pedantic, clippy::nursery)]
```

Note that not all warnings from `clippy::nursery` are meant to be followed. Some of them are quite useful and will at some point make it into `all` or `pedantic`, but some are just meant as suggestions and can safely be ignored.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/164)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`.bazelrc` line 46 at r1](https://reviewable.io/reviews/allada/turbo-cache/164#-N_1gGHB2lFY0RYnyerj:-N_1gGHB2lFY0RYnyerk:b-s6qdh3) ([raw file](https://github.com/allada/turbo-cache/blob/341e3e7349910c6b8a8ab936b152d284b85a46f6/.bazelrc#L46)):*
> ```
> #       When all warnings are fixed the config gate will be removed and clippy
> #       will be enabled by default during tests.
> build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
> ```

nit: Maybe put this also behind `build:clippy` since it will not be used in normal builds yet?

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/164#-:-N_1gNO40rkTZmhtzpb2:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/164)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

PR #166, opened 2023-07-11
Add ability to create low watermark to avoid thrashing

chrisstaite-menlo:
We had some issues that when we were hitting the eviction cap (of 200Gb) the system was thrashing as it tried to keep up with evicting data and then populating new.  This was around the time of the dodgy Mutex, so I'm not sure how much that had to do with the severe degradation we saw.

Anyway, this allows the user to configure a low watermark so when eviction starts it frees up a large amount of space to avoid constantly having to evict things.


allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/166)* status: all files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/evicting_map.rs` line 202 at r1](https://reviewable.io/reviews/allada/turbo-cache/166#-N_5xhH_307c8FNDAule:-N_5xhH_307c8FNDAulf:bktrw8x) ([raw file](https://github.com/allada/turbo-cache/blob/5fb0efa6e210e805ddda6926bea5e0b75aecdd29/util/evicting_map.rs#L202)):*
> ```XML
>             return;
>         };
>         let mut evicting = false;
> ```

nit: Suggest instead passing in the `max_bytes`:

```rs
let mut should_evict = self.should_evict(state.lru.len(), peek_entry, state.sum_store_size, self.max_bytes);
while should_evict {
  ...
  should_evict = self.should_evict(state.lru.len(), peek_entry, state.sum_store_size, self.max_bytes - self.evict_bytes);
}
```
___
*[`util/tests/evicting_map_test.rs` line 422 at r1](https://reviewable.io/reviews/allada/turbo-cache/166#-N_5ySyY2ttgcDKpQ1yZ:-N_5ySyY2ttgcDKpQ1y_:bkv55ef) ([raw file](https://github.com/allada/turbo-cache/blob/5fb0efa6e210e805ddda6926bea5e0b75aecdd29/util/tests/evicting_map_test.rs#L422)):*
> ```XML
>                 max_seconds: 0,
>                 max_bytes: 0,
>                 evict_bytes: 0,
> ```

nit: Can we also write a test for this new feature?

PR #167, opened 2023-07-11
Add support for backpressure from workers.

chrisstaite-menlo:
Our main issue is that some parts of our code base eat all of the RAM.  This is one of the mitigations against that.

This allows workers to be configured with minimum resource limits separately to the the platform properties.  This allows them to return a ResourceExhausted state which pauses the worker on the scheduler until an action has completed (if it has one).

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)

### DIFF
--- a/cas/scheduler/simple_scheduler.rs
+++ b/cas/scheduler/simple_scheduler.rs
@@ -127,16 +127,15 @@ impl Workers {
         assert!(matches!(awaited_action.current_state.stage, ActionStage::Queued));
         let action_properties = &awaited_action.action_info.platform_properties;
         let mut workers_iter = self.workers.iter_mut();
-        let workers_iter = match self.allocation_strategy {
-            // Use rfind to get the least recently used that satisfies the properties.
-            config::schedulers::WorkerAllocationStrategy::LeastRecentlyUsed => {
-                workers_iter.rfind(|(_, w)| action_properties.is_satisfied_by(&w.platform_properties))
-            }
-            // Use find to get the most recently used that satisfies the properties.
-            config::schedulers::WorkerAllocationStrategy::MostRecentlyUsed => {
-                workers_iter.find(|(_, w)| action_properties.is_satisfied_by(&w.platform_properties))
-            }
-        };
+        let workers_iter =
+            match self.allocation_strategy {
+                // Use rfind to get the least recently used that satisfies the properties.
+                config::schedulers::WorkerAllocationStrategy::LeastRecentlyUsed => workers_iter
+                    .rfind(|(_, w)| !w.is_paused && action_properties.is_satisfied_by(&w.platform_properties)),
+                // Use find to get the most recently used that satisfies the properties.
+                config::schedulers::WorkerAllocationStrategy::MostRecentlyUsed => workers_iter
+                    .find(|(_, w)| !w.is_paused && action_properties.is_satisfied_by(&w.platform_properties)),
+            };
         let worker_id = workers_iter.map(|(_, w)| &w.id);
         // We need to "touch" the worker to ensure it gets re-ordered in the LRUCache, since it was selected.
         if let Some(&worker_id) = worker_id {
@@ -247,10 +246,14 @@ impl SimpleSchedulerImpl {
         Ok(rx)
     }
 
-    fn retry_action(&mut self, action_info: &Arc<ActionInfo>, worker_id: &WorkerId) {
+    fn retry_action(&mut self, action_info: &Arc<ActionInfo>, worker_id: &WorkerId, due_to_backpressure: bool) {
         match self.active_actions.remove(action_info) {
             Some(running_action) => {
                 let mut awaited_action = running_action.action;
+                // Don't count a backpressure failure as an attempt for an action.
+                if due_to_backpressure {
+                    awaited_action.attempts -= 1;
+                }
                 let send_result = if awaited_action.attempts >= self.max_job_retries {
                     let mut default_action_result = ActionResult::default();
                     default_action_result.execution_metadata.worker = format!("{}", worker_id);
@@ -298,7 +301,7 @@ impl SimpleSchedulerImpl {
             // We create a temporary Vec to avoid doubt about a possible code
             // path touching the worker.running_action_infos elsewhere.
             for action_info in worker.running_action_infos.drain() {
-                self.retry_action(&action_info, &worker_id);
+                self.retry_action(&action_info, &worker_id, false);
             }
         }
         // Note: Calling this many time is very cheap, it'll only trigger `do_try_match` once.
@@ -390,6 +393,8 @@ impl SimpleSchedulerImpl {
             }
         };
 
+        let due_to_backpressure = err.code == Code::ResourceExhausted;
+
         if running_action.worker_id != *worker_id {
             log::error!(
                 "Got a result from a worker that should not be running the action, Removing worker. Expected worker {} got worker {}",
@@ -406,11 +411,18 @@ impl SimpleSchedulerImpl {
 
         // Clear this action from the current worker.
         if let Some(worker) = self.workers.workers.get_mut(worker_id) {
+            let was_paused = worker.is_paused;
+            // This unpauses, but since we're completing with an error, don't
+            // unpause unless all actions have completed.
             worker.complete_action(&action_info);
+            // Only pause if there's an action still waiting that will unpause.
+            if (was_paused || due_to_backpressure) && worker.has_actions() {
+                worker.is_paused = true;
+            }
         }
 
         // Re-queue the action or fail on max attempts.
-        self.retry_action(&action_info, &worker_id);
+        self.retry_action(&action_info, &worker_id, due_to_backpressure);
     }
 
     fn update_action(
--- a/cas/scheduler/worker.rs
+++ b/cas/scheduler/worker.rs
@@ -89,6 +89,9 @@ pub struct Worker {
     // Warning: Do not update this timestamp without updating the placement of the worker in
     // the LRUCache in the Workers struct.
     pub last_update_timestamp: WorkerTimestamp,
+
+    /// Whether the worker rejected the last action due to back pressure.
+    pub is_paused: bool,
 }
 
 impl Worker {
@@ -104,6 +107,7 @@ impl Worker {
             tx,
             running_action_infos: HashSet::new(),
             last_update_timestamp: timestamp,
+            is_paused: false,
         }
     }
 
@@ -148,7 +152,12 @@ impl Worker {
 
     pub fn complete_action(&mut self, action_info: &Arc<ActionInfo>) {
         self.running_action_infos.remove(action_info);
-        self.restore_platform_properties(&action_info.platform_properties)
+        self.restore_platform_properties(&action_info.platform_properties);
+        self.is_paused = false;
+    }
+
+    pub fn has_actions(&self) -> bool {
+        !self.running_action_infos.is_empty()
     }
 
     /// Reduces the platform properties available on the worker based on the platform properties provided.
--- a/cas/worker/local_worker.rs
+++ b/cas/worker/local_worker.rs
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 use std::pin::Pin;
+use std::process::Stdio;
+use std::str;
 use std::sync::Arc;
 use std::time::Duration;
 
 use futures::{future::BoxFuture, select, stream::FuturesUnordered, FutureExt, StreamExt, TryFutureExt};
+use tokio::process;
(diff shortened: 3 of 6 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

___
*[`cas/worker/local_worker.rs` line 94 at r1](https://reviewable.io/reviews/allada/turbo-cache/167#-N_5ul72BRsDLtVCe-Yb:-N_5ul72BRsDLtVCe-Yc:bojzo71) ([raw file](https://github.com/allada/turbo-cache/blob/4b220db02e1eb5829a496f751093dbce61a7b71e/cas/worker/local_worker.rs#L94)):*
> ```XML
>     }
> 
>     fn preconditions_met(&self) -> Result<(), Error> {
> ```

Can we instead put this logic into a script and run this command (set in configuration)?

This would easily allow the user to check for other resources too (like cpu, gpu, ram, network, or atomic tasks [like downloading docker images]).

I suggest exit code 0 success and anything else triggers the worker to be paused, then re-run the same command on completion of every task if the worker is paused to send a message to the scheduler to unpause it.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/worker/local_worker.rs` line 94 at r1](https://reviewable.io/reviews/allada/turbo-cache/167#-N_5ul72BRsDLtVCe-Yb:-N_DpcVsBKcTSjUfUvDD:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/4b220db02e1eb5829a496f751093dbce61a7b71e/cas/worker/local_worker.rs#L94)):*

Can we instead put this logic into a script and run this command (set in configuration)?

This would easily allow the user to check for other resources too (like cpu, gpu, ram, network, or atomic tasks [like downloading docker images]).

I suggest exit code 0 success and anything else triggers the worker to be paused, then re-run the same command on completion of every task if the worker is paused to send a message to the scheduler to unpause it.
</blockquote></details>

Done.

allada (commented):
> Reviewed 1 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: all files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: all files reviewed, 5 unresolved discussions (waiting on @allada)

___
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_EdMBtCLLYrQqo6SZU:bo36ldt) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

Idea: Invert this to a post-condition and use `entrypoint_cmd` to abort the task and then the `post-condition` to return a special `internal_error` to the scheduler with a "pause" instruction.

The reason is because it will play really well into cases like I outlined before where we want to terminate the worker on an error. So the user could get a job failure, run post-checker script, see some kernel/driver/whatever error in the sys logs, tell the scheduler to pause all jobs, wait a few seconds, then terminate.

We can also use this script instead of using a special exit code (-176 iirc).
</blockquote></details>

That would mean I'd have to do the check in the entrypoint and in the post-condition, right?  During the two, what happens if the condition cleared itself?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/simple_scheduler.rs` line 417 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EY1kR1p1PFAVa5qjm:-N_EhF4L5wc6DGYovAjq:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/scheduler/simple_scheduler.rs#L417)):*

It looks like we never unpause on all future jobs succeeding, resulting in the node being paused forever.
</blockquote></details>

Done.
___
*[`cas/scheduler/simple_scheduler.rs` line 419 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EYL4c-OjYlPnTii-h:-N_EhEdwB_Pymwbhj8bx:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/scheduler/simple_scheduler.rs#L419)):*

I don't think we should ever un-pause on job failure unless there are no more jobs running on the worker.
</blockquote></details>

Done.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_EgGWY6Z0JGphoBSwg:b569ic7) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

That would mean I'd have to do the check in the entrypoint and in the post-condition, right?  During the two, what happens if the condition cleared itself?
</blockquote></details>

I was thinking we'd inject a few magic variables to the running job (if the config file permitted it). I was planning on injecting stuff to the task anyway, it's even semi-documented `PlatformPropertyValue::Priority`. The idea is we inject a few environs with meta data about the task that is running, the runner can use these to configure the job appropriately. For example, the worker might run everything under docker and the `PlatformPropertyValue::Minimum`, might be used to tell docker how many cores or memory a job can use or `PlatformPropertyValue::Exact` might be used to tell docker what image should be used to run the task.

If the user is worried about these magic environs poisoning the run, a config variable would enable it to be turned off, and/or they could an ``entrypoint_cmd` script to remove the environs before running the command.

I am struggling here because I don't want to block you on this, but I also think it will give a lot more flexibility to other use cases.


To be specific, your use case would look like this:
```sh
# entry_point.sh has TURBO_CACHE_JOB_ID environ set.
if ! /root/memory_check.sh; then
  echo "-175" > "./$TURBO_CACHE_JOB_ID"
fi
exec $0
```

```sh
# post_checker.sh same environs are injected with maybe the exit code too.
if [ -f "./$TURBO_CACHE_JOB_ID" ] ;  then
  exit $(cat ./$TURBO_CACHE_JOB_ID)
fi
exit 0
```

```json5
// Config file
{
  "worker": {
    // Tell turbo-cache to inject environs (defaults to false [yes-inject]).
    "no_env_injection": false,
    "entrypoint_cmd": "/root/entry_point.sh",
    "post_job_cmd": "/root/post_checker.sh",
  }
}
```

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_Eml3i7reCUwO-yMzV:b-y80t0c) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

I was thinking we'd inject a few magic variables to the running job (if the config file permitted it). I was planning on injecting stuff to the task anyway, it's even semi-documented `PlatformPropertyValue::Priority`. The idea is we inject a few environs with meta data about the task that is running, the runner can use these to configure the job appropriately. For example, the worker might run everything under docker and the `PlatformPropertyValue::Minimum`, might be used to tell docker how many cores or memory a job can use or `PlatformPropertyValue::Exact` might be used to tell docker what image should be used to run the task.

If the user is worried about these magic environs poisoning the run, a config variable would enable it to be turned off, and/or they could an ``entrypoint_cmd` script to remove the environs before running the command.

I am struggling here because I don't want to block you on this, but I also think it will give a lot more flexibility to other use cases.


To be specific, your use case would look like this:
```sh
# entry_point.sh has TURBO_CACHE_JOB_ID environ set.
if ! /root/memory_check.sh; then
  echo "-175" > "./$TURBO_CACHE_JOB_ID"
fi
exec $0
```

```sh
# post_checker.sh same environs are injected with maybe the exit code too.
if [ -f "./$TURBO_CACHE_JOB_ID" ] ;  then
  exit $(cat ./$TURBO_CACHE_JOB_ID)
fi
exit 0
```

```json5
// Config file
{
  "worker": {
    // Tell turbo-cache to inject environs (defaults to false [yes-inject]).
    "no_env_injection": false,
    "entrypoint_cmd": "/root/entry_point.sh",
    "post_job_cmd": "/root/post_checker.sh",
  }
}
```
</blockquote></details>

note: -175 is supposed to represent some magic status codes to pass up, or whatever `ResourceExausted` status code is.

allada (commented):
> Reviewed 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: all files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

chrisstaite-menlo:
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_FA4V_74uql25vNrBA:b88mi74) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

note: -175 is supposed to represent some magic status codes to pass up, or whatever `ResourceExausted` status code is.
</blockquote></details>

You'd need to filter the files probably?

chrisstaite-menlo:
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_FAVF090q93GSok_Q1:b-pgxzp3) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

You'd need to filter the files probably?
</blockquote></details>

I still have issues with it downloading all of the resources for an action when it's not going to perform it, I like running the script before the action is performed.

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/167)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

___
*[`cas/worker/local_worker.rs` line 174 at r2](https://reviewable.io/reviews/allada/turbo-cache/167#-N_EVcR4BoooQC3ShyAm:-N_FB30y7VECn1ogMk2q:bx701lb) ([raw file](https://github.com/allada/turbo-cache/blob/e8082efd445c1788fac14d6a7c37d046325d54b2/cas/worker/local_worker.rs#L174)):*

I still have issues with it downloading all of the resources for an action when it's not going to perform it, I like running the script before the action is performed.
</blockquote></details>

Oh interesting point, I forgot about the fact it'll need to download everything first. It sounds like to support all cases we'd need potentially 3 scripts running. I guess it's not too bad, and they become noops unless the user needs/wants to do this.

In that case, yeah I agree we should have a pre-condition script.

chrisstaite-menlo:
*[`cas/scheduler/simple_scheduler.rs` line 396 at r3](https://reviewable.io/reviews/allada/turbo-cache/167#-N_FDN6F0ISDyEXQImMt:-N_FIayc08kANGgwh3x0:b9oa65w) ([raw file](https://github.com/allada/turbo-cache/blob/8c1d0fc859b97902470775bb12592fbf17da3cb7/cas/scheduler/simple_scheduler.rs#L396)):*

nit: Should we here apply our `-1`, I think it'd make the code a bit more clean?

```rs
                // Don't count a backpressure failure as an attempt for an action.
                if err.code == Code::ResourceExhausted {
                    awaited_action.attempts -= 1;
                }

```
</blockquote></details>

Came in after merge, I'll pop it on the cache branch.

PR #168, opened 2023-07-12
Run integration tests on 18.04 but not build on anything but 22.04

allada:
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/168)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         # Which OS versions we will test on.
-        os_version: [ 20.04, 22.04 ]
+        os_version: [ 22.04 ]
     steps:
     - uses: actions/checkout@v3.5.3
       with:
@@ -41,7 +41,7 @@ jobs:
     strategy:
       matrix:
         # Which OS versions we will test on.
-        os_version: [ 20.04, 22.04 ]
+        os_version: [ 22.04 ]
     steps:
     - uses: actions/checkout@v3.5.3
       with:
@@ -52,9 +52,11 @@ jobs:
       with:
         context: .
         file: ./deployment-examples/docker-compose/Dockerfile
+        # For integration tests we only ensure the binary can work on specifc
+        # OS versions, not built on these versions.
         build-args: |
           OPT_LEVEL=fastbuild
-          OS_VERSION=${{ matrix.os_version }}
+          OS_VERSION_EXEC=${{ matrix.os_version }}
         # The `builder` stage is pointless to be cached because any change to the
         # PR invalidates it. So to save the limited cache space we skip it.
         no-cache-filters: |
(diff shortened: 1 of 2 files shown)


aaronmondal (commented):
I like this change a lot. It's way better if we can freely use the latest toolchains to build the repo as long as it still runs on older systems. This also reduces total CI runtime significantly as running just the tests is much simipler than building AND testing for every supported platform :rocket: 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/168#-:-N_6nLX3AUp7VoZgzwk4:bdm3bnu)* status: 0 of 2 files reviewed, all discussions resolved

aaronmondal (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/168#-:-N_6no2v8IKwrTy4g0ld:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

allada (commented):
Sadly, I'm not seeing good results with this, so I think I'm going to abandoned this concept.

It will compile on 18.04 if I use a newer version of clang/gcc, but it's not easy to install and complicates the docker files a lot...

I'm thinking of just abandoning the idea completely.

I would like to just build it on one machine and run it on another, but it ends up not working because the glibc/libm and openssl libraries between the versions cannot be cross-linked :frowning: 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/168#-:-N_6ybxdDNVlVNGk8X9f:b-ucqgnz)* status: 0 of 2 files reviewed, all discussions resolved

aaronmondal (commented):
Oh I see. I mean it's surely *possible* to cross-compile it, but is it really worth the effor? For something like new ARM CPUs or WASM I'd totally get it, but for backwards compatibility with an OS that's reached EOL I'm not sure.

An option would be to statically link against LibMusl and a static OpenSSL. This could also be interesting for Cloud users because of more enterprise-friendly licensing and in better use with minimalistic container images.

The downside with musl is that we'd have to set up a fairly custom toolchain for that. If this could be useful I'd be happy to do this, but it'll be rather complicated and at the moment I'm not sure it's required "enough".

We might want to do this at a later point in time though. I've experimented with a stacic Musl-Bazel previously. In the end we went for a simpler wrapper because linking against musl means you also have to *build* against musl, i.e. you have to build the world before you get your final binary :rofl: But it's something that I'm still quite interested in. WASI-libc (the libc impl for WASM) is also a modified musl, so some of this could be useful there as well.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/168#-:-N_70JpnCtZJx4M7cho8:b-fyz5ji)* status: 0 of 2 files reviewed, all discussions resolved

allada:
Yeah, in this case, I think we should hold off to see how things grow.

I think static linking would be good for certain applications, but like you said I'm not sure we are ready yet.

I think investing in WASI seems like we'd get more returns than investing in backwards compatibility. If we push WASI to the limits, I could reach out to some of my former chrome-WASM team members and see if they can use this project as a base-case for requested features.

PR #171, opened 2023-07-12
Fix clippy messages in cas/store

allada:
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/171)

### DIFF
--- a/cas/store/ac_utils.rs
+++ b/cas/store/ac_utils.rs
@@ -63,23 +63,21 @@ pub async fn get_and_decode_digest<T: Message + Default>(
 }
 
 /// Takes a proto message and will serialize it and upload it to the provided store.
-pub fn serialize_and_upload_message<'a, T: Message>(
+pub async fn serialize_and_upload_message<'a, T: Message>(
     message: &'a T,
     cas_store: Pin<&'a dyn Store>,
-) -> impl Future<Output = Result<DigestInfo, Error>> + 'a {
-    async move {
-        let mut buffer = BytesMut::new();
-        let digest = {
-            message
-                .encode(&mut buffer)
-                .err_tip(|| "Could not encode directory proto")?;
-            let mut hasher = Sha256::new();
-            hasher.update(&buffer);
-            DigestInfo::new(hasher.finalize().into(), buffer.len() as i64)
-        };
-        upload_to_store(cas_store, digest.clone(), &mut Cursor::new(buffer)).await?;
-        Ok(digest)
-    }
+) -> Result<DigestInfo, Error> {
+    let mut buffer = BytesMut::new();
+    let digest = {
+        message
+            .encode(&mut buffer)
+            .err_tip(|| "Could not encode directory proto")?;
+        let mut hasher = Sha256::new();
+        hasher.update(&buffer);
+        DigestInfo::new(hasher.finalize().into(), buffer.len() as i64)
+    };
+    upload_to_store(cas_store, digest.clone(), &mut Cursor::new(buffer)).await?;
+    Ok(digest)
 }
 
 /// Given a bytestream computes the digest for the data.
@@ -137,7 +135,7 @@ pub fn upload_to_store<'a, R: AsyncRead + Unpin>(
                 .read_buf(&mut chunk)
                 .await
                 .err_tip(|| "Could not read chunk during upload_to_store")?;
-            if chunk.len() == 0 {
+            if chunk.is_empty() {
                 break; // EOF.
             }
             tx.send(chunk.freeze())
--- a/cas/store/compression_store.rs
+++ b/cas/store/compression_store.rs
@@ -174,7 +174,7 @@ impl UploadState {
             indexes: vec![SliceIndex { ..Default::default() }; max_index_count],
             index_count: max_index_count as u32,
             uncompressed_data_size: 0, // Updated later.
-            config: header.config.clone(),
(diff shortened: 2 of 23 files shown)


aaronmondal (approved):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/171)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/store/compression_store.rs` line 333 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_9yVzh6OjYTTyxXigV:-N_9yVzi5VSJP1h42OUz:bm3q4wt) ([raw file](https://github.com/allada/turbo-cache/blob/e6430900ef21ad4bc651eb0076060b513ca8c3b3/cas/store/compression_store.rs#L333)):*
> ```XML
>             // Note: We need to be careful that if we don't have any data (zero bytes) it
>             // doesn't go to -1.
>             if index_count > 0 {
> ```

Ah yes this one is interesting.
___
*[`cas/store/default_store_factory.rs` line 36 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_9ynWF0GpVmryuuOZf:-N_9ynWF0GpVmryuuOZg:b-dth0y3) ([raw file](https://github.com/allada/turbo-cache/blob/e6430900ef21ad4bc651eb0076060b513ca8c3b3/cas/store/default_store_factory.rs#L36)):*
> ```XML
>     backend: &'a StoreConfig,
>     store_manager: &'a Arc<StoreManager>,
> ) -> Pin<Box<dyn Future<Output = Result<Arc<dyn Store>, Error>> + 'a>> {
> ```

Doesn't this raise something like "type too complicated please declare this as separate type"? May have only been in the pedantic or nursery sets though. Not blocking.
___
*[`cas/store/tests/fast_slow_store_test.rs` line 84 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_9zpmDClwtu55ChFS-:-N_9zpmECwkxHPKwmPzw:b-hibg2i) ([raw file](https://github.com/allada/turbo-cache/blob/e6430900ef21ad4bc651eb0076060b513ca8c3b3/cas/store/tests/fast_slow_store_test.rs#L84)):*
> ```XML
> 
>         let original_data = make_random_data(20 * MEGABYTE_SZ);
>         let digest = DigestInfo::try_new(&VALID_HASH, 100).unwrap();
> ```

Amazing how many borrows we can remove here :relaxed: 
___
*[`cas/store/tests/filesystem_store_test.rs` line 506 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_A-3EuEgi11YvXc-Vl:-N_A-3EuEgi11YvXc-Vm:b-cnoblq) ([raw file](https://github.com/allada/turbo-cache/blob/a095c3c8b9892cd1245086b2c52f05320cb2d1c0/cas/store/tests/filesystem_store_test.rs#L506)):*
> ```XML
>             while let Some(temp_dir_entry) = read_dir_stream.next().await {
>                 let path = temp_dir_entry?.path();
>                 panic!("No files should exist in temp directory, found: {:?}", path);
> ```

I'm fine with adding it like this, but pedantic warnings would probably suggest inlining `path` into the string.
___
*[`cas/store/tests/filesystem_store_test.rs` line 743 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_A-Mo4-URvAnV8RVFE:-N_A-Mo4-URvAnV8RVFF:brkx4qu) ([raw file](https://github.com/allada/turbo-cache/blob/a095c3c8b9892cd1245086b2c52f05320cb2d1c0/cas/store/tests/filesystem_store_test.rs#L743)):*
> ```XML
>     }
> 
>     #[allow(clippy::await_holding_refcell_ref)]
> ```

Do we really want to allow this? AFAIK the only false positive with this warning is if the ref is explicitly dropped. Are we doing that here or is this an actual problem?

https://rust-lang.github.io/rust-clippy/master/index.html#/await_holding_refcell_ref

aaronmondal (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/171)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

___
*[`cas/store/tests/filesystem_store_test.rs` line 743 at r1](https://reviewable.io/reviews/allada/turbo-cache/171#-N_A-Mo4-URvAnV8RVFE:-N_A7GlMD0vrMuTK4rqH:b-hw5hbx) ([raw file](https://github.com/allada/turbo-cache/blob/a095c3c8b9892cd1245086b2c52f05320cb2d1c0/cas/store/tests/filesystem_store_test.rs#L743)):*

Yeah as I outlined here: https://github.com/allada/turbo-cache/issues/153#issuecomment-1631867694

There's no real easy way to do this. In the case here we need to make sure the `.poll()` returns `Polling` (ie: the future is not resolved). Internally `poll!()` creates a future and awaits it so it can get the ctx to run `.poll()`.

The test is actually a regression test for a bug that was found a long time ago and this was the best way to check it.

Since this is a test and not in the production code, I'm less strict on this kind of stuff.

Also, I tried to use this function on just the line or block that triggers it and it didn't work, so i did it to the entire function.
</blockquote></details>

Oh I overlooked that comment. Ok looks like this *might* be something we'd want to look at at some point, but "later" :sweat_smile:

PR #174, opened 2023-07-12
Removes needless overoptimization of strings for DigestInfo

allada:
We make so many copies of DigestInfo and it's not trivial-copyable because of an optimization that adds around 40 bytes of data to each struct which tries to preserve the string representation of the digest when possible. This optimization seems to be pointless and likely ends up costing more than it saves.

On local testing, I found that the overhead of the LazyInit was slightly more than a single invocation of `.str()`. Since we usually only call `.str()` once per DigestInfo, it does not seem like a good idea to assume we would need the hex string more than once per struct.

There is also a significant chance this will give secondary optimizations because without this optimization the struct is trivially-copyable. This enables the optimizer to do even more clever tricks for inlining since it knows it does not need to touch the heap.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/174)

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "ec1aaea79db6c8962a1718248f7b28b82eb70a7dee255529ea671dfed22e58a2",
+  "checksum": "4f921e8c24fd74362f4fa660db203c02dfdc564ae44a23e8d4526d28aab7aafb",
   "crates": {
     "addr2line 0.20.0": {
       "name": "addr2line",
@@ -2499,10 +2499,6 @@
             {
               "id": "json5 0.4.1",
               "target": "json5"
-            },
-            {
-              "id": "lazy-init 0.5.1",
-              "target": "lazy_init"
             },
             {
               "id": "lazy_static 1.4.0",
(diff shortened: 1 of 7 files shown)


chrisstaite-menlo (approved):
:lgtm:

> Reviewed 8 of 8 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/174#-:-N_9qSl85s8iKMIoaS2y:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`util/common.rs` line 116 at r1](https://reviewable.io/reviews/allada/turbo-cache/174#-N_9q8ehCGtCR3tj18BL:-N_9q8ehCGtCR3tj18BM:b-t6u9c1) ([raw file](https://github.com/allada/turbo-cache/blob/c8f27514b27c780e0cce3ceb0c574cea743acfee/util/common.rs#L116)):*
> ```XML
>         let packed_hash = val.packed_hash;
>         Digest {
>             hash: hex::encode(packed_hash),
> ```

Re-implementing `val.str()`.
___
*[`util/common.rs` line 41 at r3](https://reviewable.io/reviews/allada/turbo-cache/174#-N_9qdBZAiwhUE5Q5dYH:-N_9qdBZAiwhUE5Q5dYI:bxs3jvi) ([raw file](https://github.com/allada/turbo-cache/blob/434d487661bc2bb88b9cb7294ad7399723b32746/util/common.rs#L41)):*
> ```XML
> 
>     /// Possibly the size of the digest in bytes. This should only be trusted
>     /// if `truest_size` is true.
> ```

nit: typo

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/174)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/common.rs` line 116 at r1](https://reviewable.io/reviews/allada/turbo-cache/174#-N_9q8ehCGtCR3tj18BL:-N_9qrVn3XRPJsD4_GTG:bvetvgu) ([raw file](https://github.com/allada/turbo-cache/blob/c8f27514b27c780e0cce3ceb0c574cea743acfee/util/common.rs#L116)):*

Re-implementing `val.str()`.
</blockquote></details>

Yep, in reviewing my own code I noticed this too.

chrisstaite-menlo (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/174)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

aaronmondal (approved):
I like this a lot. Much simpler.

Some discussion on this was at https://github.com/allada/turbo-cache/issues/153#issuecomment-1631877495

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/174#-:-N_A1cJW22jv5ANWu4Mh:bf3256c)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #177, opened 2023-07-14
Implement the WaitExecution

chrisstaite-menlo:
These functions are used by Goma.

This also resolves the issue of holding on to abandoned queued jobs.

I probably could break this down into a few more PRs, but I'd likely break something along the way.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "4f921e8c24fd74362f4fa660db203c02dfdc564ae44a23e8d4526d28aab7aafb",
+  "checksum": "5fb139361c531ebf8263b94f2cf9d4ac159c5ae3616a08fc92e6e0089a8c6c19",
   "crates": {
     "addr2line 0.20.0": {
       "name": "addr2line",
@@ -189,6 +189,42 @@
         "version": "1.0.2"
       },
       "license": "Unlicense OR MIT"
+    },
+    "allocator-api2 0.2.16": {
+      "name": "allocator-api2",
+      "version": "0.2.16",
+      "repository": {
+        "Http": {
+          "url": "https://crates.io/api/v1/crates/allocator-api2/0.2.16/download",
+          "sha256": "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
+        }
+      },
+      "targets": [
+        {
+          "Library": {
+            "crate_name": "allocator_api2",
+            "crate_root": "src/lib.rs",
+            "srcs": [
+              "**/*.rs"
+            ]
+          }
+        }
+      ],
+      "library_target_name": "allocator_api2",
+      "common_attrs": {
+        "compile_data_glob": [
+          "**"
+        ],
+        "crate_features": {
+          "common": [
+            "alloc"
+          ],
+          "selects": {}
+        },
+        "edition": "2018",
+        "version": "0.2.16"
+      },
+      "license": "MIT OR Apache-2.0"
     },
     "android-tzdata 0.1.1": {
       "name": "android-tzdata",
@@ -2489,6 +2525,10 @@
               "target": "futures"
             },
             {
+              "id": "hashbrown 0.14.0",
+              "target": "hashbrown"
+            },
+            {
               "id": "hex 0.4.3",
               "target": "hex"
             },
@@ -4109,6 +4149,58 @@
         },
         "edition": "2021",
         "version": "0.13.2"
+      },
+      "license": "MIT OR Apache-2.0"
+    },
+    "hashbrown 0.14.0": {
+      "name": "hashbrown",
+      "version": "0.14.0",
+      "repository": {
+        "Http": {
+          "url": "https://crates.io/api/v1/crates/hashbrown/0.14.0/download",
+          "sha256": "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
+        }
+      },
+      "targets": [
+        {
+          "Library": {
+            "crate_name": "hashbrown",
+            "crate_root": "src/lib.rs",
+            "srcs": [
+              "**/*.rs"
+            ]
+          }
+        }
+      ],
+      "library_target_name": "hashbrown",
+      "common_attrs": {
+        "compile_data_glob": [
+          "**"
+        ],
+        "crate_features": {
+          "common": [
+            "ahash",
+            "allocator-api2",
+            "default",
+            "inline-more"
+          ],
+          "selects": {}
+        },
+        "deps": {
+          "common": [
+            {
+              "id": "ahash 0.8.3",
+              "target": "ahash"
+            },
+            {
+              "id": "allocator-api2 0.2.16",
+              "target": "allocator_api2"
+            }
+          ],
+          "selects": {}
+        },
+        "edition": "2021",
+        "version": "0.14.0"
       },
       "license": "MIT OR Apache-2.0"
     },
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
+
+[[package]]
 name = "android-tzdata"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -482,6 +488,7 @@ dependencies = [
  "filetime",
  "fixed-buffer",
  "futures",
+ "hashbrown 0.14.0",
  "hex",
  "http",
  "hyper",
@@ -776,6 +783,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "43a3c133739dddd0d2990f9a4bdf8eb4b21ef50e4851ca85ab661199821d510e"
 dependencies = [
  "ahash",
+]
+
+[[package]]
+name = "hashbrown"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
+dependencies = [
+ "ahash",
+ "allocator-api2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ uuid = { version = "1.4.0", features = ["v4"] }
 shlex = "1.1.0"
 relative-path = "1.8.0"
 parking_lot = "0.12.1"
+hashbrown = "0.14"
 
 [dev-dependencies]
 stdext = "0.3.1"
--- a/cas/grpc_service/execution_server.rs
+++ b/cas/grpc_service/execution_server.rs
@@ -19,11 +19,12 @@ use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use futures::{Stream, StreamExt};
 use rand::{thread_rng, Rng};
+use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
 use tonic::{Request, Response, Status};
 
 use ac_utils::get_and_decode_digest;
-use action_messages::{ActionInfo, ActionInfoHashKey, DEFAULT_EXECUTION_PRIORITY};
+use action_messages::{ActionInfo, ActionInfoHashKey, ActionState, DEFAULT_EXECUTION_PRIORITY};
 use common::{log, DigestInfo};
 use config::cas_server::{ExecutionConfig, InstanceName};
 use error::{make_input_err, Error, ResultExt};
@@ -163,6 +164,14 @@ impl ExecutionServer {
         Server::new(self)
     }
 
+    fn to_execute_stream(receiver: watch::Receiver<Arc<ActionState>>) -> Response<ExecuteStream> {
+        let receiver_stream = Box::pin(WatchStream::new(receiver).map(|action_update| {
+            log::info!("\x1b[0;31mexecute Resp Stream\x1b[0m: {:?}", action_update);
+            Ok(action_update.as_ref().clone().into())
+        }));
+        tonic::Response::new(receiver_stream)
+    }
+
     async fn inner_execute(&self, request: Request<ExecuteRequest>) -> Result<Response<ExecuteStream>, Error> {
         let execute_req = request.into_inner();
         let instance_name = execute_req.instance_name;
@@ -194,11 +203,22 @@ impl ExecutionServer {
             .await
             .err_tip(|| "Failed to schedule task")?;
 
-        let receiver_stream = Box::pin(WatchStream::new(rx).map(|action_update| {
-            log::info!("\x1b[0;31mexecute Resp Stream\x1b[0m: {:?}", action_update);
-            Ok(action_update.as_ref().clone().into())
-        }));
-        Ok(tonic::Response::new(receiver_stream))
+        Ok(Self::to_execute_stream(rx))
+    }
+
+    async fn inner_wait_execution(
+        &self,
+        request: Request<WaitExecutionRequest>,
+    ) -> Result<Response<ExecuteStream>, Status> {
+        let unique_qualifier = ActionInfoHashKey::try_from(request.into_inner().name.as_str())
+            .err_tip(|| "Decoding operation name into ActionInfoHashKey")?;
+        let Some(instance_info) = self.instance_infos.get(&unique_qualifier.instance_name) else {
+            return Err(Status::not_found(format!("No scheduler with the instance name {}", unique_qualifier.instance_name)));
+        };
+        let Some(rx) = instance_info.scheduler.find_existing_action(&unique_qualifier).await else {
+            return Err(Status::not_found("Failed to find existing task"));
+        };
+        Ok(Self::to_execute_stream(rx))
     }
 }
 
@@ -224,14 +244,11 @@ impl Execution for ExecutionServer {
         resp
     }
 
-    type WaitExecutionStream = Pin<Box<dyn Stream<Item = Result<Operation, Status>> + Send + Sync + 'static>>;
-    async fn wait_execution(
-        &self,
-        request: Request<WaitExecutionRequest>,
-    ) -> Result<Response<Self::WaitExecutionStream>, Status> {
-        use stdext::function_name;
-        let output = format!("{} not yet implemented", function_name!());
-        println!("{:?}", request);
-        Err(Status::unimplemented(output))
+    type WaitExecutionStream = ExecuteStream;
+    async fn wait_execution(&self, request: Request<WaitExecutionRequest>) -> Result<Response<ExecuteStream>, Status> {
+        self.inner_wait_execution(request)
+            .await
+            .err_tip(|| "Failed on wait_execution() command")
+            .map_err(|e| e.into())
     }
 }
--- a/cas/scheduler/BUILD
+++ b/cas/scheduler/BUILD
@@ -83,6 +83,7 @@ rust_library(
         "//util:error",
         "@crate_index//:parking_lot",
         "@crate_index//:futures",
+        "@crate_index//:hashbrown",
         "@crate_index//:lru",
         "@crate_index//:tokio",
     ],
@@ -109,6 +110,7 @@ rust_library(
         "//util:common",
         "//util:error",
         "@crate_index//:futures",
+        "@crate_index//:parking_lot",
         "@crate_index//:tonic",
         "@crate_index//:tokio",
         "@crate_index//:tokio-stream",
@@ -176,6 +178,7 @@ rust_library(
         "//config",
         "//util:error",
         "@crate_index//:futures",
+        "@crate_index//:tokio",
     ],
 )
 
--- a/cas/scheduler/cache_lookup_scheduler.rs
+++ b/cas/scheduler/cache_lookup_scheduler.rs
@@ -12,26 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
 
 use async_trait::async_trait;
 use futures::stream::{FuturesUnordered, StreamExt};
+use tokio::select;
 use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
 use tonic::Request;
 
 use ac_utils::get_and_decode_digest;
-use action_messages::{ActionInfo, ActionResult, ActionStage, ActionState};
+use action_messages::{ActionInfo, ActionInfoHashKey, ActionResult, ActionStage, ActionState};
 use common::DigestInfo;
 use error::Error;
 use grpc_store::GrpcStore;
+use parking_lot::{Mutex, MutexGuard};
 use platform_property_manager::PlatformPropertyManager;
 use proto::build::bazel::remote::execution::v2::{
     ActionResult as ProtoActionResult, FindMissingBlobsRequest, GetActionResultRequest,
 };
 use scheduler::ActionScheduler;
 use store::Store;
+
+/// Actions that are having their cache checked or failed cache lookup and are
+/// being forwarded upstream.  Missing the skip_cache_check actions which are
+/// forwarded directly.
+type CheckActions = HashMap<ActionInfoHashKey, Arc<watch::Sender<Arc<ActionState>>>>;
 
 pub struct CacheLookupScheduler {
     /// A reference to the CAS which is used to validate all the outputs of a
@@ -42,6 +50,8 @@ pub struct CacheLookupScheduler {
     /// The "real" scheduler to use to perform actions if they were not found
     /// in the action cache.
     action_scheduler: Arc<dyn ActionScheduler>,
+    /// Actions that are currently performing a CacheCheck.
+    cache_check_actions: Arc<Mutex<CheckActions>>,
 }
 
 async fn get_action_from_store(
@@ -112,6 +122,21 @@ async fn validate_outputs_exist(
     }
 }
 
+fn subscribe_to_existing_action(
+    cache_check_actions: &MutexGuard<CheckActions>,
+    unique_qualifier: &ActionInfoHashKey,
+) -> Option<watch::Receiver<Arc<ActionState>>> {
+    cache_check_actions.get(unique_qualifier).map(|tx| {
+        let current_value = tx.borrow();
+        // Subscribe marks the current value as seen, so we have to
+        // re-send it to all receivers.
+        // TODO: Fix this when fixed upstream tokio-rs/tokio#5871
+        let rx = tx.subscribe();
+        let _ = tx.send(current_value.clone());
+        rx
+    })
+}
+
 impl CacheLookupScheduler {
     pub fn new(
         cas_store: Arc<dyn Store>,
@@ -122,6 +147,7 @@ impl CacheLookupScheduler {
             cas_store,
             ac_store,
             action_scheduler,
+            cache_check_actions: Default::default(),
         })
     }
 }
@@ -142,11 +168,22 @@ impl ActionScheduler for CacheLookupScheduler {
             stage: ActionStage::CacheCheck,
         });
         let (tx, rx) = watch::channel(current_state.clone());
+        let tx = Arc::new(tx);
+        {
+            let mut cache_check_actions = self.cache_check_actions.lock();
+            // Check this isn't a duplicate request first.
+            if let Some(rx) = subscribe_to_existing_action(&cache_check_actions, &action_info.unique_qualifier) {
+                return Ok(rx);
+            }
+            cache_check_actions.insert(action_info.unique_qualifier.clone(), tx.clone());
+        }
         let ac_store = self.ac_store.clone();
         let cas_store = self.cas_store.clone();
         let action_scheduler = self.action_scheduler.clone();
+        let cache_check_actions = self.cache_check_actions.clone();
         tokio::spawn(async move {
             let instance_name = action_info.instance_name().clone();
+            let unique_qualifier = action_info.unique_qualifier.clone();
             if let Some(proto_action_result) =
                 get_action_from_store(ac_store, current_state.action_digest(), instance_name.clone()).await
             {
@@ -154,24 +191,50 @@ impl ActionScheduler for CacheLookupScheduler {
                     // Found in the cache, return the result immediately.
                     Arc::make_mut(&mut current_state).stage = ActionStage::CompletedFromCache(proto_action_result);
                     let _ = tx.send(current_state);
+                    cache_check_actions.lock().remove(&unique_qualifier);
                     return;
                 }
             }
             // Not in cache, forward to upstream and proxy state.
-            let mut watch_stream = match action_scheduler.add_action(action_info).await {
-                Ok(rx) => WatchStream::new(rx),
+            match action_scheduler.add_action(action_info).await {
+                Ok(rx) => {
+                    let mut watch_stream = WatchStream::new(rx);
+                    loop {
+                        select!(
+                            Some(action_state) = watch_stream.next() => {
+                                if tx.send(action_state).is_err() {
+                                    break;
+                                }
+                            }
+                            _ = tx.closed() => {
+                                break;
+                            }
+                        )
+                    }
+                }
                 Err(err) => {
                     Arc::make_mut(&mut current_state).stage = ActionStage::Error((err, ActionResult::default()));
                     let _ = tx.send(current_state);
-                    return;
-                }
-            };
-            while let Some(action_state) = watch_stream.next().await {
-                if tx.send(action_state).is_err() {
-                    break;
                 }
             }
+            cache_check_actions.lock().remove(&unique_qualifier);
         });
         Ok(rx)
     }
+
+    async fn find_existing_action(
+        &self,
+        unique_qualifier: &ActionInfoHashKey,
+    ) -> Option<watch::Receiver<Arc<ActionState>>> {
+        {
+            let cache_check_actions = self.cache_check_actions.lock();
+            if let Some(rx) = subscribe_to_existing_action(&cache_check_actions, unique_qualifier) {
+                return Some(rx);
+            }
+        }
+        // Cache skipped may be in the upstream scheduler.
+        self.action_scheduler.find_existing_action(unique_qualifier).await
+    }
+
+    async fn clean_recently_completed_actions(&self) {}
 }
--- a/cas/scheduler/default_scheduler_factory.rs
+++ b/cas/scheduler/default_scheduler_factory.rs
@@ -14,8 +14,10 @@
 
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 
 use futures::Future;
+use tokio::time::interval;
 
 use cache_lookup_scheduler::CacheLookupScheduler;
 use config::schedulers::SchedulerConfig;
@@ -37,7 +39,7 @@ pub fn scheduler_factory<'a>(
                 let scheduler = Arc::new(SimpleScheduler::new(config));
                 (Some(scheduler.clone()), Some(scheduler))
             }
-            SchedulerConfig::grpc(config) => (Some(Arc::new(GrpcScheduler::new(config).await?)), None),
+            SchedulerConfig::grpc(config) => (Some(Arc::new(GrpcScheduler::new(config)?)), None),
             SchedulerConfig::cache_lookup(config) => {
                 let cas_store = store_manager
                     .get_store(&config.cas_store)
@@ -54,6 +56,26 @@ pub fn scheduler_factory<'a>(
                 (Some(cache_lookup_scheduler), worker_scheduler)
             }
         };
+
+        if let Some(action_scheduler) = &scheduler.0 {
+            start_cleanup_timer(action_scheduler);
+        }
+
         Ok(scheduler)
     })
+}
+
+fn start_cleanup_timer(action_scheduler: &Arc<dyn ActionScheduler>) {
+    let weak_scheduler = Arc::downgrade(action_scheduler);
+    tokio::spawn(async move {
+        let mut ticker = interval(Duration::from_secs(1));
+        loop {
+            ticker.tick().await;
+            match weak_scheduler.upgrade() {
+                Some(scheduler) => scheduler.clean_recently_completed_actions().await,
+                // If we fail to upgrade, our service is probably destroyed, so return.
+                None => return,
+            }
+        }
+    });
 }
--- a/cas/scheduler/grpc_scheduler.rs
+++ b/cas/scheduler/grpc_scheduler.rs
@@ -17,36 +17,72 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use parking_lot::Mutex;
+use tokio::select;
 use tokio::sync::watch;
-use tonic::{transport, Request};
+use tonic::{transport, Request, Streaming};
 
-use action_messages::{ActionInfo, ActionState, DEFAULT_EXECUTION_PRIORITY};
+use action_messages::{ActionInfo, ActionInfoHashKey, ActionState, DEFAULT_EXECUTION_PRIORITY};
 use common::log;
 use error::{make_err, Code, Error, ResultExt};
 use platform_property_manager::PlatformPropertyManager;
 use proto::build::bazel::remote::execution::v2::{
     capabilities_client::CapabilitiesClient, execution_client::ExecutionClient, ExecuteRequest, ExecutionPolicy,
-    GetCapabilitiesRequest,
+    GetCapabilitiesRequest, WaitExecutionRequest,
 };
+use proto::google::longrunning::Operation;
 use scheduler::ActionScheduler;
 
 pub struct GrpcScheduler {
-    endpoint: transport::Channel,
+    capabilities_client: CapabilitiesClient<transport::Channel>,
+    execution_client: ExecutionClient<transport::Channel>,
     platform_property_managers: Mutex<HashMap<String, Arc<PlatformPropertyManager>>>,
 }
 
 impl GrpcScheduler {
-    pub async fn new(config: &config::schedulers::GrpcScheduler) -> Result<Self, Error> {
-        let endpoint = transport::Endpoint::new(config.endpoint.clone())
-            .err_tip(|| format!("Could not parse {} in GrpcScheduler", config.endpoint))?
-            .connect()
-            .await
-            .err_tip(|| format!("Could not connect to {} in GrpcScheduler", config.endpoint))?;
+    pub fn new(config: &config::schedulers::GrpcScheduler) -> Result<Self, Error> {
+        let endpoint = transport::Channel::balance_list(std::iter::once(
+            transport::Endpoint::new(config.endpoint.clone())
+                .err_tip(|| format!("Could not parse {} in GrpcScheduler", config.endpoint))?,
+        ));
 
         Ok(Self {
-            endpoint,
+            capabilities_client: CapabilitiesClient::new(endpoint.clone()),
+            execution_client: ExecutionClient::new(endpoint),
             platform_property_managers: Mutex::new(HashMap::new()),
         })
+    }
+
+    async fn stream_state(mut result_stream: Streaming<Operation>) -> Result<watch::Receiver<Arc<ActionState>>, Error> {
+        if let Some(initial_response) = result_stream
+            .message()
+            .await
+            .err_tip(|| "Recieving response from upstream scheduler")?
+        {
+            let (tx, rx) = watch::channel(Arc::new(initial_response.try_into()?));
+            tokio::spawn(async move {
+                loop {
+                    select!(
+                        _ = tx.closed() => {
+                            log::info!("Client disconnected in GrpcScheduler");
+                            return;
+                        }
+                        Ok(Some(response)) = result_stream.message() => {
+                            match response.try_into() {
+                                Ok(response) => {
+                                    if let Err(err) = tx.send(Arc::new(response)) {
+                                        log::info!("Client disconnected in GrpcScheduler: {}", err);
+                                        return;
+                                    }
+                                }
+                                Err(err) => log::error!("Error converting response to ActionState in GrpcScheduler: {}", err),
+                            }
+                        }
+                    )
+                }
+            });
+            return Ok(rx);
+        }
+        Err(make_err!(Code::Internal, "Upstream scheduler didn't accept action."))
     }
 }
 
@@ -58,8 +94,9 @@ impl ActionScheduler for GrpcScheduler {
         }
 
         // Not in the cache, lookup the capabilities with the upstream.
-        let mut capabilities_client = CapabilitiesClient::new(self.endpoint.clone());
-        let capabilities = capabilities_client
+        let capabilities = self
+            .capabilities_client
+            .clone()
             .get_capabilities(GetCapabilitiesRequest {
                 instance_name: instance_name.to_string(),
             })
@@ -97,32 +134,34 @@ impl ActionScheduler for GrpcScheduler {
             // TODO: Get me from the original request, not very important as we ignore it.
             results_cache_policy: None,
         };
-        let mut result_stream = ExecutionClient::new(self.endpoint.clone())
+        let result_stream = self
+            .execution_client
+            .clone()
             .execute(Request::new(request))
             .await
             .err_tip(|| "Sending action to upstream scheduler")?
             .into_inner();
-        if let Some(initial_response) = result_stream
-            .message()
-            .await
-            .err_tip(|| "Recieving response from upstream scheduler")?
-        {
-            let (tx, rx) = watch::channel(Arc::new(initial_response.try_into()?));
-            tokio::spawn(async move {
-                while let Ok(Some(response)) = result_stream.message().await {
-                    match response.try_into() {
-                        Ok(response) => {
-                            if let Err(err) = tx.send(Arc::new(response)) {
-                                log::info!("Client disconnected in GrpcScheduler: {}", err);
-                                return;
-                            }
-                        }
-                        Err(err) => log::error!("Error converting response to watch in GrpcScheduler: {}", err),
-                    }
-                }
-            });
-            return Ok(rx);
-        }
-        Err(make_err!(Code::Internal, "Upstream scheduler didn't accept action."))
+        Self::stream_state(result_stream).await
     }
+
+    async fn find_existing_action(
+        &self,
+        unique_qualifier: &ActionInfoHashKey,
+    ) -> Option<watch::Receiver<Arc<ActionState>>> {
+        let request = WaitExecutionRequest {
+            name: unique_qualifier.action_name(),
+        };
+        let result_stream = self
+            .execution_client
+            .clone()
+            .wait_execution(Request::new(request))
+            .await;
+        if let Err(err) = result_stream {
+            log::info!("Error response looking up action with upstream scheduler: {}", err);
+            return None;
+        }
+        Self::stream_state(result_stream.unwrap().into_inner()).await.ok()
+    }
+
+    async fn clean_recently_completed_actions(&self) {}
 }
--- a/cas/scheduler/scheduler.rs
+++ b/cas/scheduler/scheduler.rs
@@ -31,6 +31,15 @@ pub trait ActionScheduler: Sync + Send + Unpin {
 
     /// Adds an action to the scheduler for remote execution.
     async fn add_action(&self, action_info: ActionInfo) -> Result<watch::Receiver<Arc<ActionState>>, Error>;
+
+    /// Find an existing action by its name.
+    async fn find_existing_action(
+        &self,
+        unique_qualifier: &ActionInfoHashKey,
+    ) -> Option<watch::Receiver<Arc<ActionState>>>;
+
+    /// Cleans up the cache of recently completed actions.
+    async fn clean_recently_completed_actions(&self);
 }
 
 /// WorkerScheduler interface is responsible for interactions between the scheduler
--- a/cas/scheduler/simple_scheduler.rs
+++ b/cas/scheduler/simple_scheduler.rs
@@ -12,12 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Borrow;
 use std::cmp;
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::BTreeMap;
+use std::hash::{Hash, Hasher};
 use std::sync::Arc;
+use std::time::SystemTime;
 
 use async_trait::async_trait;
 use futures::Future;
+use hashbrown::{HashMap, HashSet};
 use lru::LruCache;
 use parking_lot::Mutex;
 use tokio::sync::{watch, Notify};
@@ -34,6 +38,10 @@ use worker::{Worker, WorkerId, WorkerTimestamp, WorkerUpdate};
 /// Default timeout for workers in seconds.
 /// If this changes, remember to change the documentation in the config.
 const DEFAULT_WORKER_TIMEOUT_S: u64 = 5;
+
+/// Default timeout for recently completed actions in seconds.
+/// If this changes, remember to change the documentation in the config.
+const DEFAULT_RETAIN_COMPLETED_FOR_S: u64 = 60;
 
 /// Default times a job can retry before failing.
 /// If this changes, remember to change the documentation in the config.
@@ -144,6 +152,32 @@ impl Workers {
     }
 }
 
+struct CompletedAction {
+    completed_time: SystemTime,
+    state: Arc<ActionState>,
+}
+
+impl Hash for CompletedAction {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        ActionInfoHashKey::hash(&self.state.unique_qualifier, state);
+    }
+}
+
+impl PartialEq for CompletedAction {
+    fn eq(&self, other: &Self) -> bool {
+        ActionInfoHashKey::eq(&self.state.unique_qualifier, &other.state.unique_qualifier)
+    }
+}
+
+impl Eq for CompletedAction {}
+
+impl Borrow<ActionInfoHashKey> for CompletedAction {
+    #[inline]
+    fn borrow(&self) -> &ActionInfoHashKey {
+        &self.state.unique_qualifier
+    }
+}
+
 struct SimpleSchedulerImpl {
     // BTreeMap uses `cmp` to do it's comparisons, this is a problem because we want to sort our
     // actions based on priority and insert timestamp but also want to find and join new actions
@@ -159,6 +193,13 @@ struct SimpleSchedulerImpl {
     queued_actions: BTreeMap<Arc<ActionInfo>, AwaitedAction>,
     workers: Workers,
     active_actions: HashMap<Arc<ActionInfo>, RunningAction>,
+    // These actions completed recently but had no listener, they might have
+    // completed while the caller was thinking about calling wait_execution, so
+    // keep their completion state around for a while to send back.
+    // TODO(#192) Revisit if this is the best way to handle recently completed actions.
+    recently_completed_actions: HashSet<CompletedAction>,
+    /// The duration that actions are kept in recently_completed_actions for.
+    retain_completed_for: Duration,
     /// Timeout of how long to evict workers if no response in this given amount of time in seconds.
     worker_timeout_s: u64,
     /// Default times a job can retry before failing.
@@ -168,6 +209,16 @@ struct SimpleSchedulerImpl {
 }
 
 impl SimpleSchedulerImpl {
+    fn subscribe_to_channel(awaited_action: &AwaitedAction) -> watch::Receiver<Arc<ActionState>> {
+        let rx = awaited_action.notify_channel.subscribe();
+        // TODO: Fix this when fixed upstream tokio-rs/tokio#5871
+        awaited_action
+            .notify_channel
+            .send(awaited_action.current_state.clone())
+            .unwrap();
+        rx
+    }
+
     /// Attempts to find a worker to execute an action and begins executing it.
     /// If an action is already running that is cacheable it may merge this action
     /// with the results and state changes of the already running action.
@@ -177,13 +228,7 @@ impl SimpleSchedulerImpl {
     fn add_action(&mut self, action_info: ActionInfo) -> Result<watch::Receiver<Arc<ActionState>>, Error> {
         // Check to see if the action is running, if it is and cacheable, merge the actions.
         if let Some(running_action) = self.active_actions.get_mut(&action_info) {
-            let rx = running_action.action.notify_channel.subscribe();
-            running_action
-                .action
-                .notify_channel
-                .send(running_action.action.current_state.clone())
-                .unwrap();
-            return Ok(rx);
+            return Ok(Self::subscribe_to_channel(&running_action.action));
         }
 
         // Check to see if the action is queued, if it is and cacheable, merge the actions.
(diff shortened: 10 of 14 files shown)


chrisstaite-menlo:
*[`cas/grpc_service/execution_server.rs` line 216 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KDlWCAgyn33vz2bCl:-N_KDlWCAgyn33vz2bCm:b-amfk3) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L216)):*
> ```XML
>     }
> 
>     async fn lookup_action<'a, Fut, F, G, R, S>(&'a self, operation: F, mut filter: G) -> Result<S, Status>
> ```

If you think it's better to copy this code into the below two methods rather than this abomination of generics, let me know.  However, I kinda just wanted to do it to see if I could...

allada (changes requested):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 166 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KEYeq50_o7KdBrFUv:-N_KEYeq50_o7KdBrFUw:b-g3kdhq) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L166)):*
> ```XML
>     }
> 
>     pub fn into_services(self) -> (Server<ExecutionServer>, OperationsServer<ExecutionServer>) {
> ```

At first I was thinking, this should be a separate service and not share with ExecutionServer. Then I was like... "Nah, it makes sense to put them together"... Then I started reading the API and now I think they should be separate.

ExecutionServer only serves execution requests. OperationServer is designed to serve other kinds of operations too like "WaitForWorkerToComeOnline" or "DrainAndShutdown".

Because of this and unlikely, but possible additional features, lets break this into it's own `grpc_service`. All our operations will happen on the scheduler anyway, so interacting should be easy, since the scheduler can be referenced in both services.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_KGyWXAwpch7ThQKN-:b-cs7e8b) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*
> ```XML
>         // Very innefficient having to search through all instances, however
>         // the protobuf spec doesn't state that the instance_name can be
>         // retrieved from the path.
> ```

Actually it can be retrieved from the path if it matches as outlined here:
https://github.com/allada/turbo-cache/blob/master/proto/build/bazel/remote/execution/v2/remote_execution.proto#L124

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_KHW6pA2-OBtUJKBTD:b-l1m29g) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*

Actually it can be retrieved from the path if it matches as outlined here:
https://github.com/allada/turbo-cache/blob/master/proto/build/bazel/remote/execution/v2/remote_execution.proto#L124
</blockquote></details>

Wait, no... We choose the name, so we choose the filter. In our case we can require any kind of filter format we want.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 216 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KoJsHEUR3iJzEWj4j:-N_KoJsHEUR3iJzEWj4k:bdzn1z8) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L216)):*
> ```XML
>     }
> 
>     async fn lookup_action<'a, Fut, F, G, R, S>(&'a self, operation: F, mut filter: G) -> Result<S, Status>
> ```

nit: Lets rename `filter_actions()` and add a `limit` argument. This will make it trivial to implement `ListOperations()` in the future.
___
*[`cas/grpc_service/execution_server.rs` line 225 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KoklO7J1zto8vFX9x:-N_KoklO7J1zto8vFX9y:btgn8pg) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L225)):*
> ```XML
>         // the protobuf spec doesn't state that the instance_name can be
>         // retrieved from the path.
>         self.instance_infos
> ```

nit: This is super confusing, the `Stream` api makes it hard to understand what is happening, can we expand this into a `while let Some(item) = stream.next().await {` ?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/177#-N_L7BVy4y8ebVUUbCYm:-N_L7BVy4y8ebVUUbCYn:b-s79yb2) (no related file):*

Give me some time on this, I'm going to hack up your branch to get a better feel for it. There's a lot going on and I want to see how it goes.

Right now I'm leaning towards the operation service being completely separate from the execution service. Requiring a configuration and all.

___

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 166 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KEYeq50_o7KdBrFUv:-N_MsGgK8y9WDL7mD5v_:b-2p8ixr) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L166)):*

At first I was thinking, this should be a separate service and not share with ExecutionServer. Then I was like... "Nah, it makes sense to put them together"... Then I started reading the API and now I think they should be separate.

ExecutionServer only serves execution requests. OperationServer is designed to serve other kinds of operations too like "WaitForWorkerToComeOnline" or "DrainAndShutdown".

Because of this and unlikely, but possible additional features, lets break this into it's own `grpc_service`. All our operations will happen on the scheduler anyway, so interacting should be easy, since the scheduler can be referenced in both services.
</blockquote></details>

Disregard. I experimented with this idea for a few hours and came to the conclusion we should not try to recycle `Operation` for anything but BRE required stuff. It is a decent interface, but tonic doesn't really support HTTP1 + json endpoints yet, it's close, but I did some testing with tonic-web and it didn't really work for what I wanted, since it just encodes everything to a proto and requires input as a proto.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 7 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/cas_main.rs` line 217 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MsqTf10c0-4lLZBKO:-N_MsqTf10c0-4lLZBKP:ba6io91) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/cas_main.rs#L217)):*
> ```XML
>                     .map_or(Ok(None), |cfg| {
>                         ExecutionServer::new(&cfg, &action_schedulers, &store_manager).map(|v| {
>                             let (mut execution_service, mut operations_service) = v.into_services();
> ```

I tried really really hard to make this interface not suck. Well I couldn't figure out a way to abstract this out without making a new config, which is pointless for us.
___
*[`cas/grpc_service/execution_server.rs` line 241 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MujwQ9kzEFBhq8Kp3:-N_MujwQ9kzEFBhq8Kp4:b7lyvoz) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L241)):*
> ```XML
>         let name = request.into_inner().name;
>         self.lookup_action(
>             |instance_info| instance_info.scheduler.find_existing_action(&name),
> ```

Since this file is the gateway to all things execution related, lets rewrite all output `Operation`s so it has the instance name as prefixed to the operation name. Then just lookup what scheduler it should be on here.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 8 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 237 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MwNyr4HQ8dVbN4GaZ:-N_MwNyr4HQ8dVbN4Ga_:b-d5wmm9) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L237)):*
> ```XML
>     async fn inner_wait_execution(
>         &self,
>         request: Request<WaitExecutionRequest>,
> ```

nit: Lets just pass in name as a string here and then this function is 100% compatible with: `wait_operation()`.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 8 files reviewed, 9 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 308 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MwaUY8wmhDBsNpR8d:-N_MwaUY8wmhDBsNpR8e:b-w2q467) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L308)):*
> ```XML
>     }
> 
>     async fn cancel_operation(&self, request: Request<CancelOperationRequest>) -> Result<Response<()>, Status> {
> ```

I'm really struggling on this one. Do we need/want to implement `cancel_operation`? I understand removing `delete_operation`, since it allows the scheduler to stop the action if it wants. But from what I can tell in the documentation, `cancel_operation` pretty much says: "Kill it, don't ask questions" and gives little flexibility.

The reason I'm worried here is because the scheduler multiplexes the actions. Meaning if two requests from two different clients come in for the same action, it will only execute the action one time and send the results to both clients. If we implement `cancel_action` we are allowing any client to terminate another client's action.

If the argument is made to say that's an implementation detail and the client should not be able to see other client's jobs and it should only terminate the job if no-one else is waiting on that job, I feel that's more of what `DeleteOperation` is for. But then again `DeleteOperation` might be more about actions that change external state (like upload files, install new software, exc...), which is why the client would not care about the results.

chrisstaite-menlo:


Give me some time on this, I'm going to hack up your branch to get a better feel for it. There's a lot going on and I want to see how it goes.

Right now I'm leaning towards the operation service being completely separate from the execution service. Requiring a configuration and all.
</blockquote></details>

I think I'm going to drop the operations service.  It was mostly for the use case that someone changes a header that's used in a lot of files and it's bad, so cancels the build job but we still execute a bunch of jobs that are going to fail.  However, I don't think it really gains us all that much, so I think I'm going to drop it and we can bring it in a separate branch later if we want to.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @allada)

___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_OPSXd6_-lX0DmRLSx:b-3ktop1) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*

Wait, no... We choose the name, so we choose the filter. In our case we can require any kind of filter format we want.
</blockquote></details>

Think that falls down with the GrpcScheduler, right?
___
*[`cas/grpc_service/execution_server.rs` line 237 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MwNyr4HQ8dVbN4GaZ:-N_OQCUPCnok4proxL2Z:b-9jrwbo) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L237)):*

nit: Lets just pass in name as a string here and then this function is 100% compatible with: `wait_operation()`.
</blockquote></details>

Got rid of operation server, so probably not worth doing.
___
*[`cas/grpc_service/execution_server.rs` line 241 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MujwQ9kzEFBhq8Kp3:-N_OQJ8gCEW2TBxben4y:biiifc3) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L241)):*

Since this file is the gateway to all things execution related, lets rewrite all output `Operation`s so it has the instance name as prefixed to the operation name. Then just lookup what scheduler it should be on here.
</blockquote></details>

Got rid of the operations server...
___
*[`cas/grpc_service/execution_server.rs` line 308 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_MwaUY8wmhDBsNpR8d:-N_OQMXc50VNCr-KrN2K:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L308)):*

I'm really struggling on this one. Do we need/want to implement `cancel_operation`? I understand removing `delete_operation`, since it allows the scheduler to stop the action if it wants. But from what I can tell in the documentation, `cancel_operation` pretty much says: "Kill it, don't ask questions" and gives little flexibility.

The reason I'm worried here is because the scheduler multiplexes the actions. Meaning if two requests from two different clients come in for the same action, it will only execute the action one time and send the results to both clients. If we implement `cancel_action` we are allowing any client to terminate another client's action.

If the argument is made to say that's an implementation detail and the client should not be able to see other client's jobs and it should only terminate the job if no-one else is waiting on that job, I feel that's more of what `DeleteOperation` is for. But then again `DeleteOperation` might be more about actions that change external state (like upload files, install new software, exc...), which is why the client would not care about the results.
</blockquote></details>

Done.

allada (changes requested):
> Reviewed 1 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 9 files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`-- commits` line 1 at r4](https://reviewable.io/reviews/allada/turbo-cache/177#-N_Ok46bBqDJO55yU_t2):*
I think something went wrong in your rebase.

chrisstaite-menlo:
*[`-- commits` line 1 at r4](https://reviewable.io/reviews/allada/turbo-cache/177#-N_Ok46bBqDJO55yU_t2):*

I think something went wrong in your rebase.
</blockquote></details>

Whoops....

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/177#-N_L7BVy4y8ebVUUbCYm:-N_QLucEDlzUIu_YD7X0:b-i7yh46) (no related file):*

I think I'm going to drop the operations service.  It was mostly for the use case that someone changes a header that's used in a lot of files and it's bad, so cancels the build job but we still execute a bunch of jobs that are going to fail.  However, I don't think it really gains us all that much, so I think I'm going to drop it and we can bring it in a separate branch later if we want to.
</blockquote></details>

Yeah actually this brings up a feature I was planning on doing but haven't yet... A config that times out jobs. This is common for these kinds of services, where after some config, lets say 60 seconds after all clients disconnect from a job it will terminate the job.

To do this we'll need to wrap `add_action()`'s `watch::Receiver<Arc<ActionState>`  with a wrapper that if dropped will notify the scheduler that a client disconnected, so it can start a counter if it was the last one to connect.

___
___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_Ol5Az36RN7dk2tpAg:bh79oki) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*

Think that falls down with the GrpcScheduler, right?
</blockquote></details>

I don't think so. If this is a problem in `GrpcScheduler` it'll just need to do an `instance_name` -> `instance_name` mapping.

We need to make the name of the operation contain the instance name. Something like:

```
{intance_name}/{action_info_hash}
```

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_Qa6eh66EbXRIrRO9-:b5zeitx) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*

I don't think so. If this is a problem in `GrpcScheduler` it'll just need to do an `instance_name` -> `instance_name` mapping.

We need to make the name of the operation contain the instance name. Something like:

```
{intance_name}/{action_info_hash}
```
</blockquote></details>

I think #186 will solve much of this.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 21 files reviewed, 2 unresolved discussions (waiting on @allada)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/177#-N_L7BVy4y8ebVUUbCYm:-N_U4i3O0TUD5r_NIEv1:b-fwqmkd) (no related file):*

Yeah actually this brings up a feature I was planning on doing but haven't yet... A config that times out jobs. This is common for these kinds of services, where after some config, lets say 60 seconds after all clients disconnect from a job it will terminate the job.

To do this we'll need to wrap `add_action()`'s `watch::Receiver<Arc<ActionState>`  with a wrapper that if dropped will notify the scheduler that a client disconnected, so it can start a counter if it was the last one to connect.
</blockquote></details>

It will also need to be cancelled on a new `subscribe()`.

___
___
*[`cas/grpc_service/execution_server.rs` line 224 at r3](https://reviewable.io/reviews/allada/turbo-cache/177#-N_KGyWWAyv7wXz32lmn:-N_U4z4o8KkO2A8xKnWJ:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/c91aaa584683906657600811fb99e3257ed00101/cas/grpc_service/execution_server.rs#L224)):*

I think #186 will solve much of this.
</blockquote></details>

Done.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 21 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/177#-N_L7BVy4y8ebVUUbCYm:-N_U5lNAAyNbd1KRdnT0:b9m4nye) (no related file):*

It will also need to be cancelled on a new `subscribe()`.
</blockquote></details>

Does it? If we ever get to a point where we a user wants to `subscribe()` and the task happens to be running. We always feed the results of the running job.

___

chrisstaite-menlo:

Does it? If we ever get to a point where we a user wants to `subscribe()` and the task happens to be running. We always feed the results of the running job.
</blockquote></details>

This situation:
User starts job and disconnects
Timer started as no connections
User uses WaitExecution to subscribe
Timer fires and kills job being listened to

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 21 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/177#-N_L7BVy4y8ebVUUbCYm:-N_U7sCA8_SHo1KKjbZa:br5semq) (no related file):*

This situation:
User starts job and disconnects
Timer started as no connections
User uses WaitExecution to subscribe
Timer fires and kills job being listened to
</blockquote></details>

Oh, you mean timer `canceled` on new `subscribe()`... I thought you mean job cancelled on new `subscribe()`. Yeah of course!

___

chrisstaite (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 21 files reviewed, 3 unresolved discussions (waiting on @allada and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 218 at r7](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UAaUaD143MwkpqvaM:-N_UAaUaD143MwkpqvaN:bmzh769) ([raw file](https://github.com/allada/turbo-cache/blob/a675e662f8f6a281c7618252c162f1f27b36761b/cas/grpc_service/execution_server.rs#L218)):*
> ```XML
>         // the protobuf spec doesn't state that the instance_name can be
>         // retrieved from the path.
>         let mut result_stream = self
> ```

Still needs to get instance_name from ActionInfoHashKey

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 1 of 21 files reviewed, 3 unresolved discussions (waiting on @allada, @chrisstaite, and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 218 at r7](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UAaUaD143MwkpqvaM:-N_UT32Z8QuZxdsY7T4J:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/a675e662f8f6a281c7618252c162f1f27b36761b/cas/grpc_service/execution_server.rs#L218)):*
<details><summary><i>Previously, chrisstaite (Chris) wrote…</i></summary><blockquote>

Still needs to get instance_name from ActionInfoHashKey
</blockquote></details>

Done.

chrisstaite-menlo:
Tested this branch with Goma and it appears to be working.  Still doing the annoying thing where it forgets all the blobs exist on restart (could be because I don't have a Redis running for Goma).

allada (changes requested):
> Reviewed 4 of 8 files at r7, 12 of 12 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 17 of 21 files reviewed, 10 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 213 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UYODg3yCLWa4cb8Hi:-N_UYODg3yCLWa4cb8Hj:br91r24) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L213)):*
> ```XML
>         request: Request<WaitExecutionRequest>,
>     ) -> Result<Response<ExecuteStream>, Status> {
>         let unique_qualifier = ActionInfoHashKey::try_from(request.into_inner().name.as_str())
> ```

nit:  `&request.into_inner().name` looks better.
___
*[`cas/grpc_service/execution_server.rs` line 215 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UYTvV5ouekURrp-iS:-N_UYTvV5ouekURrp-iT:bk6i5fz) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L215)):*
> ```XML
>         let unique_qualifier = ActionInfoHashKey::try_from(request.into_inner().name.as_str())
>             .err_tip(|| "Decoding operation name into ActionInfoHashKey")?;
>         let Some(instance_info) = self.instance_infos.get(unique_qualifier.instance_name.as_str()) else {
> ```

nit: `&unique_qualifier.instance_name` looks better.
___
*[`cas/grpc_service/execution_server.rs` line 219 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UZDcjCStLmxJZG5Cn:-N_UZDcjCStLmxJZG5Co:bhzfbok) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L219)):*
> ```XML
>         };
>         let Some(rx) = instance_info.scheduler.find_existing_action(&unique_qualifier).await else {
>             return Err(Status::not_found("Failed to find existing task"));
> ```

I wounder if it'd be a good idea to then try and find the item in the action cache before rejecting it?

I don't think the spec says anything about this, but I think it makes sense. From looking at the code, we'd need to modify the scheduler so before it fully drops `ActionResult`, it'll upload it to the CAS (not AC).

No need to do this now, but if you don't want to do it, lets make a TODO. The cool part is that it'll allow users to run an action, go away, and come back later to get the results (assuming the timeout is properly permissive).
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 158 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_Umoj-A6pz_Z9RZanY:-N_UgC7tCRuZh25XjvjI:b-55cxg) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L158)):*
> ```XML
>         self.cache_check_actions
>             .lock()
>             .insert(action_info.unique_qualifier.clone(), tx.clone());
> ```

There's a bit of a race condition here because we can have two clients at this point trying add the same `unique_qualifier`. This would cause some very bizarre things to happen. We should do a cache lookup here first and feed the existing item if it exists within the same lock.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 181 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UkDCT56V_YznElJCD:-N_UkDCT56V_YznElJCE:b8xymhj) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L181)):*
> ```XML
>                 Ok(rx) => {
>                     let mut watch_stream = WatchStream::new(rx);
>                     while let Some(action_state) = watch_stream.next().await {
> ```

Hmmm, minor bug here, where we only notify the downstream scheduler that the client has disconnected when the worker sends a notification to the client and `tx.send(action_state).is_err()` triggers an error and drops the `Sender`, which then causes the downstream scheduler to now be able to know the client disconnected.

This currently is not a big problem, since it only causes a very minor amount of additional needless memory, but if/when we implement the ability for the scheduler to timeout tasks because no listeners are active it'll become an issue.

This can be fixed by using a `select!{}` over `tx.closed()` and `watch_stream.next()`. Not sure if we should do this in this PR, but wanted to make sure it was known.

This is a good example on why I think we should try and avoid using spawns as much as possible, granted here we need to use a spawn (unless we change function signatures), this would auto-propagate everything down because the future would get dropped which would call `Drop` on everything downstream.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 208 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UuxH9COBDUbY2jZBx:-N_UuxH9COBDUbY2jZBy:bdu2z6j) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L208)):*
> ```XML
>                 // re-send it to all receivers.
>                 let rx = tx.subscribe();
>                 let _ = tx.send(current_value.clone());
> ```

I created an upstream ticket for this:
https://github.com/tokio-rs/tokio/issues/5871

Can we make a TODO referencing this ticket?
___
*[`cas/scheduler/simple_scheduler.rs` line 166 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UvdbR6if15-0rySV9:-N_UvdbR6if15-0rySVA:b-q4kpie) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L166)):*
> ```XML
>     // completed while the caller was thinking about calling wait_execution, so
>     // keep their completion state around for a while to send back.
>     recently_completed_actions: Vec<Arc<ActionState>>,
> ```

Can we do this part of the change in a followup PR? I want to think about this one a bit because it plays into both debugging and caching. Also, I'm not sure if the scheduler should do this here or if maybe some kind of logging service might be better to solve this for the debugging case.

We also can probably solve the non-debug use case by doing what is commented on above (store the `ActionResult` in CAS). I am pretty sure buildbarn does this.
___
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_UwraG2CZAq3eAOkoY:b6dyo64) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*
> ```XML
>     // Gets an existing action by its name, very noddy implementation that just searches for it everywhere.
>     fn find_existing_action(&self, unique_qualifier: &ActionInfoHashKey) -> Option<watch::Receiver<Arc<ActionState>>> {
>         self.queued_actions
> ```

Uug, these values area already indexed by the `ActionInfoHashKey`, but rust's `HashMap` does not support getting it by this value without: https://github.com/rust-lang/rust/issues/56167

It does look like we can use `hashbrown` though to solve it, because it uses `Equivalent<K>` for lookups instead of forcing `K: Borrow<Q>`, but untested:
https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.get

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 17 of 21 files reviewed, 11 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/grpc_scheduler.rs` line 42 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V-vC171kjd2vfYxBb:-N_V-vC2EhJSMZCGfh7L:b-nu1f9w) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L42)):*
> ```XML
> impl GrpcScheduler {
>     pub async fn new(config: &config::schedulers::GrpcScheduler) -> Result<Self, Error> {
>         let endpoint = transport::Channel::balance_list(std::iter::once(
> ```

nit: Looks unrelated (possibly future PR?).

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 17 of 21 files reviewed, 13 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/grpc_scheduler.rs` line 62 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V1JNc8UHFA2vRKG_1:-N_V1JNc8UHFA2vRKG_2:bq9ijux) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L62)):*
> ```XML
>             let (tx, rx) = watch::channel(Arc::new(initial_response.try_into()?));
>             tokio::spawn(async move {
>                 while let Ok(Some(response)) = result_stream.message().await {
> ```

nit: We also need to watch for the `tx.closed()` so we can terminate the grpc stream if our client disconnects.
___
*[`cas/scheduler/grpc_scheduler.rs` line 151 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V0yv3EtcIV3K_DcDy:-N_V0yv3EtcIV3K_DcDz:bi0bfjw) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L151)):*
> ```XML
>             .await;
>         if result_stream.is_err() {
>             return None;
> ```

nit: Lets add a `log::info!()` here with the error message.

allada (commented):
> Reviewed 1 of 8 files at r7.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 18 of 21 files reviewed, 13 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 18 of 21 files reviewed, 14 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/grpc_scheduler.rs` line 70 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V1h7-7gUvftmjtx3P:-N_V1h7-7gUvftmjtx3Q:b-nw60ri) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L70)):*
> ```XML
>                             }
>                         }
>                         Err(err) => log::error!("Error converting response to watch in GrpcScheduler: {}", err),
> ```

nit: 
```
Error converting response to ActionState in GrpcScheduler
```

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 18 of 21 files reviewed, 8 unresolved discussions (waiting on @allada and @chrisstaite)

___
*[`cas/grpc_service/execution_server.rs` line 213 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UYODg3yCLWa4cb8Hi:-N_V2Ho426lN2SXUCfKk:b-db0ooo) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L213)):*

nit:  `&request.into_inner().name` looks better.
</blockquote></details>

`the trait \`From<&String>\` is not implemented for \`ActionInfoHashKey\``
___
*[`cas/grpc_service/execution_server.rs` line 219 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UZDcjCStLmxJZG5Cn:-N_V2aINBRv0-Lbhilu2:b-g7v3gy) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L219)):*

I wounder if it'd be a good idea to then try and find the item in the action cache before rejecting it?

I don't think the spec says anything about this, but I think it makes sense. From looking at the code, we'd need to modify the scheduler so before it fully drops `ActionResult`, it'll upload it to the CAS (not AC).

No need to do this now, but if you don't want to do it, lets make a TODO. The cool part is that it'll allow users to run an action, go away, and come back later to get the results (assuming the timeout is properly permissive).
</blockquote></details>

Isn't the idea that if you do that you should just look it up in the Action Cache yourself?
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 158 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_Umoj-A6pz_Z9RZanY:-N_V4Mmh7NyzW363zx_3:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L158)):*

There's a bit of a race condition here because we can have two clients at this point trying add the same `unique_qualifier`. This would cause some very bizarre things to happen. We should do a cache lookup here first and feed the existing item if it exists within the same lock.
</blockquote></details>

Done.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 181 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UkDCT56V_YznElJCD:-N_V58C83T1OGgKuSfGt:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L181)):*

Hmmm, minor bug here, where we only notify the downstream scheduler that the client has disconnected when the worker sends a notification to the client and `tx.send(action_state).is_err()` triggers an error and drops the `Sender`, which then causes the downstream scheduler to now be able to know the client disconnected.

This currently is not a big problem, since it only causes a very minor amount of additional needless memory, but if/when we implement the ability for the scheduler to timeout tasks because no listeners are active it'll become an issue.

This can be fixed by using a `select!{}` over `tx.closed()` and `watch_stream.next()`. Not sure if we should do this in this PR, but wanted to make sure it was known.

This is a good example on why I think we should try and avoid using spawns as much as possible, granted here we need to use a spawn (unless we change function signatures), this would auto-propagate everything down because the future would get dropped which would call `Drop` on everything downstream.
</blockquote></details>

Done.
___
*[`cas/scheduler/cache_lookup_scheduler.rs` line 208 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UuxH9COBDUbY2jZBx:-N_V5Vx3-kvHVfWF75Iy:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/cache_lookup_scheduler.rs#L208)):*

I created an upstream ticket for this:
https://github.com/tokio-rs/tokio/issues/5871

Can we make a TODO referencing this ticket?
</blockquote></details>

Done.
___
*[`cas/scheduler/grpc_scheduler.rs` line 42 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V-vC171kjd2vfYxBb:-N_V5XBt5ec05p9Olzis:b-g73dxo) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L42)):*

nit: Looks unrelated (possibly future PR?).
</blockquote></details>

This made more sense when I was adding in the `OperationsClient` too, but now I've taken that out it makes a little less sense, but I think this is worth adding in as I believe the `balance_list` will re-try a failed connection whereas the existing code won't.  Happy to move this out to another PR, but it's a bit mixed in now...
___
*[`cas/scheduler/simple_scheduler.rs` line 166 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UvdbR6if15-0rySV9:-N_V7TW3CZwWJZMw7fWJ:bocdzpj) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L166)):*

Can we do this part of the change in a followup PR? I want to think about this one a bit because it plays into both debugging and caching. Also, I'm not sure if the scheduler should do this here or if maybe some kind of logging service might be better to solve this for the debugging case.

We also can probably solve the non-debug use case by doing what is commented on above (store the `ActionResult` in CAS). I am pretty sure buildbarn does this.
</blockquote></details>

For fast build actions this is required for Goma otherwise the action disappears between it calling `add_action` and calling `wait_execution`.  That's why I had to implement it.  We could pull it out, but both are necessary for this code to work, so it doesn't make much sense to I don't think?

chrisstaite-menlo:
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_V98DV2IW9cGFEmMwZ:b-8z4ig3) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*

Uug, these values area already indexed by the `ActionInfoHashKey`, but rust's `HashMap` does not support getting it by this value without: https://github.com/rust-lang/rust/issues/56167

It does look like we can use `hashbrown` though to solve it, because it uses `Equivalent<K>` for lookups instead of forcing `K: Borrow<Q>`, but untested:
https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.get
</blockquote></details>

Looking at the source of `HashSet` it calls `self.base.get` where `base` is `hashbrown`...

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 14 of 22 files reviewed, 8 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/grpc_service/execution_server.rs` line 213 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UYODg3yCLWa4cb8Hi:-N_V8wvq2C_oAita4too:b-b0sw51) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L213)):*

`the trait \`From<&String>\` is not implemented for \`ActionInfoHashKey\``
</blockquote></details>

blah. Probably needs `Borrow<&T>` magic in the `TryFrom` trait.
___
*[`cas/grpc_service/execution_server.rs` line 219 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UZDcjCStLmxJZG5Cn:-N_V971v1caO_JSVBHX4:b-583sb4) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L219)):*

Isn't the idea that if you do that you should just look it up in the Action Cache yourself?
</blockquote></details>

Yes and no. If caching was enabled 100% of the time yes, but in our case we can send requests that will not be saved in action cache. In the event the `no_cache` flag is sent on the initial request, by design the `ActionInfoHashKey` will be different for every request, but if the `no_cache` is not set, then it may use the same `ActionInfoHashKey` and thus the same Operation Id.

Lets not get caught up on this in this PR and if possible lets split off this part into a different PR. I will look into how other projects do this. I do remember buildbarn having a flag somewhere saying something like: `Store action results for non-cachable items in the CAS. This breaks the "HASH(value) -> value" guarantee though`.

allada (changes requested):
> Reviewed 3 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 5 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/grpc_scheduler.rs` line 42 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V-vC171kjd2vfYxBb:-N_VBkLl3PRr-mb8Wg7A:bxi96rd) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L42)):*

This made more sense when I was adding in the `OperationsClient` too, but now I've taken that out it makes a little less sense, but I think this is worth adding in as I believe the `balance_list` will re-try a failed connection whereas the existing code won't.  Happy to move this out to another PR, but it's a bit mixed in now...
</blockquote></details>

Yeah, I agree we should use `balance_list`, but we need to modify the config file too, which is technically a breaking change, so it needs to be documented in the commit title.

chrisstaite-menlo:
*[`cas/scheduler/grpc_scheduler.rs` line 42 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_V-vC171kjd2vfYxBb:-N_VCCUV9f0IYJeW5tkO:b-f9p2xq) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/grpc_scheduler.rs#L42)):*

Yeah, I agree we should use `balance_list`, but we need to modify the config file too, which is technically a breaking change, so it needs to be documented in the commit title.
</blockquote></details>

Not at all, I just add a single endpoint to the `balance_list` as shown here...

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 5 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 166 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UvdbR6if15-0rySV9:-N_VCY4FAcLCUFs4oQJk:b50sdj8) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L166)):*

For fast build actions this is required for Goma otherwise the action disappears between it calling `add_action` and calling `wait_execution`.  That's why I had to implement it.  We could pull it out, but both are necessary for this code to work, so it doesn't make much sense to I don't think?
</blockquote></details>

I dont want to block this PR on this especially since you are able to get Goma working. So I created a ticket:
https://github.com/allada/turbo-cache/issues/192

Can you just make a note in the code referencing this ticket here?

Something like:
```
// TOOD(#192) Revisit if this is the best way to handle recently completed actions.
```

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 5 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_VD0C_EA54kBx9VZJF:bwsywq9) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*

Looking at the source of `HashSet` it calls `self.base.get` where `base` is `hashbrown`...
</blockquote></details>

Can we add a comment here with something like:
```
// TODO: We can lookup the action without needing to iterate the maps but there's no API for it.
// Maybe when: https://github.com/rust-lang/rust/issues/56167 is done?
```

chrisstaite-menlo:
*[`cas/grpc_service/execution_server.rs` line 219 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UZDcjCStLmxJZG5Cn:-N_VDITH-q9Oh0-3jQ_t:b8w7k9o) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L219)):*

Yes and no. If caching was enabled 100% of the time yes, but in our case we can send requests that will not be saved in action cache. In the event the `no_cache` flag is sent on the initial request, by design the `ActionInfoHashKey` will be different for every request, but if the `no_cache` is not set, then it may use the same `ActionInfoHashKey` and thus the same Operation Id.

Lets not get caught up on this in this PR and if possible lets split off this part into a different PR. I will look into how other projects do this. I do remember buildbarn having a flag somewhere saying something like: `Store action results for non-cachable items in the CAS. This breaks the "HASH(value) -> value" guarantee though`.
</blockquote></details>

Worth a new issue?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 4 unresolved discussions (waiting on @allada and @chrisstaite)

___
*[`cas/grpc_service/execution_server.rs` line 213 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UYODg3yCLWa4cb8Hi:-N_VDCuZEP2tqLPbLaI7:b-9ygzv0) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/grpc_service/execution_server.rs#L213)):*

blah. Probably needs `Borrow<&T>` magic in the `TryFrom` trait.
</blockquote></details>

Did some messing and Googling and found there's not a nice way to do this...
___
*[`cas/scheduler/simple_scheduler.rs` line 166 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UvdbR6if15-0rySV9:-N_VDBet6csn3Ejg-6sC:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L166)):*

I dont want to block this PR on this especially since you are able to get Goma working. So I created a ticket:
https://github.com/allada/turbo-cache/issues/192

Can you just make a note in the code referencing this ticket here?

Something like:
```
// TOOD(#192) Revisit if this is the best way to handle recently completed actions.
```
</blockquote></details>

Done.

chrisstaite-menlo:
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_VDVwO7MdiLW0iE-rA:b-tzo4nb) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*

Can we add a comment here with something like:
```
// TODO: We can lookup the action without needing to iterate the maps but there's no API for it.
// Maybe when: https://github.com/rust-lang/rust/issues/56167 is done?
```
</blockquote></details>

I already swapped to `hashbrown`.  Do you think I should put it back with an issue?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 4 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_VEAuC9-PFx-KnQB0r:bgpes69) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*

I already swapped to `hashbrown`.  Do you think I should put it back with an issue?
</blockquote></details>

Wait, it was really that easy!?!?!
___
*[`cas/scheduler/simple_scheduler.rs` line 168 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VEOKp2rQh_B03HG2t:-N_VEOKp2rQh_B03HG2u:b4wc3p1) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L168)):*
> ```XML
>     // keep their completion state around for a while to send back.
>     // TODO(#192) Revisit if this is the best way to handle recently completed actions.
>     recently_completed_actions: Vec<Arc<ActionState>>,
> ```

Can't this be a `HashSet` now so we can just grab it without iterating the vector?
___
*[`cas/scheduler/simple_scheduler.rs` line 255 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFGIB0Oa6pNaAShzp:-N_VFGIB0Oa6pNaAShzq:b-yloju4) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L255)):*
> ```XML
>         self.recently_completed_actions.retain(|action| match &action.stage {
>             ActionStage::Completed(state) => {
>                 if state.execution_metadata.worker_completed_timestamp > expiry_time {
> ```

I think we should wrap `ActionStage` so we can store the timestamp because I don't think we should blindly trust workers and I think we should hold on to non `ActionStage::Completed` (like errors) for some set duration too.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 5 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 252 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VF_XJ6L97x_p-NCup:-N_VF_XJ6L97x_p-NCuq:b-6yqjvr) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L252)):*
> ```XML
>     fn clean_recently_completed_actions(&mut self) {
>         // Keep for a minute maximum.
>         let expiry_time = SystemTime::now().checked_sub(Duration::new(60, 0)).unwrap();
> ```

Can we move this setting to a config?

chrisstaite-menlo:
*[`cas/scheduler/simple_scheduler.rs` line 283 at r8](https://reviewable.io/reviews/allada/turbo-cache/177#-N_UwraG2CZAq3eAOkoX:-N_VFbFG43wEyEm-mr59:b-sikik1) ([raw file](https://github.com/allada/turbo-cache/blob/f4643203fd95df4d3a40d88a231c83a08fd1d491/cas/scheduler/simple_scheduler.rs#L283)):*

Wait, it was really that easy!?!?!
</blockquote></details>

Seemingly?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 25 files reviewed, 7 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 280 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFpeg8lDFas2A5zHc:-N_VFpehE_KpiUnxLrvV:bio3o3b) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L280)):*
> ```XML
> 
>         // Clean up the actions list.
>         self.clean_recently_completed_actions();
> ```

lets not do this here, we'll do it similar to how we deal with `remove_timedout_workers()`.
___
*[`cas/scheduler/simple_scheduler.rs` line 521 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFyze-n-glcnVp14-:-N_VFyze-n-glcnVp140:bio3o3b) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L521)):*
> ```XML
>                 .push(running_action.action.current_state);
>             // Clean up any old actions.
>             self.clean_recently_completed_actions();
> ```

lets not do this here, we'll do it similar to how we deal with `remove_timedout_workers()`.

chrisstaite-menlo:
*[`cas/scheduler/simple_scheduler.rs` line 168 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VEOKp2rQh_B03HG2t:-N_VGpPM4ub6FaWJrwlv:b-ja6l4h) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L168)):*

Can't this be a `HashSet` now so we can just grab it without iterating the vector?
</blockquote></details>

No, because it's `ActionState` not `ActionInfo` :(. I suppose we could update `ActionState` to also be hashed in the same way as `ActionInfo`...

allada (changes requested):
> Reviewed 2 of 8 files at r7, 2 of 5 files at r9, 4 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 23 of 25 files reviewed, 7 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 516 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VH6wo6hwPEF6eE0JF:-N_VH6wo6hwPEF6eE0JG:b-iw1q2g) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L516)):*
> ```XML
>             self.active_actions.insert(action_info, running_action);
>             return Ok(());
>         } else if send_result.is_err() {
> ```

shouldn't we add the item to `recently_complted_actions` regardless of if the result was unable to be sent? Ie: there may be multiple clients listening, the first client was success, but the second client then calls `WaitExecution` and fails because the task was multiplexed.

allada (commented):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 24 of 25 files reviewed, 7 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

allada (commented):
Sorry about doing this in phases. It was a pretty difficult one to review. I finished going over everything. The things commented on are the only things left. Before it was: "I can't really review this until the last comment is addressed".

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177#-:-N_VHW5_AorYlcKFRVJt:b-i1b1cs)* status: 24 of 25 files reviewed, 7 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 24 of 25 files reviewed, 7 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/simple_scheduler.rs` line 168 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VEOKp2rQh_B03HG2t:-N_VI6LN-N5K9djmMQGC:b8xgss2) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L168)):*

No, because it's `ActionState` not `ActionInfo` :(. I suppose we could update `ActionState` to also be hashed in the same way as `ActionInfo`...
</blockquote></details>

Yeah actually it can. `ActionState` just needs the same hash function that `ActionInfo` has. Ironically `ActionState` is owns the properties that `ActionInfo` uses to hash.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177)* status: 16 of 26 files reviewed, 7 unresolved discussions (waiting on @allada and @chrisstaite)

___
*[`cas/scheduler/simple_scheduler.rs` line 168 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VEOKp2rQh_B03HG2t:-N_VOPow5M1T1mUOvTj7:baq8zuh) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L168)):*

Yeah actually it can. `ActionState` just needs the same hash function that `ActionInfo` has. Ironically `ActionState` is owns the properties that `ActionInfo` uses to hash.
</blockquote></details>

I actually did this in a different way since I was wrapping the completed time in anyway and implemented it for `CompletedAction`...
___
*[`cas/scheduler/simple_scheduler.rs` line 252 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VF_XJ6L97x_p-NCup:-N_VOIOW37XVrGFEM-CL:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L252)):*

Can we move this setting to a config?
</blockquote></details>

Done.
___
*[`cas/scheduler/simple_scheduler.rs` line 255 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFGIB0Oa6pNaAShzp:-N_VOIcz2lf0Sc7UVLwi:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L255)):*

I think we should wrap `ActionStage` so we can store the timestamp because I don't think we should blindly trust workers and I think we should hold on to non `ActionStage::Completed` (like errors) for some set duration too.
</blockquote></details>

Done.
___
*[`cas/scheduler/simple_scheduler.rs` line 280 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFpeg8lDFas2A5zHc:-N_VLaot7y2ZQHWPnXsU:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L280)):*

lets not do this here, we'll do it similar to how we deal with `remove_timedout_workers()`.
</blockquote></details>

Done.
___
*[`cas/scheduler/simple_scheduler.rs` line 516 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VH6wo6hwPEF6eE0JF:-N_VL_Xe1lyYVXZW-5dP:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L516)):*

shouldn't we add the item to `recently_complted_actions` regardless of if the result was unable to be sent? Ie: there may be multiple clients listening, the first client was success, but the second client then calls `WaitExecution` and fails because the task was multiplexed.
</blockquote></details>

Done.
___
*[`cas/scheduler/simple_scheduler.rs` line 521 at r12](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VFyze-n-glcnVp14-:-N_VLZl-4KgcBLxsRi4x:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0cbd85d0ec37c175c7d37734f8b0ef4e2b9ad5cb/cas/scheduler/simple_scheduler.rs#L521)):*

lets not do this here, we'll do it similar to how we deal with `remove_timedout_workers()`.
</blockquote></details>

Done.

chrisstaite-menlo:
I knew it was going to be a horrid one, because it really should have been more than one PR, but it was a mangled mess!

allada (approved):
Woot! This was a tough one.

:lgtm_strong:  

> Reviewed 1 of 3 files at r5, 9 of 9 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/177#-:-N_VVUD00LYUJGC_dKwr:bbk9ny)* status: all files reviewed, 2 unresolved discussions (waiting on @chrisstaite and @chrisstaite-menlo)

___
*[`cas/scheduler/default_scheduler_factory.rs` line 70 at r13](https://reviewable.io/reviews/allada/turbo-cache/177#-N_VSz3x6JuC5wrKmOZj:-N_VSz3x6JuC5wrKmOZk:bru9cky) ([raw file](https://github.com/allada/turbo-cache/blob/436642db80f7073a80718bceda3a7421b51bcb6f/cas/scheduler/default_scheduler_factory.rs#L70)):*
> ```XML
> fn start_cleanup_timer(action_scheduler: &Arc<dyn ActionScheduler>) {
>     let weak_scheduler = Arc::downgrade(action_scheduler);
>     tokio::spawn(async move {
> ```

:frowning_face:  It looks like it's time for a `SchedulerManager` so we don't need these spawns to be running all over the place.

No action needed.

PR #178, opened 2023-07-14
Add highly experimental local k8s cluster

aaronmondal:
This uses rules_ll to expose `ll up` and `ll down` commands to create and delete a k8s cluster for testing S3 storage and cloud deployments in general.

Replaces local Bazel with a reproducible RBE-compatible wrapper that exposes OpenSSL in a reproducible, hermetic manner.

Requires x86_64 linux, a standard rules_ll setup and eBPF JIT compilation enabled in the kernel. Known to work on Linux 6.4.3-gentoo. If the cluster doesn't start properly it's probably an issue with missing kernel config flags.

Only works with the ll using nix and breaks all other build setups. Not meant for merging.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/178)


aaronmondal:
Ah and this requires a clone of upstream rules_ll adjacent to the `turbo-cache` directory :rofl:

allada (commented):
There's so much here that I don't understand, nix is new to me :frowning_face: 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/178#-:-N_qRCiZ6V65lLxVRHZb:b3q5ws0)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`pre-commit-hooks.nix` line 1 at r2](https://reviewable.io/reviews/allada/turbo-cache/178#-N_qRN863_9U0qHyHk6J:-N_qRN863_9U0qHyHk6K:b-gck03) ([raw file](https://github.com/allada/turbo-cache/blob/840bddda9ee2d732cd11ac85d097dcea65653c3f/pre-commit-hooks.nix#L1)):*
> ```Nix
> { pkgs, ... }:
> ```

do we really want pre-commit hooks here?

aaronmondal (commented):
IMO nix is basically the best dev experience possible for users at the price of the worst dev experience possible for those writing the actual nix code :rofl:

For instance, the `ll up` command is a nix-templated wrapper that invokes another nix-templated wrapper around bazel to build go, add a pulumi exe to the PATH, build kind and the pulumi sdk and then execute that to set up k8s. As end user a bunch of magic happens and you get a nice cluster but on a maintenance level it quickly gets tricky to get everything working properly :sweat_smile: 

The flake also adds additional tooling like `kubectl` so that one doesn't need to install it manually.

The language feels incredibly hacky (and in this flake it's even less transparent what actually happens since we're overriding devenv instead of the regular nix shell) but somehow it produces some of the most robust, flexible and reusable systems I've ever seen :laughing: The way Nix lets you glue together vastly different projects is IMO actually quite similar to what bazel does, but on a coarser package level rather than on a build target level.

Potential downside is that once you start using it is slowly creeps into all your workflows until you can't live without it :laughing: AFAIK my colleagues went from "ugh such an ugly unintuitive language" to "ok pretty cool dev experience" to "the first thing I do in a new project is set up a flake" :rofl: 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/178#-:-N_t_AaN4f8qArsj9LVp:b-hgded4)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`pre-commit-hooks.nix` line 1 at r2](https://reviewable.io/reviews/allada/turbo-cache/178#-N_qRN863_9U0qHyHk6J:-N_tXrV70v4dyzwDNXcs:b-epaqqa) ([raw file](https://github.com/allada/turbo-cache/blob/840bddda9ee2d732cd11ac85d097dcea65653c3f/pre-commit-hooks.nix#L1)):*

do we really want pre-commit hooks here?
</blockquote></details>

We had issues where some hooks didn't work consistently on all machines and sometimes different formatter versions produce slightly different results. With a regular installation, some hooks might depend on e.g. python and a pre-commit installation. This can lead to surprisingly unpredictable behavior.

So at some point we just decided to hard-lock the hooks down so that we know that they exactly reproduce what is tested in CI. The downside of this is that sometimes hooks need to be a bit more customized (e.g. rust and go linters tend to require some degree of knowledge of the entire repo and can't operate on standalone files). But once it's set up it trivializes these checks and the hook installations can be cached by caching the entire nix store that we'd cache anyways.

I didn't set up rust hooks here because we already have reproducible rustfmt and clippy via bazel and using those via nix seemed like a pessimization of the current workflow.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/178)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`pre-commit-hooks.nix` line 1 at r2](https://reviewable.io/reviews/allada/turbo-cache/178#-N_qRN863_9U0qHyHk6J:-N_xuEs20I69Dp6s-f8g:bcl7p75) ([raw file](https://github.com/allada/turbo-cache/blob/840bddda9ee2d732cd11ac85d097dcea65653c3f/pre-commit-hooks.nix#L1)):*

We had issues where some hooks didn't work consistently on all machines and sometimes different formatter versions produce slightly different results. With a regular installation, some hooks might depend on e.g. python and a pre-commit installation. This can lead to surprisingly unpredictable behavior.

So at some point we just decided to hard-lock the hooks down so that we know that they exactly reproduce what is tested in CI. The downside of this is that sometimes hooks need to be a bit more customized (e.g. rust and go linters tend to require some degree of knowledge of the entire repo and can't operate on standalone files). But once it's set up it trivializes these checks and the hook installations can be cached by caching the entire nix store that we'd cache anyways.

I didn't set up rust hooks here because we already have reproducible rustfmt and clippy via bazel and using those via nix seemed like a pessimization of the current workflow.
</blockquote></details>

I was more thinking to instead have these hooks as tests and if the user wants them as pre-commit hooks they do it themselves.

I've never been a fan of checking in pre-commit hooks... I felt like people were always fighting them.

aaronmondal:
This has been integrated in some other form :partying_face:

PR #179, opened 2023-07-14
Fix potential bug where scheduer could drop action and remove Arc

allada:
In the code as it is written right now it is not possible for this to happen, but it could easily be implemented on the worker without needing to touch the scheduler code.

The case if the worker looked up the action in the cache and found a hit, it could return a result stating it was served from cache. In this event the scheduler would never flag the action as completed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)

### DIFF
--- a/cas/grpc_service/tests/worker_api_server_test.rs
+++ b/cas/grpc_service/tests/worker_api_server_test.rs
@@ -365,7 +365,7 @@ pub mod execution_response_tests {
             // Ensure our client thinks we are executing.
             client_action_state_receiver.changed().await?;
             let action_state = client_action_state_receiver.borrow();
-            assert_eq!(action_state.as_ref().stage, ActionStage::Executing);
+            assert_eq!(action_state.stage, ActionStage::Executing);
         }
 
         // Now send the result of our execution to the scheduler.
--- a/cas/scheduler/action_messages.rs
+++ b/cas/scheduler/action_messages.rs
@@ -558,9 +558,10 @@ pub enum ActionStage {
     /// Result was found from cache, don't decode the proto just to re-encode it.
     CompletedFromCache(ProtoActionResult),
     /// Error or action failed with an exit code on the worker.
-    /// This means that the job might have finished executing, but returned a non-zero
-    /// exit code (for example a test failing or file not compilable). The `ActionResult`
-    /// may contain better results than the `Error` message.
+    /// This means that the job might have finished executing, but the worker had an
+    /// internal error. This might have happened if the worker timed out, crashed,
+    /// action cleanup failure, out of memory or other kind of errors that are not
+    /// related to the action, but rather the environment.
     Error((Error, ActionResult)),
 }
 
@@ -570,6 +571,13 @@ impl ActionStage {
             Self::Unknown | Self::CacheCheck | Self::Queued | Self::Executing => false,
             Self::Completed(_) | Self::CompletedFromCache(_) | Self::Error(_) => true,
         }
+    }
+
+    /// Returns true if the worker considers the action done and no longer needs to be tracked.
+    // Note: This function is separate from `has_action_result()` to not mix the concept of
+    //       "finished" with "has a result".
+    pub const fn is_finished(&self) -> bool {
+        self.has_action_result()
     }
 }
 
--- a/cas/scheduler/simple_scheduler.rs
+++ b/cas/scheduler/simple_scheduler.rs
@@ -467,17 +467,12 @@ impl SimpleSchedulerImpl {
             return Err(make_input_err!("{}", msg));
         }
 
-        let did_complete = match action_stage {
-            ActionStage::Completed(_) => true,
-            _ => false,
-        };
-
         Arc::make_mut(&mut running_action.action.current_state).stage = action_stage;
 
         let send_result = running_action
             .action
             .notify_channel
-            .send(running_action.action.current_state);
+            .send(running_action.action.current_state.clone());
         if send_result.is_err() {
             log::warn!(
                 "Action {} has no more listeners during update_action()",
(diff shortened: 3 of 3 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)* status: 0 of 10 files reviewed, all discussions resolved

___
*[`cas/scheduler/simple_scheduler.rs` line 46 at r1](https://reviewable.io/reviews/allada/turbo-cache/179#-N_KhqRg1TEQrqfA0GN7:-N_KhqRg1TEQrqfA0GN8:b-phlqlh) ([raw file](https://github.com/allada/turbo-cache/blob/d1a5fd6026cdf3871404e7fdd34475f12b1e8052/cas/scheduler/simple_scheduler.rs#L46)):*
> ```XML
> struct AwaitedAction {
>     action_info: Arc<ActionInfo>,
>     current_state: ActionState,
> ```

I also considered making this a `Cow`, This struct might be big (hundreds of bytes), I felt keeping it simple was worth the effort.

allada (commented):
+@chrisstaite-menlo 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179#-:-N_Ki1e36pVABL3WE1hW:b96vhr1)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

allada (commented):
-@chrisstaite-menlo 
+@aaronmondal 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179#-:-N_LEXHf5bdSSHmk195z:b-h8s7tc)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @aaronmondal)

aaronmondal (changes requested):
> Reviewed 2 of 10 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/scheduler/simple_scheduler.rs` line 483 at r2](https://reviewable.io/reviews/allada/turbo-cache/179#-N_LMqLS212AsCH4UKAH:-N_LMqLS212AsCH4UKAI:bi6b9ci) ([raw file](https://github.com/allada/turbo-cache/blob/775ef8005dfd22f23b7ee6ef1480100bfcf86e59/cas/scheduler/simple_scheduler.rs#L483)):*
> ```XML
>         }
> 
>         if !running_action.action.current_state.stage.is_finished() {
> ```

Hmm I get that we generally might want to separate "being done" with "having a result", but why do we need `is_finished` here when we could query `has_action_result` directly and omit `is_finished` entirely?

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`cas/scheduler/simple_scheduler.rs` line 483 at r2](https://reviewable.io/reviews/allada/turbo-cache/179#-N_LMqLS212AsCH4UKAH:-N_LOt1MA7zzzuG8j8Uo:ba6vl0f) ([raw file](https://github.com/allada/turbo-cache/blob/775ef8005dfd22f23b7ee6ef1480100bfcf86e59/cas/scheduler/simple_scheduler.rs#L483)):*

Hmm I get that we generally might want to separate "being done" with "having a result", but why do we need `is_finished` here when we could query `has_action_result` directly and omit `is_finished` entirely?
</blockquote></details>

The idea is that `is_finished` is supposed to signal "I have no more work to do" from the worker.

`has_action_result` current means that, but I was worried that in the future there might be something added like: "Here's your stderr/stdout stream, but i'm still running" or: "I ran the test, but had to terminate the job for {reason}, but here's the logs/results/state of the last execution"... This example might happen for ML model training for example.

aaronmondal (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

___
*[`cas/scheduler/simple_scheduler.rs` line 483 at r2](https://reviewable.io/reviews/allada/turbo-cache/179#-N_LMqLS212AsCH4UKAH:-N_LRZG-0Cj5AwnYdb1Q:bzej0vv) ([raw file](https://github.com/allada/turbo-cache/blob/775ef8005dfd22f23b7ee6ef1480100bfcf86e59/cas/scheduler/simple_scheduler.rs#L483)):*

The idea is that `is_finished` is supposed to signal "I have no more work to do" from the worker.

`has_action_result` current means that, but I was worried that in the future there might be something added like: "Here's your stderr/stdout stream, but i'm still running" or: "I ran the test, but had to terminate the job for {reason}, but here's the logs/results/state of the last execution"... This example might happen for ML model training for example.
</blockquote></details>

I wonder whether `done` would be less confusing than `is_finished` then. Anyways, no strong opinion. :lgtm: :laughing:

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

___
*[`cas/scheduler/simple_scheduler.rs` line 483 at r2](https://reviewable.io/reviews/allada/turbo-cache/179#-N_LMqLS212AsCH4UKAH:-N_LSSsW0eXVivg7WbSa:b-v55bbu) ([raw file](https://github.com/allada/turbo-cache/blob/775ef8005dfd22f23b7ee6ef1480100bfcf86e59/cas/scheduler/simple_scheduler.rs#L483)):*
Well, I went down that rabbit hole:
https://grammarhow.com/done-vs-finished/

> **“Done” and “finished” are nearly identical synonyms. This means that in nearly all contexts, they mean the same thing. The difference between the two is that “done” means that you aren’t doing any more of the action, while “finished” means that you’ve finished the action.**

So I guess, "Finished" is more appropriate here :laughing:

PR #183, opened 2023-07-14
Remove need for spawn in BytestreamServer

allada:
Spawns are very expensive to make and one of the most common API points is BystreamServer::read. From a long time ago it was really hard to get lifetimes to live long enough here for the stream API. Armed with more experience the lifetimes are now resolved, so we do not need to use this pointless spawn.

Now we will pipe the data from the store to the client in the same spawn, which means we will have a very high chance of being on the same thread when shipping bytes to the client.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)

### DIFF
--- a/cas/grpc_service/bytestream_server.rs
+++ b/cas/grpc_service/bytestream_server.rs
@@ -18,7 +18,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Instant;
 
-use futures::{stream::unfold, Stream};
+use futures::{future::pending, stream::unfold, Future, Stream};
 use parking_lot::Mutex;
 use proto::google::bytestream::{
     byte_stream_server::ByteStream, byte_stream_server::ByteStreamServer as Server, QueryWriteStatusRequest,
@@ -34,12 +34,6 @@ use grpc_store::GrpcStore;
 use resource_info::ResourceInfo;
 use store::{Store, StoreManager, UploadSizeInfo};
 use write_request_stream_wrapper::WriteRequestStreamWrapper;
-
-struct ReaderState {
-    max_bytes_per_stream: usize,
-    rx: DropCloserReadHalf,
-    reading_future: tokio::task::JoinHandle<Result<(), Error>>,
-}
 
 type ReadStream = Pin<Box<dyn Stream<Item = Result<ReadResponse, Status>> + Send + 'static>>;
 
@@ -77,14 +71,14 @@ impl ByteStreamServer {
             usize::try_from(read_request.read_limit).err_tip(|| "read_limit has is not convertible to usize")?;
         let resource_info = ResourceInfo::new(&read_request.resource_name)?;
         let instance_name = resource_info.instance_name;
-        let store_clone = self
+        let store = self
             .stores
             .get(instance_name)
             .err_tip(|| format!("'instance_name' not configured for '{}'", instance_name))?
             .clone();
 
         // If we are a GrpcStore we shortcut here, as this is a special store.
-        let any_store = store_clone.clone().as_any();
+        let any_store = store.clone().as_any();
         let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
         if let Some(grpc_store) = maybe_grpc_store {
             let stream = grpc_store.read(Request::new(read_request)).await?.into_inner();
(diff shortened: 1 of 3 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @aaronmondal and @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/183#-N_Lhwg_COI1kTldRP8C:-N_Lhwg_COI1kTldRP8D:blv0ogn) (no related file):*
This has a high chance of breaking things unintentionally. In theory it's a 100% non-noticable change, but because of some subtle details on the way errors are propagated it is possible for me to have got it wrong here.

Therefore, can I get a review from both of you:
+@aaronmondal 
+@chrisstaite-menlo 

___

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @allada and @chrisstaite-menlo)

___
*[`cas/grpc_service/bytestream_server.rs` line 110 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_M5bkE5zO88SnKnjBK:-N_M5bkE5zO88SnKnjBL:b7qw3xe) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L110)):*
> ```XML
> 
>         Ok(Response::new(Box::pin(unfold(state, move |state| async {
>             let mut state = state?; // If None our stream is done.
> ```

Is this shadowing safe to do? If we get cancelled while awaiting `state.get_part_fut.await` below, could this lead to a loss of data? I'm basically just reciting https://smallcultfollowing.com/babysteps/blog/2022/06/13/async-cancellation-a-case-study-of-pub-sub-in-mini-redis/#ways-to-write-a-broken-read_frame here, but I'm not sure whether it's applicable in this case.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/grpc_service/bytestream_server.rs` line 110 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_M5bkE5zO88SnKnjBK:-N_MFpZ93kZ7y33a3mGk:b6srrfe) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L110)):*

Is this shadowing safe to do? If we get cancelled while awaiting `state.get_part_fut.await` below, could this lead to a loss of data? I'm basically just reciting https://smallcultfollowing.com/babysteps/blog/2022/06/13/async-cancellation-a-case-study-of-pub-sub-in-mini-redis/#ways-to-write-a-broken-read_frame here, but I'm not sure whether it's applicable in this case.
</blockquote></details>

In any case, it'd be up to the store implementation to protect against this.

In code that matters we use special `Drop` objects with callbacks to correct for these edge cases. This is a bit of a problem, but it's not something we can do anything against.

Here's an example on how we cleanup files in the event they are no longer needed (including temp files used when the client is uploading):
https://github.com/allada/turbo-cache/blob/f211ef23a1836f2e0ae25e04832175df87ab23e7/cas/store/filesystem_store.rs#L98

Here's an example on how we handle cases where we want to terminate a spawn in the event the handle is dropped:
https://github.com/allada/turbo-cache/blob/f211ef23a1836f2e0ae25e04832175df87ab23e7/util/common.rs#L144

chrisstaite-menlo:
*[`cas/grpc_service/bytestream_server.rs` line 155 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_OgCpbCcI_ySVPMV1u:-N_OgCpbCcI_ySVPMV1v:bw7lnzo) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L155)):*
> ```XML
>                     },
>                     result = &mut state.get_part_fut => {
>                         state.maybe_get_part_result = Some(result);
> ```

Could you just use a `tokio::sync::oneshot::channel` here instead of all this complexity?

chrisstaite-menlo (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: all files reviewed, 2 unresolved discussions (waiting on @aaronmondal)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: all files reviewed, 2 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/grpc_service/bytestream_server.rs` line 155 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_OgCpbCcI_ySVPMV1u:-N_OikjW9dBW8AVWcXRC:b-dsfy27) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L155)):*

Could you just use a `tokio::sync::oneshot::channel` here instead of all this complexity?
</blockquote></details>

I'm not sure how `oneshot::channel` would help. The reason it's complicated is because I'm trying avoid creating a new spawn, but need to poll two futures.

chrisstaite-menlo:
*[`cas/grpc_service/bytestream_server.rs` line 155 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_OgCpbCcI_ySVPMV1u:-N_Q0IdC3sGoxz_FSiy6:byw18tp) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L155)):*

I'm not sure how `oneshot::channel` would help. The reason it's complicated is because I'm trying avoid creating a new spawn, but need to poll two futures.
</blockquote></details>

Yeah, I realised what it was doing after I wrote it...
I just thought, could you get rid of `maybe_get_part_result` and simply set `get_part_fut` to `Box::pin(future::ready(result))` instead and then whichever wins will be the result of `state.get_part_fut.await`.

aaronmondal (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: all files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/183)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/grpc_service/bytestream_server.rs` line 155 at r3](https://reviewable.io/reviews/allada/turbo-cache/183#-N_OgCpbCcI_ySVPMV1u:-N_QK2-gEv5DuTQ0yPot:bvj7hp3) ([raw file](https://github.com/allada/turbo-cache/blob/1d36726d6af01d99cdfb80f0b2f2bc77c6b72763/cas/grpc_service/bytestream_server.rs#L155)):*

Yeah, I realised what it was doing after I wrote it...
I just thought, could you get rid of `maybe_get_part_result` and simply set `get_part_fut` to `Box::pin(future::ready(result))` instead and then whichever wins will be the result of `state.get_part_fut.await`.
</blockquote></details>

Sadly no because it would cause the `select!()` to keep hitting on that future and spin lock until the other future is done.

PR #184, opened 2023-07-15
Add store type to GrpcStore.

chrisstaite-menlo:
When forwarding AC to an upstream GrpcStore. the store type is lost along the way.  If the GrpcStore is used directly then there's optimisations, however, if it's wrapped in another store (such as a fast-slow store) then the context is lost and queries go to the CAS server.

Instead, we make it possible to specify the upstream server type and enforce the upstream request to go via the AC server endpoint.  This is a little less efficient as tonic decodes the message then we have to re-encode it so it can be used by other stores, but it's the simplest solution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)

### DIFF
--- a/cas/store/BUILD
+++ b/cas/store/BUILD
@@ -257,6 +257,7 @@ rust_library(
         "//cas:__subpackages__",
     ],
     deps = [
+        ":ac_utils",
         ":traits",
         "//config",
         "//proto",
@@ -264,8 +265,10 @@ rust_library(
         "//util:common",
         "//util:error",
         "//util:write_request_stream_wrapper",
+        "@crate_index//:bytes",
         "@crate_index//:futures",
         "@crate_index//:parking_lot",
+        "@crate_index//:prost",
         "@crate_index//:shellexpand",
         "@crate_index//:tokio",
         "@crate_index//:tonic",
--- a/cas/store/grpc_store.rs
+++ b/cas/store/grpc_store.rs
@@ -16,10 +16,13 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use bytes::BytesMut;
 use futures::{stream::unfold, Stream};
+use prost::Message;
 use tonic::{transport, IntoRequest, Request, Response, Streaming};
 use uuid::Uuid;
 
+use ac_utils::ESTIMATED_DIGEST_SIZE;
 use buf_channel::{DropCloserReadHalf, DropCloserWriteHalf};
 use common::{log, DigestInfo};
 use error::{error_if, make_input_err, Error, ResultExt};
@@ -45,6 +48,7 @@ pub struct GrpcStore {
     cas_client: ContentAddressableStorageClient<transport::Channel>,
     bytestream_client: ByteStreamClient<transport::Channel>,
     ac_client: ActionCacheClient<transport::Channel>,
+    store_type: config::stores::StoreType,
 }
 
 impl GrpcStore {
@@ -71,6 +75,7 @@ impl GrpcStore {
             cas_client: ContentAddressableStorageClient::new(conn.clone()),
             bytestream_client: ByteStreamClient::new(conn.clone()),
             ac_client: ActionCacheClient::new(conn),
+            store_type: config.store_type,
         })
     }
 
@@ -78,6 +83,11 @@ impl GrpcStore {
         &self,
         grpc_request: Request<FindMissingBlobsRequest>,
     ) -> Result<Response<FindMissingBlobsResponse>, Error> {
+        error_if!(
+            matches!(self.store_type, config::stores::StoreType::AC),
+            "CAS operation on AC store"
+        );
+
         let mut request = grpc_request.into_inner();
         request.instance_name = self.instance_name.clone();
         let mut client = self.cas_client.clone();
@@ -91,6 +101,11 @@ impl GrpcStore {
         &self,
         grpc_request: Request<BatchUpdateBlobsRequest>,
     ) -> Result<Response<BatchUpdateBlobsResponse>, Error> {
+        error_if!(
+            matches!(self.store_type, config::stores::StoreType::AC),
+            "CAS operation on AC store"
+        );
+
         let mut request = grpc_request.into_inner();
         request.instance_name = self.instance_name.clone();
         let mut client = self.cas_client.clone();
@@ -104,6 +119,11 @@ impl GrpcStore {
         &self,
         grpc_request: Request<BatchReadBlobsRequest>,
     ) -> Result<Response<BatchReadBlobsResponse>, Error> {
+        error_if!(
+            matches!(self.store_type, config::stores::StoreType::AC),
+            "CAS operation on AC store"
+        );
+
         let mut request = grpc_request.into_inner();
         request.instance_name = self.instance_name.clone();
         let mut client = self.cas_client.clone();
@@ -117,6 +137,11 @@ impl GrpcStore {
         &self,
         grpc_request: Request<GetTreeRequest>,
     ) -> Result<Response<Streaming<GetTreeResponse>>, Error> {
+        error_if!(
+            matches!(self.store_type, config::stores::StoreType::AC),
+            "CAS operation on AC store"
+        );
+
         let mut request = grpc_request.into_inner();
         request.instance_name = self.instance_name.clone();
         let mut client = self.cas_client.clone();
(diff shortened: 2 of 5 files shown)


allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184#-:-N_QJ6ee4IPnHLLDTkr2:bnfp4nl)* status: all files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 86 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QBAaq7vxLw6h5-nX1:-N_QBAaq7vxLw6h5-nX2:b-q6cv3h) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L86)):*
> ```XML
>         grpc_request: Request<FindMissingBlobsRequest>,
>     ) -> Result<Response<FindMissingBlobsResponse>, Error> {
>         if matches!(self.store_type, config::stores::StoreType::ac) {
> ```

nit: Try:
```rs
error_if!(matches!(self.store_type, config::stores::StoreType::ac), "CAS operation on AC store");
```

(and others below)
___
*[`cas/store/grpc_store.rs` line 311 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QHATG-KcHw346trmJ:-N_QHATG-KcHw346trmK:bi93961) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L311)):*
> ```XML
>             .ok()
>             .err_tip(|| "Action result not found")?;
>         let mut value = BytesMut::new();
> ```

nit: Very sad we need to do this re-encoding of the data considering we don't actually care anything about it.

We should add a comment that it could be optimized in the future, but currently tonic does not offer a way to send a request and receive the raw bytes back without decoding them (at least not by using the generated code from the protos).
___
*[`cas/store/grpc_store.rs` line 327 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QI2e93VUYPkMQG8c9:-N_QI2e93VUYPkMQG8cA:b-7w8moy) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L327)):*
> ```XML
>             .send_eof()
>             .await
>             .err_tip(|| "Failed to write EOF in grpc store get_part")?;
> ```

nit: `s/get_part/get_action_result_as_part/`
___
*[`cas/store/grpc_store.rs` line 357 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QCTZbAIUklDyZn1Iy:-N_QCTZbAIUklDyZn1Iz:bvc044f) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L357)):*
> ```XML
>         if matches!(self.store_type, config::stores::StoreType::ac) {
>             // The length of an AC is incorrect, so we don't figure out the length, just return 1.
>             return self.get_action_result_from_digest(digest).await.map(|_| Some(1));
> ```

Is this code path even possible with AC?

Also, maybe return a size of `usize::MAX` or `0x8888888888888888` so for debugging we see a special number and can look it up.
___
*[`config/stores.rs` line 395 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QAKniAB3OQDTOprHv:-N_QAKniAB3OQDTOprHw:b-s63awr) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/config/stores.rs#L395)):*
> ```XML
> pub enum StoreType {
>     /// The store is content addressable storage.
>     #[default]
> ```

nit: Lets not default this (or above in the `derive`, or in `store_type`). The user should know what they are doing, or they might configure it wrong.
___
*[`config/stores.rs` line 396 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_Q9jy149Jv5u1L7gAW:-N_Q9jy149Jv5u1L7gAX:bmqkbvf) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/config/stores.rs#L396)):*
> ```XML
>     /// The store is content addressable storage.
>     #[default]
>     cas,
> ```

nit: Can we caps these? For example `CompressionAlgorithm::LZ4` is used for compression, as I find it looks prettier.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 357 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QCTZbAIUklDyZn1Iy:-N_U6V-i5NNKjSiUK_SZ:b29o1mu) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L357)):*

Is this code path even possible with AC?

Also, maybe return a size of `usize::MAX` or `0x8888888888888888` so for debugging we see a special number and can look it up.
</blockquote></details>

I was a little worried about using a large figure here in case it ended up getting used as a buffer size and an attempt to allocate all memory.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)* status: all files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`config/stores.rs` line 395 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QAKniAB3OQDTOprHv:-N_U6dqA7K8RRCVrcobI:b-oabh19) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/config/stores.rs#L395)):*

nit: Lets not default this (or above in the `derive`, or in `store_type`). The user should know what they are doing, or they might configure it wrong.
</blockquote></details>

I did it this way to be backwards compatible, are you happy with the breaking change?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 357 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QCTZbAIUklDyZn1Iy:-N_U73hF7VlHSW0I32oD:b-3i0tl7) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/cas/store/grpc_store.rs#L357)):*

I was a little worried about using a large figure here in case it ended up getting used as a buffer size and an attempt to allocate all memory.
</blockquote></details>

That's why I was thinking to use `usize::MAX` because it has a 100% chance of causing the program to crash if they try to trust the size info out of this response. I personally prefer the idea of "crash the program if the code is trusting something it shouldn't". In this case it'd crash on the first iteration of it trying to use the data.

allada (commented):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

___
*[`config/stores.rs` line 395 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QAKniAB3OQDTOprHv:-N_U8pdL9MQL-4iUuCY2:but4uoo) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/config/stores.rs#L395)):*

I did it this way to be backwards compatible, are you happy with the breaking change?
</blockquote></details>

In this case yes. I'd rather have users know what they are doing when they write the config file than to have them not realize there's a config that is important.

As long as `[Breaking Change]` is in the commit, I don't think breaking changes are a big deal, especially since there's no official release channels.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

___
*[`config/stores.rs` line 395 at r1](https://reviewable.io/reviews/allada/turbo-cache/184#-N_QAKniAB3OQDTOprHv:-N_U9lPm8iY3y0wKD01o:bcb9y07) ([raw file](https://github.com/allada/turbo-cache/blob/d0dba2aeff676b0b960498fde4ec4dba1be7c4d3/config/stores.rs#L395)):*

In this case yes. I'd rather have users know what they are doing when they write the config file than to have them not realize there's a config that is important.

As long as `[Breaking Change]` is in the commit, I don't think breaking changes are a big deal, especially since there's no official release channels.
</blockquote></details>

If we had official versions, I'd be more hesitant on this decision though.

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/184)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #189, opened 2023-07-16
Add TSAN suppression and harness

allada:
Bazel does not make it easy to add a file to every test as a dependency. But it does have --run_under, so we hijack this in order to make our suppression file. Later we may want to make a macro override for our test rules to add the suppression file if it grows too large.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/189)

Issue #188, opened 2023-07-16
Sanitizer tracking issue


aaronmondal:
Sanitizer integration in rust is still quite experimental and tends to produce false positives. I went through a bunch of logs and think the issues below could be real bugs. I've added a few points of interest to key points in the codebase after some initial disentangling of the error logs.

AddressSanitizer:

- [ ] A lot of leaked memory during `cas/store:ref_store_test`. Seems to occur in `get_test` and `update_test`. POI:
    - The `setup_stores` call in `update_test`, in `ref_store_owned`. Seemingly during some clone operations.
    - The `add_store` call in `setup_stores`.
    - The first scope in each of the tests.
    - The `ref_store` field in the `RefStore::new` implementation.
    - The `unsafe impl Sync for StoreReference {}`.
    - The `RefStore` struct.
    - The `name.to_string` call in `stores.insert` in `cas/store/lib.rs` in `add_store`.
    - Seems like everything points to the string keys of the hashmap. in the `stores.insert` call.
    - Related issues:
      - https://users.rust-lang.org/t/understand-memory-leak-with-parkinglot-and-hashmap/92519
      - https://github.com/rust-lang/rust/issues/73307

ThreadSanitizer:
- [ ] Data race in `cas/worker:running_actions_manager_test` in `cleanup_happens_on_job_failure`. POI:
    - Creation of `fast_store` in the `setup_stores` for these tests.
    - The `try_join` in `running_actions_manager.rs` in `upload_results`.
 
- [ ] Data race in `cas/worker:local_worker_test` in `new_local_worker_removes_work_directory_before_start_test`. POI:
    - The `new_local_worker` in the failing test.
    - The `fs::canonicalize` call in `new_local_worker` in `cas/worker/local_worker.rs`.

- [ ] Data race in `cas/store:filesystem_store_test` in `oldest_entry_evicted_with_access_times_loaded_from_disk`. POI:
    - The `match` statement at the end of the test on `store.get_file_entry_for_digest`.
    - The `fs::create_dir_all` call in the test.
    - The `write_file` calls in that test.

allada:
I fixed all the asan bugs here and enabled it in CI: #187

allada:
For reference those errors had nothing to do with the "unsafe" nature. It was purely in a circular reference where the `RefStore` owned an Arc of the `StoreManager` and the `StoreManager` owned an Arc of the `RefStore`.

allada:
[...]

PR #190, opened 2023-07-16
Fix false positives for `--config=tsan`.

aaronmondal:
Run tests under tsan in serial, and introduce a `.tsanignore` to selectively disable warnings on unfixable symbols.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/190)


aaronmondal:
Also for #188, but probably superceded by #189

allada:
Yeah, this is a better long term solution, but I'd rather be simple for now. I don't think the suppression file will get that large.

PR #198, opened 2023-07-17
Change StoreTrait::has to take HashSet.

chrisstaite-menlo:
Currently the `has` method in the StoreTrait takes a single DigestInfo which means that it only has to do a single lookup at a time.  This is optimised in places for the GrpcStore to ensure that efficient upstream lookups are performed.

However, if the GrpcStore is wrapped in another store (e.g. FastSlowStore for a local cache) then the `has` call is performed for each DigestInfo which is very inefficient on the GrpcStore call.

This optimises that use case by having each Store implementation pass through all DigestInfos that are being requested meaning the by-pass is no longer required for GrpcStore::has and all other stores can make use of the optimised route.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)

### DIFF
--- a/cas/grpc_service/cas_server.rs
+++ b/cas/grpc_service/cas_server.rs
@@ -19,7 +19,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use bytes::Bytes;
-use futures::{stream::FuturesUnordered, stream::Stream, StreamExt, TryStreamExt};
+use futures::{stream::FuturesUnordered, stream::Stream, TryStreamExt};
 use proto::google::rpc::Status as GrpcStatus;
 use tonic::{Request, Response, Status};
 
@@ -31,7 +31,7 @@ use proto::build::bazel::remote::execution::v2::{
     batch_read_blobs_response, batch_update_blobs_response,
     content_addressable_storage_server::ContentAddressableStorage,
     content_addressable_storage_server::ContentAddressableStorageServer as Server, BatchReadBlobsRequest,
-    BatchReadBlobsResponse, BatchUpdateBlobsRequest, BatchUpdateBlobsResponse, Digest, FindMissingBlobsRequest,
+    BatchReadBlobsResponse, BatchUpdateBlobsRequest, BatchUpdateBlobsResponse, FindMissingBlobsRequest,
     FindMissingBlobsResponse, GetTreeRequest, GetTreeResponse,
 };
 use store::{Store, StoreManager};
@@ -71,39 +71,19 @@ impl CasServer {
             .err_tip(|| format!("'instance_name' not configured for '{}'", instance_name))?
             .clone();
 
-        // If we are a GrpcStore we shortcut here, as this is a special store.
-        let any_store = store.clone().as_any();
-        let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
-        if let Some(grpc_store) = maybe_grpc_store {
-            return grpc_store.find_missing_blobs(Request::new(inner_request)).await;
+        let mut requested_blobs = Vec::with_capacity(inner_request.blob_digests.len());
+        for digest in inner_request.blob_digests.iter() {
+            requested_blobs.push(DigestInfo::try_from(digest.clone())?);
         }
-
-        let store_pin = Pin::new(store.as_ref());
-        let check_futures: FuturesUnordered<_> = inner_request
-            .blob_digests
+        let sizes = Pin::new(store.as_ref())
+            .has_many(&requested_blobs)
+            .await
+            .err_tip(|| "In find_missing_blobs")?;
+        let missing_blob_digests = sizes
             .into_iter()
-            .map(|digest| async move {
-                let digest_info = match DigestInfo::try_from(digest.clone()) {
-                    Ok(digest_info) => digest_info,
-                    Err(err) => return Some(Err(err)),
-                };
-                match store_pin.has(digest_info).await {
-                    Ok(maybe_size) => maybe_size.map_or(Some(Ok(digest)), |_| None),
-                    Err(err) => {
-                        log::error!(
-                            "Error during .has() call in .find_missing_blobs() : {:?} - {}",
-                            err,
-                            digest_info.hash_str()
-                        );
-                        Some(Ok(digest))
-                    }
-                }
-            })
+            .zip(inner_request.blob_digests)
+            .filter_map(|(maybe_size, digest)| maybe_size.map_or_else(|| Some(digest), |_| None))
             .collect();
-        let missing_blob_digests = check_futures
-            .filter_map(|result| async move { result })
-            .try_collect::<Vec<Digest>>()
-            .await?;
 
         Ok(Response::new(FindMissingBlobsResponse { missing_blob_digests }))
     }
--- a/cas/grpc_service/tests/bytestream_server_test.rs
+++ b/cas/grpc_service/tests/bytestream_server_test.rs
@@ -140,7 +140,10 @@ pub mod write_tests {
             assert_eq!(committed_size, raw_data.len());
 
             // Now lets check our store to ensure it was written with proper data.
-            store.has(DigestInfo::try_new(HASH1, raw_data.len())?).await?;
+            assert!(
+                store.has(DigestInfo::try_new(HASH1, raw_data.len())?).await?.is_some(),
+                "Not found in store",
+            );
             let store_data = store
                 .get_part_unchunked(DigestInfo::try_new(HASH1, raw_data.len())?, 0, None, None)
                 .await?;
--- a/cas/scheduler/cache_lookup_scheduler.rs
+++ b/cas/scheduler/cache_lookup_scheduler.rs
@@ -17,7 +17,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use futures::stream::{FuturesUnordered, StreamExt};
+use futures::stream::StreamExt;
 use tokio::select;
 use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
@@ -30,9 +30,7 @@ use error::Error;
 use grpc_store::GrpcStore;
 use parking_lot::{Mutex, MutexGuard};
 use platform_property_manager::PlatformPropertyManager;
-use proto::build::bazel::remote::execution::v2::{
-    ActionResult as ProtoActionResult, FindMissingBlobsRequest, GetActionResultRequest,
-};
+use proto::build::bazel::remote::execution::v2::{ActionResult as ProtoActionResult, GetActionResultRequest};
 use scheduler::ActionScheduler;
 use store::Store;
 
@@ -82,13 +80,11 @@ async fn get_action_from_store(
     }
 }
 
-async fn validate_outputs_exist(
-    cas_store: Arc<dyn Store>,
-    action_result: &ProtoActionResult,
-    instance_name: String,
-) -> bool {
+async fn validate_outputs_exist(cas_store: Arc<dyn Store>, action_result: &ProtoActionResult) -> bool {
     // Verify that output_files and output_directories are available in the cas.
-    let required_digests = action_result
+    let mut required_digests =
+        Vec::with_capacity(action_result.output_files.len() + action_result.output_directories.len());
+    for digest in action_result
         .output_files
         .iter()
         .filter_map(|output_file| output_file.digest.clone())
@@ -98,28 +94,17 @@ async fn validate_outputs_exist(
                 .iter()
                 .filter_map(|output_directory| output_directory.tree_digest.clone()),
         )
-        .collect();
-
-    // If the CAS is a GrpcStore store we can check all the digests in one message.
-    let any_store = cas_store.clone().as_any();
-    let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
-    if let Some(grpc_store) = maybe_grpc_store {
-        grpc_store
-            .find_missing_blobs(Request::new(FindMissingBlobsRequest {
-                instance_name,
-                blob_digests: required_digests,
-            }))
-            .await
-            .is_ok_and(|find_result| find_result.into_inner().missing_blob_digests.is_empty())
-    } else {
-        let cas_pin = Pin::new(cas_store.as_ref());
-        required_digests
-            .into_iter()
-            .map(|digest| async move { cas_pin.has(DigestInfo::try_from(digest)?).await })
-            .collect::<FuturesUnordered<_>>()
-            .all(|result| async { result.is_ok_and(|result| result.is_some()) })
-            .await
+    {
+        let Ok(digest) = DigestInfo::try_from(digest) else {
+            return false;
+        };
+        required_digests.push(digest);
     }
+
+    let Ok(sizes) = Pin::new(cas_store.as_ref()).has_many(&required_digests).await else {
+        return false;
+    };
+    sizes.into_iter().all(|size| size.is_some())
 }
 
 fn subscribe_to_existing_action(
@@ -187,7 +172,7 @@ impl ActionScheduler for CacheLookupScheduler {
             if let Some(proto_action_result) =
                 get_action_from_store(ac_store, current_state.action_digest(), instance_name.clone()).await
             {
-                if validate_outputs_exist(cas_store, &proto_action_result, instance_name).await {
+                if validate_outputs_exist(cas_store, &proto_action_result).await {
                     // Found in the cache, return the result immediately.
                     Arc::make_mut(&mut current_state).stage = ActionStage::CompletedFromCache(proto_action_result);
                     let _ = tx.send(current_state);
--- a/cas/store/BUILD
+++ b/cas/store/BUILD
@@ -83,6 +83,7 @@ rust_library(
         "//util:buf_channel",
         "//util:common",
         "//util:error",
+        "@crate_index//:tokio",
     ],
 )
 
--- a/cas/store/compression_store.rs
+++ b/cas/store/compression_store.rs
@@ -233,8 +233,14 @@ impl CompressionStore {
 
 #[async_trait]
 impl StoreTrait for CompressionStore {
-    async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
-        Pin::new(self.inner_store.as_ref()).has(digest).await
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
+        Pin::new(self.inner_store.as_ref())
+            .has_with_results(digests, results)
+            .await
     }
 
     async fn update(
--- a/cas/store/dedup_store.rs
+++ b/cas/store/dedup_store.rs
@@ -22,7 +22,7 @@ use bincode::{
     config::{FixintEncoding, WithOtherIntEncoding},
     DefaultOptions, Options,
 };
-use futures::stream::{self, StreamExt};
+use futures::stream::{self, FuturesOrdered, StreamExt, TryStreamExt};
 use futures::{future::try_join_all, FutureExt};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
@@ -96,10 +96,7 @@ impl DedupStore {
     fn pin_index_store(&self) -> Pin<&dyn StoreTrait> {
         Pin::new(self.index_store.as_ref())
     }
-}
 
-#[async_trait]
-impl StoreTrait for DedupStore {
     async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
         // First we need to load the index that contains where the individual parts actually
         // can be fetched from.
@@ -133,31 +130,47 @@ impl StoreTrait for DedupStore {
             }
         };
 
-        let mut stream = stream::iter(index_entries.entries)
-            .map(move |index_entry| {
-                let content_store = self.content_store.clone();
-                async move {
-                    let digest = DigestInfo::new(index_entry.packed_hash, index_entry.size_bytes);
-                    Pin::new(content_store.as_ref())
-                        .has(digest)
-                        .await
-                        .err_tip(|| "Failed to check .has() on content_store in dedup store")
-                }
-            })
-            .buffer_unordered(self.max_concurrent_fetch_per_get);
-
+        let digests: Vec<DigestInfo> = index_entries
+            .entries
+            .into_iter()
+            .map(|index_entry| DigestInfo::new(index_entry.packed_hash, index_entry.size_bytes))
+            .collect();
         let mut sum = 0;
-        while let Some(result) = stream.next().await {
-            let maybe_size = result?;
-            if let Some(size) = maybe_size {
-                sum += size;
-                continue;
-            }
-            // A part is missing so return None meaning not-found.
-            // This will abort all in-flight queries related to this request.
-            return Ok(None);
+        for size in Pin::new(self.content_store.as_ref()).has_many(&digests).await? {
+            let Some(size) = size else {
+                // A part is missing so return None meaning not-found.
+                // This will abort all in-flight queries related to this request.
+                return Ok(None);
+            };
+            sum += size;
         }
         Ok(Some(sum))
+    }
+}
+
+#[async_trait]
+impl StoreTrait for DedupStore {
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
+        digests
+            .iter()
+            .zip(results.iter_mut())
+            .map(|(digest, result)| async move {
+                match self.has(*digest).await {
+                    Ok(maybe_size) => {
+                        *result = maybe_size;
+                        Ok(())
+                    }
+                    Err(err) => Err(err),
+                }
+            })
+            .collect::<FuturesOrdered<_>>()
+            .try_collect()
+            .await?;
+        Ok(())
     }
 
     async fn update(
--- a/cas/store/fast_slow_store.rs
+++ b/cas/store/fast_slow_store.rs
@@ -84,13 +84,31 @@ impl FastSlowStore {
 
 #[async_trait]
 impl StoreTrait for FastSlowStore {
-    async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
-        // TODO(blaise.bruer) Investigate if we should maybe ignore errors here instead of
-        // forwarding the up.
-        if let Some(sz) = self.pin_fast_store().has(digest).await? {
-            return Ok(Some(sz));
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
+        self.pin_fast_store().has_with_results(digests, results).await?;
+        let missing_digests: Vec<DigestInfo> = results
+            .iter()
+            .zip(digests.iter())
+            .filter_map(|(result, digest)| result.map_or_else(|| Some(*digest), |_| None))
+            .collect();
+        if !missing_digests.is_empty() {
+            let mut slow_store_results = self.pin_slow_store().has_many(&missing_digests).await?.into_iter();
+            // We did not change our `result` order and the slow store's results are all the None
+            // entries in `results`. This means we can just iterate over the `results`, find the None
+            // items and put in whatever the next item in the results from `slow_store_results`.
+            for result in results.iter_mut() {
+                if result.is_none() {
+                    *result = slow_store_results
+                        .next()
+                        .err_tip(|| "slow_store_results out of sync with missing_digests")?;
+                }
+            }
         }
-        self.pin_slow_store().has(digest).await
+        Ok(())
     }
 
     async fn update(
@@ -170,16 +188,12 @@ impl StoreTrait for FastSlowStore {
         let sz_result = slow_store
             .has(digest)
             .await
-            .err_tip(|| "Failed to run has() on slow store");
-        let sz = match sz_result {
-            Ok(Some(sz)) => sz,
-            Ok(None) => {
-                return Err(make_err!(
-                    Code::NotFound,
-                    "Object not found in either fast or slow store"
-                ))
-            }
-            Err(err) => return Err(err),
+            .err_tip(|| "Failed to run has() on slow store")?;
+        let Some(sz) = sz_result else {
+            return Err(make_err!(
+                Code::NotFound,
+                "Object not found in either fast or slow store"
+            ));
         };
 
         let (mut fast_tx, fast_rx) = make_buf_channel_pair();
--- a/cas/store/filesystem_store.rs
+++ b/cas/store/filesystem_store.rs
@@ -565,8 +565,13 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
 
 #[async_trait]
 impl<Fe: FileEntry> StoreTrait for FilesystemStore<Fe> {
-    async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
-        Ok(self.evicting_map.size_for_key(&digest).await)
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
+        self.evicting_map.sizes_for_keys(digests, results).await;
+        Ok(())
     }
 
     async fn update(
--- a/cas/store/grpc_store.rs
+++ b/cas/store/grpc_store.rs
@@ -17,7 +17,8 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use bytes::BytesMut;
-use futures::{stream::unfold, Stream};
+use futures::stream::{unfold, FuturesUnordered};
+use futures::{future, Stream, TryStreamExt};
 use prost::Message;
 use tonic::{transport, IntoRequest, Request, Response, Streaming};
 use uuid::Uuid;
@@ -360,30 +361,56 @@ impl GrpcStore {
 impl StoreTrait for GrpcStore {
     // NOTE: This function can only be safely used on CAS stores. AC stores may return a size that
     // is incorrect.
-    async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
         if matches!(self.store_type, config::stores::StoreType::AC) {
-            // The length of an AC is incorrect, so we don't figure out the
-            // length, instead the biggest possible result is returned in the
-            // hope that we detect incorrect usage.
-            return self
-                .get_action_result_from_digest(digest)
+            digests
+                .iter()
+                .zip(results.iter_mut())
+                .map(|(digest, result)| async move {
+                    // The length of an AC is incorrect, so we don't figure out the
+                    // length, instead the biggest possible result is returned in the
+                    // hope that we detect incorrect usage.
+                    self.get_action_result_from_digest(*digest).await?;
+                    *result = Some(usize::MAX);
+                    Ok::<_, Error>(())
+                })
+                .collect::<FuturesUnordered<_>>()
+                .try_for_each(|_| future::ready(Ok(())))
                 .await
-                .map(|_| Some(usize::MAX));
+                .err_tip(|| "Getting upstream action cache entry")?;
+            return Ok(());
         }
 
-        let digest_size =
-            usize::try_from(digest.size_bytes).err_tip(|| "GrpcStore digest size cannot be converted to usize")?;
         let missing_blobs_response = self
             .find_missing_blobs(Request::new(FindMissingBlobsRequest {
                 instance_name: self.instance_name.clone(),
-                blob_digests: vec![digest.into()],
+                blob_digests: digests.iter().map(|digest| digest.into()).collect(),
             }))
             .await?
             .into_inner();
-        if missing_blobs_response.missing_blob_digests.is_empty() {
-            return Ok(Some(digest_size));
+
+        // Since the ordering is not guaranteed above, the matching has to check
+        // all missing blobs against all entries in the unsorted digest list.
+        // To optimise this, the missing digests are sorted and then it is
+        // efficient to perform a binary search for each digest within the
+        // missing list.
+        let mut missing_digests = Vec::with_capacity(missing_blobs_response.missing_blob_digests.len());
+        for missing_digest in missing_blobs_response.missing_blob_digests {
+            missing_digests.push(DigestInfo::try_from(missing_digest)?);
         }
-        return Ok(None);
+        missing_digests.sort_unstable();
+        for (digest, result) in digests.iter().zip(results.iter_mut()) {
+            match missing_digests.binary_search(digest) {
+                Ok(_) => *result = None,
+                Err(_) => *result = Some(usize::try_from(digest.size_bytes)?),
+            }
+        }
+
+        Ok(())
     }
 
     async fn update(
--- a/cas/store/memory_store.rs
+++ b/cas/store/memory_store.rs
@@ -67,8 +67,13 @@ impl MemoryStore {
 
 #[async_trait]
 impl StoreTrait for MemoryStore {
-    async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
-        Ok(self.map.size_for_key(&digest).await)
+    async fn has_with_results(
+        self: Pin<&Self>,
+        digests: &[DigestInfo],
+        results: &mut [Option<usize>],
+    ) -> Result<(), Error> {
+        self.map.sizes_for_keys(digests, results).await;
+        Ok(())
     }
 
     async fn update(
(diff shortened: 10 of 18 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 0 of 24 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

___
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N_Z8WU719mB5ByFatxt:b1aabs0) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*
> ```XML
>     /// digest to the size of the digest if it was in the store.  If it wasn't
>     /// in the store then it will not be in the output map.
>     async fn has(self: Pin<&Self>, digests: HashSet<DigestInfo>) -> Result<HashMap<DigestInfo, usize>, Error>;
> ```

I'm strugging with this one, here's my thoughts:
* HashSet is good if there's a high chance of collisions, but I our use case the cost of hashing is far more the overhead of possible collisions (resulting in double-lookup for stores that don't optimize it).
* Vec seems like the most logical choice, which is nice because we can just return a vec of the same size with our results.
* Passing a stream in and returning an async stream seems like an interesting idea too, however the only gain I can think of is the ability to abort the `.has()` request if the digest doesn't exist, which is a rare need and has a log of complexity.

Given this. I think passing in a (slice) `&[DigestInfo]` and returning a `Vec<Option<usize>>` seems like the best choice. The overhead of hashing `DigestInfo` is pretty high and doubt it really saves anything.

If we want to go overboard with optimizations we can this:
```rs
async fn has(self: Pin<&Self>, digests: DigestInfo) -> Result<Option<usize>, Error> {
  let mut result = None;
  self.has_with_pre_alloc_results(&[digest], &[result]).await?;
  return Some(result)
}

async fn has_many(self: Pin<&Self>, digests: &[DigestInfo]) -> Result<Vec<Option<usize>>, Error> {
  let mut results = Vec::with_capacity(digests.len());
  self.has_with_pre_alloc_results(digests, &mut results).await?;
  Some(results)
}

async fn has_with_pre_alloc_results(self: Pin<&Self>, digests: &[DigestInfo], results: &mut [Option<usize>]) -> Result<(), Error>;
```
This approach would give all the gains of both worlds, because we won't need any needless heap allocations for single `.has()` calls and a single one `.has_many()` as well as the ability for the caller to use `.has_with_alloc_results()` if they have a use case that is even more advanced.

chrisstaite-menlo:
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N_ZHmFe89oGX6Tf4BPR:bfhlnso) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*

I'm strugging with this one, here's my thoughts:
* HashSet is good if there's a high chance of collisions, but I our use case the cost of hashing is far more the overhead of possible collisions (resulting in double-lookup for stores that don't optimize it).
* Vec seems like the most logical choice, which is nice because we can just return a vec of the same size with our results.
* Passing a stream in and returning an async stream seems like an interesting idea too, however the only gain I can think of is the ability to abort the `.has()` request if the digest doesn't exist, which is a rare need and has a log of complexity.

Given this. I think passing in a (slice) `&[DigestInfo]` and returning a `Vec<Option<usize>>` seems like the best choice. The overhead of hashing `DigestInfo` is pretty high and doubt it really saves anything.

If we want to go overboard with optimizations we can this:
```rs
async fn has(self: Pin<&Self>, digests: DigestInfo) -> Result<Option<usize>, Error> {
  let mut result = None;
  self.has_with_pre_alloc_results(&[digest], &[result]).await?;
  return Some(result)
}

async fn has_many(self: Pin<&Self>, digests: &[DigestInfo]) -> Result<Vec<Option<usize>>, Error> {
  let mut results = Vec::with_capacity(digests.len());
  self.has_with_pre_alloc_results(digests, &mut results).await?;
  Some(results)
}

async fn has_with_pre_alloc_results(self: Pin<&Self>, digests: &[DigestInfo], results: &mut [Option<usize>]) -> Result<(), Error>;
```
This approach would give all the gains of both worlds, because we won't need any needless heap allocations for single `.has()` calls and a single one `.has_many()` as well as the ability for the caller to use `.has_with_alloc_results()` if they have a use case that is even more advanced.
</blockquote></details>

I originally wrote this using a `Vec`, but during implementation I found myself converting it to a `HashSet` all over the place in order to perform a `difference`.  Therefore, it made sense to simply use a `HashSet`.

chrisstaite-menlo:
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N_ZI3TC0EJonXjuYyIG:b-j09t8j) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*

I originally wrote this using a `Vec`, but during implementation I found myself converting it to a `HashSet` all over the place in order to perform a `difference`.  Therefore, it made sense to simply use a `HashSet`.
</blockquote></details>

Oh, I also decided on a `HashMap` return type to avoid having to perform the operations in order in the stores.  This gives the freedom to do things fast.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 0 of 24 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/memory_store.rs` line 71 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_ZM_Mw7omkQ_BdEio1:-N_ZM_Mw7omkQ_BdEio2:b-8krrn7) ([raw file](https://github.com/allada/turbo-cache/blob/857bf40e18ed2809f0a1bac35128261175f58da1/cas/store/memory_store.rs#L71)):*
> ```XML
> impl StoreTrait for MemoryStore {
>     async fn has(self: Pin<&Self>, digest: DigestInfo) -> Result<Option<usize>, Error> {
>         Ok(self.map.size_for_key(&digest).await)
> ```

It looks like this should be optimized too to take in multiple digests to prevent so many mutex locks.
___
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N_ZJv0LAqxoA4N4CBhK:b-j61fcg) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*

Oh, I also decided on a `HashMap` return type to avoid having to perform the operations in order in the stores.  This gives the freedom to do things fast.
</blockquote></details>

My specific concern is the most important hot path for this call is `find_missing_blobs()`, because it's called on every action before it's executed.

I just looked over all the calls in the PR to `.has()` that would be complicated. I feel most of the non-tests would prefer it as a vector. The ones that don't can easily use: `.flatten()` to remove all the `None` values or `.into_iter().collect::<HashSet<usize>>()`

My major concern here is the hashing cost of `HashMap` its quite expensive plus the way it does heap allocations is pretty bad, since it doesn't store the data contiguously in memory and can hold multiple heap pointers. Doing it this way requires two of them. And the last part is that `DigestInfo` is actually not that cheap, it's 40 bytes, since we are forcing a copy of the data as our result it means we are at least doubling the memory needs. So if we have 10,000 digests (I've seen millions of digests for a single action before), we are adding an additional 400k of memory usage just for the returned results. If we use a Vec it'll be (depending on how it optimizes `Option`), 80k (5x less).

Sorry to be such a stickler on this, I just care a lot about API calls and performance implications of them.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 0 of 24 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N_ZQ0dX8Nd8EjFpavDe:bjmfh4e) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*

That's fair, I'm happy with that evaluation.
</blockquote></details>

This might help you:

```rs
let digests: Vec<DigestInfo> = digests.into();

let missing_digests = digests
  .iter()
  .zip(store.has_many(&digests).await.into_iter())
  .filter_map(|(digest, sz)| sz.map_or_else(|| digest, || None))
```

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/store/memory_store.rs` line 71 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_ZM_Mw7omkQ_BdEio1:-N__QV4S8YLC4-OEZr7o:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/857bf40e18ed2809f0a1bac35128261175f58da1/cas/store/memory_store.rs#L71)):*

It looks like this should be optimized too to take in multiple digests to prevent so many mutex locks.
</blockquote></details>

Done.
___
*[`cas/store/store_trait.rs` line 47 at r1](https://reviewable.io/reviews/allada/turbo-cache/198#-N_Z8WU719mB5ByFatxs:-N__g2opDFA3SwQrUjBv:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/0c03fede313ed7b5ff80ab726f811230315e548f/cas/store/store_trait.rs#L47)):*

This might help you:

```rs
let digests: Vec<DigestInfo> = digests.into();

let missing_digests = digests
  .iter()
  .zip(store.has_many(&digests).await.into_iter())
  .filter_map(|(digest, sz)| sz.map_or_else(|| digest, || None))
```
</blockquote></details>

Done.

allada (changes requested):
> Reviewed 1 of 24 files at r1, 24 of 24 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: all files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/grpc_service/bytestream_server.rs` line 264 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bb5PiE-3OdccKczeP:-N_bb5Pj524edkuZ4RWk:b-34j85s) ([raw file](https://github.com/allada/turbo-cache/blob/857bf40e18ed2809f0a1bac35128261175f58da1/cas/grpc_service/bytestream_server.rs#L264)):*
> ```XML
> 
>         let digest = DigestInfo::try_new(resource_info.hash, resource_info.expected_size)?;
>         let result = tokio::spawn(async move { Pin::new(store_clone.as_ref()).has(digest).await })
> ```

I'm not 100% on this, but I believe I did this because I think all requests come in on the same spawn/thread and this was a way to force mult-threading. I'm not sure which model is better now that there's a lot more code with a more clear view. Some stores do call their own spawn.

I need to check to see if different grpc requests happen on the same spawn or get their own unique one on connection.
___
*[`cas/grpc_service/cas_server.rs` line 82 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bcexm3qC-vYcW2t0o:-N_bcexm3qC-vYcW2t0p:b-ws59nx) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/grpc_service/cas_server.rs#L82)):*
> ```XML
>             .into_iter()
>             .zip(inner_request.blob_digests.into_iter())
>             .filter_map(|(maybe_size, digest)| maybe_size.map_or_else(|| Some(digest), |_| None))
> ```

this is so much cleaner!
___
*[`cas/grpc_service/tests/bytestream_server_test.rs` line 144 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bcqMzC4nJcFRjTHdg:-N_bcqMzC4nJcFRjTHdh:b-rifp9r) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/grpc_service/tests/bytestream_server_test.rs#L144)):*
> ```XML
>             // Now lets check our store to ensure it was written with proper data.
>             assert_eq!(
>                 true,
> ```

nit: Just use `assert!()`
___
*[`cas/store/fast_slow_store.rs` line 101 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bf6DQE0QgCd6O_FLC:-N_bf6DQE0QgCd6O_FLD:bkgxol2) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/fast_slow_store.rs#L101)):*
> ```XML
>             .collect();
>         if !missing_digests.is_empty() {
>             let mut slow_store_results = self.pin_slow_store().has_many(&missing_digests).await?.into_iter();
> ```

this is one of the few cases where streaming the results out would save a bit of time.
___
*[`cas/store/fast_slow_store.rs` line 104 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bfe6l9HWp3KXlhSPc:-N_bfe6l9HWp3KXlhSPd:bpxvd40) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/fast_slow_store.rs#L104)):*
> ```XML
>             for result in results.iter_mut() {
>                 if result.is_none() {
>                     *result = slow_store_results.next().unwrap();
> ```

nit: Try to avoid using unwrap when possible, instead use `err_tip()` and question mark.

I'd rather it just return an error even rather than panic.
___
*[`cas/store/grpc_store.rs` line 384 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bgIKe5Uj8MeN1DSg4:-N_bgIKe5Uj8MeN1DSg5:bmatslh) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L384)):*
> ```XML
>                 .collect::<FuturesUnordered<_>>()
>                 .try_collect::<Vec<()>>()
>                 .await?;
> ```

nit: Lets annotate our error with `.err_tip()` saying we are in GrpcStore.
___
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_bi5zR18G0XFRQIZfh:b-sosxnx) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*
> ```XML
>             .into_inner();
> 
>         // TODO: Can we do any better than this?
> ```

Sill not that pretty, but at least now it's O(N * log(N))
```rs
if missing_blobs_response.missing_blob_digests.is_empty() {
  return Ok(());
}
let mut digests_and_results = digests.iter().zip(results.iter_mut()).collect::<Vec<_>>();
digests_and_results.sort_unstable_by_key(|&(a, _)|  a);

for missing_blobs_response.missing_blob_digests {
  let missing_digest = DigestInfo::try_from(missing_digest)?;
  let index = digests_and_results.binary_search_by_key(&missing_digest, |&(a, b)| a).map_err(|_| make_err!("Could not find digest in GrpcStore.has_with_results"))?;
  *digests_and_results[index].1 = None;
}

```
___
*[`util/evicting_map.rs` line 248 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_boOxG7OZZSZIxO7-X:-N_boOxG7OZZSZIxO7-Y:b-r7wz7a) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/util/evicting_map.rs#L248)):*
> ```XML
>             .collect();
>         drop(state);
>         for data in to_touch {
> ```

nit: use `futures::future::join_all(&mut to_touch).await` to parallelize this.

chrisstaite-menlo:
*[`cas/store/fast_slow_store.rs` line 101 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bf6DQE0QgCd6O_FLC:-N_cP_Zn7Aaj4TmXBYqc:b-sjdm29) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/fast_slow_store.rs#L101)):*

this is one of the few cases where streaming the results out would save a bit of time.
</blockquote></details>

It's probably not worth the effort though since collating the results at the end is not going to add a significant amount of effort.  It's a shame we can't pass the results in directly as a slice because they have to stay in order.  Ditto for the size partitioning store.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_cWR0w2nA_0ZY16QxL:b-4exdgx) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

The final implementation of this ends up with a lot of copying of DigestInfo.  Instead, I've tried to take a hint from this and construct two `BinaryHeap`s which then can be compared in order.
</blockquote></details>

I'm terrible at Big-O, but I believe this is now O(N)?  Constructing the two binary heaps are O(N) and then a single iteration over the two heaps.

chrisstaite-menlo:
*[`cas/grpc_service/bytestream_server.rs` line 264 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bb5PiE-3OdccKczeP:-N_c_rYC0cV_iLxIy1w6:b1sxrwj) ([raw file](https://github.com/allada/turbo-cache/blob/857bf40e18ed2809f0a1bac35128261175f58da1/cas/grpc_service/bytestream_server.rs#L264)):*

I'm not 100% on this, but I believe I did this because I think all requests come in on the same spawn/thread and this was a way to force mult-threading. I'm not sure which model is better now that there's a lot more code with a more clear view. Some stores do call their own spawn.

I need to check to see if different grpc requests happen on the same spawn or get their own unique one on connection.
</blockquote></details>

Reverted. Even if we do want to do this, should probably be a separate PR.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 430 at r5](https://reviewable.io/reviews/allada/turbo-cache/198#-N_ckASp8upKRtL_EbSJ:-N_ckASp8upKRtL_EbSK:bkgqihy) ([raw file](https://github.com/allada/turbo-cache/blob/b583215c04a014e8a72a542ecd4dd4b0d0f42615/cas/store/grpc_store.rs#L430)):*
> ```XML
>             }
>             if !found {
>                 return Err(make_input_err!("Missing digest not in request list"));
> ```

Seeing this in testing, so something appears to be wrong here.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_dCvHT6P_t94WG9KfU:bijxhmh) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

I'm terrible at Big-O, but I believe this is now O(N)?  Constructing the two binary heaps are O(N) and then a single iteration over the two heaps.
</blockquote></details>

Actually, it's a bit confusing because looking at the `BinaryHeap` documentation, pushing is somewhere between O(N) and O(1) and popping is O(log N).  So it might actually be better than I thought.

chrisstaite-menlo:
After this change, my upstream caching scheduler is working :). For some reason though, the remote action cache fast store doesn't populate :(. No idea why...

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 21 of 27 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_daYb83flN0RjOPtye:b9x91q9) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

Don't let the documentation fool you, in your example:
Constructing binary heap: `O(N * log N)`
Poping binary heap: `O(log N)`.

In your code it is creating 2 heaps and iterating them in tandem. This means we are inuring the `log N` cost twice, so it's `(construct-O(N * log(N) + M * log(M))) + iteration-O((N * M) * log(N))` (where M is the smaller of the two). Yours also creates 2 full copies.

In the example I gave it was `construction-O(N * log(N)) + iteration-O(M * log(N))` and memory of `O(M)`.

But in the simple Big-O, our implementations are the same, but in the expanded Big-O they are quite different.

I had this stuff drilled into me when I used to interview people at Google.

I can come by and change this later if you don't want to figure it out here.
</blockquote></details>

correction: mine is `and memory of `O(N)`.

allada (commented):
> Reviewed 2 of 3 files at r3, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 25 of 27 files reviewed, 4 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`util/common.rs` line 94 at r6](https://reviewable.io/reviews/allada/turbo-cache/198#-N_datXJDcXm1G4BiRyc:-N_datXJDcXm1G4BiRyd:b-gqi7ox) ([raw file](https://github.com/allada/turbo-cache/blob/88a185e1f8f9abc03dbd54194a008054cc80ad3d/util/common.rs#L94)):*
> ```XML
> impl Ord for DigestInfo {
>     fn cmp(&self, other: &Self) -> Ordering {
>         // Want the highest priority on top, but the lowest insert_timestamp.
> ```

nit: Remove this comment.
___
*[`util/evicting_map.rs` line 254 at r6](https://reviewable.io/reviews/allada/turbo-cache/198#-N_dcQBY0a4GcvZzvC7t:-N_dcQBY0a4GcvZzvC7u:b36bmvy) ([raw file](https://github.com/allada/turbo-cache/blob/88a185e1f8f9abc03dbd54194a008054cc80ad3d/util/evicting_map.rs#L254)):*
> ```XML
>             .map(|data| data.touch())
>             .collect::<FuturesUnordered<_>>()
>             .collect::<Vec<()>>()
> ```

nit: use `.for_each(|_|{}).await`

This will discard the data as it is received reducing a needless heap allocation.

allada (commented):
> Reviewed 1 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/fast_slow_store.rs` line 92 at r6](https://reviewable.io/reviews/allada/turbo-cache/198#-N_dddVM48DM5BNrm0lS:-N_dddVM48DM5BNrm0lT:b-i77h43) ([raw file](https://github.com/allada/turbo-cache/blob/88a185e1f8f9abc03dbd54194a008054cc80ad3d/cas/store/fast_slow_store.rs#L92)):*
> ```XML
>         results: &mut [Option<usize>],
>         // TODO(blaise.bruer) Investigate if we should maybe ignore errors here instead of
> ```

nit: We can remove this now. It's clear we need to forward errors.
___
*[`cas/store/fast_slow_store.rs` line 102 at r6](https://reviewable.io/reviews/allada/turbo-cache/198#-N_deyYf4Fkhm4wgKnVe:-N_deyYf4Fkhm4wgKnVf:bjxd8zi) ([raw file](https://github.com/allada/turbo-cache/blob/88a185e1f8f9abc03dbd54194a008054cc80ad3d/cas/store/fast_slow_store.rs#L102)):*
> ```XML
>         if !missing_digests.is_empty() {
>             let mut slow_store_results = self.pin_slow_store().has_many(&missing_digests).await?.into_iter();
>             for result in results.iter_mut() {
> ```

nit: Make a comment here saying:

```
// We did not change our `result` order and the slow store's results are all the None
// entries in `results`. This means we can just iterate over the `results`, find the None
// items and put in whatever the next item in the results from `slow_store_results`.
```

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198#-:-N_dfZMzDFzKaL2YFsGu:bnfp4nl)* status: 26 of 27 files reviewed, 6 unresolved discussions (waiting on @chrisstaite-menlo)

chrisstaite-menlo:

Interesting, are we sure it's from this PR?

Do you happen to have a simple example show casing it?
</blockquote></details>

I never had the action cache in a fast-slow store before now because it caused too many errors.  So it's probably some other incompatibility that I haven't figured out yet.

The simple repro is:

```
"UPSTREAM_AC": {
  "grpc": {
    "instance_name": "main",
    "endpoints": ["grpc://upstream"],
    "store_type": "AC"
  }
},
"LOCAL_AC": {
  "fast_slow": {
    "fast": {
      "filesystem": {
        ...
      }
    }.
    "slow": {
       "ref_store": {
          "name": "UPSTREAM_AC"
       }
    }
  }
}
```

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_dj9iWCMyCk9ZJwpZq:b-y8v921) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

correction: mine is `and memory of `O(N)`.
</blockquote></details>

The iteration on mine is not `O(N*M * log(N))` because it only iterates each heap once, right?  Plus it uses `into` so the memory usage is only double during construction.

The reason I wasn't happy with the binary search solution was that it required a copy of the digest for each iteration of the search algorithm because:
`digests\_and\_results.binary\_search\_by\_key(&missing\_digest, |&(a, b)| a)`
is actually:
`digests\_and\_results.binary\_search\_by\_key(&missing\_digest, |(a, b)| *a)`
if you want it to compile...

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`util/common.rs` line 94 at r6](https://reviewable.io/reviews/allada/turbo-cache/198#-N_datXJDcXm1G4BiRyc:-N_dkK8W7hwCVi5q0Uf8:b-ys21qo) ([raw file](https://github.com/allada/turbo-cache/blob/88a185e1f8f9abc03dbd54194a008054cc80ad3d/util/common.rs#L94)):*

nit: Remove this comment.
</blockquote></details>

It's almost like I copied this implementation from elsewhere :eyes:

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_dmB845WNyCcJ8JpRF:b-z6pmkx) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

The iteration on mine is not `O(N*M * log(N))` because it only iterates each heap once, right?  Plus it uses `into` so the memory usage is only double during construction.

The reason I wasn't happy with the binary search solution was that it required a copy of the digest for each iteration of the search algorithm because:
`digests\_and\_results.binary\_search\_by\_key(&missing\_digest, |&(a, b)| a)`
is actually:
`digests\_and\_results.binary\_search\_by\_key(&missing\_digest, |(a, b)| *a)`
if you want it to compile...
</blockquote></details>

Yeah i meant: `O((N+M) * log(N))`

As for the copy. Digests are cheap to copy now they are 40 bytes and that binary search make a single copy while it's iterating (so the `log(N)` part). The `|(a, b)| *a` is just a visitor and once it sees it, it throws it away. The entire thing is on the stack, so the compiler re-uses the same memory.

I can fix it later.

chrisstaite-menlo:
GitHub Actions appears to be having a terrible time downloading from crates.io

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 17 of 27 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 406 at r8](https://reviewable.io/reviews/allada/turbo-cache/198#-N_dq3zp1llCV3InGOmf:-N_dq3zp1llCV3InGOmg:b52m47t) ([raw file](https://github.com/allada/turbo-cache/blob/4d0e78bdbfcf1bbaedcd402e0ac83bb5c9ff2897/cas/store/grpc_store.rs#L406)):*
> ```XML
>             return Ok(());
>         }
> ```

nit: Early return here if `missing_blobs_response.missing_blob_digests.is_empty()`. This is a hot path, so it'll happen a lot.
___
*[`cas/store/grpc_store.rs` line 415 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_emTU62rI3hYGCNzEg:-N_emTU62rI3hYGCNzEh:b-d3et65) ([raw file](https://github.com/allada/turbo-cache/blob/c8584cf2a28705a89a4793338a748d36b8846bef/cas/store/grpc_store.rs#L415)):*
> ```XML
>         digests_and_results.sort_unstable_by_key(|(digest, _result)| *digest);
> 
>         for missing_digest in missing_blobs_response.missing_blob_digests {
> ```

Aaah, I just realized, we are setting setting all `results` to `None` above. This means iterating over `missing_blob_digests` vs iterating over `results` is the same cost.

So, lets move that `*result = Size(digest.size)` stuff above into this loop and only set it to none if we don't find the digest in `missing_blob_digests`. 

This will ensure our `N * log(N)` is always on `N = min(results, missing_digests)`. Currently it's on the `max()` of the two.

So something like:

```rs
for result in &mut results {
            let missing_digest = DigestInfo::try_from(missing_digest)?;
            if let Ok(index) = missing_digests.binary_search_by(|missing_digest| missing_digest.cmp(digest)) {
               *digest = missing_digests[index];
            } else {
                *digest = None;
            }
}
```

(and sort the `missing_digests` instead of all the results.

allada (commented):
> Reviewed 1 of 4 files at r7, 3 of 8 files at r9.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 21 of 27 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

allada (commented):
Yeah their site was also having issues early today too.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198#-:-N_eovpS0Arb08nYknwg:b-qlohuw)* status: 21 of 27 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

allada (commented):
> Reviewed 5 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 1 unresolved discussion (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_eqeyYDZjPC7zyINoE:brfdht0) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

Thanks for the explanation, happy to put it in, just wondered if there was a better way than sorting and then searching.  Seemed like two sorts would be faster in my head...

Please check my implementation of your above...  I'm worried this file doesn't have tests due to #200
</blockquote></details>

Yep seems to be good. I'm a bit of a stickler for performance in this project.
___
*[`cas/store/grpc_store.rs` line 406 at r8](https://reviewable.io/reviews/allada/turbo-cache/198#-N_dq3zp1llCV3InGOmf:-N_eqndO-JcEbrcBdzoQ:bxffls1) ([raw file](https://github.com/allada/turbo-cache/blob/4d0e78bdbfcf1bbaedcd402e0ac83bb5c9ff2897/cas/store/grpc_store.rs#L406)):*

That's already there??
</blockquote></details>

Ooops.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 415 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_emTU62rI3hYGCNzEg:-N_eqWLB5z4S7VrWzmDa:b-yszh9w) ([raw file](https://github.com/allada/turbo-cache/blob/c8584cf2a28705a89a4793338a748d36b8846bef/cas/store/grpc_store.rs#L415)):*

Aaah, I just realized, we are setting setting all `results` to `None` above. This means iterating over `missing_blob_digests` vs iterating over `results` is the same cost.

So, lets move that `*result = Size(digest.size)` stuff above into this loop and only set it to none if we don't find the digest in `missing_blob_digests`. 

This will ensure our `N * log(N)` is always on `N = min(results, missing_digests)`. Currently it's on the `max()` of the two.

So something like:

```rs
for result in &mut results {
            let missing_digest = DigestInfo::try_from(missing_digest)?;
            if let Ok(index) = missing_digests.binary_search_by(|missing_digest| missing_digest.cmp(digest)) {
               *digest = missing_digests[index];
            } else {
                *digest = None;
            }
}
```

(and sort the `missing_digests` instead of all the results.
</blockquote></details>

I've read that a few times and I don't understand what it's getting at.

We set all results to `Some(size)` and then in this loop we set them to `None`.

You're suggesting we iterate over all the results and then search for the missing ones and set them to `None`?  Isn't that what we're already doing...

I think you're suggesting something like:

```
missing_blobs_response.missing_blob_digests.sort_unstable();
for (digest, result) in digests.iter().zip(results.iter_mut()) {
    let missing_digest = DigestInfo::try_from(missing_digest)?;
    match missing_digests.binary_search_by(|missing_digest| missing_digest.cmp(digest)) {
        Ok(_) => *result = None,
        Err(_) => *result = Some(usize::try_from(digest.size_bytes)?),
    }
}
```

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`-- commits` line 2 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_es57k4x3bs1wl50ZE):*
Can we also update our commit to say it takes a slice now?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 415 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_emTU62rI3hYGCNzEg:-N_esLhG2mJGZgVjR-9w:b-x0kva1) ([raw file](https://github.com/allada/turbo-cache/blob/c8584cf2a28705a89a4793338a748d36b8846bef/cas/store/grpc_store.rs#L415)):*

I've read that a few times and I don't understand what it's getting at.

We set all results to `Some(size)` and then in this loop we set them to `None`.

You're suggesting we iterate over all the results and then search for the missing ones and set them to `None`?  Isn't that what we're already doing...

I think you're suggesting something like:

```
missing_blobs_response.missing_blob_digests.sort_unstable();
for (digest, result) in digests.iter().zip(results.iter_mut()) {
    let missing_digest = DigestInfo::try_from(missing_digest)?;
    match missing_digests.binary_search_by(|missing_digest| missing_digest.cmp(digest)) {
        Ok(_) => *result = None,
        Err(_) => *result = Some(usize::try_from(digest.size_bytes)?),
    }
}
```
</blockquote></details>

Yeah that's exactly what I'm suggesting.

It makes sure we are paying for the `N * log(N)` on the smaller count (only cache misses), and our iteration on the total size we are spending anyway.

It also means we don't need to make a copy of anything, currently we are making a copy of all digests and the pointer to the results. Now zero copies are needed.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 415 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_emTU62rI3hYGCNzEg:-N_etDfN67B7_op4wQok:b-g1xku) ([raw file](https://github.com/allada/turbo-cache/blob/c8584cf2a28705a89a4793338a748d36b8846bef/cas/store/grpc_store.rs#L415)):*

Yeah that's exactly what I'm suggesting.

It makes sure we are paying for the `N * log(N)` on the smaller count (only cache misses), and our iteration on the total size we are spending anyway.

It also means we don't need to make a copy of anything, currently we are making a copy of all digests and the pointer to the results. Now zero copies are needed.
</blockquote></details>

Not quite zero, we need to map all the `missing_blob_digests` to `DigestInfo`, see commit...

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 402 at r2](https://reviewable.io/reviews/allada/turbo-cache/198#-N_bjFnRC4av3Ce5I8T9:-N_etRo2BVKTVDpbBtLN:bpacb0y) ([raw file](https://github.com/allada/turbo-cache/blob/a5be086c9e0db852a7c3c1bd6323ac1bf332a225/cas/store/grpc_store.rs#L402)):*

Yep seems to be good. I'm a bit of a stickler for performance in this project.
</blockquote></details>

It is kinda the point of it...

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 25 of 27 files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`-- commits` line 2 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_es57k4x3bs1wl50ZE):*

Can we also update our commit to say it takes a slice now?
</blockquote></details>

Done.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 25 of 27 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 415 at r9](https://reviewable.io/reviews/allada/turbo-cache/198#-N_emTU62rI3hYGCNzEg:-N_etxVN92WoxpjFr7Kv:b-dk4wcg) ([raw file](https://github.com/allada/turbo-cache/blob/c8584cf2a28705a89a4793338a748d36b8846bef/cas/store/grpc_store.rs#L415)):*

Not quite zero, we need to map all the `missing_blob_digests` to `DigestInfo`, see commit...
</blockquote></details>

Aaah yeah. Almost zero  :laughing:

allada (approved):
:lgtm:

> Reviewed 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198#-:-N_euMmEAzAny9FgldFa:bnfp4nl)* status: 26 of 27 files reviewed, 2 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/store/grpc_store.rs` line 407 at r10](https://reviewable.io/reviews/allada/turbo-cache/198#-N_eu91F4G54srLAv-PW:-N_eu91F4G54srLAv-PX:b-sm5iks) ([raw file](https://github.com/allada/turbo-cache/blob/85161105077b935e6d84035ae97a50e2e5447a0b/cas/store/grpc_store.rs#L407)):*
> ```XML
>         missing_digests.sort_unstable();
>         for (digest, result) in digests.iter().zip(results.iter_mut()) {
>             match missing_digests.binary_search_by(|missing_digest| missing_digest.cmp(digest)) {
> ```

I'm not sure if this is tested, but we need to be very careful of which direction this `cmp()` goes, (up or down) if it is the wrong direction it'll say nearly everything is missing.

chrisstaite-menlo:
*[`cas/store/grpc_store.rs` line 407 at r10](https://reviewable.io/reviews/allada/turbo-cache/198#-N_eu91F4G54srLAv-PW:-N_euRJFC9wt3a1zOWY7:bpiuoyt) ([raw file](https://github.com/allada/turbo-cache/blob/85161105077b935e6d84035ae97a50e2e5447a0b/cas/store/grpc_store.rs#L407)):*

Oh great, you removed it!
</blockquote></details>

Yeah, I haven't got a test for this, but I had the same thought... hence the change to using `binary_search` rather than `binary_search_by` so it should work in the same order!

allada (commented):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: 26 of 27 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

chrisstaite-menlo (commented):
> Reviewed 13 of 24 files at r2, 1 of 3 files at r3, 1 of 4 files at r4, 3 of 3 files at r5, 1 of 4 files at r7, 7 of 8 files at r9, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/198)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

PR #213, opened 2023-07-19
Add metrics to LocalWorker and RunningActionsManager

allada:
Adds the states to LocalWorker and RunningActionManager and adds a bunch of framework to make publishing stats about futures and other related items easier.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213)


allada (commented):
I suspect this one will be a tough one, but here it is:

+@chrisstaite-menlo 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213#-:-N_kJG--9PlWGPSAEkXN:b-m3rnu4)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @chrisstaite-menlo)

*[a discussion](https://reviewable.io/reviews/allada/turbo-cache/213#-N_kW211EaYbPM-owIee:-N_kW211EaYbPM-owIef:b9lp20g) (no related file):*
Tested with remote execution locally and it appears to be working. Bazel reported `949` actions were ran remote and the stats said it processed `949` items with zero active.

___

chrisstaite-menlo (changes requested):
Since workers don't usually expose a server, should they not publish their metrics to the scheduler?

> Reviewed 7 of 10 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213#-:-N_miqtg0y8rus1IAnEd:bz7x0i9)* status: all files reviewed, 5 unresolved discussions (waiting on @allada)

___
*[`cas/cas_main.rs` line 68 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_miWdT93xo_wkP3bIU:-N_miWdT93xo_wkP3bIV:b1fcey4) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/cas_main.rs#L68)):*
> ```XML
>     let mut root_metrics_registry = <Registry>::with_prefix("turbo_cache");
> 
>     let cfg: CasConfig = {
> ```

Kinda wonder if these scoped blocks ought to be functions?
___
*[`cas/worker/running_actions_manager.rs` line 458 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mjpy_89xfSSL-Y0YC:-N_mjpy_89xfSSL-Y0YD:b-z0nvya) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L458)):*
> ```XML
>     /// up to the stores to rate limit if needed.
>     async fn prepare_action(self: Arc<Self>) -> Result<Arc<Self>, Error> {
>         self.metrics()
> ```

That's a lot of indent.  perhaps split to `prepare_action` and `prepare_action_impl`?
___
*[`cas/worker/running_actions_manager.rs` line 597 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mk5ZDBMNvrIPzELKR:-N_mk5ZDBMNvrIPzELKS:b-8km8v3) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L597)):*
> ```XML
>             let mut killed_action = false;
> 
>             let timer = self.metrics().child_process.begin_timer();
> ```

Seems like we're doing the timing twice, could we use the existing action timer?  Or populate the action timing fields with the ones recorded for metrics?
___
*[`cas/worker/running_actions_manager.rs` line 620 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mkJY4AkLS7nDLko3s:-N_mkJY4AkLS7nDLko3t:bajgbsc) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L620)):*
> ```XML
>                                 exit_code
>                             } else {
>                                 EXIT_CODE_FOR_SIGNAL
> ```

Should be have a metric for this case too?  Or perhaps increment failure?
___
*[`cas/worker/running_actions_manager.rs` line 1022 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mm9lc2qMpxrcSvdZK:-N_mm9lc2qMpxrcSvdZL:b2hh64n) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L1022)):*
> ```XML
>         start_execute: StartExecute,
>     ) -> Result<Arc<RunningActionImpl>, Error> {
>         // Note: We have to `.clone()` something here because we capture `self` in the closure
> ```

That's sad, I wish the borrow checker was less finikity about something you `.await`.  It feels like it shouldn't be unsafe if it is cancelled, right?
___
*[`util/prometheus_utils.rs` line 249 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mnrMV3EihXyM-dQDE:-N_mnrMV3EihXyM-dQDF:bb70w76) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/util/prometheus_utils.rs#L249)):*
> ```XML
>         // By default `drop_counter` will increment the drop counter when it goes out of scope.
>         // This will ensure we don't increment the counter if we make it here with a zero cost.
>         forget(drop_counter);
> ```

Nice

allada (commented):
In such a case, you'd expose a private port that only has the `prometheus` service enabled.

Prometheus is a pulling system, so upon request the stats are delivered, it doesn't push the metrics anywhere.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213#-:-N_n_5f3-ZlXgtvf5mSB:b6r0yhd)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/cas_main.rs` line 68 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_miWdT93xo_wkP3bIU:-N_n_U7U5Hfdp4mKLveN:b-dd5nnf) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/cas_main.rs#L68)):*

Kinda wonder if these scoped blocks ought to be functions?
</blockquote></details>

I use this a lot as inline functions. I don't like making functions for things that are only called once (with some exceptions).
___
*[`cas/worker/running_actions_manager.rs` line 597 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mk5ZDBMNvrIPzELKR:-N_naBbLE824Mc6_LSnJ:b-je7bcl) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L597)):*

Seems like we're doing the timing twice, could we use the existing action timer?  Or populate the action timing fields with the ones recorded for metrics?
</blockquote></details>

The internals of `begin_timer()` do other things like count how many times it was ran.
We only store the accumulated running time, not the individual timing stats.

I also plan on making metrics gathering disablable via ENV or config, which would cause weird things in this code path if we tried to combine the two.
___
*[`cas/worker/running_actions_manager.rs` line 620 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mkJY4AkLS7nDLko3s:-N_nbYB22DRu7c2n2R7s:b-efeznq) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L620)):*

Should be have a metric for this case too?  Or perhaps increment failure?
</blockquote></details>

We kinda track this in the `kill_all()` method. I'm kinda torn on this one. I'd say we add it if we find we could use it. I'm imagining what the documented differences between the two metrics would be.
___
*[`cas/worker/running_actions_manager.rs` line 1022 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mm9lc2qMpxrcSvdZK:-N_nbtSz8zB7hqs87baS:br78bo8) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L1022)):*

That's sad, I wish the borrow checker was less finikity about something you `.await`.  It feels like it shouldn't be unsafe if it is cancelled, right?
</blockquote></details>

Actually, after re-reading the code, I can just change the function signature to a reference since we already clone the Arc anyway.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213)* status: 12 of 14 files reviewed, 3 unresolved discussions (waiting on @chrisstaite-menlo)

___
*[`cas/worker/running_actions_manager.rs` line 458 at r2](https://reviewable.io/reviews/allada/turbo-cache/213#-N_mjpy_89xfSSL-Y0YC:-N_ni87x-LLmJKo68nJA:b-896fix) ([raw file](https://github.com/allada/turbo-cache/blob/8fc4f736301a11794c1b63d8fa00306550d511ee/cas/worker/running_actions_manager.rs#L458)):*

That's a lot of indent.  perhaps split to `prepare_action` and `prepare_action_impl`?
</blockquote></details>

Done.

chrisstaite-menlo (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/213)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #216, opened 2023-07-21
Add metrics to SimpleScheduler and Worker

allada:
Adds metrics for running tasks and such to prometheus.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)


allada (commented):
+@aaronmondal do you want to take a swing at reviewing this one? Might be the biggest one of the batch.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216#-:-N_pts5sCPQGKqSoI9vF:bmb68g1)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @aaronmondal)

allada:
cc: @chrisstaite-menlo, this should give you the metrics you were wanting.

aaronmondal (commented):
Only got through the first half. Will get back to you on the rest in a bit.

> Reviewed 6 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216#-:-N_thrIICX-BGu35s_8_:b-ump284)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @allada)

___
*[`cas/scheduler/default_scheduler_factory.rs` line 102 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tf4D81LH6xqJZObnR:-N_tf4D9-cWhunxoWNyf:bnjmel1) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/default_scheduler_factory.rs#L102)):*
> ```XML
>             // has already been visited, not just the trait. `Any` could be used, but that'd require some rework
>             // of all the schedulers. This is the most simple way to do it. Rust's uintptr_t is usize.
>             let action_scheduler_uintptr: usize = Arc::as_ptr(action_scheduler) as *const () as usize;
> ```

I get why it's done this way but we should probably track this in an issue if there isn't one already.
___
*[`cas/scheduler/simple_scheduler.rs` line 237 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tg3Em3cXc-j7Yk61y:-N_tg3Em3cXc-j7Yk61z:bfu4ugy) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L237)):*
> ```XML
>         // Check to see if the action is running, if it is and cacheable, merge the actions.
>         if let Some(running_action) = self.active_actions.get_mut(&action_info) {
>             self.metrics.add_action_joined_running_action.inc();
> ```

Nit: I think it would be nice to name this `increment`. I wasn't sure whether this was something like "include" and had to look up the implementation :sweat_smile: 
___
*[`cas/scheduler/simple_scheduler.rs` line 686 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tgllP0nzYU4STr4rq:-N_tgllP0nzYU4STr4rr:b8cffr) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L686)):*
> ```XML
>     }
> 
>     fn get_inner_lock(&self) -> MutexGuard<'_, SimpleSchedulerImpl> {
> ```

I'll get back to you on this one and the wrapping calls. I need to think about those for a bit.
___
*[`cas/scheduler/simple_scheduler.rs` line 832 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_thQAG-QexxosOFIr7:-N_thQAG-QexxosOFIr8:b-ovo6v9) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L832)):*
> ```XML
> }
> 
> impl MetricsComponent for SimpleScheduler {
> ```

Since this code is so "uniform" would it make sense to move all those MetricsComponent impls into a separate file? I like the way it's done in general a lot though.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @aaronmondal)

___
*[`cas/scheduler/default_scheduler_factory.rs` line 102 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tf4D81LH6xqJZObnR:-N_xsswFCUXRhZ0AGzl4:b-xs2jjm) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/default_scheduler_factory.rs#L102)):*

I get why it's done this way but we should probably track this in an issue if there isn't one already.
</blockquote></details>

It's not really an issue. I'm not aware of a way to do this without using the raw pointers because they are two completely different interfaces, but might be the same object.
___
*[`cas/scheduler/simple_scheduler.rs` line 237 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tg3Em3cXc-j7Yk61y:-N_xt8_0CKfu_MCUvWYi:bnqn6wt) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L237)):*

Nit: I think it would be nice to name this `increment`. I wasn't sure whether this was something like "include" and had to look up the implementation :sweat_smile: 
</blockquote></details>

You could make the argument `.inc()` could mean `.include()`. I thought about `.measure()`, but i liked how short this function is to reduce the chance it'd get put on 2 lines.
___
*[`cas/scheduler/simple_scheduler.rs` line 686 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_tgllP0nzYU4STr4rq:-N_xtRlw4kDIgYO_4FFP:b-pp1b2d) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L686)):*

I'll get back to you on this one and the wrapping calls. I need to think about those for a bit.
</blockquote></details>

FWIW, 99% of the time those wrappers should get inlined.
___
*[`cas/scheduler/simple_scheduler.rs` line 832 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-N_thQAG-QexxosOFIr7:-N_xsEd9EcE6KAZAfWxP:bejxfkx) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/simple_scheduler.rs#L832)):*

Since this code is so "uniform" would it make sense to move all those MetricsComponent impls into a separate file? I like the way it's done in general a lot though.
</blockquote></details>

I thought about it, but then I realized that because the code is so uniform it is easy for people to realize they are in metrics publication code and scroll back to where they want.

I like the idea of moving it to a new file, but I also don't like the idea of  making people question where the metrics for this file are. This file is touched a lot, so we will see if the developer experience sucks because of this. I personally use cold-folding IDEs, so it is easy for me to ignore large sections of code.

chrisstaite-menlo (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)* status: all files reviewed, 3 unresolved discussions (waiting on @aaronmondal and @allada)

___
*[`cas/scheduler/default_scheduler_factory.rs` line 105 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-NaBW7tc9cH42jSKzunf:-NaBW7tc9cH42jSKzung:buw9ui6) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/default_scheduler_factory.rs#L105)):*
> ```XML
>             if !visited_schedulers.contains(&action_scheduler_uintptr) {
>                 visited_schedulers.insert(action_scheduler_uintptr);
>                 action_scheduler.clone().register_metrics(scheduler_metrics);
> ```

You could just call `register_metrics` on the scheduler when it's constructed above instead of all of this...  That wouldn't require a map or anything.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)* status: all files reviewed, 3 unresolved discussions (waiting on @aaronmondal and @chrisstaite-menlo)

___
*[`cas/scheduler/default_scheduler_factory.rs` line 105 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-NaBW7tc9cH42jSKzunf:-NaCkjkz89vAnWe42-0r:bdad7f3) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/default_scheduler_factory.rs#L105)):*

You could just call `register_metrics` on the scheduler when it's constructed above instead of all of this...  That wouldn't require a map or anything.
</blockquote></details>

Yeah, I did consider that. The major reason I liked doing it this way is because it is more standard to how large projects manage it. Because there's a decent chance in the future we may make `unregister_metrics()` to allow temporary attachment of metrics for debugging. It's more about long-term flexibility.

chrisstaite-menlo:
*[`cas/scheduler/default_scheduler_factory.rs` line 105 at r1](https://reviewable.io/reviews/allada/turbo-cache/216#-NaBW7tc9cH42jSKzunf:-NaHj4tn1hjw8-D7vaks:b-w3wkrm) ([raw file](https://github.com/allada/turbo-cache/blob/ebcde037b6034d155e073b920b4d17ff61ddf00a/cas/scheduler/default_scheduler_factory.rs#L105)):*

Yeah, I did consider that. The major reason I liked doing it this way is because it is more standard to how large projects manage it. Because there's a decent chance in the future we may make `unregister_metrics()` to allow temporary attachment of metrics for debugging. It's more about long-term flexibility.
</blockquote></details>

Is it worth complexity now for possible future expansion?  I don't mind either way, and it's only during start-up, so it's not like it's a major issue.

chrisstaite-menlo (approved):
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)* status: all files reviewed, 2 unresolved discussions (waiting on @aaronmondal)

allada (commented):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/216)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

PR #235, opened 2023-08-19
Add windows support

allada:
Officially windows is now supported.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/235)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,18 @@ on:
     branches: [ main ]
 
 jobs:
+  windows-tests:
+    # The type of runner that the job will run on.
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v3.5.3
+      with:
+        fetch-depth: 0
+    - name: Compile on windows
+      run: cargo build --all
+    - name: Test on windows
+      run: cargo test --all
+
   cargo-tests:
     # The type of runner that the job will run on.
     runs-on: ubuntu-22.04
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Turbo Cache is a project that implements the [Bazel Remote Execution protocol](h
 
 When properly configured this project will provide extremely fast and efficient build cache for any systems that communicate using the [BRE protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto) and/or extremely fast, efficient and low foot-print remote execution capability.
 
+Unix based operating systems and Windows are fully supported.
+
 ## TL;DR
 To compile and run the server:
 ```sh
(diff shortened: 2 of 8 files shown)


allada (commented):
+@aaronmondal 

Uuuf, this was not fun testing this, but finally got all tests working. Luckily most of the ugly stuff is in the tests.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/235#-:-NcBNGZy9FRarBKsbHcX:bs7smdd)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @aaronmondal)

aaronmondal (approved):
I can't really tell whether the Windows things are correct, but since tests pass modulo some CI job this is fine i guess :sweat_smile:

Although this makes me think: Is it actually desirable to have the added maintenance burden of native windows support in the first place? I might be biased, but everyone i know that develops on windows uses WSL2, i.e. runs something like Ubuntu.

I'm not familiar with Goma, but at least for Bazel, I'd expect users that build prominent projects like Tensorflow/Jax/Torch for windows in a remote exec/caching setup do so on linux servers and just run tests on windows. This use case only requires setting the `--remote_cache`/`--remote_execution` grpc endpoint in a config file but doesn't require running a turbo-cache instance on the windows machine. Bazel rules for "windows"y languages like C#/.NET aren't up to par with native tooling and I'd expect a very small minority of users to actually use those.

I'd argue that users running turbocache in e.g. a cloud with *actual windows servers* probably have more pressing issues than efficient caching :rofl:
:lgtm:

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/235#-:-NcDRZ-Y874lussBbKB8:bi18rzi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

aaronmondal (approved):
Hmm I might be wrong after all. Looks like users try to avoid crosscompilation and instead use actual windows runners for windows builds :smiling_face_with_tear: 

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/235#-:-NcDV64C3wOMyXwEroSq:b-y37nwz)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

allada (commented):
Cross compiling for windows is complicated because of licensing issues, so most companies avoid it. Most devs cross-compile to make it easier to develop.

But the major reason is because with remote execution you don't need to cross-compile, you just specify the platform_properties to windows and your cross compiling is actually doing the true compiling on native hardware.

And finally, I've been talking to Mongo about using TurboCache and this was a requirement. They want to run all tests on native windows. The compiling is not the big thing, it's the test execution.

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/235#-:-NcEvrLp4JCEgi_eGnsL:b-evyopg)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #253, opened 2023-09-08
add a few depdencies.

MarcusSorealheis:
From last night's conversation. Seems like these two would be important.

- - -
This change is [Reviewable](https://reviewable.io/reviews/allada/turbo-cache/253)


allada (commented):
I was planning on pushing some changes to the dockerfile that account for this problem...

see:
https://reviewable.io/reviews/allada/turbo-cache/254

> *[Reviewable](https://reviewable.io/reviews/allada/turbo-cache/253#-:-Ndqeyhg3KySQSaWsWCc:br10rzs)* status: 0 of 1 files reviewed, all discussions resolved

chrisstaite-menlo:
By default these should not be part of the build image as that does not create hermetic builds.   It is sometimes necessary for some environments, but it should not be the standard.

MarcusSorealheis:
Yea, I agree completely. We were discussing the project, stepping through some issues. 

There are other hermeticity considerations as well. This PR is definitely close-able.

PR #292, opened 2023-09-19
Introduce get_part_ref() and migrate primary use to .get_part()

allada:
Due to DropCloserWriteHalf being so fragile in how it is dropped, it was decided that we should prefer to pass an owned version to the store api and introduce `.get_part_ref()` to handle cases where the writer should not be dropped on the return.

This will protect from cases where users join!() the reader and writer sides, but a deadlock happens because the writer errors but doesn't drop, notifying the reader side.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/292)

### DIFF
--- a/cas/grpc_service/bytestream_server.rs
+++ b/cas/grpc_service/bytestream_server.rs
@@ -244,7 +244,7 @@ impl ByteStreamServer {
 
         let digest = DigestInfo::try_new(resource_info.hash, resource_info.expected_size)?;
 
-        let (mut tx, rx) = buf_channel::make_buf_channel_pair();
+        let (tx, rx) = buf_channel::make_buf_channel_pair();
 
         struct ReaderState {
             max_bytes_per_stream: usize,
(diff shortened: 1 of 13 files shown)


allada (commented):
+@MarcusSorealheis 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/292#-:-NeiLd0AEv3H4metPT1S:b2iyrky)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @MarcusSorealheis)

MarcusSorealheis (commented):
I could see how this could introduce race conditions or deadlocks. So that makes sense. 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/292#-:-NeiLtKo1UyCZjx8H6Lf:b-8hg7kc)* status: 0 of 13 files reviewed, all discussions resolved

MarcusSorealheis (approved):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/292)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @allada)

PR #315, opened 2023-10-17
Make memory_store_test compatible with Windows

aaronmondal:
Increase the precision of this test from 32 to 64 bit, prevent potential conversion issues and make the test compatible with Windows.

This commit doesn't fix the flakiness of the test. We still expect a maximum resident set size increase of 10%.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/315)


aaronmondal:
FYI this is the `memory-stats` implementation: https://github.com/Arc-blroth/memory-stats/blob/main/src/linux.rs
(and docs: https://docs.rs/memory-stats/latest/memory_stats/)

The linux variant now uses `/proc/self/smaps` instead of `/proc/self/statm` by default and only uses the latter as fallback. This has performance implications but it's also more precise and could actually lead to less flakiness in this test (https://stackoverflow.com/questions/30744333/linux-os-proc-pid-smaps-vs-proc-pid-statm).

PR #319, opened 2023-10-19
Add nix flake

aaronmondal:
This setup introduces a reproducible development environment with Cargo, Bazel and a few cloud CLIs for AWS and GCP.

It also introduces a fully static build for the cas executable that can be installed by nix users without having to clone the repository.


aaronmondal:
Missing CI and Docs. Will split this into smaller changes to make it easier to see what is going on.

allada:
We landed a variation of this, right?

aaronmondal:
Yep. Addressed by #442

PR #341, opened 2023-10-25
Add optimistic channels

### LINKED ISSUE #304: S3 store closes connections early
There's a bug in the S3 store where if the digest is small it will upload it as a one-shot for efficiency, however it requires the size to be known. Once that exact size of bytes is received it drops stream causing an upstream error to the sender.

aaronmondal:
This variation on make_buf_channel_pair omits the EOF check. This works around the fact that S3 drops connections as soon as all data is sent or received.

Fixes #304 and fixes the same issue for non-oneshot cases where a similar issue occurred as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341)

### DIFF
--- a/cas/store/s3_store.rs
+++ b/cas/store/s3_store.rs
@@ -110,10 +110,7 @@ where
         // HTTP-level errors. Sometimes can retry.
         Err(RusotoError::Unknown(e)) => match e.status {
             StatusCode::NOT_FOUND => RetryResult::Err(make_err!(Code::NotFound, "{}", e.status.to_string())),
-            StatusCode::INTERNAL_SERVER_ERROR => {
-                RetryResult::Retry(make_err!(Code::Unavailable, "{}", e.status.to_string()))
-            }
-            StatusCode::SERVICE_UNAVAILABLE => {
+            StatusCode::INTERNAL_SERVER_ERROR | StatusCode::SERVICE_UNAVAILABLE => {
                 RetryResult::Retry(make_err!(Code::Unavailable, "{}", e.status.to_string()))
             }
             StatusCode::CONFLICT => RetryResult::Retry(make_err!(Code::Unavailable, "{}", e.status.to_string())),
@@ -153,7 +150,7 @@ impl S3Store {
             S3Client::new_with(dispatcher, credentials_provider, region)
         };
         let jitter_amt = config.retry.jitter;
-        S3Store::new_with_client_and_jitter(
+        Self::new_with_client_and_jitter(
             config,
             s3_client,
             Box::new(move |delay: Duration| {
@@ -172,12 +169,12 @@ impl S3Store {
         s3_client: S3Client,
         jitter_fn: Box<dyn Fn(Duration) -> Duration + Send + Sync>,
     ) -> Result<Self, Error> {
-        Ok(S3Store {
+        Ok(Self {
             s3_client: Arc::new(s3_client),
             bucket: config.bucket.to_string(),
-            key_prefix: config.key_prefix.as_ref().unwrap_or(&"".to_string()).to_owned(),
+            key_prefix: config.key_prefix.as_ref().unwrap_or(&String::new()).clone(),
             jitter_fn,
-            retry: config.retry.to_owned(),
+            retry: config.retry.clone(),
             retrier: Retrier::new(Box::new(|duration| Box::pin(sleep(duration)))),
         })
     }
@@ -196,8 +193,8 @@ impl S3Store {
                 retry_config,
                 unfold((), move |state| async move {
                     let head_req = HeadObjectRequest {
-                        bucket: self.bucket.to_owned(),
-                        key: s3_path.to_owned(),
+                        bucket: self.bucket.clone(),
+                        key: s3_path.clone(),
                         ..Default::default()
                     };
 
@@ -267,8 +264,7 @@ impl StoreTrait for S3Store {
         let s3_path = &self.make_s3_path(&digest);
 
         let max_size = match upload_size {
-            UploadSizeInfo::ExactSize(sz) => sz,
-            UploadSizeInfo::MaxSize(sz) => sz,
+            UploadSizeInfo::ExactSize(sz) | UploadSizeInfo::MaxSize(sz) => sz,
         };
         // NOTE(blaise.bruer) It might be more optimal to use a different heuristic here, but for
         // simplicity we use a hard codded value. Anything going down this if-statement will have
@@ -296,8 +292,8 @@ impl StoreTrait for S3Store {
             };
 
             let put_object_request = PutObjectRequest {
-                bucket: self.bucket.to_owned(),
-                key: s3_path.to_owned(),
+                bucket: self.bucket.clone(),
+                key: s3_path.clone(),
                 content_length,
                 body,
                 ..Default::default()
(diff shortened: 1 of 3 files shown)


MarcusSorealheis (commented):
My question is should these `ObjectStore` or `S3Store` . I'm torn because I don't want us to be wed to a vendor in the implementation on the one hand, but on the other hand the other object stores in the market tend to implement the S3 API because they are a leader. My hesitation feels a little nit picky and pedantic and can probably be deferred. 

Explicitly keeping S3 in the codebase as a store type is something we probably want to move away from.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-:-Nhd65Bu-_9OyS1KWzPf:b-nb226g)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @allada)

allada (changes requested):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @aaronmondal)

___
*[`util/buf_channel.rs` line 58 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd50TZ3A0Ncft8BGYL:-Nhd50TZ3A0Ncft8BGYM:b-twjjao) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L58)):*
> ```XML
> /// the receiver can be trusted to handle failing streams this is safe to do.
> #[must_use]
> pub fn make_buf_channel_pair_optimistic_writer() -> (DropCloserWriteHalf, DropCloserReadHalf) {
> ```

Why not just add `.disable_eof_check(&mut self)` on the writer?
___
*[`util/buf_channel.rs` line 88 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd7BfI4JsNT6zJKT4z:-Nhd7BfI4JsNT6zJKT5-:byf8xyc) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L88)):*
> ```XML
>     /// receiver end's drop status (ie: if the receiver dropped unexpectedly).
>     close_rx: oneshot::Receiver<Result<(), Error>>,
>     disable_eof_check: bool,
> ```

It looks like we already have `close_after_size` on `DropCloserReadHalf`. I think we can use this flag to ensure the read half can get into this configuration.

As for the Writer, would it make more sense to simply add a similar kind of flag check it in`send_eof()` and not emit an error if the flag is set (ie: `set_ignore_eof()`)?
___
*[`util/buf_channel.rs` line 131 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd5KaaCsju9LItndLG:-Nhd5KaaCsju9LItndLH:bq85xel) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L131)):*
> ```XML
>     {
>         loop {
>             if let Some(maybe_chunk) = reader.next().await {
> ```

nit: I personally prefer the `match` syntax here, it feels easier to read to me. but just my opinion.

allada (changes requested):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @aaronmondal)

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-:-NhdLCESEaVYm0p8kxHe:bnfp4nl)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @aaronmondal and @allada)

aaronmondal (commented):
@MarcusSorealheis  I agree that we don't want to name this S3. Something like `ObjectStore` or `BlobStore` seems more appropriate and lets us use more vendor-specific libraries to interact with AWS, GCP, Azure and custom solutions like CubeFS, Ceph and Minio.

I'll soon have a new implementation for S3 via the AWS SDK instead of rusoto ready. I still need to look into others. Then we can generalize the current implementation.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-:-NhdKmWvAzjUbMh5RMz2:bskhubw)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @allada)

___
*[`util/buf_channel.rs` line 58 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd50TZ3A0Ncft8BGYL:-NhdHtidBc-6Mu8-Layu:b-7ltkoe) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L58)):*

Why not just add `.disable_eof_check(&mut self)` on the writer?
</blockquote></details>

Done. Yes that's much better :+1:
___
*[`util/buf_channel.rs` line 88 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd7BfI4JsNT6zJKT4z:-NhdI27pEZlG4xNEtw7T:b69x8kt) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L88)):*

It looks like we already have `close_after_size` on `DropCloserReadHalf`. I think we can use this flag to ensure the read half can get into this configuration.

As for the Writer, would it make more sense to simply add a similar kind of flag check it in`send_eof()` and not emit an error if the flag is set (ie: `set_ignore_eof()`)?
</blockquote></details>

Not sure I understood you correctly here. Is the new version what you envision?
___
*[`util/buf_channel.rs` line 131 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd5KaaCsju9LItndLG:-NhdI8MfAo3wF09mZVEV:bxr80cu) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L131)):*

nit: I personally prefer the `match` syntax here, it feels easier to read to me. but just my opinion.
</blockquote></details>

That was flagged by clippy::single_match_else I'd prefer the new variant since it reduces indentation.

allada (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-:-NhdNQxH5q5lT4FUwvzG:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

___
*[`util/buf_channel.rs` line 88 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/341#-Nhd7BfI4JsNT6zJKT4z:-NhdN-gd0-iFzMSuvaC5:b-rq7u4t) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/d8bf8c0670a8171beeecf58b4c1e4c033aa16f4f/util/buf_channel.rs#L88)):*

Not sure I understood you correctly here. Is the new version what you envision?
</blockquote></details>

Yep, this is exactly what I was referring :-) ... To me this is so much easier to read.

PR #369, opened 2023-11-01
Migrate S3Store from rusoto to AWS SDK for Rust

### LINKED ISSUE #131: Look for rusoto alternatives
Unfortunately rusoto is in "maintenance mode".

Fortunately, this only seems to affect `cas/store/s3_store.rs` and its tests.

It seems that https://github.com/awslabs/aws-sdk-rust could be a good alternative. That's also used by `attic`, which is basically turbo-cache for nix: https://github.com/zhaofengli/attic/blob/main/server/src/storage/s3.rs

aaronmondal:
Fixes #131

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/369)


allada (commented):
> Reviewed 6 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/369)* status: 6 of 10 files reviewed, 16 unresolved discussions (waiting on @aaronmondal)

___
*[`flake.nix` line 30 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibrjNPDUEnr-ho2pZW:-NibrjNPDUEnr-ho2pZX:b-jw453j) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/flake.nix#L30)):*
> ```Nix
>               openssl_static # Required explicitly for cargo test support.
>               bazel
>               pkgs.awscli2
> ```

This is needed?
___
*[`cas/store/BUILD` line 406 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibrdmJ5XNpkN-hWEvS:-NibrdmJ5XNpkN-hWEvT:b-ky2d3r) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/BUILD#L406)):*
> ```Starlark
> )
> 
> # rust_test(
> ```

reminder to fix this.
___
*[`cas/store/s3_store.rs` line 15 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibs1Jp1BgzqhZawJMW:-Nibs1Jp1BgzqhZawJMX:b-jluord) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L15)):*
> ```XML
> // limitations under the License.
> 
> use std::{
> ```

nit: The rest of the repo only uses `{}` style for the end names. I personally find it easier to read spelled out (but last one can be joined together if the rest of the prefix matches).
___
*[`cas/store/s3_store.rs` line 67 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibsbu714qFBvrg28sF:-Nibsbu714qFBvrg28sG:b-41rqgt) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L67)):*
> ```XML
>             .with_webpki_roots()
>             .https_only()
>             .enable_http2()
> ```

hmmm, might be worth benchmarking http1 vs http2. From what I remember http2 will handle small objects better, but http1 will handle large objects better because large objects will have multiple streams opened in this implementation.
___
*[`cas/store/s3_store.rs` line 75 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibsxPO0WFYdtjH4_Sp:-NibsxPO0WFYdtjH4_Sq:b-ro8jje) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L75)):*
> ```XML
>     async fn simple_retry(
>         &mut self,
>         req: Uri,
> ```

nit, pass a ref, since we are making a copy internally.
___
*[`cas/store/s3_store.rs` line 85 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibtQr26VffXSn4AMLQ:-NibtQr26VffXSn4AMLR:b-ialoat) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L85)):*
> ```XML
>             match self.connector.call(req.clone()).await {
>                 Ok(stream) => return Ok(stream),
>                 Err(_) if retries < max_retries => {
> ```

nit: This is hard to read, can we make this more traditional logic instead of very-rust-specific style?
___
*[`cas/store/s3_store.rs` line 85 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibtYVL1bcPJ6Y_g5U6:-NibtYVL1bcPJ6Y_g5U7:bj4ftbw) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L85)):*
> ```XML
>             match self.connector.call(req.clone()).await {
>                 Ok(stream) => return Ok(stream),
>                 Err(_) if retries < max_retries => {
> ```

What about 404? Isnt' that an error? This would be a do-not-retry.
___
*[`cas/store/s3_store.rs` line 119 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic-AEs3iF-_QEY3nrj:-Nic-AEs3iF-_QEY3nrk:bt81zc4) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L119)):*
> ```XML
> 
>     fn call(&mut self, req: Uri) -> Self::Future {
>         const MAX_RETRIES: usize = 10;
> ```

nit: These should be configs.
___
*[`cas/store/s3_store.rs` line 350 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic0XNj3oItd0g0MCmG:-Nic0XNj3oItd0g0MCmH:bf1ksu0) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L350)):*
> ```XML
>                 let stream = ByteStream::new(SdkBody::from(write_buf));
>                 let s3_client = self.s3_client.clone();
>                 let inner_s3_path = s3_path.clone();
> ```

nit: Just inline these.
___
*[`cas/store/s3_store.rs` line 355 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic0eMzCRufNn03f-i-:-Nic0eMzCRufNn03f-i0:b-dcl6xt) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L355)):*
> ```XML
> 
>                 completed_part_futs.push(
>                     JoinHandleDropGuard::new(tokio::spawn(async move {
> ```

nit: While we are here lets remove these spawns As a bonus we won't need `JoinHandleDropGuard`.
___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic11NK5wstMM49K4IO:-Nic11NK5wstMM49K4IP:b-bkbv8t) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*
> ```XML
>             Ok(())
>         };
>         if complete_result.is_err() {
> ```

nit: This should probably be in:
https://docs.rs/scopeguard/latest/scopeguard/fn.guard.html

So if the future ever gets dropped without `complete_result` ever finishing it also cleans up.
___
*[`cas/store/s3_store.rs` line 458 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic1V642CcXDM2W5i6l:-Nic1V642CcXDM2W5i6m:bc0cway) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L458)):*
> ```XML
>                             // doesn't require a bunch of hackery around poor UX
>                             // in the s3 sdk.
>                             RetryResult::Err(make_err!(Code::NotFound, "Couldn't get object.")),
> ```

nit: add `in s3 store`, so it's string searchable in the code.
___
*[`cas/store/s3_store.rs` line 487 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic1sZU77EUNk2XgJMJ:-Nic1sZU77EUNk2XgJMK:b6frenr) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L487)):*
> ```XML
>                                 if let Err(e) = writer.send(bytes).await {
>                                     return Some((
>                                         RetryResult::Err(make_input_err!("Error sending bytes to consumer: {e}")),
> ```

nit: In all these errors say `in s3 store`.
___
*[`cas/store/s3_store.rs` line 494 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic2Aqw2sb07L1jDIdl:-Nic2Aqw2sb07L1jDIdm:b-dg7ix6) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/22abf900fa6a6ca53e340d6a5a4ad3279a3bdeb3/cas/store/s3_store.rs#L494)):*
> ```XML
>                     // Copy data from s3 input stream to the writer stream.
>                     let result = writer
>                         .forward(s3_in_stream, true /* Forward EOF */)
> ```

nit: We might be able to remove the `.forward()` function. Not sure, but i'd be suprised if it's used anywhere else.
___
*[`config/stores.rs` line 412 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic2XveBkZ-8ZVaLiQ2:-Nic2XveBkZ-8ZVaLiQ3:bb9qi2u) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/22abf900fa6a6ca53e340d6a5a4ad3279a3bdeb3/config/stores.rs#L412)):*
> ```XML
>     /// Default: 20.
>     #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
>     pub additional_max_concurrent_requests: usize,
> ```

we could add this in the `global` config now.
___
*[`tools/cargo_shared.bzl` line 72 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibrru26zRpxugf3ugG:-Nibrru26zRpxugf3ugH:b-gparbq) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/tools/cargo_shared.bzl#L72)):*
> ```Starlark
>         "version": "0.8.5",
>     },
>     "rusoto_s3": {
> ```

Reminder: Remove these.

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/369)* status: 2 of 13 files reviewed, 7 unresolved discussions (waiting on @allada)

___
*[`flake.nix` line 30 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibrjNPDUEnr-ho2pZW:-Nix8kx1-8RLlGuEByww:b42lg9t) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/flake.nix#L30)):*

This is needed?
</blockquote></details>

This is only used when entering the dev environment. This way users don't need to install the aws cli manually.
___
*[`cas/store/BUILD` line 406 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibrdmJ5XNpkN-hWEvS:-Nix92gu95qvg4mb5Tje:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/BUILD#L406)):*

reminder to fix this.
</blockquote></details>

Done.
___
*[`cas/store/s3_store.rs` line 67 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibsbu714qFBvrg28sF:-Nix956jAm1pDloc66nc:b5cu648) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L67)):*

hmmm, might be worth benchmarking http1 vs http2. From what I remember http2 will handle small objects better, but http1 will handle large objects better because large objects will have multiple streams opened in this implementation.
</blockquote></details>

I agree, but propose to postpone this to a later commit when we have better tooling/setup for benchmarking these things. I don't think there is an easy way to keep both http1 and http2 happy, so I'd argue that the performance difference needs to be significant for this to justify the additional complexity.

The current variant works for AWS and if it works for GCP as well I think this should be fine.
___
*[`cas/store/s3_store.rs` line 85 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibtYVL1bcPJ6Y_g5U6:-Nix9YIU31KEgi8souRc:bs2pwx3) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L85)):*

What about 404? Isnt' that an error? This would be a do-not-retry.
</blockquote></details>

Interestingly, this is the thing we're catching here. S3 tends to lose connections quite frequently with a 404 error that resolves itself after a few seconds. An unrelated instance where you can notice this is when running `aws s3 rm s3://somebucket --recursive` on a sufficiently large bucket. You'll get connection errors after a while. This retry mechanism counteracts that.
___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic11NK5wstMM49K4IO:-NixA8vb4fjZEUGpzEJJ:b87whr0) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

nit: This should probably be in:
https://docs.rs/scopeguard/latest/scopeguard/fn.guard.html

So if the future ever gets dropped without `complete_result` ever finishing it also cleans up.
</blockquote></details>

I double-checked with clippy, and I think this is fine. There are no panics raised in (the latest version). Or maybe I misunderstood the `guard`?
___
*[`config/stores.rs` line 412 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic2XveBkZ-8ZVaLiQ2:-NixAWpcEK6axrKuxOEx:b-k6k3c) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/22abf900fa6a6ca53e340d6a5a4ad3279a3bdeb3/config/stores.rs#L412)):*

we could add this in the `global` config now.
</blockquote></details>

Let's do this in a different PR. There are some other things we can do with S3 specifically since it has it's own retry mechanisms (including exponential backoff-type retries) in a way that is potentially better than what we have at the moment. This seems to require additional efforts to evaluate though.
___
*[`tools/cargo_shared.bzl` line 72 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibrru26zRpxugf3ugG:-NixAtJ29OArnkoAyIVb:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/tools/cargo_shared.bzl#L72)):*

Reminder: Remove these.
</blockquote></details>

Done.

blakehatch (commented):
> Reviewed 1 of 10 files at r1, 8 of 10 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/369)* status: all files reviewed, 8 unresolved discussions (waiting on @aaronmondal, @allada, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nic11NK5wstMM49K4IO:-NjCyqIo2iu9IoFQCIst:bzf7vox) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

I double-checked with clippy, and I think this is fine. There are no panics raised in (the latest version). Or maybe I misunderstood the `guard`?
</blockquote></details>

From my understanding there are cases like a panic where this or if the execution stops unexpectedly this cleanup logic might not be triggered. The guard will make sure that the cleanup operation runs no matter how this scope is exited even with a panic. If a panic here were ever re-introduced accidentally at least we wouldn't have to worry about this cleanup being affected.
___
*[`cas/store/tests/s3_store_test.rs` line 28 at r4](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NjD1b64AHVJR9tKCyZl:-NjD1b64AHVJR9tKCyZm:br0w0ry) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/1128b9f5d15df7447f4f6ced37c73d990fc429f7/cas/store/tests/s3_store_test.rs#L28)):*
> ```XML
> use traits::StoreTrait;
> 
> // TODO(aaronmondal): Figure out how to test the connector retry mechanism.
> ```

nit: Seems like this is being tested in fn simple_has_retries()? Could be incorrect but if it is just remove the TODO

allada (changes requested):
> Reviewed 1 of 10 files at r2, 1 of 2 files at r3, 5 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/369)* status: 12 of 13 files reviewed, 8 unresolved discussions (waiting on @aaronmondal, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 67 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-Nibsbu714qFBvrg28sF:-NjT1INB40-NVq8Kfsqa:bmdxzct) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L67)):*

I agree, but propose to postpone this to a later commit when we have better tooling/setup for benchmarking these things. I don't think there is an easy way to keep both http1 and http2 happy, so I'd argue that the performance difference needs to be significant for this to justify the additional complexity.

The current variant works for AWS and if it works for GCP as well I think this should be fine.
</blockquote></details>

To be honest, I'd rather not take the risk. I'd rather stick to http1 and support http2 after we have benchmarking. I have had some bad experiences with http2 when dealing with high throughput & lots of streams.

Later we can add http2 as a config option once we have time to benchmark the differences.
___
*[`cas/store/s3_store.rs` line 85 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NibtYVL1bcPJ6Y_g5U6:-NjT27SQ4YPSX7vun0Fu:b-6q72mb) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L85)):*

Interestingly, this is the thing we're catching here. S3 tends to lose connections quite frequently with a 404 error that resolves itself after a few seconds. An unrelated instance where you can notice this is when running `aws s3 rm s3://somebucket --recursive` on a sufficiently large bucket. You'll get connection errors after a while. This retry mechanism counteracts that.
</blockquote></details>

Hmmm, did you get the 404 on S3? Because AWS-S3 & GCP-buckets should no longer be eventual-consistent, but other providers are.
___
*[`cas/store/s3_store.rs` line 74 at r5](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NjT142o1I5gbCGsH94_:-NjT142o1I5gbCGsH94a:boytifb) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/cas/store/s3_store.rs#L74)):*
> ```XML
>         let connector = hyper_rustls::HttpsConnectorBuilder::new()
>             .with_webpki_roots()
>             .https_only()
> ```

nit: I don't think we should require https. I could see someone wanting to use their own internal S3 api that is inside their network on their non-ssl port.
___
*[`cas/store/s3_store.rs` line 80 at r5](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NjT1sIh99__vfnWZuic:-NjT1sIh99__vfnWZuid:biaauqm) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/cas/store/s3_store.rs#L80)):*
> ```XML
>         let connector_max_retries = config
>             .connector_max_retries
>             .map_or(DEFAULT_CONNECTOR_MAX_RETRIES, |retries| retries);
> ```

nit: I think you want `.unwrap_or()`
___
*[`config/stores.rs` line 416 at r5](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NjT0CSQ2HRWcf4DCnk2:-NjT0CSQ2HRWcf4DCnk3:b-qpnfmg) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/config/stores.rs#L416)):*
> ```XML
> 
>     /// Maximum number of concurrent UploadPart requests per MultipartUpload.
>     /// Default: 10.
> ```

nit: I usually put a full new line between the comment and the `Default` line. (just for consistency)
___
*[`config/stores.rs` line 421 at r5](https://reviewable.io/reviews/TraceMachina/turbo-cache/369#-NjT0gLq3lETODY2UTMu:-NjT0gLq3lETODY2UTMv:bsrf1cz) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/config/stores.rs#L421)):*
> ```XML
>     /// Maximum number of retries in case the connection to S3 drops.
>     /// Default: 10.
>     pub connector_max_retries: Option<usize>,
> ```

Can the connector logic be part of the `retry` config above? Like make them part of the same internal-retry object?

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 10 of 13 files reviewed, 6 unresolved discussions (waiting on @allada, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 67 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-Nibsbu714qFBvrg28sF:-NjZAFqmDrdUClLRWfxz:bow2rw6) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L67)):*

To be honest, I'd rather not take the risk. I'd rather stick to http1 and support http2 after we have benchmarking. I have had some bad experiences with http2 when dealing with high throughput & lots of streams.

Later we can add http2 as a config option once we have time to benchmark the differences.
</blockquote></details>

I've now enabled both HTTP1 and HTTP2. This way the descision is handled by the aws sdk.
___
*[`cas/store/s3_store.rs` line 85 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-NibtYVL1bcPJ6Y_g5U6:-NjZANN95U762cJzKX1x:bt0hv4t) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L85)):*

Hmmm, did you get the 404 on S3? Because AWS-S3 & GCP-buckets should no longer be eventual-consistent, but other providers are.
</blockquote></details>

This is not a 404 "object not found in bucket" error, but a DNS resolution error as in "I can't find the bucket". Sometimes the endpoint for the entire bucket fails to resolve.
___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-Nic11NK5wstMM49K4IO:-NjZBCVlA84nBZvQFTGp:bw3zttw) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

From my understanding there are cases like a panic where this or if the execution stops unexpectedly this cleanup logic might not be triggered. The guard will make sure that the cleanup operation runs no matter how this scope is exited even with a panic. If a panic here were ever re-introduced accidentally at least we wouldn't have to worry about this cleanup being affected.
</blockquote></details>

Seems to add unnecessary complexity, and with the current state of the code it's not an issue. If we want to handle such things we should add global clippy configs to forbid undocumented panics.
___
*[`cas/store/s3_store.rs` line 74 at r5](https://reviewable.io/reviews/tracemachina/native-link/369#-NjT142o1I5gbCGsH94_:-NjZ9s196UZezPI8Ykca:bqh77kg) ([raw file](https://github.com/tracemachina/native-link/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/cas/store/s3_store.rs#L74)):*

nit: I don't think we should require https. I could see someone wanting to use their own internal S3 api that is inside their network on their non-ssl port.
</blockquote></details>

We have to default to HTTPS-only because of "safe default" requirements. I've added an `insecure_allow_http` flag to make HTTP opt-in.
___
*[`cas/store/tests/s3_store_test.rs` line 28 at r4](https://reviewable.io/reviews/tracemachina/native-link/369#-NjD1b64AHVJR9tKCyZl:-NjZBbwTC4eFQnLIPdle:bkg9kr6) ([raw file](https://github.com/tracemachina/native-link/blob/1128b9f5d15df7447f4f6ced37c73d990fc429f7/cas/store/tests/s3_store_test.rs#L28)):*

nit: Seems like this is being tested in fn simple_has_retries()? Could be incorrect but if it is just remove the TODO
</blockquote></details>

This is still correct. We're only testing the HTTP traffic here but not the underlying TCP connector. I still didn't figure out how to test it properly and we probably want to put the entire TlsConnector into some separate file and make it handle all connections of native link globally.
___
*[`config/stores.rs` line 421 at r5](https://reviewable.io/reviews/tracemachina/native-link/369#-NjT0gLq3lETODY2UTMu:-NjZC3lrBtSutXoiIRPM:b-ieumm3) ([raw file](https://github.com/tracemachina/native-link/blob/ae0c958667a2e2f1d7c83a7b2233f05b93ff6341/config/stores.rs#L421)):*

Can the connector logic be part of the `retry` config above? Like make them part of the same internal-retry object?
</blockquote></details>

I tried this, but it turned out to be misleading since we're configuring two different retriers for the S3 store. At the moment, the `Retry` config is reusable (for instance in some grpc services) and its `max_retries` configures global retries in some cases. This is not true for the S3 implementation which has distinct limits for Http failures and tcp failures.

This is certainly not optimal, but I think we should keep it this way and improve the retry logic globally in a way that doesn't lead to different "meanings" in different stores.

allada (changes requested):
> Reviewed 4 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 9 of 13 files reviewed, 4 unresolved discussions (waiting on @aaronmondal, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 96 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_geX26qp0MhM8h7eq:-Nj_geX26qp0MhM8h7er:b-h40ta2) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L96)):*
> ```XML
>     }
> 
>     async fn simple_retry(&mut self, req: &Uri) -> Result<MaybeHttpsStream<TcpStream>, Error> {
> ```

After thinking about it and hacking a bit, I really do think we should use `Retrier`. Even though the API may not be super beautiful. It has the advantage that it is tested, and other parts of the code use it.

In the future if/when we upgrade it to use an external version, finding all parts of the code that need retry ability is easy, but if we use custom logic like this, it makes it harder to code search.

I created a commit on top of this branch that shows how it can be added without too much mess...

The only down side is it requires us to use more Arcs instead of Box, but I think its a minimal cost (~2x atomic add/sub calls on average).

https://github.com/TraceMachina/native-link/commit/6c2a27a3fafdde29fa1100eeed4ce0184e61789d

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 9 of 17 files reviewed, 2 unresolved discussions (waiting on @allada, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 96 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_geX26qp0MhM8h7eq:-Nja-6JT9xNL3AHGkM-q:b-896fix) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L96)):*

After thinking about it and hacking a bit, I really do think we should use `Retrier`. Even though the API may not be super beautiful. It has the advantage that it is tested, and other parts of the code use it.

In the future if/when we upgrade it to use an external version, finding all parts of the code that need retry ability is easy, but if we use custom logic like this, it makes it harder to code search.

I created a commit on top of this branch that shows how it can be added without too much mess...

The only down side is it requires us to use more Arcs instead of Box, but I think its a minimal cost (~2x atomic add/sub calls on average).

https://github.com/TraceMachina/native-link/commit/6c2a27a3fafdde29fa1100eeed4ce0184e61789d
</blockquote></details>

Done.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 7 of 17 files reviewed, 11 unresolved discussions (waiting on @aaronmondal, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 30 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_hUXu3NPoS940R3Ys:-Nj_hUXu3NPoS940R3Yt:b-lsjvio) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L30)):*
> ```XML
> use aws_sdk_s3::primitives::{ByteStream, SdkBody};
> use aws_sdk_s3::types::builders::{CompletedMultipartUploadBuilder, CompletedPartBuilder};
> // use aws_sdk_s3::types::CompletedPart;
> ```

nit: Remove.
___
*[`cas/store/s3_store.rs` line 273 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_iC4v3ZWQ1lbzZbRa:-Nj_iC4v3ZWQ1lbzZbRb:b-vn7bpk) ([raw file](https://github.com/tracemachina/native-link/blob/7d9deaf9ec113b4984dfd7061d1616c1d029d859/cas/store/s3_store.rs#L273)):*
> ```XML
>         // for multipart upload requests.
>         if max_size < MIN_MULTIPART_SIZE {
>             let (body, content_length) = if let UploadSizeInfo::ExactSize(sz) = upload_size {
> ```

By removing this code path we are forcing a copy of our data into a temp buffer when the content size is known.

Keep in mind that most uploads are small and size is known, so most uploads do hit this code path.
___
*[`cas/store/s3_store.rs` line 316 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_jZlK74s4855uTE19:-Nj_jZlK74s4855uTE1A:b-82ekpq) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L316)):*
> ```XML
>             .create_multipart_upload()
>             .bucket(&self.bucket)
>             .key(s3_path.clone())
> ```

super nit: I don't think you need to call `clone()` here, since it's an `Into<String>` and will do it internally.
___
*[`cas/store/s3_store.rs` line 331 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_lac1AVHLN6tQeODO:-Nj_lac20ftLERrx2K-L:b-lsjvio) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L331)):*
> ```XML
>             // the worst case.
>             let mut completed_parts = Vec::with_capacity((max_size / bytes_per_upload_part) + 1);
>             // let mut completed_part_futs = Vec::with_capacity(completed_parts.len());
> ```

nit: Remove.
___
*[`cas/store/s3_store.rs` line 333 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_lfUhA0nu_tQfUgSb:-Nj_lfUhA0nu_tQfUgSc:bxurbyb) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L333)):*
> ```XML
>             // let mut completed_part_futs = Vec::with_capacity(completed_parts.len());
> 
>             let max_concurrent_requests = self.multipart_max_concurrent_uploads;
> ```

nit: Inline this.
___
*[`cas/store/s3_store.rs` line 335 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_mDCCEehPvLB-xm-F:-Nj_mDCCEehPvLB-xm-G:bxkdo5x) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L335)):*
> ```XML
>             let max_concurrent_requests = self.multipart_max_concurrent_uploads;
> 
>             let mut futures = futures::stream::FuturesUnordered::new();
> ```

nit: Import this and use the name directly.
___
*[`cas/store/s3_store.rs` line 367 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_krUWACazq9PzV7lW:-Nj_krUWACazq9PzV7lX:b-frij3y) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L367)):*
> ```XML
>                     }
> 
>                     let current_part_number = part_number;
> ```

nit: This is redundant, just use `part_number` below
___
*[`cas/store/s3_store.rs` line 396 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_lQ6a3smMe1MSLbYa:-Nj_lQ6a3smMe1MSLbYb:b-exxipg) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L396)):*
> ```XML
>             // Await remaining results.
>             {
>                 use futures::StreamExt;
> ```

nit: Move this to top of file.
___
*[`cas/store/s3_store.rs` line 334 at r8](https://reviewable.io/reviews/tracemachina/native-link/369#-Nja6v1P5ukNnEsq6fHi:-Nja6v1P5ukNnEsq6fHj:b7vvpv) ([raw file](https://github.com/tracemachina/native-link/blob/1d0cabe1deabfeccca7b04ac6d523ecca2b7ebb3/cas/store/s3_store.rs#L334)):*
> ```XML
>             // CompletedMultipartUploadBuilder keeps track of the response
>             // order and lets S3 know about it.
>             loop {
> ```

I don't think we should complicate things here. Instead we should just pub-sub into a stream that is bounded then in another future use `buffered_unordered()` to extract them into a vector then sort them.

I created a sample of what I mean here based on this PR:
https://github.com/TraceMachina/native-link/commit/0dc01335dc0157be37025819737c3f1792cd7f56

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 6 of 17 files reviewed, 6 unresolved discussions (waiting on @allada, @blakehatch, and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 30 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_hUXu3NPoS940R3Ys:-Njd6Ufm2xvjlgNimweR:b4mlygk) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L30)):*

nit: Remove.
</blockquote></details>

I think there was some commented out code here, but it's no longer present, potentially due your changes I merged in.
___
*[`cas/store/s3_store.rs` line 273 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_iC4v3ZWQ1lbzZbRa:-Njd8vWe-NarcDgsh0HV:b-v218he) ([raw file](https://github.com/tracemachina/native-link/blob/7d9deaf9ec113b4984dfd7061d1616c1d029d859/cas/store/s3_store.rs#L273)):*

By removing this code path we are forcing a copy of our data into a temp buffer when the content size is known.

Keep in mind that most uploads are small and size is known, so most uploads do hit this code path.
</blockquote></details>

I've verified that the removed branch indeed was used when added back. I added it back but had to change it a bit to make it compatible with `SdkBody`. I think this directly pipes our data through now.
___
*[`cas/store/s3_store.rs` line 316 at r6](https://reviewable.io/reviews/tracemachina/native-link/369#-Nj_jZlK74s4855uTE19:-Njd9jGHApsYUJ8650qG:b-896fix) ([raw file](https://github.com/tracemachina/native-link/blob/324b076753cae0086694c5a83d68ccda42736760/cas/store/s3_store.rs#L316)):*

super nit: I don't think you need to call `clone()` here, since it's an `Into<String>` and will do it internally.
</blockquote></details>

Done.
___
*[`cas/store/s3_store.rs` line 334 at r8](https://reviewable.io/reviews/tracemachina/native-link/369#-Nja6v1P5ukNnEsq6fHi:-Njd9sVUBDb6DBr-RTBd:bc8q8dj) ([raw file](https://github.com/tracemachina/native-link/blob/1d0cabe1deabfeccca7b04ac6d523ecca2b7ebb3/cas/store/s3_store.rs#L334)):*

I don't think we should complicate things here. Instead we should just pub-sub into a stream that is bounded then in another future use `buffered_unordered()` to extract them into a vector then sort them.

I created a sample of what I mean here based on this PR:
https://github.com/TraceMachina/native-link/commit/0dc01335dc0157be37025819737c3f1792cd7f56
</blockquote></details>

Interesting. I think I wasn't aware of the difference between the future-based and stream-based concurrency models. I'll need to look into the stream-based approach more.

Merged your proposed changes into the branch.

allada (commented):
> Reviewed 2 of 7 files at r7, 5 of 8 files at r9.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 10 of 17 files reviewed, 1 unresolved discussion (waiting on @blakehatch and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-Nic11NK5wstMM49K4IO:-NjecX1e-iAScCzrFxsN:ba8fnz3) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

Seems to add unnecessary complexity, and with the current state of the code it's not an issue. If we want to handle such things we should add global clippy configs to forbid undocumented panics.
</blockquote></details>

Sorry, I should elaborate. What this branch is doing is telling AWS it can discard all the uploaded parts, so we don't get billed for it.

It has minimally to do with panics and more around if a stream is interrupted (upload or download) and the upload gets cancelled somewhere (all those `?` can trigger a cancel). `ScopeGuard` is an inline wrapper for of `Drop`, similar to go's `defer` statement, so you can run a function when the variable goes out of scope. This is often used so if a future is ever dropped for any reason it'll run the logic. In this case it should just launch a spawn to abort the upload unless the upload succeeded.

An example of it being used for similar stuff:
https://github.com/TraceMachina/native-link/blob/aea37682dbed261c401e5025ffd77dff2711f699/cas/worker/running_actions_manager.rs#L746

allada (approved):
:lgtm:

Sorry it took so long on this one. Was a bit of logic to review and was pretty busy last week.

> Reviewed 1 of 10 files at r2, 1 of 8 files at r6, 3 of 7 files at r7, 3 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369#-:-Njegdro1IE7oPy0G9yi:b-lt6fzx)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronmondal and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 498 at r9](https://reviewable.io/reviews/tracemachina/native-link/369#-Njef5wa2TTdXKIIRDeZ:-Njef5wa2TTdXKIIRDe_:b-bmlne0) ([raw file](https://github.com/tracemachina/native-link/blob/afc14dc93ea71718050e88e62b701a00512bce84/cas/store/s3_store.rs#L498)):*
> ```XML
>                                 // Handle the error, log it, or break the loop
>                                 return Some((
>                                     RetryResult::Err(make_input_err!("Bad bytestream element in S3: {e}")),
> ```

nit: Shouldn't this be a `RetryResult::Retry` case?

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 16 of 17 files reviewed, all discussions resolved (waiting on @allada and @MarcusSorealheis)

___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-Nic11NK5wstMM49K4IO:-Njeuvsj68psz8-inBmT:b-mq7pyo) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

Sorry, I should elaborate. What this branch is doing is telling AWS it can discard all the uploaded parts, so we don't get billed for it.

It has minimally to do with panics and more around if a stream is interrupted (upload or download) and the upload gets cancelled somewhere (all those `?` can trigger a cancel). `ScopeGuard` is an inline wrapper for of `Drop`, similar to go's `defer` statement, so you can run a function when the variable goes out of scope. This is often used so if a future is ever dropped for any reason it'll run the logic. In this case it should just launch a spawn to abort the upload unless the upload succeeded.

An example of it being used for similar stuff:
https://github.com/TraceMachina/native-link/blob/aea37682dbed261c401e5025ffd77dff2711f699/cas/worker/running_actions_manager.rs#L746
</blockquote></details>

The only time we can call the abort, is when we know the upload id. so the only thing we can guard is the scope constructing the `complete_result`. Since all `?` occurrences would cause the `complete_result` to contain an `Err` value (and not exit the entire function) the current logic seems sound to me. So the `?` might trigger stream cancelling, but it doesn't matter because the error value will be captured in the `complete_result` and handled by the (IMO easier to understand) `if complete_result.is_err()` block.

If this is actually an issue we can fix it in a future PR.

MarcusSorealheis:
I was too late and was still reviewing this one. Let's discuss in the next ticket. Large PR. Nice work @aaronmondal and @allada (on the review)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/369)* status: 16 of 17 files reviewed, all discussions resolved

___
*[`cas/store/s3_store.rs` line 398 at r1](https://reviewable.io/reviews/tracemachina/native-link/369#-Nic11NK5wstMM49K4IO:-Njh_7O-5LETRmpS4u09:bxdwiit) ([raw file](https://github.com/tracemachina/native-link/blob/4fee1ef857ecb52e519c789cb56d1e8475ea9602/cas/store/s3_store.rs#L398)):*

The only time we can call the abort, is when we know the upload id. so the only thing we can guard is the scope constructing the `complete_result`. Since all `?` occurrences would cause the `complete_result` to contain an `Err` value (and not exit the entire function) the current logic seems sound to me. So the `?` might trigger stream cancelling, but it doesn't matter because the error value will be captured in the `complete_result` and handled by the (IMO easier to understand) `if complete_result.is_err()` block.

If this is actually an issue we can fix it in a future PR.
</blockquote></details>

Yes, but if the client stream disconnects this future will abruptly be destroyed and no error will be returned, resulting in a partial upload. `ScopeGuard` would protect against that case.

PR #370, opened 2023-11-02
Redis Store Implementation

### LINKED ISSUE #302: Implement Redis store
Looking at some stats it looks like a very significant portion of data is small and might be most optimally stored in redis.

We should implement a redis backend. This would pair well with a SizeDistinguishing store.

blakehatch:
# Description

Added a Redis Store to Turbo Cache's store options.

Fixes #302 

## Type of change

## How Has This Been Tested?

Testing all basic functionality of Redis Store, need to make sure unique Redis functionality has full coverage. Also might be worth considering an integration test of the SizePartitioningStore paired with the Redis store, gonna try to mock the behavior in a unit test.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/370)

### DIFF
--- a/cas/store/grpc_store.rs
+++ b/cas/store/grpc_store.rs
@@ -25,6 +25,7 @@ use prost::Message;
 use proto::build::bazel::remote::execution::v2::digest_function;
 use rand::{rngs::OsRng, Rng};
 use tokio::time::sleep;
+use tonic::transport::Channel;
 use tonic::{transport, IntoRequest, Request, Response, Streaming};
 use uuid::Uuid;
 
@@ -111,6 +112,12 @@ impl GrpcStore {
         })
     }
 
+    fn get_retry_config(&self) -> impl Iterator<Item = Duration> + '_ {
+        ExponentialBackoff::new(Duration::from_millis(self.retry.delay as u64))
+            .map(|d| (self.jitter_fn)(d))
+            .take(self.retry.max_retries) // Remember this is number of retries, so will run max_retries + 1.
+    }
+
     async fn perform_request<F, Fut, R, I>(&self, input: I, mut request: F) -> Result<R, Error>
     where
         F: FnMut(I) -> Fut + Send + Copy,
@@ -118,9 +125,7 @@ impl GrpcStore {
         R: Send,
         I: Send + Clone,
     {
-        let retry_config = ExponentialBackoff::new(Duration::from_millis(self.retry.delay as u64))
-            .map(|d| (self.jitter_fn)(d))
-            .take(self.retry.max_retries); // Remember this is number of retries, so will run max_retries + 1.
+        let retry_config = self.get_retry_config();
         self.retrier
             .retry(
                 retry_config,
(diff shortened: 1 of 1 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/370)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @blakehatch)

___
*[`cas/store/redis_store.rs` line 34 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFVrOF6IFe6eJ44i8S:-NiFVrOF6IFe6eJ44i8T:b-54trgi) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L34)):*
> ```XML
> impl RedisStore {
>     pub async fn new(config: &config::stores::RedisStore) -> Result<Self, Error> {
>         let client = redis::Client::open(format!("redis://{}:{}", config.host, config.port))?;
> ```

nit: Try to declare your dependencies above. In this case I suggest: `use redis::{AsyncCommands, Client as RedisClient}`
___
*[`cas/store/redis_store.rs` line 34 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFW0f20gxr2k5bdtCp:-NiFW0f20gxr2k5bdtCq:bc0xg2y) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L34)):*
> ```XML
> impl RedisStore {
>     pub async fn new(config: &config::stores::RedisStore) -> Result<Self, Error> {
>         let client = redis::Client::open(format!("redis://{}:{}", config.host, config.port))?;
> ```

nit: Can we instead have users provide the full redis path instead of just host/port? It seems like we are giving more power to the users if we do.
___
*[`cas/store/redis_store.rs` line 46 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFWKiuCr-h8r082Mxi:-NiFWKiuCr-h8r082Mxj:bq917jp) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L46)):*
> ```XML
>         results: &mut [Option<usize>],
>         let mut hasher = Sha256::new();
> ```

nit: We don't need to hash. The `DigestInfo` is already a packed hash.
___
*[`cas/store/redis_store.rs` line 47 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFqy2F3Rt3WBsRuxrF:-NiFqy2F3Rt3WBsRuxrG:bpuvgpz) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L47)):*
> ```XML
>         let mut hasher = Sha256::new();
>         let mut conn = self.client.get_async_connection().await?;
> ```

I think we should be using the tokio multiplex connection. If we use this connection we could end up exhausting the open file descriptor limits.

Using the tokio version uses the async tcp library in tokio.

One thing we need to be careful of here is to use whatever API gives us more throughput. multiplex might be a single connection to a single node, where non-multiplex might allow multiple uploads/reads to hit multiple redis nodes at once.
___
*[`cas/store/redis_store.rs` line 51 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFrEhD2yNXc49dh6Oi:-NiFrEhD2yNXc49dh6Oj:bg4kjmh) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L51)):*
> ```XML
>             hasher.update(digest.packed_hash);
>             let hash = format!("{:x}", hasher.finalize_reset());
>             *result = conn.exists(&hash).await?;
> ```

nit: This will happen in serial, we should make this generate a bunch of futures then await all the futures at once, so we can ask for lots of requests at once.
___
*[`cas/store/redis_store.rs` line 78 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiGBauwC_sp9UfQKxaB:-NiGBauwC_sp9UfQKxaC:bf58wg5) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L78)):*
> ```XML
>             .err_tip(|| "Failed to collect all bytes from reader in redis_store::update")?;
> 
>         let _: () = conn.set(hash, buffer.to_vec()).await?;
> ```

nit: Use `&buff[..]`. This way you don't need to copy the data again.

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/370)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/store/redis_store.rs` line 34 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFVrOF6IFe6eJ44i8S:-NiGfvnDEmK26QuOlB2X:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L34)):*

nit: Try to declare your dependencies above. In this case I suggest: `use redis::{AsyncCommands, Client as RedisClient}`
</blockquote></details>

Done.
___
*[`cas/store/redis_store.rs` line 34 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFW0f20gxr2k5bdtCp:-NiGk72o6yRYdLUb3fRm:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L34)):*

nit: Can we instead have users provide the full redis path instead of just host/port? It seems like we are giving more power to the users if we do.
</blockquote></details>

Done.
___
*[`cas/store/redis_store.rs` line 46 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFWKiuCr-h8r082Mxi:-NiGlTtsAX8mMCVTy7bd:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L46)):*

nit: We don't need to hash. The `DigestInfo` is already a packed hash.
</blockquote></details>

Done.
___
*[`cas/store/redis_store.rs` line 47 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFqy2F3Rt3WBsRuxrF:-NiGsRyJ4-8YTxtF60o9:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L47)):*

I think we should be using the tokio multiplex connection. If we use this connection we could end up exhausting the open file descriptor limits.

Using the tokio version uses the async tcp library in tokio.

One thing we need to be careful of here is to use whatever API gives us more throughput. multiplex might be a single connection to a single node, where non-multiplex might allow multiple uploads/reads to hit multiple redis nodes at once.
</blockquote></details>

Done.
___
*[`cas/store/redis_store.rs` line 51 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiFrEhD2yNXc49dh6Oi:-NiGnBRnA-Zsg2-paBYh:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L51)):*

nit: This will happen in serial, we should make this generate a bunch of futures then await all the futures at once, so we can ask for lots of requests at once.
</blockquote></details>

Done.
___
*[`cas/store/redis_store.rs` line 78 at r3](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiGBauwC_sp9UfQKxaB:-NiGtLor0I7S4YCd3yX-:b-896fix) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/cce8e8a72b4719b81a43bc3bac9023f4e5c7fa03/cas/store/redis_store.rs#L78)):*

nit: Use `&buff[..]`. This way you don't need to copy the data again.
</blockquote></details>

Done.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/370)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @blakehatch)

___
*[`cas/store/redis_store.rs` line 33 at r4](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiKvO7f3WAivb5rgbCS:-NiKvO7f3WAivb5rgbCT:bzen1ko) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/e6af8b879040e389ddc5515a9f428342be823631/cas/store/redis_store.rs#L33)):*
> ```XML
> impl RedisStore {
>     pub async fn new(config: &config::stores::RedisStore) -> Result<Self, Error> {
>         let url = config.url.as_ref().ok_or_else(|| Error::new(Code::InvalidArgument, "URL is not set".to_string()))?;
> ```

nit:
```rs
let client = RedisClient::open(config.url.err_tip(|| "URL is not set")?).err_tip(|| "Failed to create redis client")?;
Ok(RedisStore { client });
```
___
*[`cas/store/redis_store.rs` line 57 at r4](https://reviewable.io/reviews/TraceMachina/turbo-cache/370#-NiKxF8Q9tvseaU9Q9lC:-NiKxF8Q9tvseaU9Q9lD:bi0db6p) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/e6af8b879040e389ddc5515a9f428342be823631/cas/store/redis_store.rs#L57)):*
> ```XML
>         }).collect();
> 
>         let results_vec: Vec<_> = try_join_all(futures).await?;
> ```

Avoid collecting results as much as possible.

In this case, what you should be doing is creating a `futures::stream::FuturesUnordered` and fill it with your futures.

Then use `.buffer_unordered(SOME_N_LIMIT)` and await each item in a stream and extract each value one-at-a-time.

This will be extremely memory efficient, we won't overload redis, since we are limiting our concurrent requests, and only requires 1 allocation.

PR #372, opened 2023-11-02
Fix never looping loops

aaronmondal:
Clippy complained with `never_loop`, and indeed these loops never loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/372)

### DIFF
--- a/cas/store/tests/filesystem_store_test.rs
+++ b/cas/store/tests/filesystem_store_test.rs
@@ -41,7 +41,7 @@ use traits::UploadSizeInfo;
 
 use buf_channel::{make_buf_channel_pair, DropCloserReadHalf};
 use common::{fs, DigestInfo};
-use error::{Code, Error, ResultExt};
+use error::{make_input_err, Code, Error, ResultExt};
 use filesystem_store::{digest_from_filename, EncodedFilePath, FileEntry, FileEntryImpl, FilesystemStore};
 use traits::StoreTrait;
 
@@ -308,9 +308,16 @@ mod filesystem_store_tests {
             .into_inner();
         let mut read_dir_stream = ReadDirStream::new(temp_dir_handle);
 
-        while let Some(temp_dir_entry) = read_dir_stream.next().await {
-            let path = temp_dir_entry?.path();
-            panic!("No files should exist in temp directory, found: {path:?}");
+        while let Some(entry) = read_dir_stream.next().await {
+            match entry {
+                Ok(temp_dir_entry) => {
+                    let path = temp_dir_entry.path();
+                    panic!("No files should exist in temp directory, found: {path:?}");
+                }
+                Err(e) => {
+                    return Err(make_input_err!("Couldn't read directory entry: {e:?}"));
+                }
+            }
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/372)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aaronmondal)

___
*[`cas/store/tests/filesystem_store_test.rs` line 312 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/372#-NiGDfLzBS8_ZnHy14xY:-NiGDfLzBS8_ZnHy14xZ:b-caxnkm) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/88182ea1bd7f87a5d66e449134abe858fa6aa874/cas/store/tests/filesystem_store_test.rs#L312)):*
> ```XML
> 
>         while let Some(temp_dir_entry) = read_dir_stream.next().await {
>             if let Ok(dir_entry) = temp_dir_entry {
> ```

I don't think we can do it this way. If `temp_dir_entry` had an error the test would pass.

I think you can also resolve this by using `return Err(make_input_error!("No files should exist in temp directory, found {:?}", ...));`

Instead.

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/372)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/store/tests/filesystem_store_test.rs` line 312 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/372#-NiGDfLzBS8_ZnHy14xY:-NiHBn_oCiglXpr1QS02:btpkoct) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/88182ea1bd7f87a5d66e449134abe858fa6aa874/cas/store/tests/filesystem_store_test.rs#L312)):*

I don't think we can do it this way. If `temp_dir_entry` had an error the test would pass.

I think you can also resolve this by using `return Err(make_input_error!("No files should exist in temp directory, found {:?}", ...));`

Instead.
</blockquote></details>

Updated this. A bit verbose, but a `temp_dir_entry?` doesn't work.

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/372#-:-NiI3u6b8pKf0_H3o3u8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronmondal)

___
*[`cas/store/tests/filesystem_store_test.rs` line 312 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/372#-NiGDfLzBS8_ZnHy14xY:-NiI3r1sBkHt3kMAMTj_:brqtavm) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/88182ea1bd7f87a5d66e449134abe858fa6aa874/cas/store/tests/filesystem_store_test.rs#L312)):*

Updated this. A bit verbose, but a `temp_dir_entry?` doesn't work.
</blockquote></details>

Yeah that is pretty verbose :-(

PR #378, opened 2023-11-04
Fixed flakiness in integration tests

### LINKED ISSUE #375: Fix Integration test flakiness by listening for connection string echo

blakehatch:
Added server readiness string listening to integration tests to reduce flakiness

# Description

Listens for string from docker-compose logs to indicate that the servers are ready and listening before running integration tests.

Fixes #375 

## Type of change

## How Has This Been Tested?

Ran all tests individually and together. TLS test was flaky when running with all tests before change and had no failed runs afterwards.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378)

### DIFF
--- a/run_integration_tests.sh
+++ b/run_integration_tests.sh
@@ -102,6 +102,12 @@ for pattern in "${TEST_PATTERNS[@]}"; do
     FILENAME=$(basename $fullpath)
     echo "Running test $FILENAME"
     sudo docker-compose up -d
+    if perl -e 'alarm shift; exec @ARGV' 30 bash -c 'until sudo docker-compose logs | grep -q "Ready, listening on"; do sleep 1; done'    
+    then
+      echo "String 'Ready, listening on' found in the logs."
+    else
+      echo "String 'Ready, listening on' not found in the logs within the given time."
+    fi
     set +e
     bash -euo pipefail "$fullpath"
     EXIT_CODE="$?"


aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378)* status: all files reviewed, 1 unresolved discussion (waiting on @blakehatch)

___
*[`run_integration_tests.sh` line 111 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-NiUk39Z2dcZl4JlRXyg:-NiUk39Z2dcZl4JlRXyh:b-shq6w4) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/2e5f1c501f236a177dad4a588bd5232b1f52c15b/run_integration_tests.sh#L111)):*
> ```Shell
>       echo "String 'Ready, listening on' not found in the logs within the given time."
>     fi
>     set +e
> ```

Hmm would there be a way to do this without the perl wrapper?

Also somewhat related question: WDYT about migrating these tests to rust in the long run? A tricky thing here seems to be getting cross-platform compatibility right. Rust might help with that.

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378)* status: all files reviewed, 1 unresolved discussion (waiting on @MarcusSorealheis)

___
*[`run_integration_tests.sh` line 111 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-NiUk39Z2dcZl4JlRXyg:-NiV01BJBdxkPW_2-ATG:bct0os2) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/2e5f1c501f236a177dad4a588bd5232b1f52c15b/run_integration_tests.sh#L111)):*

Hmm would there be a way to do this without the perl wrapper?

Also somewhat related question: WDYT about migrating these tests to rust in the long run? A tricky thing here seems to be getting cross-platform compatibility right. Rust might help with that.
</blockquote></details>

I originally tried timeout 30 bash -c 'until sudo docker-compose logs | grep -q "Ready, listening on"; do sleep 1; done' which is cleaner and works on Linux but the integration tests recently had some functionality added to make it support MacOS so I decided to use a command that works on all Unix-like OSes.

@MarcusSorealheis and I were discussing converting these integration tests to Python instead of Bash to fix the cross-platform compatibility while making the translation from Bash a bit simpler than rust. I do agree high level that they should definitely be converted in some form to have better cross-platform compatibility than the currently do.

aaronmondal (approved):
I guess let's just do it this way then. I'm not a huge fan, but it doesn't seem to really matter if we plan to port this to some other language anyways :dancer:

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-:-NiV6HUB--Tee8v8_sSn:boph53g)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @blakehatch)

blakehatch (commented):
Yeah agree not a huge fan either but best not to break functionality in the meantime, esp in the context of the likely switch as you said.🕺

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-:-NiV6yGj9D9vY2vuzbRS:brd3ev0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @blakehatch)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378)* status: all files reviewed, 1 unresolved discussion

___
*[`run_integration_tests.sh` line 105 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-NiWhMzJ-VI0U6dwj6DD:-NiWhMzK0YeXxiV1xeBz:biwfcy2) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/2e5f1c501f236a177dad4a588bd5232b1f52c15b/run_integration_tests.sh#L105)):*
> ```Shell
>     echo "Running test $FILENAME"
>     sudo docker-compose up -d
>     if perl -e 'alarm shift; exec @ARGV' 30 bash -c 'until sudo docker-compose logs | grep -q "Ready, listening on"; do sleep 1; done'    
> ```

nit: spaces at end of line.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378)* status: all files reviewed, 2 unresolved discussions

___
*[`run_integration_tests.sh` line 105 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-NiWhMzJ-VI0U6dwj6DD:-NiWhn5F1EBKE04Ml_L8:b52zanh) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/2e5f1c501f236a177dad4a588bd5232b1f52c15b/run_integration_tests.sh#L105)):*

nit: spaces at end of line.
</blockquote></details>

Btw, this could have been done with a bash `for loop` in a function.

```sh
function wait_for_ready() {
  for i in $(seq 1 10); do
    if sudo docker-compose logs | grep -q "Ready, listening on"; then
      return
    fi
    sleep 1
  done
  echo "Timed out"
  exit 1
}

wait_for_ready
```
___
*[`run_integration_tests.sh` line 109 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-NiWiUk87H1-bfDoqSsW:-NiWiUk87H1-bfDoqSsX:b-kun0sb) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/2e5f1c501f236a177dad4a588bd5232b1f52c15b/run_integration_tests.sh#L109)):*
> ```Shell
>       echo "String 'Ready, listening on' found in the logs."
>     else
>       echo "String 'Ready, listening on' not found in the logs within the given time."
> ```

nit: Also we don't exit fail here so this message will be buried deep in the logs.

MarcusSorealheis (commented):
It's merged and is totally fine. A small nit

In the future, I'd like to avoid Perl. There were some exceptions to this rule a few years ago, but I don't know if even those remain valid reasons to insist on using Perl. I would to need to look at some references.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/378#-:-NiWxxl83t7qK3qWi7WL:b97t2yv)* status: all files reviewed, 2 unresolved discussions

PR #382, opened 2023-11-07
Added error forwarding to client for failed command executions

### LINKED ISSUE #257: Errors should be forwarded to client if it could not execute command
Currently if a command cannot be executed on the worker the client is never notified and according to the client it hangs.

It is likely not passing the error code to the proper field.

TripleKai:
# Description

Added error forwarding to client upon any command execution failures. Interrupts client hang and displays the forwarded error.

Fixes #257 

## Type of change

## How Has This Been Tested?

Tested with plain `bazel test //...` to ensure all pass as well as `bazel test //...` after adding deliberate modifications to deployment-examples/docker-compose/worker.json to ensure error forwarding to client (ex. `"entrypoint_cmd": "foo"`).

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/382)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/382)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @TripleKai)

___
*[`cas/scheduler/simple_scheduler.rs` line 468 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/382#-Nif18rDCVpLFcyNG9_r:-Nif18rDCVpLFcyNG9_s:b32rx9e) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/589056bdafd798e514eec3d51aefbc43410d5146/cas/scheduler/simple_scheduler.rs#L468)):*
> ```XML
>         }
> 
>         if running_action.worker_id == *worker_id {
> ```

Something seems weird here. In theory this error should not even hit this condition. The worker should be forwarding the error back to the scheduler and the scheduler should be thinking the worker owns the action.

Can you explain a bit of the code path that is happening here?

TripleKai:
*[`cas/scheduler/simple_scheduler.rs` line 468 at r1](https://reviewable.io/reviews/TraceMachina/turbo-cache/382#-Nif18rDCVpLFcyNG9_r:-Nim_bCr8j3n1LJdV7N7:b8tg7bt) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/589056bdafd798e514eec3d51aefbc43410d5146/cas/scheduler/simple_scheduler.rs#L468)):*

Something seems weird here. In theory this error should not even hit this condition. The worker should be forwarding the error back to the scheduler and the scheduler should be thinking the worker owns the action.

Can you explain a bit of the code path that is happening here?
</blockquote></details>

The path I see is:

`LocalWorker::run()` → `WorkerApi::execution_response()` *(under Err response case)* → `WorkerApiServer::inner_execution_response()` → `SimpleSchedulerImpl::update_action_with_internal_error()` → Reaches the mentioned condition

My understanding is the `SimpleSchedulerImpl::update_action_with_internal_error()` is always called for this type of error because the `execute_result` in `WorkerApiServer::inner_execution_response()` is only ever classified as an `InternalError` for commands that fail to execute. Should this error be handled earlier in the call path?

TripleKai:
*[`cas/worker/local_worker.rs` line 278 at r2](https://reviewable.io/reviews/TraceMachina/turbo-cache/382#-NjEf5Z00LKSLnaYjfLF:-NjEf5Z00LKSLnaYjfLG:b-etr1v0) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ce6c14ef3a62d728071c0221c246d0a76e1547ea/cas/worker/local_worker.rs#L278)):*
> ```XML
>                                             "Internal Error: {:?}",
>                                             error
>                                         ));
> ```

Hi @allada, I updated the implementation so that the error is actually forwarded earlier in the call path. I believe this is cleaner and closer to what you described. Let me know if this needs further changes.



_Code quote:_
```XML
                                        return Err(make_err!(
                                            Code::Internal,
                                            "Internal Error: {:?}",
                                            error
                                        ));
```

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/382)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @TripleKai)

___
*[`cas/worker/local_worker.rs` line 278 at r2](https://reviewable.io/reviews/TraceMachina/turbo-cache/382#-NjEf5Z00LKSLnaYjfLF:-NjOcw3J2sPQ6vru0nmE:b-zbylew) ([raw file](https://github.com/TraceMachina/turbo-cache/blob/ce6c14ef3a62d728071c0221c246d0a76e1547ea/cas/worker/local_worker.rs#L278)):*
<details><summary><i>Previously, TripleKai (Kailash Saravanan) wrote…</i></summary><blockquote>

Hi @allada, I updated the implementation so that the error is actually forwarded earlier in the call path. I believe this is cleaner and closer to what you described. Let me know if this needs further changes.
</blockquote></details>

Hmmm, interesting. This seems much closer but still seems off...

The reason is because here I think the reason it is working is because by returning an error here is forcing the worker to disconnect from the scheduler then it auto-reconnects, which then makes the scheduler re-send the action.

The larger question is why the scheduler stops sending actions if this code path is reached?

TripleKai:
Closing PR since the commit history became too complex unnecessarily. Refer to the new PR for this issue's resolution: 
https://github.com/TraceMachina/native-link/pull/432

Issue #358, opened 2023-10-31
Implement Completeness checking store


allada:
This store is for Action Cache only. It is designed to ensure all referenced items exist in the CAS before forwarding the results.

PR #386, opened 2023-11-09
Added completeness checking store

### LINKED ISSUE #362: Implement Completeness Checking Store
Implement other store (#357 is other store) for full Action Cache Validator functionality
- Checks does this exist in the item recently and caches existence bool
	- Doesn't store item in memory
	- Checks if requested recently
	- Pairs nicely with the existence caching
- Might be able to configure with digest -> boolean hashmap in memory store
	- Make optional in memory store

blakehatch:
# Description
Added a completeness checking store to verify all referenced items exist in the CAS before forwarding the items.

Fixes #362 #358

## Type of change

## How Has This Been Tested?
- Tests whether the CAS tree walk works
- Tests whether the CAS tree walk caches when an item exists in the CAS to optimize later checks

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/386)

### DIFF
--- a/cas/store/BUILD
+++ b/cas/store/BUILD
@@ -126,17 +126,42 @@ rust_library(
 )
 
 rust_library(
+    name = "completeness_checking_store",
+    srcs = ["completeness_checking_store.rs"],
+    proc_macro_deps = ["@crate_index//:async-trait"],
+    visibility = ["//cas:__pkg__"],
+    deps = [
+        ":traits",
+        ":ac_utils",
+        "//proto",
+        "//config",
+        "//util:buf_channel",
+        "//util:common",
+        "//util:error",
+        "//util:metrics_utils",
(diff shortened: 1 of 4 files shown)


blakehatch:
Recursive method no longer freezing up, however having trouble checking the result of the boxed future after it's awaited.

Also is the forwarding of the items happening with has_with_results? Seems as though that happens in GrpcStore for example, however the cas walking function does not interact nicely with has_with_results might have to restructure it if that's where it needs to be used.

blakehatch:
Switched to an iterative approach to walk the directory tree to avoid excessive heap allocations from Box and was able to make the method interoperable with has_with_results to prune the results before forwarding them.

PR #397, opened 2023-11-16
Added completeness checking store

### LINKED ISSUE #358: Implement Completeness checking store

blakehatch:
# Description
The proto promises that all results will exist in the CAS when requested using the associated actions. However we currently allow stores to prune themselves which violates this condition, because the action cache and CAS are different. Therefore we need this completeness checking store to check that all results exist in the CAS before we allow the the action result to be returned to the client.

Fixes #358 #362

## Type of change

## How Has This Been Tested?
Added test where all files are present in the cas causing completeness to succeed and another where the cas does not have all of the files causing completeness checking to fail.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/turbo-cache/397)


blakehatch:
The algorithm is has_with_results does parallelize but may be sub-optimal based on how has checks of digests is done.

PR #405, opened 2023-11-17
Add default pre-commit hooks

aaronmondal:
These have mostly no effect and just make sure that trailing newlines are used properly, linebreaks are linux-like and trailing whitespace is removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/405)

PR #407, opened 2023-11-18
Fix subtraction bug in start_turbo_cache.sh

cormacrelf:
- Accept shellcheck SC2155 recommendations in start_turbo_cache.sh
- Remove SCHEDULER_URL as it is both unused & refers to a nonexistent key
- Accept shellcheck SC2004
- Fix double negative when computing remaining memory % in terraform deployment

# Description

I saw this while browsing through the code. There was a double negative sign.
While I was there, I accepted some shellcheck recommendations, which you can 
see in the individual commits.

## Type of change

## How Has This Been Tested?

It has not been tested.

## Checklist

- [ ] PR is contained in a single commit (I left it as separate commits for 
  reviewability. You are welcome to squash it at merge time.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/407)

### DIFF
--- a/deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh
+++ b/deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh
@@ -47,7 +47,7 @@ elif [ "$TYPE" == "cas" ]; then
     # 5% goes to the index memory cache.
     export NATIVE_LINK_CAS_INDEX_CACHE_LIMIT=$(( $TOTAL_AVAIL_MEMORY / 20 ))
     # 85% goes to the content memory cache.
-    export NATIVE_LINK_CAS_CONTENT_LIMIT=$(( $TOTAL_AVAIL_MEMORY - - $NATIVE_LINK_AC_CONTENT_LIMIT - $NATIVE_LINK_CAS_INDEX_CACHE_LIMIT ))
+    export NATIVE_LINK_CAS_CONTENT_LIMIT=$(( $TOTAL_AVAIL_MEMORY - $NATIVE_LINK_AC_CONTENT_LIMIT - $NATIVE_LINK_CAS_INDEX_CACHE_LIMIT ))
 
     /usr/local/bin/native-link /root/cas.json
 elif [ "$TYPE" == "scheduler" ]; then
@@ -56,7 +56,7 @@ elif [ "$TYPE" == "scheduler" ]; then
     # 5% goes to the index memory cache.
     export NATIVE_LINK_CAS_INDEX_CACHE_LIMIT=$(( $TOTAL_AVAIL_MEMORY / 20 ))
     # 85% goes to the content memory cache.
-    export NATIVE_LINK_CAS_CONTENT_LIMIT=$(( $TOTAL_AVAIL_MEMORY - - $NATIVE_LINK_AC_CONTENT_LIMIT - $NATIVE_LINK_CAS_INDEX_CACHE_LIMIT ))
+    export NATIVE_LINK_CAS_CONTENT_LIMIT=$(( $TOTAL_AVAIL_MEMORY - $NATIVE_LINK_AC_CONTENT_LIMIT - $NATIVE_LINK_CAS_INDEX_CACHE_LIMIT ))
 
     /usr/local/bin/native-link /root/scheduler.json
 elif [ "$TYPE" == "worker" ]; then


aaronmondal (changes requested):
Thanks a lot for finding this :heart:

I like these changes a lot, but they overlap with ongoing work to integrate shellcheck in a regression-safe manner via pre-commit hooks.

I'd like to suggest only making the `- -` -> `-` change in this commit and remove all the shellcheck fixes.

Then we can fix all shellcheck issues in a single commit, similar to #405 by adding this to the global `tools/pre-commit-hooks.nix` file:

```
shellcheck.enable = true;
```

This way the `nix flake check` that runs in CI will ensure that we don't regress shellcheck issues. @cormacrelf WDYT? Would you be interested in tackling this?

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/407#-:-NjYcW_K7gfIRQ7DXX5w:b-slltrw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cormacrelf)

allada (commented):
FYI: We have been doing a lot of work behind the scenes to support Azure and GCP. We found some issues in the AWS example in the repo while supporting the other providers but haven't upstreamed them yet.

Thank you very much for your contributions ;-)

> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/407#-:-NjZdL0S-0EsPvraAeej:b2xpe9w)* status: all files reviewed, 1 unresolved discussion (waiting on @cormacrelf)

___
*[`deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh` line 35 at r4](https://reviewable.io/reviews/tracemachina/native-link/407#-NjZe2ab4lHQSR8T06_g:-NjZe2ab4lHQSR8T06_h:bpfwcrc) ([raw file](https://github.com/tracemachina/native-link/blob/9e1cf3422c369099126713ee86b5bcda899242a5/deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh#L35)):*
> ```Shell
> # These environmental variables are used inside the json file resolution.
> export NATIVE_LINK_AWS_REGION="$AWS_REGION"
> NATIVE_LINK_AWS_S3_CAS_BUCKET=$(echo "$TAGS"| jq -r '.Tags[] | select(.Key == "native_link:s3_cas_bucket").Value')
> ```

nit: Any reason why you prefer define then export rather than doing both in one line?
___
*[`deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh` line 63 at r4](https://reviewable.io/reviews/tracemachina/native-link/407#-NjZeHIh4R5UNg1osFtq:-NjZeHIi38qwkZhB_kV4:bumydxv) ([raw file](https://github.com/tracemachina/native-link/blob/7d9deaf9ec113b4984dfd7061d1616c1d029d859/deployment-examples/terraform/AWS/scripts/start_turbo_cache.sh#L63)):*
> ```Shell
>     /usr/local/bin/native-link /root/scheduler.json
> elif [ "$TYPE" == "worker" ]; then
>     SCHEDULER_URL=$(echo "$TAGS"| jq -r '.Tags[] | select(.Key == "native_link:scheduler_url").Value')
> ```

good catch :-)

cormacrelf:
It seems one of you wants me to remove the shellcheck fixes, the other wants me to do them slightly differently. I'm going to get rid of them. You'll see them again when you get CI to check it for you.

I'm not using the AWS deployment at this stage. I might file an issue for the thing I'm thinking of doing (Nix shenanigans).

aaronmondal (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tracemachina/native-link/407#-:-Nja70foDO13OhSS6bR1:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @cormacrelf)

allada:
> It seems one of you wants me to remove the shellcheck fixes, the other wants me to do them slightly differently. I'm going to get rid of them. You'll see them again when you get CI to check it for you.
> 
> I'm not using the AWS deployment at this stage. I might file an issue for the thing I'm thinking of doing (Nix shenanigans).

I see now, I was confused because some things where shellchecked and some were not. I'm in support of adding shellcheck, however I personally go for readability/simplicity, so I'd prefer to never allow our code base to use `export VAR=...` and force the definition and `export` to be separate rather than the alternative. I further debate can be had about it. I was more confused why some lines where changed and others were not, but then realized at a second pass because only the lines with commands were modified. I'd just rather not have other people go through the same thought process and rather just blindly follow a rule.

cormacrelf:
Fair. The reason for the rule is that the export builtin always returns 0, so a $() command can fail silently. So it does not warn about `export VAR="... $whatever"`, but will complain about the actual problematic `export VAR=$(...)` ones. You can blindly "follow the rule" of just doing what shellcheck says, if it's running.

I recommend shellcheck in the editor, e.g. via null-ls for Neovim.

Issue #409, opened 2023-11-18
Sparse files gets de-sparsified 


lukts30:
After execution when output files are copied from the worker to the filesystem CAS directory, they lose their sparse file properties and get fully de-sparsified. This obviously increases disk usage, and the entire process of copying the file & hashing takes considerably more time than a simple move operation followed by a sha256sum calculation on a sparse file.

(byte-by-byte copy and not moved or copied via reflink/copy_file_range).

https://github.com/TraceMachina/native-link/blob/f989e612715a7fe645e69c4c78a50e9b7262ad17/config/examples/basic_cas.json

```
genrule(
  name = "create_sparse_file",
  outs = ["re_large_file"],
  cmd = "fallocate -l 5G $(OUTS) "
)

lukas@PC6061B ~/Downloads/turbo-cache $ filefrag -v tmp/turbo_cache/work/30d5d89de2794c8b81c8ceac8c02d0cc96085888988a72075d39e8b939202650/bazel-out/k8-fastbuild/bin/re_large_file
Filesystem type is: 9123683e
File size of tmp/turbo_cache/work/30d5d89de2794c8b81c8ceac8c02d0cc96085888988a72075d39e8b939202650/bazel-out/k8-fastbuild/bin/re_large_file is 5368709120 (1310720 blocks of 4096 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..   65535: 1863472064..1863537599:  65536:             unwritten
   1:    65536..  196607: 1872450496..1872581567: 131072: 1863537600: unwritten
   2:   196608..  327679: 1881560000..1881691071: 131072: 1872581568: unwritten
   3:   327680..  458751: 1882974994..1883106065: 131072: 1881691072: unwritten
   4:   458752..  524287: 1883657152..1883722687:  65536: 1883106066: unwritten
   5:   524288..  589823: 1883741638..1883807173:  65536: 1883722688: unwritten
   6:   589824..  655359: 1890763862..1890829397:  65536: 1883807174: unwritten
   7:   655360..  720895: 2028229568..2028295103:  65536: 1890829398: unwritten
   8:   720896..  917503: 2028884928..2029081535: 196608: 2028295104: unwritten
   9:   917504.. 1114111: 2030457792..2030654399: 196608: 2029081536: unwritten
  10:  1114112.. 1310719: 2030982080..2031178687: 196608: 2030654400: last,unwritten,eof
[...]

PR #410, opened 2023-11-19
Add function to Store API to get the inner store when possible

allada:
Adds `inner_store()` function to all stores that enables the resolution of inner stores recursively to get an underlying store. This is mostly for places that can perform optimizations when specific code paths can be optimized with specific stores.

towards: #409

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410)

### DIFF
--- a/native-link-scheduler/src/cache_lookup_scheduler.rs
+++ b/native-link-scheduler/src/cache_lookup_scheduler.rs
@@ -57,11 +57,11 @@ pub struct CacheLookupScheduler {
 
 async fn get_action_from_store(
     ac_store: &Arc<dyn Store>,
-    action_digest: &DigestInfo,
+    action_digest: DigestInfo,
     instance_name: String,
 ) -> Option<ProtoActionResult> {
     // If we are a GrpcStore we shortcut here, as this is a special store.
-    let any_store = ac_store.clone().as_any();
+    let any_store = ac_store.clone().inner_store(Some(action_digest)).as_any();
     let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
     if let Some(grpc_store) = maybe_grpc_store {
         let action_result_request = GetActionResultRequest {
@@ -78,7 +78,7 @@ async fn get_action_from_store(
             .map(|response| response.into_inner())
             .ok()
     } else {
-        get_and_decode_digest::<ProtoActionResult>(Pin::new(ac_store.as_ref()), action_digest)
+        get_and_decode_digest::<ProtoActionResult>(Pin::new(ac_store.as_ref()), &action_digest)
             .await
             .ok()
     }
@@ -185,7 +185,7 @@ impl ActionScheduler for CacheLookupScheduler {
             // Perform cache check.
             let action_digest = current_state.action_digest();
             let instance_name = action_info.instance_name().clone();
-            if let Some(action_result) = get_action_from_store(&ac_store, action_digest, instance_name).await {
+            if let Some(action_result) = get_action_from_store(&ac_store, *action_digest, instance_name).await {
                 if validate_outputs_exist(&cas_store, &action_result).await {
                     // Found in the cache, return the result immediately.
                     Arc::make_mut(&mut current_state).stage = ActionStage::CompletedFromCache(action_result);
--- a/native-link-service/src/ac_server.rs
+++ b/native-link-service/src/ac_server.rs
@@ -63,18 +63,19 @@ impl AcServer {
             .get(instance_name)
             .err_tip(|| format!("'instance_name' not configured for '{}'", instance_name))?;
 
+        // TODO(blaise.bruer) We should write a test for these errors.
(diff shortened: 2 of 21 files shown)


allada (commented):
+@aaronmondal 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410#-:-NjcqwLnCqoRKHKCRKl7:b-p12xwa)* status: 0 of 22 files reviewed, all discussions resolved (waiting on @aaronmondal)

aaronmondal (commented):
I'm not sure whether this is an API that should apply "globally" to all stores. I also find it somewhat confusing that e.g. the `FastSlowStore`'s `inner_store` is itself and not e.g. a vector of its fast and slow stores or something like that.

Actually this is something that's been bothering me for a while: In my mind something like the `RedisStore` and the `DedupStore` are quite different. The former actually stores data while the latter is just a wrapper. Our current API doesn't reflect that and I think it could make things like an `inner_store` functionality more intuitive if we split up stores into e.g. "stores" and "store-wrappers". This way it'd be clear that a `store` can't have an inner store (i.e. it's always a leaf node), and a store-wrapper always has some inner store (i.e. it's never a leaf node).

This could also simplify the API for non-wrapper stores as those only need to care about how to read/update data.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410#-:-NjdBRoz2-wXGvdKghSn:b-ami80i)* status: 0 of 22 files reviewed, all discussions resolved

allada (commented):
I partially agree. In the case of `DedupStore`, there is no actual inner store, the inner store is itself. What I'm trying to do is create a case where `inner_store` will resolve any stores that pass through the data but do no complex logic on it. This will allow us to use things like:
```
    "GRPC_CAS_STORE": {
      "size_partitioning": {
        "size": 10000,
        "lower": {
          "grpc": {
            // ...
          }
        },
        "upper": {
          "s3_store": {
            // ...
          }
        }
      }
    },
    "WORKER_FAST_SLOW_STORE": {
      "fast_slow": {
        "fast": {
          "filesystem": {
            // ...
          }
        },
        "slow": {
          "ref_store": {
            "ref": "CAS_STORE"
          }
        }
      }
    }
```

This config would allow the GRPC optimizations we have laying around significantly more frequent.

The specific use case that I am doing this for is for #409, what I plan on doing is checking to see if the store I'm uploading to when a worker is uploading it's result is a filesystem store. If it is, it'll simply downcast it to the filesystem store and then move the file directly instead of coping the data. The reason I wanted to do this first is because `ref_store` is frequently used when the worker lives on the same machine as the CAS.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410#-:-Njeh5Oa4_fngBMS0kne:bvixmpk)* status: 0 of 22 files reviewed, all discussions resolved (waiting on @aaronmondal)

aaronmondal (approved):
:lgtm:

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410#-:-Njiqw8JC5xksS5lI4i1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

___
*[`cas/store/shard_store.rs` line 178 at r1](https://reviewable.io/reviews/TraceMachina/native-link/410#-NjiqDL6E6jlL1SHdgG9:-NjiqDL71m2MMWDjLqc6:b-elkxju) ([raw file](https://github.com/TraceMachina/native-link/blob/c98baaf184be772dfaedd823eaf7e7f68b644936/cas/store/shard_store.rs#L178)):*
> ```XML
>         };
>         let index = self.get_store_index(digest);
>         self.weights_and_stores[index].1.clone().inner_store(Some(digest))
> ```

nit: Might be worth considering some sort of struct to get rid of the `1` syntax here. Not too relevant though.

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410#-:-Nk2oJKu6Xlk_WGstkLc:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @allada)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/410)* status: 0 of 43 files reviewed, all discussions resolved (waiting on @aaronmondal)

___
*[`cas/store/shard_store.rs` line 178 at r1](https://reviewable.io/reviews/TraceMachina/native-link/410#-NjiqDL6E6jlL1SHdgG9:-NkcqifO7qbbeZTO8t7R:boc3ul1) ([raw file](https://github.com/TraceMachina/native-link/blob/c98baaf184be772dfaedd823eaf7e7f68b644936/cas/store/shard_store.rs#L178)):*

nit: Might be worth considering some sort of struct to get rid of the `1` syntax here. Not too relevant though.
</blockquote></details>

Kinda agree, but not for this PR.

PR #426, opened 2023-11-28
Fix lost file data in filesystem store

Jirixek:
# Description

This patch fixes lost data in the case when a file was closed due to a timeout.

The bug manifests itself as follows:
In the first loop pass in get_part_ref, everything works smoothly. After the writer times out however, the buffer is split again. At this point, the buffer is empty, original data is lost, and an empty buffer is passed to writer.send(), causing an assert error.

## Type of change

## How Has This Been Tested?

By building [envoy](https://github.com/envoyproxy/envoy) extensions tests using
```
bazel test //test/extensions/...
```
I've been building this using 128 core machine, so the build is very IO intensive.
`basic_cas.json` config is enough to reproduce the issue. After this patch, the errors are gone.

As for the unit testing, I tried to create a test.
```[rust]
    #[tokio::test]
    async fn get_part_timeout_test() -> Result<(), Error> {
        let digest1 = DigestInfo::try_new(HASH1, VALUE1.len())?;
        let content_path = make_temp_path("content_path");
        let temp_path = make_temp_path("temp_path");

        fs::set_idle_file_descriptor_timeout(Duration::from_millis(100))
            .err_tip(|| "Error setting idle file descriptor timeout")?;

        assert_eq!(fs::idle_file_descriptor_timeout(), Duration::from_millis(100));

        let store = Arc::new(
            FilesystemStore::<FileEntryImpl>::new(&native_link_config::stores::FilesystemStore {
                content_path: content_path.clone(),
                temp_path: temp_path.clone(),
                eviction_policy: Some(native_link_config::stores::EvictionPolicy {
                    max_count: 3,
                    ..Default::default()
                }),
                ..Default::default()
            })
            .await?,
        );

        let store_pin = Pin::new(store.as_ref());
        // Insert data into store.
        store_pin.as_ref().update_oneshot(digest1, VALUE1.into()).await?;

        let (writer, mut reader) = make_buf_channel_pair();
        let store_clone = store.clone();
        let digest1_clone = digest1;

        tokio::spawn(async move { Pin::new(store_clone.as_ref()).get(digest1_clone, writer).await });

        // Check to ensure our first byte has been received. The future should be stalled here.
        let first_byte = DropCloserReadHalf::take(&mut reader, 1)
            .await
            .err_tip(|| "Error reading first byte")?;
        assert_eq!(first_byte[0], VALUE1.as_bytes()[0], "Expected first byte to match");

        sleep(Duration::from_millis(200)).await;

        let remaining_file_data = DropCloserReadHalf::take(&mut reader, 1024)
            .await
            .err_tip(|| "Error reading remaining bytes")?;

        assert_eq!(
            &remaining_file_data,
            VALUE1[1..].as_bytes(),
            "Expected file content to match"
        );

        Ok(())
    }
```
However, the test passes even before the patch, because the future in `get_part_ref()` doesn't time out for some reason. I don't consider myself a rust programmer, do you think you could help me with this?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/426)

### DIFF
--- a/native-link-store/src/filesystem_store.rs
+++ b/native-link-store/src/filesystem_store.rs
@@ -648,6 +648,7 @@ impl<Fe: FileEntry> Store for FilesystemStore<Fe> {
             if buf.is_empty() {
                 break; // EOF.
             }
+            let buf_content = buf.split().freeze();
             // In the event it takes a while to send the data to the client, we want to close the
             // reading file, to prevent the file descriptor left open for long periods of time.
             // Failing to do so might cause deadlocks if the receiver is unable to receive data
@@ -655,7 +656,7 @@ impl<Fe: FileEntry> Store for FilesystemStore<Fe> {
             // Using `ResumeableFileSlot` will re-open the file in the event it gets closed on the
             // next iteration.
             loop {
-                match timeout(fs::idle_file_descriptor_timeout(), writer.send(buf.split().freeze())).await {
+                match timeout(fs::idle_file_descriptor_timeout(), writer.send(buf_content.clone())).await {
                     Ok(Ok(())) => break,
                     Ok(Err(err)) => {
                         return Err(err).err_tip(|| "Failed to send chunk in filesystem store get_part");


allada (commented):
Thank you very much for this PR. I can confirm that this is an issue. I have asked @adam-singer to review and get the test working.

We have a soft-rule that we don't let a "fix" committed without a regression test as well, so this will be stalled on Adam helping get the test working.

Thanks so much :smile: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/426#-:-NkR-OHwDdUMMpXrmKbV:b-wrptwe)* status: 0 of 1 files reviewed, all discussions resolved

allada (commented):
+@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/426#-:-NkR-nllEt_Llpuov7PP:b-bn1n1c)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @adam-singer)

Jirixek:
Superseded by #439.

PR #442, opened 2023-12-04
Make Native Link installable via nix

aaronmondal:
This commit makes Native Link directly buildable/runnable via nix:

```bash
nix run github:TraceMachina/native-link /path/to/config.json
```

To reduce closure size for upcoming container images the environment that builds the derivation is a fairly customized LLVM-based toolchain. This toolchain should make it easier to migrate to more sophisticated approaches in the future.

Test infrastructure is not yet enabled as it'll be a larger effort to make our testsuite compatible with strictly controlled environments like the nix build sandbox.

Importantly, the updated README now makes extensive use of emojis ✨

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/442)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,19 @@ various Cloud CLIs for you:
     @rules_rust//:rustfmt
   ```
 
+## Updating Rust dependencies
+
+After modifying the corresponding `Cargo.toml` file in either the top level or
+one of the crate subdirectories run the following command to rebuild the crate
+index:
+
+```
+CARGO_BAZEL_REPIN=1 bazel sync --only=crate_index
+```
+
+This updates `Cargo.lock` and `Cargo.Bazel.lock` with the new dependency
+information.
+
 ## Conduct
 
 Native Link Code of Conduct is available in the
--- a/README.md
+++ b/README.md
@@ -2,42 +2,88 @@
 
 [![CI](https://github.com/tracemachina/native-link/workflows/CI/badge.svg)](https://github.com/tracemachina/native-link/actions/workflows/main.yml)
 
-An extremely fast and efficient bazel cache service (CAS) written in rust.
+Native link is an extremely (blazingly?) fast and efficient build cache and
+remote executor for systems that communicate using the [Remote execution
+protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto) such as [Bazel](https://bazel.build), [Buck2](https://buck2.build), [Goma](https://chromium.googlesource.com/infra/goma/client/) and
+[Reclient](https://github.com/bazelbuild/reclient).
 
-The goals of this project are:
-1. Stability - Things should work out of the box as expected
-2. Efficiency - Don't waste time on inefficiencies &amp; low resource usage
-3. Tested - Components should have plenty of tests &amp; each bug should be regression tested
-4. Customers First - Design choices should be optimized for what customers want
+Supports Unix based operating systems and Windows.
 
-## Overview
+## ❄️ Installing with Nix
 
-Native Link is a project that implements the Bazel's [Remote Execution protocol](https://github.com/bazelbuild/remote-apis) (both CAS/Cache and remote execution portion).
+**Installation requirements:**
 
-When properly configured this project will provide extremely fast and efficient build cache for any systems that communicate using the [RBE protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto) and/or extremely fast, efficient and low foot-print remote execution capability.
+* Nix with [flakes](https://nixos.wiki/wiki/Flakes) enabled
 
-Unix based operating systems and Windows are fully supported.
+This build does not require cloning the repository, but you need to provide a
+config file, for instance the one at [native-link-config/examples/basic_cas.json](./native-link-config/examples/basic_cas.json).
 
-## TL;DR
+The following command builds and runs Native Link in release (optimized) mode:
 
-If you have not updated Rust or Cargo recently, run:
-
-`rustup update`
-
-To compile and run the server:
 ```sh
-# Install dependencies needed to compile Native Link with bazel on
-# worker machine (which is this machine).
-apt install -y gcc g++ lld python3
+nix run github:TraceMachina/native-link ./basic_cas.json
+```
 
-# Install cargo (if needed).
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+For use in production pin the executable to a specific revision:
 
-# --release causes link-time-optmization to be enabled, which can take a while
-# to compile, but will result in a much faster binary.
+```sh
+nix run github:TraceMachina/native-link/<revision> ./basic_cas.json
+```
+
+## 🌱 Building with Bazel
+
+**Build requirements:**
+
+* Bazel 6.4.0+
+* A recent C++ toolchain with LLD as linker
+
+> [!TIP]
+> This build supports Nix/direnv which provides Bazel but no C++ toolchain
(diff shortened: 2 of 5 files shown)


aaronmondal (commented):
+@allada @MarcusSorealheis @adam-singer @blakehatch 

The updated readme is more readable at https://github.com/aaronmondal/turbo-cache/tree/nix-package

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/442#-:-Nkqh-Gb2xpdN_O_qNDP:b461k2w)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @allada)

aaronmondal (commented):
+@MarcusSorealheis +@adam-singer +@blakehatch 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/442#-:-NkqhN2aBVE2p6pBZfIJ:b-s7wjj5)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @adam-singer, @allada, @blakehatch, and @MarcusSorealheis)

MarcusSorealheis (approved):
I've already reviewed this change based on the other PR I just reviewed, but I am reserving my overarching comments for this PR. It's very large and proposes a lot of changes. I think we still need to have a way to download the library via Cargo and the Docker container registry.

allada (approved):
Not gunna lie, I laughed a lot at the emojis :laughing: 
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/442#-:-NksF2kl-LKvRmvHwjbf:bwpigqw)* status: all files reviewed, 4 unresolved discussions (waiting on @aaronmondal, @adam-singer, @blakehatch, and @MarcusSorealheis)

___
*[`flake.nix` line 45 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksPXlZ5f3kInXMXujH:-NksPXlZ5f3kInXMXujI:b-pf4n3e) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/flake.nix#L45)):*
> ```Nix
> 
>           native-link = craneLib.buildPackage (commonArgs
>             // {
> ```

nit: needless?
___
*[`README.md` line 7 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksPnfcAtK_9o8VYDFY:-NksPnfcAtK_9o8VYDFZ:b-npsg6o) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/README.md#L7)):*
> ```Markdown
> Native link is an extremely (blazingly?) fast and efficient build cache and
> remote executor for systems that communicate using the [Remote execution
> protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto) such as [Bazel](https://bazel.build), [Buck2](https://buck2.build) and [Goma](https://chromium.googlesource.com/infra/goma/client/).
> ```

nit: Maybe add ReClient?
___
*[`README.md` line 54 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksQUm56J5aUETLPzfn:-NksQUm56J5aUETLPzfo:bkad03e) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/README.md#L54)):*
> ```Markdown
> 
> # Unoptimized development build on Windows
> bazel run --config=windows cas
> ```

nit: the Cargo example uses `-- ./native-link-config/examples/basic_cas.json`, should we do the same here?
___
*[`README.md` line 65 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksQDFrBrJgbi6BHOaT:-NksQDFrBrJgbi6BHOaU:bxh1jpw) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/README.md#L65)):*
> ```Markdown
> 
> * Cargo 1.73.0+
> * A recent C++ toolchain with LLD as linker
> ```

Does Cargo need a C++ toolchain? I thought building protoc was the only C++ code required, which is not needed by cargo.

aaronmondal (commented):
@MarcusSorealheis Agreed that we want `cargo install` and direct `docker run` support for images tagged by a semantic version. Both of those require a tagged release. Regarding Containers #443 gives us almost everything except for SemVer tag handling. Regarding Cargo support I'll add another workflow that publishes nlink on crates.io.

@allada :seedling: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/442#-:-NktwaIi3WPq9sNxzMDs:b-u11uw3)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @adam-singer, @allada, @blakehatch, and @MarcusSorealheis)

___
*[`flake.nix` line 45 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksPXlZ5f3kInXMXujH:-NktzRdE9xUnQtVav13R:b40och8) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/flake.nix#L45)):*

nit: needless?
</blockquote></details>

For the currently enabled code we could indeed say

```nix
native-link = craneLib.buildPackage (commonArgs
  // {
  cargoArtifacts = craneLib.buildDepsOnly commonArgs;
});
```

Which also suggests using an immediately invoked lambda:

```nix
native-link = (args:
  (craneLib.buildPackage (args // {
    cargoArtifacts = craneLib.buildDepsOnly args;
  })
commonArgs)
```

But I think the way it's now is fine as it makes it easier to uncomment the `tests` block while ensuring that we're always using the same `cargoArtifacts`.

The merge operator `//` doesn't recurse and `inherit x` is the same as saying `x = x`. Expanded this code says something like:

```nix
native-link = craneLib.buildPackage {
  inherit src;
  buildInputs = [];
  nativeBuildInputs = [ pkgs.autoPatchelfHook pkgs.cacert ];
  stdenv = customStdenv;
  cargoArtifacts = craneLib.buildDepsOnly {
    inherit src;
    buildInputs = [ ];
    nativeBuildInputs = [ pkgs.autoPatchelfHook pkgs.cacert ];
    stdenv = customStdenv;
  };
};
```

Note that `buildDepsOnly` and `buildPackage` are technically entirely different functions that "just happen" to both accept the `commonArgs` attrset as argument.

Without this the two functions would end up using default arguments and for instance use a different stdenv/toolchain for the deps and the `native-link-*` crates.
___
*[`README.md` line 65 at r1](https://reviewable.io/reviews/TraceMachina/native-link/442#-NksQDFrBrJgbi6BHOaT:-Nktt1yY9GsrJfde0UPL:bww59pl) ([raw file](https://github.com/TraceMachina/native-link/blob/bb5db4f3449c0f9ef1ff03074f7aad9dc6147f85/README.md#L65)):*

Does Cargo need a C++ toolchain? I thought building protoc was the only C++ code required, which is not needed by cargo.
</blockquote></details>

This is a tricky one. Technically you're correct in that Cargo only depends on the linker but not a full C++ toolchain. However, in most configurations Cargo won't use the linker directly and instead use `clang++` or `g++` as "wrapped linker" since those are preconfigured to link dynamic dependencies like `libc`, `libm` and the dynamic linker such as`ld-linux-x86-64.so.2`.

It's possible to configure Cargo with a "raw" linker and ideally this would be the default behavior. But at the moment this would require users to explicitly configure linking of the non-rust shared objects.

I'll keep this as it is for now and will look into whether we can configure our Cargo build in a way that makes it depend on just a raw linker for all users so that we can safely remove this requirement. Using musl by default likely makes this dependency obsolete as it'll allow us to default to fully static builds where we don't need to worry about distro-specific locations of shared objects.

PR #453, opened 2023-12-07
Add docs to monorepo

### LINKED ISSUE #452: Add Docs Folder to Monorepo
Add Mintlify docs to a new folder called Native-Link-Docs.

blakehatch:
# Description

Adds Mintlify docs to monorepo

Fixes #452 
## Type of change

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/453)


aaronmondal (commented):
After more thinking and researching it seems to me that Docusaurus (https://docusaurus.io/) is the lowest-risk choice.

As migration path we can then switch to Astro (https://docs.astro.build/de/guides/migrate-to-astro/from-docusaurus/) in the future and can swap components to qwik over time (https://qwik.builder.io/docs/integrations/astro/) for further optimization (for instance, aided by mitosis (https://github.com/BuilderIO/mitosis)).

This way we can start off with something simple, mature and stable, but keep the flexibility to migrate to more modern approaches in the future.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/453#-:-Nl3YK2a8jnHe4Y96ga0:b-ljktqx)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MarcusSorealheis)

blakehatch (commented):
These are good concerns to bring up. The biggest deal-breaker that would make the decision simple is if we want to improve lighthouse scores we will need to move off of Mintlify, it's a highly abstracted framework that's quick to develop on but offers no way to optimize further than it already is. Also the ecosystem of tooling allowing us to not have to maintain documentation in two separate places between the internal docs and the website.

I'm happy to re-build this in something like docusaurus and if we decide that's something worth doing it's probably best to do at this early stage while the docs are lean.

As far as the logo I'm personally not too concerned as the rocket taking off is such a common design element but it's an understandable worry.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/453#-:-Nl3yN9e2cnDIeAobunv:b20798v)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MarcusSorealheis)

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/453)* status: all files reviewed, 4 unresolved discussions (waiting on @aaronmondal, @blakehatch, and @MarcusSorealheis)

___
*[`native-link-docs/quickstart.mdx` line 6 at r2](https://reviewable.io/reviews/TraceMachina/native-link/453#-NlA7dXP5ELkVZ2i5K0c:-NlAWDAB9iYcp2c6bAks:b-gqvp6l) ([raw file](https://github.com/TraceMachina/native-link/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/quickstart.mdx#L6)):*

Section out of sync with current readme.

I see that you made some customizations to the previous readme though, so let me comment on those changes as well:

1. Users on MacOS shouldn't be using GCC. They should use Clang as C++ toolchain instead which is supports MacOS natively.
2. There is no need for custom setups since we have the nix flake.
3. Don't use `apt` in any installation commands as it's not cross-platform. These commands won't work on non-debian distros and don't even work on all Ubuntu systems. They're also irreproducible, meaning that on e.g. Ubuntu 18 you'll get a toolchain that can't build our project while on Ubuntu 22 it'll work. These kinds of issues are incredibly frustrating to users, so let's prevent this and just recommend the flake which is the exact same environment regardless of whether you run it on Ubuntu 18, upstream Gentoo or MacOS.
</blockquote></details>

opinion: 

I think popularizing nix flake is a good thing, maybe that can be in the header section of this document as one way to bootstrap. That being said I don't think it is very mainstream. It could be a turn off if thats the only instructions they see to bootstrap / run tests of this project. One way to look at it, someone is having to research this project for work related requirements, usually have a work issued laptop/environment, generally introducing non-standard (or commonly known) tooling add a layer of intimidation / fear. I would personally in work environments take the `apt` commands over `nix` on local or development machines because they would be familiar and commonly seen with other setups. If I was working on my personal computer, I'd probably use `nix` as another fun tool to try out, messing up a personal laptop doesn't come with the same type of associated risk as a work laptop environment (assuming we are talking about local machines).

For each of the sections on this doc, I would also cross link to the general owning tools websites that provide bootstrap docs, https://rustup.rs/ / https://bazel.build/install/bazelisk / etc.. it is nice to have quick copy and paste in case something is missing, but these things are usually so commonly available on most systems that the point of documentation here is to know you need them (if you don't have them already) vs installing them.

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/453)* status: all files reviewed, 4 unresolved discussions (waiting on @adam-singer, @blakehatch, and @MarcusSorealheis)

___
*[`native-link-docs/quickstart.mdx` line 6 at r2](https://reviewable.io/reviews/TraceMachina/native-link/453#-NlA7dXP5ELkVZ2i5K0c:-NlAkxUQAhNTJ9CeHVmi:ba65h2t) ([raw file](https://github.com/TraceMachina/native-link/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/quickstart.mdx#L6)):*

opinion: 

I think popularizing nix flake is a good thing, maybe that can be in the header section of this document as one way to bootstrap. That being said I don't think it is very mainstream. It could be a turn off if thats the only instructions they see to bootstrap / run tests of this project. One way to look at it, someone is having to research this project for work related requirements, usually have a work issued laptop/environment, generally introducing non-standard (or commonly known) tooling add a layer of intimidation / fear. I would personally in work environments take the `apt` commands over `nix` on local or development machines because they would be familiar and commonly seen with other setups. If I was working on my personal computer, I'd probably use `nix` as another fun tool to try out, messing up a personal laptop doesn't come with the same type of associated risk as a work laptop environment (assuming we are talking about local machines).

For each of the sections on this doc, I would also cross link to the general owning tools websites that provide bootstrap docs, https://rustup.rs/ / https://bazel.build/install/bazelisk / etc.. it is nice to have quick copy and paste in case something is missing, but these things are usually so commonly available on most systems that the point of documentation here is to know you need them (if you don't have them already) vs installing them.
</blockquote></details>

Agreed that it can be daunting to install nix. It's just one command, but that can be off-putting.

On the other hand, I'd also never recommend doing anything with `apt` though, since I've spent literal days of my life helping users figure out that their toolchains were too old because they were using outdated OSes. I'd also argue that users more open to trying out new things tend to also have custom distros like Arch, Gentoo and void. From personal experience it's often quite annoying if installation commands don't actually contain all the dependencies necessary. For instance, on Ubuntu the `clang` package will also make some libcxx available. On Gentoo that's not the case and the installation of a full toolchain requires more than one package. Translating commands like that is poor UX.

The way to go might actually be neither: We have prebuilt images, so we could just tell users to use those via `docker run ghcr.io/tracemachina/native-link:xxx`. That's something that should work for everyone and doesn't require any additional tooling besides docker which we can everyone to know how to install.

The other dependencies are technically only required to build from source. I'd argue that those building from source will be open to more exotic tooling as it's common that every project uses custom build infra and users building from source will generally be aware that they'll need to adjust to the project's build system.

I think it makes sense to have two sections. One using just something like 

```
git clone xxx && \
  cd xxx && \
  docker run ghcr.io/xxx \
    -v native-link-config/examples/basic_cas.json:/config.json \
    -p 50051:50051 \
    -p 50052:50052 \
    /config.json
```
for ease of use and one "Building from source" section that contains more info on the more elaborate setups.

blakehatch (commented):
Gonna keep open, was able to fix the github submodule issue.

> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/453#-:-NlR_Q2D0PfyTJ4aPSJa:biy7dl5)* status: 6 of 12 files reviewed, 4 unresolved discussions (waiting on @aaronmondal and @MarcusSorealheis)

___
*[`native-link-docs/favicon.svg` line 21 at r2](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlA7Hj7EUPVncfNvEIO:-NlRVDH7Aj57YV62wx_6:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/favicon.svg#L21)):*

Missing newline. The pre-commit hooks in the nix flake should autoformat this for you.
</blockquote></details>

Done.
___
*[`native-link-docs/introduction.mdx` line 17 at r2](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlA7Rwo9sJoU5vdedH9:-NlRZY2X3mXQ1Wc8Ybm3:b-histo6) ([raw file](https://github.com/tracemachina/nativelink/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/introduction.mdx#L17)):*

Section out of date with current readme.
</blockquote></details>

Updated to match
___
*[`native-link-docs/quickstart.mdx` line 6 at r2](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlA7dXP5ELkVZ2i5K0c:-NlRZcAc6YfMUyEnVqBA:b-nb9b2s) ([raw file](https://github.com/tracemachina/nativelink/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/quickstart.mdx#L6)):*

Agreed that it can be daunting to install nix. It's just one command, but that can be off-putting.

On the other hand, I'd also never recommend doing anything with `apt` though, since I've spent literal days of my life helping users figure out that their toolchains were too old because they were using outdated OSes. I'd also argue that users more open to trying out new things tend to also have custom distros like Arch, Gentoo and void. From personal experience it's often quite annoying if installation commands don't actually contain all the dependencies necessary. For instance, on Ubuntu the `clang` package will also make some libcxx available. On Gentoo that's not the case and the installation of a full toolchain requires more than one package. Translating commands like that is poor UX.

The way to go might actually be neither: We have prebuilt images, so we could just tell users to use those via `docker run ghcr.io/tracemachina/native-link:xxx`. That's something that should work for everyone and doesn't require any additional tooling besides docker which we can everyone to know how to install.

The other dependencies are technically only required to build from source. I'd argue that those building from source will be open to more exotic tooling as it's common that every project uses custom build infra and users building from source will generally be aware that they'll need to adjust to the project's build system.

I think it makes sense to have two sections. One using just something like 

```
git clone xxx && \
  cd xxx && \
  docker run ghcr.io/xxx \
    -v native-link-config/examples/basic_cas.json:/config.json \
    -p 50051:50051 \
    -p 50052:50052 \
    /config.json
```
for ease of use and one "Building from source" section that contains more info on the more elaborate setups.
</blockquote></details>

I decided to go with apt while including Nix in an "Advanced Setup" section. From my understanding most of our users will be debian-based with context switching being fairly simple from apt to other package managers however I can include instructions for others (Pacman, etc.) if there is a compelling reason. Seems like a lot of the users in alternative Distros will prefer/be comfortable using Nix. 
___
*[`native-link-docs/README.md` line 1 at r2](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlA7s0e2D52pZmyRxJU:-NlR_xyBD9PNS-5AmYos:b-n6ut6e) ([raw file](https://github.com/tracemachina/nativelink/blob/a5baa9d7e9c66fdbbbbb844e76943c3991b4dadf/native-link-docs/README.md#L1)):*

Let's use our own readme here.
</blockquote></details>

Modified to just include necessary instructions for local development and changed header.

aaronmondal (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/453)* status: all files reviewed, 2 unresolved discussions (waiting on @blakehatch and @MarcusSorealheis)

___
*[`native-link-docs/quickstart.mdx` line 26 at r3](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlTZ4Qt1sYuNnogb32G:-NlTZ4Qt1sYuNnogb32H:bjwpbqw) ([raw file](https://github.com/tracemachina/nativelink/blob/1d7ff74aa6185d585303a1671e3b36f9bb9e4820/native-link-docs/quickstart.mdx#L26)):*
> ```MDX
> 
> * Bazel 6.4.0+
> * gcc (exc. MacOS)
> ```

I'd recommend clang instead of gcc/g++. That's what rust uses under the hood (and where LLD comes from as well https://github.com/llvm/llvm-project/tree/main/lld) and works on Linux, MacOS and Windows.
___
*[`native-link-docs/quickstart.mdx` line 227 at r3](https://reviewable.io/reviews/tracemachina/nativelink/453#-NlTZyaHBw9we5Fm1eEp:-NlTZyaHBw9we5Fm1eEq:b-4uso8c) ([raw file](https://github.com/tracemachina/nativelink/blob/1d7ff74aa6185d585303a1671e3b36f9bb9e4820/native-link-docs/quickstart.mdx#L227)):*
> ```MDX
>   completely clean and aligned with the commit you want to reproduce. 
>   Otherwise the image will be tainted with a `"dirty"` revision label. 
> </Tip>
> ```

nit: Newline.

aaronmondal (approved):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/453)* status: all files reviewed, 2 unresolved discussions (waiting on @blakehatch and @MarcusSorealheis)

PR #458, opened 2023-12-08
Remove caching from native cargo action

aaronmondal:
If we share caches between workflows, the runners run out of storage. If we separate the caches we have too many of them and keep stepping over GitHub's cache limitations. Remove caching entirely as this workflows doesn't take *that* long and we need the caches for Bazel and Nix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/458)

### DIFF
--- a/.github/workflows/native-cargo.yaml
+++ b/.github/workflows/native-cargo.yaml
@@ -30,21 +30,6 @@ jobs:
           && rustup default ${{ matrix.toolchain }}
         shell: bash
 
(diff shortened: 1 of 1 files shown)


adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/458)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #503, opened 2023-12-18
Integrate Google Analytics

blakehatch:
# Description
Integrates google analytics into the documentation site to monitor user activity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/503)


MarcusSorealheis (approved):
should be fine to house this detail here. It is available on the client side anyway, and i bet GA requires `txt` verification or something

blakehatch:
> should be fine to house this detail here. It is available on the client side anyway, and i bet GA requires `txt` verification or something

Yes Google Analytics verifies that the URL the Measurement ID is linked with a URL that we control and therefore cannot be used externally.

PR #526, opened 2023-12-19
Add small sleep in some tests to reduce flakes in CI

allada:
This is to get around a docker-in-ci bug with how the host machine
is not able to sync files to disk properly.
    
see: #527

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/526)


MarcusSorealheis (approved):
lgtm

MarcusSorealheis:
Does anyone know what is the cause of the panic`k` from `memory_store`?

MarcusSorealheis (approved):
looks better and is temporary

PR #575, opened 2023-12-22
Add Blake3 support for Verify Store

### LINKED ISSUE #444: VerifyStore should support Blake3
Currently VerifyStore is Sha256 specific. We should make this store support different digest functions.

see: https://github.com/TraceMachina/native-link/issues/191#issuecomment-1839543501
and:
https://github.com/TraceMachina/native-link/blob/f9e537c3f9b5656be6251902640ff003a5b8cc48/native-link-store/src/verify_store.rs#L140

steed924:
# Description

I've added Blake3 hash support Verify Store.
I just used the `digest_hasher` utiliy module to keep consistency.
For simplicity, I just used the default hash function which is set from the global config.
I've done integration tests on my machine to verify blake3 function actually works.
But since the global hash function is `OnceLock` and it can't be set twice, I couldn't add specific unit tests for blake3 and sha256 separately.
May it would be better if we add hash function config variable for the Verify Store Config?
Or is it just adding extra complexity?
I'd like to make this clear.

Fixes #444 

## Type of change

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)

### DIFF
--- a/nativelink-config/src/cas_server.rs
+++ b/nativelink-config/src/cas_server.rs
@@ -20,7 +20,7 @@ use crate::schedulers::SchedulerConfig;
 use crate::serde_utils::{
     convert_numeric_with_shellexpand, convert_optinoal_numeric_with_shellexpand, convert_string_with_shellexpand,
 };
-use crate::stores::{StoreConfig, StoreRefName};
+use crate::stores::{ConfigDigestHashFunction, StoreConfig, StoreRefName};
 
 /// Name of the scheduler. This type will be used when referencing a
 /// scheduler in the `CasConfig::schedulers`'s map key.
@@ -541,18 +541,6 @@ pub struct LocalWorkerConfig {
 pub enum WorkerConfig {
     /// A worker type that executes jobs locally on this machine.
     local(LocalWorkerConfig),
-}
-
-#[allow(non_camel_case_types)]
-#[derive(Deserialize, Debug, Clone, Copy)]
-pub enum ConfigDigestHashFunction {
-    /// Use the sha256 hash function.
-    /// <https://en.wikipedia.org/wiki/SHA-2>
-    sha256,
-
-    /// Use the blake3 hash function.
-    /// <https://en.wikipedia.org/wiki/BLAKE_(hash_function)>
-    blake3,
 }
 
 #[derive(Deserialize, Debug, Clone, Copy)]
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -21,6 +21,18 @@ use crate::serde_utils::{convert_numeric_with_shellexpand, convert_string_with_s
 pub type StoreRefName = String;
 
 #[allow(non_camel_case_types)]
+#[derive(Serialize, Deserialize, Debug, Clone, Copy)]
+pub enum ConfigDigestHashFunction {
+    /// Use the sha256 hash function.
+    /// <https://en.wikipedia.org/wiki/SHA-2>
+    sha256,
+
+    /// Use the blake3 hash function.
+    /// <https://en.wikipedia.org/wiki/BLAKE_(hash_function)>
+    blake3,
+}
+
+#[allow(non_camel_case_types)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum StoreConfig {
     /// Memory store will store all data in a hashmap in memory.
@@ -329,12 +341,13 @@ pub struct VerifyStore {
     #[serde(default)]
     pub verify_size: bool,
 
-    /// If set this store will hash the contents and verify it matches the
-    /// digest hash before writing the entry to underlying store.
+    /// The digest hash function to hash the contents and to verify if the digest hash is
+    /// matching before writing the entry to underlying store.
     ///
-    /// This should be set to false for AC, but true for CAS stores.
-    #[serde(default)]
-    pub verify_hash: bool,
+    /// If None, the hash verification will be disabled.
+    ///
+    /// This should be set to None for AC, but hashing function like `sha256` for CAS stores.
+    pub hash_verification_function: Option<ConfigDigestHashFunction>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
--- a/nativelink-store/src/verify_store.rs
+++ b/nativelink-store/src/verify_store.rs
@@ -17,17 +17,18 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use nativelink_config::stores::ConfigDigestHashFunction;
 use nativelink_error::{make_input_err, Error, ResultExt};
 use nativelink_util::buf_channel::{make_buf_channel_pair, DropCloserReadHalf, DropCloserWriteHalf};
 use nativelink_util::common::DigestInfo;
+use nativelink_util::digest_hasher::{DigestHasher, DigestHasherFunc};
 use nativelink_util::metrics_utils::{Collector, CollectorState, CounterWithTime, MetricsComponent, Registry};
 use nativelink_util::store_trait::{Store, UploadSizeInfo};
-use sha2::{Digest, Sha256};
 
 pub struct VerifyStore {
     inner_store: Arc<dyn Store>,
     verify_size: bool,
-    verify_hash: bool,
+    hash_verification_function: Option<ConfigDigestHashFunction>,
 
     // Metrics.
     size_verification_failures: CounterWithTime,
@@ -39,7 +40,7 @@ impl VerifyStore {
         VerifyStore {
             inner_store,
             verify_size: config.verify_size,
-            verify_hash: config.verify_hash,
+            hash_verification_function: config.hash_verification_function,
             size_verification_failures: CounterWithTime::default(),
             hash_verification_failures: CounterWithTime::default(),
         }
@@ -54,7 +55,8 @@ impl VerifyStore {
         mut tx: DropCloserWriteHalf,
         mut rx: DropCloserReadHalf,
         size_info: UploadSizeInfo,
-        mut maybe_hasher: Option<([u8; 32], Sha256)>,
+        original_hash: [u8; 32],
+        mut maybe_hasher: Option<DigestHasher>,
     ) -> Result<(), Error> {
         let mut sum_size: u64 = 0;
         loop {
@@ -76,8 +78,9 @@ impl VerifyStore {
                         ));
                     }
                 }
-                if let Some((original_hash, hasher)) = maybe_hasher {
-                    let hash_result: [u8; 32] = hasher.finalize().into();
+                if let Some(hasher) = maybe_hasher.as_mut() {
+                    // We are passing -1 here because we just need to get the hashing result not the size.
+                    let hash_result: [u8; 32] = hasher.finalize_digest(-1).packed_hash;
                     if original_hash != hash_result {
                         self.hash_verification_failures.inc();
                         return Err(make_input_err!(
@@ -94,7 +97,7 @@ impl VerifyStore {
             // This will allows us to hash while sending data to another thread.
             let write_future = tx.send(chunk.clone());
 
-            if let Some((_, hasher)) = maybe_hasher.as_mut() {
+            if let Some(hasher) = maybe_hasher.as_mut() {
                 hasher.update(chunk.as_ref());
             }
 
@@ -135,15 +138,14 @@ impl Store for VerifyStore {
             }
         }
 
-        let mut hasher = None;
-        if self.verify_hash {
-            hasher = Some((digest.packed_hash, Sha256::new()));
-        }
+        let hasher = self
+            .hash_verification_function
+            .map(|v| DigestHasher::from(DigestHasherFunc::from(v)));
 
         let (tx, rx) = make_buf_channel_pair();
 
         let update_fut = self.pin_inner().update(digest, rx, size_info);
-        let check_fut = self.inner_check_update(tx, reader, size_info, hasher);
+        let check_fut = self.inner_check_update(tx, reader, size_info, digest.packed_hash, hasher);
 
         let (update_res, check_res) = tokio::join!(update_fut, check_fut);
 
@@ -181,9 +183,9 @@ impl MetricsComponent for VerifyStore {
             "If the verification store is verifying the size of the data",
         );
         c.publish(
-            "verify_hash_enabled",
-            &self.verify_hash,
-            "If the verification store is verifying the hash of the data",
+            "hash_verification_function",
+            &format!("{:?}", self.hash_verification_function),
+            "Hash verification function to verify the contents of the data",
         );
         c.publish(
             "size_verification_failures_total",
--- a/deployment-examples/docker-compose/local-storage-cas.json
+++ b/deployment-examples/docker-compose/local-storage-cas.json
@@ -25,7 +25,7 @@
           }
         },
         "verify_size": true,
-        "verify_hash": true
+        "hash_verification_function": "sha256"
       }
     },
     "AC_MAIN_STORE": {
--- a/deployment-examples/terraform/experimental_AWS/scripts/cas.json
+++ b/deployment-examples/terraform/experimental_AWS/scripts/cas.json
@@ -108,7 +108,7 @@
           }
         },
         "verify_size": true,
-        "verify_hash": true
+        "hash_verification_function": "sha256"
       }
     }
   },
--- a/nativelink-config/README.md
+++ b/nativelink-config/README.md
@@ -291,9 +291,10 @@ the rest will be stored in AWS's S3:
 
 This store is special. It's only job is to verify the content as it is fetched
 and uploaded to ensure it meets some criteria or errors. This store should only
-be added to the CAS. If `verify_hash` is set to true, it will apply a sha256
-algorithm on the data as it is sent/received and at the end if it does not match
-the name of the digest it will cancel the upload/download and return an error.
+be added to the CAS. If `hash_verification_function` is set, it will apply the
+hashing algorithm on the data as it is sent/received and at the end if it does
+not match the name of the digest it will cancel the upload/download and return
+an error. If it's not set, the hashing verification will be disabled.
 If `verify_size` is set, a similar item will happen, but count the bytes sent
 and check it against the digest instead.
 
@@ -311,7 +312,8 @@ and check it against the digest instead.
           }
         },
         "verify_size": true,
-        "verify_hash": true,
+        // sha256 or blake3
+        "hash_verification_function": "sha256",
       }
     },
     "AC_MAIN_STORE": {
--- a/nativelink-config/examples/filesystem_cas.json
+++ b/nativelink-config/examples/filesystem_cas.json
@@ -76,7 +76,7 @@
           }
         },
         "verify_size": true,
-        "verify_hash": true
+        "hash_verification_function": "sha256"
       }
     },
     "AC_MAIN_STORE": {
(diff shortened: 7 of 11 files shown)


steed924:
Hope you to review this. cc: @allada, @MarcusSorealheis, @aaronmondal

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmHHvmk0z-NriHBUVTD:-NmHHvmk0z-NriHBUVTE:blhimt) (no related file):*
Sadly I don't think we should do it this way. The original request knows what hash function to use and here we just assume it's the default.

I'd rather do it the right way.

I'm not sure what the best way to do this is. On first glance it feels like adding the hash enum type to `DigestInfo` seems like the right way, but we frequently cast `DigestInfo` into the proto's `Digest` which looses this information.

Another possible option is to change the store api so it passes the digest function along with the requests. This will require a lot of pluming though and should be done in multiple PRs if this is the right way.

As for right now I'm not sure what the best option is.

___

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmHHvmk0z-NriHBUVTD:-NmHUO_UEgO4Evvx2srX:bp14ta2) (no related file):*

Sadly I don't think we should do it this way. The original request knows what hash function to use and here we just assume it's the default.

I'd rather do it the right way.

I'm not sure what the best way to do this is. On first glance it feels like adding the hash enum type to `DigestInfo` seems like the right way, but we frequently cast `DigestInfo` into the proto's `Digest` which looses this information.

Another possible option is to change the store api so it passes the digest function along with the requests. This will require a lot of pluming though and should be done in multiple PRs if this is the right way.

As for right now I'm not sure what the best option is.
</blockquote></details>

I thought about this a bit more and did some rough tests to see how difficult it would be for a refactor in this area, and believe that adding a config for what hash function to use on the config for this store is the right way for now.

So if we change the config to support a hash function I think we can proceed.

___

steed924 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmHHvmk0z-NriHBUVTD:-NmHh3eND6TUACu8LitS:b-it4a14) (no related file):*

I thought about this a bit more and did some rough tests to see how difficult it would be for a refactor in this area, and believe that adding a config for what hash function to use on the config for this store is the right way for now.

So if we change the config to support a hash function I think we can proceed.
</blockquote></details>

Thanks for your feedback, @allada . Let me do work in that way. 

___

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmHsGT9BWfmvM3Dnctm:-NmHsGT9BWfmvM3Dnctn:b-1d0yws) (no related file):*
Also needs tests.

___

allada (changes requested):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

___
*[`nativelink-config/src/stores.rs` line 17 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmaqLqT1mkvv1pb5OBi:-NmaqLqU0pRgnOjl3cDW:b-s7twh3) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L17)):*
> ```XML
> use serde::{Deserialize, Serialize};
> 
> use crate::cas_server::ConfigDigestHashFunction;
> ```

Lets move this into this crate. By doing this, it creates a circular reference, which is bad practice.
___
*[`nativelink-config/src/stores.rs` line 338 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmasdGc6eZ9oCHEoGIx:-NmasdGc6eZ9oCHEoGIy:b-kidtiu) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L338)):*
> ```XML
>     /// This should be set to false for AC, but true for CAS stores.
>     #[serde(default)]
>     pub verify_hash: bool,
> ```

nit: Lets rename this:
`hash_verification_function` and make it an `Option<ConfigDigestHashFunction>`

Also add a comment above saying that `None` means it won't verify the hash.
___
*[`nativelink-store/src/verify_store.rs` line 60 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmarIRx-QbCWqzz6MJs:-NmarIRx-QbCWqzz6MJt:b-n474mq) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-store/src/verify_store.rs#L60)):*
> ```XML
>         mut rx: DropCloserReadHalf,
>         size_info: UploadSizeInfo,
>         mut maybe_hasher: Option<([u8; 32], DigestHasher)>,
> ```

nit: Lets split this up. Les pass 
___
*[`nativelink-store/src/verify_store.rs` line 83 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#-Nmar1_i2E4d8mCyFBRi:-Nmar1_i2E4d8mCyFBRj:bq61y34) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-store/src/verify_store.rs#L83)):*
> ```XML
>                 }
>                 if let Some((original_hash, hasher)) = maybe_hasher.as_mut() {
>                     let hash_result: [u8; 32] = hasher.finalize_digest(i64::try_from(sum_size)?).packed_hash;
> ```

nit: We can't trust this size. But we also don't need it. So instead, lets pass `-1` here and make a comment that this digest is only validating the hash.

nativelink-config/src/stores.rs
```diff
@@ -335,6 +336,11 @@ pub struct VerifyStore {
     /// This should be set to false for AC, but true for CAS stores.
     #[serde(default)]
     pub verify_hash: bool,
+
+    /// Digest hash function to use for hashing contents in the verify store
+    ///
+    /// Default: ConfigDigestHashFunction::sha256
```
MarcusSorealheis: maybe specify what another option might be here in the comments. `Blake3`

nativelink-config/src/cas_server.rs
```diff
@@ -544,7 +544,7 @@ pub enum WorkerConfig {
 }
 
 #[allow(non_camel_case_types)]
-#[derive(Deserialize, Debug, Clone, Copy)]
+#[derive(Serialize, Deserialize, Debug, Clone, Copy)]
 pub enum ConfigDigestHashFunction {
```
MarcusSorealheis: oh, just below the changed line, we have the enum we need. to build upon to support sha256 or Blake3.

nativelink-store/src/verify_store.rs
```diff
@@ -137,7 +140,12 @@ impl Store for VerifyStore {
 
         let mut hasher = None;
         if self.verify_hash {
-            hasher = Some((digest.packed_hash, Sha256::new()));
+            hasher = Some((
+                digest.packed_hash,
+                DigestHasher::from(DigestHasherFunc::from(
+                    self.digest_hash_function.unwrap_or(ConfigDigestHashFunction::sha256),
```
MarcusSorealheis: why is this here? Im confused.

nativelink-store/src/verify_store.rs
```diff
@@ -76,9 +79,9 @@ impl VerifyStore {
                         ));
                     }
                 }
-                if let Some((original_hash, hasher)) = maybe_hasher {
-                    let hash_result: [u8; 32] = hasher.finalize().into();
-                    if original_hash != hash_result {
+                if let Some((original_hash, hasher)) = maybe_hasher.as_mut() {
```
MarcusSorealheis: I think the `if let` conditional is fine, but computing the hash should live elsewhere. That way, here you can still have the comparator on current line 84 (`if *original_hash != hash_result`), with which type of hash being used is outside of the block. It should work for either hashing algorithm.

Right now, `hash_result` depends on Blake3 and that would be bad.

MarcusSorealheis (changes requested):
I'll defer to  Blaise on this one, as I think his comments point in the right direction and you have done some important work to get started. At a high level:

Tests, as Blaise said
Config, as Blaise said. You already have a lot in place for that direction.
The only thing I would think about on the code is making the implementation in `verify_store.rs` less coupled to one hashing algo or another. It should be straightforward but happy to help (or ask someone else) if needed.

Nice work and thank you for the contribution.

steed924 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmHsGT9BWfmvM3Dnctm:-Nmex6n94ZBd4_KuGXiw:b-896fix) (no related file):*

Also needs tests.
</blockquote></details>

Done.

___
___
*[`nativelink-config/src/cas_server.rs` line 548 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436654169:-NmeyCDs3W6iAI5N0sNb:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/cas_server.rs#L548)):*
<details><summary><i>Previously, MarcusSorealheis (Marcus Eagan) wrote…</i></summary><blockquote>

oh, just below the changed line, we have the enum we need. to build upon to support sha256 or Blake3.
</blockquote></details>

Done.
___
*[`nativelink-config/src/stores.rs` line 17 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmaqLqT1mkvv1pb5OBi:-NmeyCsM4MRbEVKO9wUX:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L17)):*

Lets move this into this crate. By doing this, it creates a circular reference, which is bad practice.
</blockquote></details>

Done.
___
*[`nativelink-config/src/stores.rs` line 342 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436640937:-NmeyDph0B3VwTtLwOFz:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L342)):*
<details><summary><i>Previously, MarcusSorealheis (Marcus Eagan) wrote…</i></summary><blockquote>

maybe specify what another option might be here in the comments. `Blake3`
</blockquote></details>

Done.
___
*[`nativelink-store/src/verify_store.rs` line 82 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436690298:-NmeyH7PEtISj2QlrpFu:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-store/src/verify_store.rs#L82)):*
<details><summary><i>Previously, MarcusSorealheis (Marcus Eagan) wrote…</i></summary><blockquote>

I think the `if let` conditional` is fine, but computing the hash should live elsewhere. That way, here you can still have the comparator on current line 84 (`if *original_hash != hash_result`), with which type of hash being used is outside of the block. It should work for either hashing algorithm.

Right now, `hash_result` depends on Blake3 and that would be bad. 
</blockquote></details>

Done.
___
*[`nativelink-store/src/verify_store.rs` line 146 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436688467:-NmeyI3Y-GkpBobxwxoC:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-store/src/verify_store.rs#L146)):*
<details><summary><i>Previously, MarcusSorealheis (Marcus Eagan) wrote…</i></summary><blockquote>

why is this here? Im confused.
</blockquote></details>

Done.

allada (commented):
> Reviewed 8 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks

___
*[`nativelink-config/src/stores.rs` line 342 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436640937:-Nmh6nh-9JEtZNmzxucL:b-ic32gw) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L342)):*
<details><summary><i>Previously, steed924 (Steed) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I don't think we should encourage `blake3`. Sha256 is what the default is for most implementations. Using blake3 requires the client to be setup for blake3 (which none to my knowledge do by default. They all use sha256 by default).

I'd rather just say "Like sha256" and not even mention blake3, since it requires a lot to go right.
___
*[`nativelink-store/src/verify_store.rs` line 141 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmhATV346006MOx54R3:-NmhATV346006MOx54R4:b-k7vhl0) ([raw file](https://github.com/TraceMachina/nativelink/blob/c6cd61642d394a019c701f4e38aa40e5e0ef625a/nativelink-store/src/verify_store.rs#L141)):*
> ```XML
>         }
> 
>         let mut hasher = None;
> ```

nit: use (you can inline this too after it's done):
```rs
let hasher = self.hash_verification_function.map(|v| DigestHasher::from(DigestHasherFunc::from(v)));
```
___
*[`nativelink-store/src/verify_store.rs` line 186 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/575#-NmhB9ED-rAuqA2fgYgX:-NmhB9ED-rAuqA2fgYgY:b-i1l87o) ([raw file](https://github.com/TraceMachina/nativelink/blob/c6cd61642d394a019c701f4e38aa40e5e0ef625a/nativelink-store/src/verify_store.rs#L186)):*
> ```XML
>             "If the verification store is verifying the size of the data",
>         );
>         if let Some(hash_verification_function) = self.hash_verification_function {
> ```

nit: Just use:
```rs
c.publish("hash_verification_function", format!("{:?}", self.hash_verification_function));
```

steed924 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

___
*[`nativelink-config/src/stores.rs` line 342 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/575#gh-1436640937:-Nmh_B5r0xXVvoS3cJ-W:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/e669c0aec0104d1ba3c3745b9c8af22fc0d62e0f/nativelink-config/src/stores.rs#L342)):*

I don't think we should encourage `blake3`. Sha256 is what the default is for most implementations. Using blake3 requires the client to be setup for blake3 (which none to my knowledge do by default. They all use sha256 by default).

I'd rather just say "Like sha256" and not even mention blake3, since it requires a lot to go right.
</blockquote></details>

Done.

allada (approved):
:lgtm:

> Reviewed 2 of 11 files at r3, 2 of 2 files at r4, all commit messages.
> Dismissed @MarcusSorealheis from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-NmhtEy4EYCVKDoOo8WM:bnfp4nl)* status: 1 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, pre-commit-checks

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-NnARkhc4oNzTITaxcLP:bnfp4nl)* status: 2 of 1 LGTMs obtained, and pending CI: pre-commit-checks

MarcusSorealheis (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575)* status: 2 of 1 LGTMs obtained, and pending CI: pre-commit-checks

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-NnMoWhHALtAxiFKt3r5:bnfp4nl)* status: 3 of 1 LGTMs obtained, and pending CI: pre-commit-checks

allada (approved):
@steed924 , can you rebase and we'll get this landed.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-Nnf1fFA4cWiD3mmZiZI:btfdaee)* status: 3 of 1 LGTMs obtained, and pending CI: pre-commit-checks

steed924:
> @steed924 , can you rebase and we'll get this landed.
> 
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-Nnf1fFA4cWiD3mmZiZI:btfdaee)_ status: 3 of 1 LGTMs obtained, and pending CI: pre-commit-checks

Just done.

aaronmondal (approved):
FYI @steed924 The failing CI job are the pre-commit hooks. I think it's trailing whitespace in the README.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-NnikLYy94HoDawf6adk:b-vrx0zk)* status: 4 of 1 LGTMs obtained, and pending CI: pre-commit-checks

steed924:
> FYI @steed924 The failing CI job are the pre-commit hooks. I think it's trailing whitespace in the README. :lgtm:
> 
> > Reviewed all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/575#-:-NnikLYy94HoDawf6adk:b-vrx0zk)_ status: 4 of 1 LGTMs obtained, and pending CI: pre-commit-checks

Thanks, @aaronmondal . Just fixed that issue. ;)

PR #583, opened 2023-12-29
Build nativelink with musl

aaronmondal:
Reduce container image sizes by over 50% to about 28 megabytes.

The nativelink executable is now fully self-contained and portable across linux systems regardless of the host's libc.

This change introduces new Rust toolchains on the nix side that mirror the Bazel toolchains. The new toolchains enable `cargo clippy` invocations without Bazel and allow invoking the same nightly `rustfmt` that is used in the Bazel build via pre-commit hooks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583)

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,3 +1,4 @@
+ignore = ["nativelink-proto/genproto/lib.rs"]
 max_width = 120
 reorder_imports = true
 imports_granularity = "Module"
--- a/flake.lock
+++ b/flake.lock
@@ -58,11 +58,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1685518550,
-        "narHash": "sha256-o2d0KcvaXzTrPRIo0kOLV0/QXHhDQ5DTi+OxcjO8xqY=",
+        "lastModified": 1701680307,
+        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "a1720a10a6cfe8234c0e93907ffe81be440f4cef",
+        "rev": "4022d587cbbfd70fe950c1e2083a02621806a725",
         "type": "github"
       },
       "original": {
@@ -145,7 +145,9 @@
     "pre-commit-hooks": {
       "inputs": {
         "flake-compat": "flake-compat",
-        "flake-utils": "flake-utils",
+        "flake-utils": [
+          "flake-utils"
+        ],
         "gitignore": "gitignore",
         "nixpkgs": [
           "nixpkgs"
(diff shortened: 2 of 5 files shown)


MarcusSorealheis:
@blakehatch can you check this on the Mac please?

MarcusSorealheis:
I will also, later today sometime

aaronmondal (commented):
@MarcusSorealheis This executable won't work on mac. The statically linked musl still calls into the Linux kernel which Mac doesn't have.

However, this should make it a lot easier to get the same thing working for mac in a future commit since musl does support mac. We'll need a separate build job for that but it should be very close to this linux build.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583#-:-Nmx9-u5BbQML4TU8D9D:be2czr4)* status: 0 of 1 LGTMs obtained

MarcusSorealheis:
I don't think we can change this in that case just yet. Or, if we do, we'd need to do a lot to offer a path that does work for users building on a Mac.  Many of our users are on Mac OS, certainly from a building from source lens.

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583)* status: 0 of 1 LGTMs obtained

___
*[`flake.nix` line 54 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/583#-Nmys6ww0Ccyp8etr-xe:-Nmys6ww0Ccyp8etr-xf:bvlcfqz) ([raw file](https://github.com/TraceMachina/nativelink/blob/beb87cf91b50c3574b75819e44beb6aa3d96da42/flake.nix#L54)):*
> ```Nix
>         };
> 
>         craneLib = crane.lib.${system};
> ```

Ah my bad. I thought we only supported the native Cargo build on mac. But we actually did have a working Mac build here that was just not tracked in CI i think. So I definitely should add mac support in this change as well.

aaronmondal (commented):
I've added Mac support back. This commit should now be a no-op change for MacOS.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583#-:-Nmz2olP-3vP8_dnxdZV:bx0a6vu)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

adam-singer (approved):
:lgtm:

Tested on MacOS

```
nix develop
nix build
cargo test --all
```

> Reviewed 3 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583#-:-Nn02j0J8_GEg4GD2g7z:bxgcpjq)* status: 1 of 1 LGTMs obtained

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/583#-:-Nn12m2RBuC3A5XpwpyT:bnfp4nl)* status: 2 of 1 LGTMs obtained

PR #589, opened 2024-01-04
Span tracing

adam-singer:
# Description

Adding feature to enable/disable builds of nativelink to include trace spans. 

Trace spans can be use for instrumenting historical call paths of the tokio workers. Implementation provided shunts enabling/disabling at the cfg/features flags level. Enabling this feature would have performance runtime impact and should be used in debugging/introspection scenarios. Applying the instrumentation updates the source tree and not expected to be checked in (atm). The span format uses chrome tracing json and can be viewed in `chrome://tracing` (or https://ui.perfetto.dev/). Trace snapshots are collected upon sever exit.

Output traces upon server exit are dumped in the cwd of the nativelink bin in the format `trace-*.json`

## Example

```bash
# Apply instrumentation
./tools/patch_tracing.sh fix
# Build/Run
RUST_LOG=INFO cargo --config 'build.rustflags=["--cfg", "tokio_unstable", "--cfg", "enable_tracing"]' run --features enable_tokio_console,enable_tracing -- nativelink-config/examples/basic_cas.json
# Unapply instrumentation
./tools/patch_tracing.sh strip
```

Screenshot 2024-01-03 at 6 24 14 PM

## Open Questions

Should trace snapshots be collected upon server exit, client requests or external signal to admin ports?  

Should `clippy-tracing` be a preinstalled expectation or configured/added in nix?

Should we checkin the tracing macro and create linting that auto lints (fix) the source tree when instrumentation is missing?

Fixes # (issue)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/589)

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "a6b16863ad5d0a1a72b20d57a7b103696e736e78f865de8cee087e4876a07970",
+  "checksum": "0537a3af4c115d803aa31489688679acfecd43a5d05b79d2e6cb5e8f1dc21951",
   "crates": {
     "addr2line 0.21.0": {
       "name": "addr2line",
@@ -8522,6 +8522,10 @@
             {
               "id": "shellexpand 3.1.0",
               "target": "shellexpand"
+            },
+            {
+              "id": "tracing 0.1.40",
+              "target": "tracing"
             }
           ],
           "selects": {}
@@ -8572,6 +8576,10 @@
             {
               "id": "tonic 0.10.2",
               "target": "tonic"
+            },
+            {
+              "id": "tracing 0.1.40",
+              "target": "tracing"
             }
           ],
           "selects": {}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1673,6 +1673,7 @@ dependencies = [
  "tonic",
  "tower",
  "tracing",
+ "tracing-chrome",
  "tracing-subscriber",
 ]
 
@@ -1682,6 +1683,7 @@ version = "0.2.0"
 dependencies = [
  "serde",
  "shellexpand",
+ "tracing",
 ]
 
 [[package]]
@@ -1694,6 +1696,7 @@ dependencies = [
  "prost-types",
  "tokio",
  "tonic",
+ "tracing",
 ]
 
 [[package]]
@@ -3072,6 +3075,17 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.46",
+]
+
+[[package]]
+name = "tracing-chrome"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "496b3cd5447f7ff527bbbf19b071ad542a000adf297d4127078b4dfdb931f41a"
+dependencies = [
+ "serde_json",
+ "tracing-core",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ name = "nativelink"
 
 [features]
 enable_tokio_console = []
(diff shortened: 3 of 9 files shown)


aaronmondal (commented):
I like this a lot.

> Should trace snapshots be collected upon server exit, client requests or external signal to admin ports?

Not sure about this one. NVIDIAs profiling tooling was pretty nice (https://docs.nvidia.com/cuda/profiler-users-guide/index.html#flush-profile-data). Maybe we could take some inspiration from it.

> Should `clippy-tracing` be a preinstalled expectation or configured/added in nix?

We should certainly add it to the flake. Crane (the cargo wrapper we use) probably supports various cargo extensions already (https://github.com/ipetkov/crane?tab=readme-ov-file#features). We might be able to add clippy-tracing there somehow.

> Should we checkin the tracing macro and create linting that auto lints (fix) the source tree when instrumentation is missing?

Yes that seems like the way to go. This would also give us a good starting point to remove regular `tracing::info` calls etc over time in favor of the `instrument` approach.

I'm a bit confused by the `enable_tracing` feature. There is the `tracing` library and the `tracing` log level. Libraries like `tokio` use a feature called `tracing` to turn the dependency on the `tracing` crate on/off. AFAIU we always want to have the library enabled though as it's also the thing that handles the `critical`, `error`, `info`, `warn`, `debug` and `tracing` log levels.

Wouldn't the runtime performance impact only be noticeable when the `tracing` or `debug` log levels are enabled? But that wouldn't have to be a feature and instead would just be a config option like `RUST_LOG=tracing`. Or maybe I'm misunderstanding some inner workings or the `tracing` crate?

If the `chrome_tracing` layer is just a subscriber I also wonder whether it would have any noticeable runtime impact. If it did, would it make sense to have a `chrome_tracing` feature instead?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/589#-:-NnZdz2I77TBOylKtD7a:bgzreyg)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, ubuntu-20.04, ubuntu-22.04

adam-singer (commented):
> I'm a bit confused by the `enable_tracing` feature.

Didn't want to conflate other features with span tracing for nativelink, https://gist.github.com/adam-singer/fe59163849cc232a7ba47d0befe48b0d seems reasonable enough that we could use `tracing` instead of a specific feature/label for nativelink. 

> Wouldn't the runtime performance impact only be noticeable when the `tracing` or `debug` log levels are enabled?

Was thinking more of the layer and guard since it can be defined before enabling tokio console. The annotations shouldn't cause any performance impact, we could move to a `chrome_tracing` feature/label.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/589#-:-NneV5E14BsJyLCeYUXw:bxmzo3j)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, ubuntu-20.04, ubuntu-22.04

palfrey:
This was an interesting idea, but we've since [dropped console-subscriber](https://github.com/TraceMachina/nativelink/pull/1683) which this needs, and [the upstream project still has a bunch of issues with old dependencies](https://github.com/tokio-rs/console/issues) so I'm not entirely comfortable bringing that back in. We can re-add if we can find another way around this.

PR #601, opened 2024-01-10
EvictingMap should evict keys on all public access

adam-singer:
# Description

Implementation change for `EvictingMap` where all public methods should attempt at evicting keys that have expired. 

Refactor `size_for_key` to call `sizes_for_keys`

Fixes # (issue)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)

### DIFF
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -20,7 +20,7 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use async_lock::Mutex;
 use async_trait::async_trait;
 use futures::stream::FuturesUnordered;
-use futures::{future, StreamExt};
+use futures::{future, join, FutureExt, StreamExt};
 use lru::LruCache;
 use nativelink_config::stores::EvictionPolicy;
 use serde::{Deserialize, Serialize};
@@ -172,7 +172,9 @@ where
     }
 
     pub async fn build_lru_index(&self) -> SerializedLRU {
-        let state = self.state.lock().await;
+        let mut state = self.state.lock().await;
+        self.evict_items(state.deref_mut()).await;
+
         let mut serialized_lru = SerializedLRU {
             data: Vec::with_capacity(state.lru.len()),
             anchor_time: self.anchor_time.unix_timestamp(),
@@ -247,43 +249,68 @@ where
         }
     }
 
+    /// Return the size of a `DigestInfo`, if not found `None` is returned.
     pub async fn size_for_key(&self, digest: &DigestInfo) -> Option<usize> {
-        let mut state = self.state.lock().await;
-        let entry = state.lru.get_mut(digest)?;
-        entry.seconds_since_anchor = self.anchor_time.elapsed().as_secs() as i32;
-        let data = entry.data.clone();
-        drop(state);
-        data.touch().await;
-        Some(data.len())
+        let mut results = [None];
+        self.sizes_for_keys(&[*digest], &mut results[..]).await;
+        results[0]
     }
 
+    /// Return the sizes of a collection of `DigestInfo`. Expects `results` collection
+    /// to be provided for storing the resulting `DigestInfo` size. Each index value in
+    /// `digests` maps directly to the size value of the `DigestInfo` in `results`.
+    /// If no digest is found in the internal map, `None` is filled in its place.
     pub async fn sizes_for_keys(&self, digests: &[DigestInfo], results: &mut [Option<usize>]) {
         let mut state = self.state.lock().await;
-        let seconds_since_anchor = self.anchor_time.elapsed().as_secs() as i32;
+        let mut remove_digests: Vec<&DigestInfo> = Vec::new();
+
         let to_touch: Vec<T> = digests
             .iter()
             .zip(results.iter_mut())
-            .filter_map(|(digest, result)| {
-                let entry = state.lru.get_mut(digest)?;
-                entry.seconds_since_anchor = seconds_since_anchor;
-                let data = entry.data.clone();
-                *result = Some(data.len());
-                Some(data)
+            .flat_map(|(digest, result)| {
+                let lru_len = state.lru.len();
+                let sum_store_size = state.sum_store_size;
+                // Determine if a digest should be evicted or data should be touched.
+                // Digests to be eviected are collected in separate vector and chained
+                // in a single future.
+                if let Some(entry) = state.lru.get(digest) {
+                    if self.should_evict(lru_len, entry, sum_store_size, self.max_bytes) {
+                        // Digest should be evicted.
+                        remove_digests.push(digest);
+                        None
+                    } else {
+                        // Extract data entry to be touched and slot length into results.
+                        let data = entry.data.clone();
+                        *result = Some(data.len());
+                        Some(data)
+                    }
+                } else {
+                    // Digest will be evicted if not in lru map, this is a pedantic case.
+                    remove_digests.push(digest);
+                    None
+                }
             })
             .collect();
-        drop(state);
-        to_touch
-            .iter()
-            .map(|data| data.touch())
-            .collect::<FuturesUnordered<_>>()
-            .for_each(|_| future::ready(()))
-            .await;
+
+        join!(
+            to_touch
+                .iter()
+                .map(|data| data.touch().map(|_| ()))
+                .collect::<FuturesUnordered<_>>()
+                .for_each(|_| future::ready(())),
+            async move {
+                for digest in remove_digests {
+                    self.inner_remove(state.deref_mut(), digest).await;
+                }
+            }
+        );
     }
 
     pub async fn get(&self, digest: &DigestInfo) -> Option<T> {
         let mut state = self.state.lock().await;
+        self.evict_items(state.deref_mut()).await;
+
         if let Some(entry) = state.lru.get_mut(digest) {
-            entry.seconds_since_anchor = self.anchor_time.elapsed().as_secs() as i32;
             let data = entry.data.clone();
             drop(state);
             data.touch().await;
@@ -354,7 +381,8 @@ where
         self.inner_remove(&mut state, digest).await
     }
 
-    async fn inner_remove(&self, state: &mut State<T>, digest: &DigestInfo) -> bool {
+    async fn inner_remove(&self, mut state: &mut State<T>, digest: &DigestInfo) -> bool {
+        self.evict_items(state.deref_mut()).await;
         if let Some(entry) = state.lru.pop(digest) {
             let data_len = entry.data.len() as u64;
             state.sum_store_size -= data_len;
--- a/nativelink-util/tests/evicting_map_test.rs
+++ b/nativelink-util/tests/evicting_map_test.rs
@@ -300,8 +300,8 @@ mod evicting_map_tests {
 
         assert_eq!(
             evicting_map.size_for_key(&DigestInfo::try_new(HASH1, 0)?).await,
-            Some(DATA.len()),
-            "Expected map to have item 1"
+            None,
+            "Expected map to not have item 1"
         );
         assert_eq!(
             evicting_map.size_for_key(&DigestInfo::try_new(HASH2, 0)?).await,
@@ -415,8 +415,8 @@ mod evicting_map_tests {
 
         assert_eq!(
             evicting_map.size_for_key(&DigestInfo::try_new(HASH1, 0)?).await,
-            Some(8),
-            "Expected map to have item 1"
+            None,
+            "Expected map to not have item 1"
         );
         assert_eq!(
             evicting_map.size_for_key(&DigestInfo::try_new(HASH2, 0)?).await,
@@ -538,6 +538,63 @@ mod evicting_map_tests {
             None,
             "HASH3/2 should be empty"
         );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn get_evicts_on_time() -> Result<(), Error> {
+        let evicting_map = EvictingMap::<BytesWrapper, MockInstantWrapped>::new(
+            &EvictionPolicy {
+                max_count: 0,
+                max_seconds: 5,
+                max_bytes: 0,
+                evict_bytes: 0,
+            },
+            MockInstantWrapped(MockInstant::now()),
+        );
+
+        const DATA: &str = "12345678";
+        let digest_info1: DigestInfo = DigestInfo::try_new(HASH1, 0)?;
+        evicting_map.insert(digest_info1, Bytes::from(DATA).into()).await;
+
+        // Getting from map before time has expired should return the value.
+        assert_eq!(evicting_map.get(&digest_info1).await, Some(Bytes::from(DATA).into()));
+
+        MockClock::advance(Duration::from_secs(10));
+
+        // Getting from map after time has expired should return None.
+        assert_eq!(evicting_map.get(&digest_info1).await, None);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn remove_evicts_on_time() -> Result<(), Error> {
+        let evicting_map = EvictingMap::<BytesWrapper, MockInstantWrapped>::new(
+            &EvictionPolicy {
+                max_count: 0,
+                max_seconds: 5,
+                max_bytes: 0,
+                evict_bytes: 0,
+            },
+            MockInstantWrapped(MockInstant::now()),
+        );
+
+        const DATA: &str = "12345678";
+        let digest_info1: DigestInfo = DigestInfo::try_new(HASH1, 0)?;
+        evicting_map.insert(digest_info1, Bytes::from(DATA).into()).await;
+
+        let digest_info2: DigestInfo = DigestInfo::try_new(HASH2, 0)?;
+        evicting_map.insert(digest_info2, Bytes::from(DATA).into()).await;
+
+        // Removing digest before time has expired should return true.
+        assert!(evicting_map.remove(&digest_info2).await);
+
+        MockClock::advance(Duration::from_secs(10));
+
+        // Removing digest after time has expired should return false.
+        assert!(!evicting_map.remove(&digest_info1).await);
 
         Ok(())
     }


allada (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), ubuntu-20.04, ubuntu-22.04

___
*[`nativelink-util/src/evicting_map.rs` line 251 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnoWoVD3xDddfrRRR6D:-NnoWoVD3xDddfrRRR6E:b-o8ay1a) ([raw file](https://github.com/TraceMachina/nativelink/blob/7ddcba57dda7998df7ed3c25c680147e8940fd22/nativelink-util/src/evicting_map.rs#L251)):*
> ```XML
> 
>     pub async fn size_for_key(&self, digest: &DigestInfo) -> Option<usize> {
>         let results = &mut vec![None];
> ```

nit, Doing it this way will reduce the need for an alloc:
```rs
let mut results = [None];
self.sizes_for_keys(&[*digest], &mut results[..]).await;
results[0]
```
___
*[`nativelink-util/src/evicting_map.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnoZHxb-7FK6OroPAzn:-NnoZHxb-7FK6OroPAzo:b-60zy9r) ([raw file](https://github.com/TraceMachina/nativelink/blob/7ddcba57dda7998df7ed3c25c680147e8940fd22/nativelink-util/src/evicting_map.rs#L266)):*
> ```XML
>             .filter_map(|(digest, result)| {
>                 let entry = state.lru.get_mut(digest)?;
>                 entry.seconds_since_anchor = seconds_since_anchor;
> ```

I don't think we can do it this way. I was looking over the code and it looks like we need to instead check to see if the item meets the criteria to be evicted before saying it exists.

We also need to remove all updates to `seconds_since_anchor` updating. 

The reason is because what can end up happening is the item keeps getting `.has()`'ed, but the underlying store it never gets touched and ends up being evicted. Then eventually someone comes along and says "do you exist", it responds "yes", then it goes to read it, forwards to the underlying store, but it was evicted.

What we want is to ensure that the `max_seconds` config is only checked on reading.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), ubuntu-20.04, ubuntu-22.04

___
*[`nativelink-util/src/evicting_map.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnoZHxb-7FK6OroPAzn:-NnpNZCG1i1addquD2iv:b-edkw06) ([raw file](https://github.com/TraceMachina/nativelink/blob/7ddcba57dda7998df7ed3c25c680147e8940fd22/nativelink-util/src/evicting_map.rs#L266)):*

I don't think we can do it this way. I was looking over the code and it looks like we need to instead check to see if the item meets the criteria to be evicted before saying it exists.

We also need to remove all updates to `seconds_since_anchor` updating. 

The reason is because what can end up happening is the item keeps getting `.has()`'ed, but the underlying store it never gets touched and ends up being evicted. Then eventually someone comes along and says "do you exist", it responds "yes", then it goes to read it, forwards to the underlying store, but it was evicted.

What we want is to ensure that the `max_seconds` config is only checked on reading.
</blockquote></details>

Are you saying that we shouldn't call `evict_items` and/or put an additional check when mapping over digests to check that `entry` should be filtered out but returning `None` if the entry is determined expired by `should_evict` ?

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Vercel, publish-image, ubuntu-20.04, ubuntu-20.04 / stable

___
*[`nativelink-util/src/evicting_map.rs` line 251 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnoWoVD3xDddfrRRR6D:-NnptiNVET3U6aE9yuMb:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/7ddcba57dda7998df7ed3c25c680147e8940fd22/nativelink-util/src/evicting_map.rs#L251)):*

nit, Doing it this way will reduce the need for an alloc:
```rs
let mut results = [None];
self.sizes_for_keys(&[*digest], &mut results[..]).await;
results[0]
```
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnoZHxb-7FK6OroPAzn:-Nnptm4v6_bl0xRW--fG:b51lqh2) ([raw file](https://github.com/TraceMachina/nativelink/blob/7ddcba57dda7998df7ed3c25c680147e8940fd22/nativelink-util/src/evicting_map.rs#L266)):*
@allada ptal, not sure if I fully understand what you mean by the last thing

> What we want is to ensure that the `max_seconds` config is only checked on reading.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 267 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu2_ZN9P_ROAWWdmxA:-Nnu2_ZN9P_ROAWWdmxB:bggvp5w) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L267)):*
> ```XML
>                 let lru_len = state.lru.len();
>                 let sum_store_size = state.sum_store_size;
>                 if let Some(entry) = state.lru.get(digest) {
> ```

nit:

```rs
let entry = state.lru.get(digest)?;
```
___
*[`nativelink-util/src/evicting_map.rs` line 269 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu24SaA6gme9CljpzG:-Nnu24SaA6gme9CljpzH:b-qotzgz) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L269)):*
> ```XML
>                 if let Some(entry) = state.lru.get(digest) {
>                     if self.should_evict(lru_len, entry, sum_store_size, self.max_bytes) {
>                         None
> ```

nit: We should clean up our memory here too, so I suggest removing the entry from the map using `self.inner_remove(digest);`

Or if we want to be even more safe, create a new function for `should_evict_from_max_time()` or something. I'm afraid of accounting being slightly off and us evicting a lot if there's a fence post.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 267 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu2_ZN9P_ROAWWdmxA:-Nnudv3n932A6Dd1a7vJ:b-m0w3bm) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L267)):*

nit:

```rs
let entry = state.lru.get(digest)?;
```
</blockquote></details>

Will do, initially I had that but then had to fight the borrow checker on `state.lru.len()` && `state.sum_store_size`, thought keeping it in a `if / let` helped with scoping before hoisting them up. 
___
*[`nativelink-util/src/evicting_map.rs` line 269 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu24SaA6gme9CljpzG:-NnueJ3H0grQBpyhQm9I:b-yqreo2) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L269)):*

nit: We should clean up our memory here too, so I suggest removing the entry from the map using `self.inner_remove(digest);`

Or if we want to be even more safe, create a new function for `should_evict_from_max_time()` or something. I'm afraid of accounting being slightly off and us evicting a lot if there's a fence post.
</blockquote></details>

Will look into `self.inner_remove(digest);`, also concerned a bit about having multiple points of accounting for eviction logic. Taking a second look to see if it is reasonable

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

___
*[`nativelink-util/src/evicting_map.rs` line 267 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu2_ZN9P_ROAWWdmxA:-Nnwa-c9CJD8TZIEP2zC:b-1tom03) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L267)):*

Will do, initially I had that but then had to fight the borrow checker on `state.lru.len()` && `state.sum_store_size`, thought keeping it in a `if / let` helped with scoping before hoisting them up.
</blockquote></details>

Changed around implementation where a tuple is returned, keeping the `if / let` unless there is a similar way to model the sequence 
___
*[`nativelink-util/src/evicting_map.rs` line 269 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu24SaA6gme9CljpzG:-NnwaJ8fBckSlIkxlWPh:bfva5gw) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L269)):*

Will look into `self.inner_remove(digest);`, also concerned a bit about having multiple points of accounting for eviction logic. Taking a second look to see if it is reasonable
</blockquote></details>

Unsure if this was a correct rust convention of dealing with async/collections, split the two data paths ("data entry to touch", "digest to delete"), each required a bit different requirement to lift into a `Future` context. I think this could have a single collections flow if the tuple was lifted into an enum or reworked as a single match statement where the async calls get collected, since its internal to method, trying this approach unless advised its some sort of squirrely rust.
___
*[`nativelink-util/src/evicting_map.rs` line 307 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnwbWEH-7Qv6uWvwUKN:-NnwbWEH-7Qv6uWvwUKO:b-lsvacf) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L307)):*
> ```XML
>             });
>         });
>         while (futures.next().await).is_some() {}
> ```

@allada in this case, how will errors be propagated out, in a safe way or panic/stop the world?

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/evicting_map.rs` line 269 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu24SaA6gme9CljpzG:-No-w6qy3q7E9JUSwclS:b-b31ato) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L269)):*

Unsure if this was a correct rust convention of dealing with async/collections, split the two data paths ("data entry to touch", "digest to delete"), each required a bit different requirement to lift into a `Future` context. I think this could have a single collections flow if the tuple was lifted into an enum or reworked as a single match statement where the async calls get collected, since its internal to method, trying this approach unless advised its some sort of squirrely rust.
</blockquote></details>

I'd use an enum instead to annotate if it's for touching or for removing, then read it out.
___
*[`nativelink-util/src/evicting_map.rs` line 302 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-No-xH0y4pprQhpfDqh2:-No-xH0y4pprQhpfDqh3:b885y09) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L302)):*
> ```XML
>         to_remove.iter().flatten().for_each(|digest| {
>             futures.push(async move {
>                 let mut state = self.state.lock().await;
> ```

nit: Just hold a lock around this entire block instead of doing a bunch of lock/unlocks.
___
*[`nativelink-util/src/evicting_map.rs` line 309 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-No-xC4E4C6gDWfMQHGz:-No-xC4E4C6gDWfMQHH-:b5tnooh) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L309)):*
> ```XML
>         while (futures.next().await).is_some() {}
> 
>         // Map/Collect calls to touch/atime update LenEntry.
> ```

nit: This map is agnostic to atime.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/evicting_map.rs` line 302 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-No-xH0y4pprQhpfDqh2:-NoIIJ0HDhZfcypNQ0bu:b-uusvr4) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L302)):*

nit: Just hold a lock around this entire block instead of doing a bunch of lock/unlocks.
</blockquote></details>

When I was doing that I kept on running into violations from the borrow checker. I'll ping you to see what I missed

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 267 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu2_ZN9P_ROAWWdmxA:-NoOX-6gArjHdOOFw4nB:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L267)):*

Changed around implementation where a tuple is returned, keeping the `if / let` unless there is a similar way to model the sequence
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 269 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/601#-Nnu24SaA6gme9CljpzG:-NoOX0Ko8bIIqZC_oF86:bpg8bxd) ([raw file](https://github.com/TraceMachina/nativelink/blob/61785cefd18d90d105aaeefd0eff22c33865ce48/nativelink-util/src/evicting_map.rs#L269)):*

I'd use an enum instead to annotate if it's for touching or for removing, then read it out.
</blockquote></details>

Done.

As we iterated in discussion around enum and for-loop, eventually went with for-loop and chaining into the to touch futures. 
 
___
*[`nativelink-util/src/evicting_map.rs` line 302 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-No-xH0y4pprQhpfDqh2:-NoOXKH48W6Ms6FzpWwT:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L302)):*

When I was doing that I kept on running into violations from the borrow checker. I'll ping you to see what I missed
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 309 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-No-xC4E4C6gDWfMQHGz:-NoOXNeI-8Xh7EExJsCr:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L309)):*

nit: This map is agnostic to atime.
</blockquote></details>

Done.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 307 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnwbWEH-7Qv6uWvwUKN:-NoOmpsFDV0AJtgXasDw:b-1ze8uc) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L307)):*

@allada in this case, how will errors be propagated out, in a safe way or panic/stop the world?
</blockquote></details>

It doesn't look like this will ever return an error, but could panic. We have lots of places around our code that could panic. In such case, I think we should just stop the "spawn"... I am pretty sure no state issues will result in panics here.
___
*[`nativelink-util/src/evicting_map.rs` line 265 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOl8hRA82EGFVAZqXL:-NoOl8hRA82EGFVAZqXM:b3bv2ke) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L265)):*
> ```XML
>     pub async fn sizes_for_keys(&self, digests: &[DigestInfo], results: &mut [Option<usize>]) {
>         let mut state = self.state.lock().await;
>         self.evict_items(state.deref_mut()).await;
> ```

nit: No need to do this here, I think it makes more sense to do most cleanup evictions on update/insert.
___
*[`nativelink-util/src/evicting_map.rs` line 272 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOlWWvEL5DPbZtKUAo:-NoOlWWvEL5DPbZtKUAp:b-u4ry9x) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L272)):*
> ```XML
>             .zip(results.iter_mut())
>             .flat_map(|(digest, result)| {
>                 let lru_len = state.lru.len();
> ```

nit: Move this down to where it's used (or inline it).
___
*[`nativelink-util/src/evicting_map.rs` line 273 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOldg0DOhanpbl69av:-NoOldg0DOhanpbl69aw:b-5v3gaq) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L273)):*
> ```XML
>             .flat_map(|(digest, result)| {
>                 let lru_len = state.lru.len();
>                 let sum_store_size = state.sum_store_size;
> ```

nit: ditto.
___
*[`nativelink-util/src/evicting_map.rs` line 296 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOmDEh-McgfT2b4bG-:-NoOmDEiEfuQbTJo8ycQ:b-wem82k) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L296)):*
> ```XML
>             .collect();
> 
>         let remove_stream = stream::iter(vec![
> ```

nit: This is a stream of one. I think this should just be a future instead of a stream. Then in the `to_touch` iteration below, something more like:

 ```rs
join!(to_touch
            .iter()
            .map(|data| data.touch().map(|_| ()))
            .collect::<FuturesUnordered<_>>()
            .chain(remove_stream)
            .for_each(|_| future::ready(())),
            async move {
                for digest in remove_digests {
                    self.inner_remove(state.deref_mut(), digest).await;
                }
            }

);
```

**Review by allada — COMMENTED** [review]
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

**Review by adam-singer — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 307 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NnwbWEH-7Qv6uWvwUKN:-NoOrrsL1AcuwAOfv_1M:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/eeed4fbe4446f082ab5b82fdf1e6beeab248929e/nativelink-util/src/evicting_map.rs#L307)):*

It doesn't look like this will ever return an error, but could panic. We have lots of places around our code that could panic. In such case, I think we should just stop the "spawn"... I am pretty sure no state issues will result in panics here.
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 265 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOl8hRA82EGFVAZqXL:-NoOrv0u6ZVjGBbFQaFK:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L265)):*

nit: No need to do this here, I think it makes more sense to do most cleanup evictions on update/insert.
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 272 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOlWWvEL5DPbZtKUAo:-NoOq5uV6NbP4QW2mb-E:b2vh9ml) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L272)):*

nit: Move this down to where it's used (or inline it).
</blockquote></details>

The borrow checker will complain with both instances being inlined

```
error[E0502]: cannot borrow `state` as immutable because it is also borrowed as mutable
   --> nativelink-util/src/evicting_map.rs:278:42
    |
277 |                 if let Some(entry) = state.lru.get(digest) {
    |                                      ----- mutable borrow occurs here
278 |                     if self.should_evict(state.lru.len(), entry, state.sum_store_size, self.max_bytes) {
    |                                          ^^^^^            ----- mutable borrow later used here
    |                                          |
    |                                          immutable borrow occurs here


   --> nativelink-util/src/evicting_map.rs:278:66
    |
277 |                 if let Some(entry) = state.lru.get(digest) {
    |                                      ----- mutable borrow occurs here
278 |                     if self.should_evict(state.lru.len(), entry, state.sum_store_size, self.max_bytes) {
    |                                                                  ^^^^^ immutable borrow occurs here
...
284 |                         let data = entry.data.clone();
    |                                    ---------- mutable borrow later used here


```
___
*[`nativelink-util/src/evicting_map.rs` line 296 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOmDEh-McgfT2b4bG-:-NoOrosS5e4MnwyJgaq0:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L296)):*

nit: This is a stream of one. I think this should just be a future instead of a stream. Then in the `to_touch` iteration below, something more like:

 ```rs
join!(to_touch
            .iter()
            .map(|data| data.touch().map(|_| ()))
            .collect::<FuturesUnordered<_>>()
            .chain(remove_stream)
            .for_each(|_| future::ready(())),
            async move {
                for digest in remove_digests {
                    self.inner_remove(state.deref_mut(), digest).await;
                }
            }

);
```
</blockquote></details>

Done.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/evicting_map.rs` line 272 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOlWWvEL5DPbZtKUAo:-NoOw43kBTblM9U4wFot:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L272)):*

The borrow checker will complain with both instances being inlined

```
error[E0502]: cannot borrow `state` as immutable because it is also borrowed as mutable
   --> nativelink-util/src/evicting_map.rs:278:42
    |
277 |                 if let Some(entry) = state.lru.get(digest) {
    |                                      ----- mutable borrow occurs here
278 |                     if self.should_evict(state.lru.len(), entry, state.sum_store_size, self.max_bytes) {
    |                                          ^^^^^            ----- mutable borrow later used here
    |                                          |
    |                                          immutable borrow occurs here


   --> nativelink-util/src/evicting_map.rs:278:66
    |
277 |                 if let Some(entry) = state.lru.get(digest) {
    |                                      ----- mutable borrow occurs here
278 |                     if self.should_evict(state.lru.len(), entry, state.sum_store_size, self.max_bytes) {
    |                                                                  ^^^^^ immutable borrow occurs here
...
284 |                         let data = entry.data.clone();
    |                                    ---------- mutable borrow later used here


```
</blockquote></details>

Done.
___
*[`nativelink-util/src/evicting_map.rs` line 273 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOldg0DOhanpbl69av:-NoOw54e-aECa8sHT03U:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L273)):*

nit: ditto.
</blockquote></details>

Done.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/evicting_map.rs` line 272 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOlWWvEL5DPbZtKUAo:-NoOwVVY8tT0Rj3CpsZN:bmma279) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L272)):*

Done.
</blockquote></details>

Hmmm, I see. This probably could be resolved using `state.as_ref().lru.get(digest)` (same with `.len()`). (or `.deref()`)

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601#-:-NoOyZ5x1tCWdz5lJgfe:bnfp4nl)* status: 1 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601)* status: 1 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

___
*[`nativelink-util/src/evicting_map.rs` line 272 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/601#-NoOlWWvEL5DPbZtKUAo:-NoOzqHs4Y6u1g7H7LpT:b2opqge) ([raw file](https://github.com/TraceMachina/nativelink/blob/d8c37e2bc7fb7d785d39c6dba67e6105bd1eef53/nativelink-util/src/evicting_map.rs#L272)):*

Hmmm, I see. This probably could be resolved using `state.as_ref().lru.get(digest)` (same with `.len()`). (or `.deref()`)
</blockquote></details>

`as_ref()`:

```
error[E0599]: no method named `as_ref` found for struct `async_lock::MutexGuard<'_, evicting_map::State<T>>` in the current scope
   --> nativelink-util/src/evicting_map.rs:276:44
    |
276 |                 if let Some(entry) = state.as_ref().lru.get(digest) {
    |                                            ^^^^^^ method not found in `MutexGuard<'_, State<T>>`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `as_ref`, perhaps you need to implement it:
            candidate #1: `AsRef`

```

`deref()`:

```
error[E0596]: cannot borrow data in a `&` reference as mutable
   --> nativelink-util/src/evicting_map.rs:277:38
    |
277 |                 if let Some(entry) = state.deref().lru.get(digest) {
    |                                      ^^^^^^^^^^^^^^^^^ cannot borrow as mutable
```

adam-singer (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/601#-:-NoP-L3ZCMaSe9KFJVTN:bnfp4nl)* status: 2 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

PR #605, opened 2024-01-12
Fix a potential bug in DropCloserReadHalf::take()

### LINKED ISSUE #578: Potential bugs in DropCloserReadHalf::take()
While looking at some other issues I noticed that the `take()` function here:
https://github.com/TraceMachina/nativelink/blob/3578d50fa78387670b7d3761396e4c26b7ee8814/nativelink-util/src/buf_channel.rs#L304

is a bit confusing and when digging deep into it, I believe it was sometimes hiding possible EOFs. I rewrote it to try and better understand it. Right now the only places that use this function are CompressionStore and S3Store.

Here's how I rewrote it:
```rs

    /// Takes exactly `size` number of bytes from the stream and returns them.
    /// This means the stream will keep polling until either an EOF is received or
    /// `size` bytes are received and concat them all together then return them.
    /// This method is optimized to reduce copies when possible.
    pub async fn take(&mut self, size: usize) -> Result<Bytes, Error> {
        fn populate_partial_if_needed(
            current_size: usize,
            desired_size: usize,
            chunk: &mut Bytes,
            partial: &mut Option<Result<Bytes, Error>>,
        ) {
            if current_size + chunk.len() <= desired_size {
                return;
            }
            assert!(partial.is_none(), "Partial should have been consumed during the recv()");
            let local_partial = chunk.split_off(desired_size - current_size);
            *partial = if local_partial.is_empty() {
                None
            } else {
                Some(Ok(local_partial))
            };
        }

        let first_chunk = {
            // This is an optimization for a relatively common case when the first chunk in the
            // stream satisfies all the requirements to fill this `take()`.
            // This will prevent us from needing to copy the data into a new buffer and instead
            // we can just forward on the original Bytes object. If we need more than the first
            // chunk we will then go the slow path and actually copy our data.

            // 1. Read some data from our stream (or self.partial).
            let mut first_chunk = self.recv().await.err_tip(|| "During first buf_channel::take()")?;
            assert!(self.partial.is_none(), "Partial should have been consumed during the recv()");
            // 2. Split our data so `first_chunk` is <= `size` and puts any remaining
            //    in `self.partial` (or set it to None).
            populate_partial_if_needed(0, size, &mut first_chunk, &mut self.partial);
            // 3a. If our `first_chunk` is EOF, we are done.
            // 3b. If our `first_chunk` is exactly `size` we can return our current state.
            if first_chunk.is_empty() || first_chunk.len() >= size {
                assert!(first_chunk.len() < size, "Length should be exactly size here");
                return Ok(first_chunk);
            }
            assert!(
                self.partial.is_none(),
                "If `first_chunk`'s len is < size and not EOF, partial should be None"
            );
            first_chunk
        };
        let mut output = BytesMut::with_capacity(size);
        output.put(first_chunk);

        loop {
            let mut chunk = self.recv().await.err_tip(|| "During buf_channel::take()")?;
            assert!(self.partial.is_none(), "Partial should have been consumed during the recv()");
            if chunk.is_empty() {
                // Forward EOF to next recv() and return our current buffer.
                self.partial = Some(Ok(chunk));
                return Ok(output.freeze());
            }

            populate_partial_if_needed(output.len(), size, &mut chunk, &mut self.partial);

            output.put(chunk);

            if output.len() == size {
                return Ok(output.freeze());
            }
            assert!(output.len() < size, "Length should never be larger than size in take()");
            assert!(self.partial.is_none(), "Partial shouldn't be set if chunk < size in take()");
        }
    }
```

It would still need a test to validate my theory.

steed924:
# Description

I fixed a potential bug in DropCloserReadHalf::take() function.
I mainly applied @allada's recommendations.
Though, I kept the previous `second_chunk` logic because we need to check `second_chunk` is empty or not in order to return `first_chunk` without copying stuff.
Detailed context can be found in the issue below.

Fixes #578 

## Type of change

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/605)

### DIFF
--- a/nativelink-util/src/buf_channel.rs
+++ b/nativelink-util/src/buf_channel.rs
@@ -303,11 +303,21 @@ impl DropCloserReadHalf {
         let (first_chunk, second_chunk) = {
             // This is an optimization for a relatively common case when the first chunk in the
             // stream satisfies all the requirements to fill this `take()`.
-            // This will us from needing to copy the data into a new buffer and instead we can
+            // This will prevent us from needing to copy the data into a new buffer and instead we can
             // just forward on the original Bytes object. If we need more than the first chunk
             // we will then go the slow path and actually copy our data.
+
+            // 1. Read some data from our stream (or self.partial).
             let mut first_chunk = self.recv().await.err_tip(|| "During first buf_channel::take()")?;
+            assert!(
+                self.partial.is_none(),
+                "Partial should have been consumed during the recv()"
+            );
+            // 2. Split our data so `first_chunk` is <= `size` and puts any remaining
+            //    in `self.partial` (or set it to None).
             populate_partial_if_needed(0, size, &mut first_chunk, &mut self.partial);
+            // 3a. If our `first_chunk` is EOF, we are done.
+            // 3b. If our `first_chunk` is exactly `size` we can return our current state.
             if first_chunk.is_empty() || first_chunk.len() >= size {
                 assert!(
                     first_chunk.is_empty() || first_chunk.len() == size,
@@ -315,6 +325,10 @@ impl DropCloserReadHalf {
                 );
                 return Ok(first_chunk);
             }
+            assert!(
+                self.partial.is_none(),
+                "If `first_chunk`'s len is < size and not EOF, partial should be None"
+            );
 
             let mut second_chunk = self.recv().await.err_tip(|| "During second buf_channel::take()")?;
             if second_chunk.is_empty() {
@@ -333,20 +347,29 @@ impl DropCloserReadHalf {
 
         loop {
             let mut chunk = self.recv().await.err_tip(|| "During buf_channel::take()")?;
+            assert!(
+                self.partial.is_none(),
+                "Partial should have been consumed during the recv()"
+            );
             if chunk.is_empty() {
-                break; // EOF.
+                // Forward EOF to next recv() and return our current buffer.
+                self.partial = Some(Ok(chunk));
+                return Ok(output.freeze());
             }
 
             populate_partial_if_needed(output.len(), size, &mut chunk, &mut self.partial);
 
             output.put(chunk);
 
-            if output.len() >= size {
-                assert!(output.len() == size); // Length should never be larger than size here.
-                break;
+            if output.len() == size {
+                return Ok(output.freeze());
             }
+            assert!(output.len() < size, "Length should never be larger than size in take()");
+            assert!(
+                self.partial.is_none(),
+                "Partial shouldn't be set if chunk < size in take()"
+            );
         }
-        Ok(output.freeze())
     }
 }
 


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/605)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/buf_channel.rs` line 321 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/605#-NnyU7OW-e2QuBT2jwhj:-NnyU7OW-e2QuBT2jwhk:b1pb7ff) ([raw file](https://github.com/TraceMachina/nativelink/blob/2584294dd4c49a0dc5bdf67f6a19d11625ca5be5/nativelink-util/src/buf_channel.rs#L321)):*
> ```XML
>             // 3a. If our `first_chunk` is EOF, we are done.
>             // 3b. If our `first_chunk` is exactly `size` we can return our current state.
>             if first_chunk.is_empty() || first_chunk.len() >= size {
> ```

I don't think this part is correct.

We need to receive the EOF, and if the first chunk returns exactly the number of bytes we won't receive our EOF.

If `first_chunk.is_empty()` is true this is fine.
If `self.partial.is_some()` is true, it is ok to go down this path but we should assert `first_chunk.len() == size`.

Id rewrite this as:
```rs
// 3a. If our `first_chunk` is EOF, we are done.
if first_chunk.is_empty() {
  return Ok(first_chunk);
}

// 3b. If our first_chunk has data and it our self.partial was filled it means our stream has more data.
if self.partial.is_some() {
  assert!(first_chunk.len() == size, "Length should be exactly size here");
  return Ok(first_chunk);
}
```
___
*[`nativelink-util/src/buf_channel.rs` line 364 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/605#-NnyXz7C5FS_7OCxRkXC:-NnyXz7C5FS_7OCxRkXD:b-uz0sp2) ([raw file](https://github.com/TraceMachina/nativelink/blob/2584294dd4c49a0dc5bdf67f6a19d11625ca5be5/nativelink-util/src/buf_channel.rs#L364)):*
> ```XML
>             output.put(chunk);
> 
>             if output.len() == size {
> ```

We can only do this if `output.len() == size && self.partial.is_some()`
___
*[`nativelink-util/src/buf_channel.rs` line 367 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/605#-NnyYHuu2WVu_d4_QULL:-NnyYHuu2WVu_d4_QULM:bq5dxtw) ([raw file](https://github.com/TraceMachina/nativelink/blob/2584294dd4c49a0dc5bdf67f6a19d11625ca5be5/nativelink-util/src/buf_channel.rs#L367)):*
> ```XML
>                 return Ok(output.freeze());
>             }
>             assert!(output.len() < size, "Length should never be larger than size in take()");
> ```

nit: can needs to be `<=`
___
*[`nativelink-util/src/buf_channel.rs` line 369 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/605#-Nny_kQp6BWObavfhkBp:-Nny_kQp6BWObavfhkBq:b-7u4rsb) ([raw file](https://github.com/TraceMachina/nativelink/blob/2584294dd4c49a0dc5bdf67f6a19d11625ca5be5/nativelink-util/src/buf_channel.rs#L369)):*
> ```XML
>             assert!(output.len() < size, "Length should never be larger than size in take()");
>             assert!(
>                 self.partial.is_none(),
> ```

nit: We can remove this, it's checked at the beginning of the loop.

PR #606, opened 2024-01-12
Fix a potential bug in DropCloserReadHalf::take()

### LINKED ISSUE #578: Potential bugs in DropCloserReadHalf::take()

steed924:
# Description

I fixed a potential bug in DropCloserReadHalf::take() function.
The regression test against the potential bug is done by `send_and_take_fuzz_test` in `nativelink-util/tests/buf_channel_test.rs`.
Detailed context can be found in the issue link below.

Fixes #578 

## Type of change

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)

### DIFF
--- a/nativelink-util/src/buf_channel.rs
+++ b/nativelink-util/src/buf_channel.rs
@@ -303,16 +303,26 @@ impl DropCloserReadHalf {
         let (first_chunk, second_chunk) = {
             // This is an optimization for a relatively common case when the first chunk in the
             // stream satisfies all the requirements to fill this `take()`.
-            // This will us from needing to copy the data into a new buffer and instead we can
+            // This will prevent us from needing to copy the data into a new buffer and instead we can
             // just forward on the original Bytes object. If we need more than the first chunk
             // we will then go the slow path and actually copy our data.
+
+            // 1. Read some data from our stream (or self.partial).
             let mut first_chunk = self.recv().await.err_tip(|| "During first buf_channel::take()")?;
+            assert!(
+                self.partial.is_none(),
+                "Partial should have been consumed during the recv()"
+            );
+            // 2. Split our data so `first_chunk` is <= `size` and puts any remaining
+            //    in `self.partial` (or set it to None).
             populate_partial_if_needed(0, size, &mut first_chunk, &mut self.partial);
-            if first_chunk.is_empty() || first_chunk.len() >= size {
-                assert!(
-                    first_chunk.is_empty() || first_chunk.len() == size,
-                    "Length should be exactly size here"
-                );
+            // 3a. If our `first_chunk` is EOF, we are done.
+            if first_chunk.is_empty() {
+                return Ok(first_chunk);
+            }
+            // 3b. If our first_chunk has data and it our self.partial was filled it means our stream has more data.
+            if self.partial.is_some() {
+                assert!(first_chunk.len() == size, "Length should be exactly size here");
                 return Ok(first_chunk);
             }
 
@@ -332,21 +342,33 @@ impl DropCloserReadHalf {
         output.put(second_chunk);
 
         loop {
+            if self.partial.is_some() {
+                assert!(
+                    output.len() == size,
+                    "If partial is set expected output length to be {size}"
+                );
+                return Ok(output.freeze());
+            }
+            assert!(
+                output.len() <= size,
+                "Length should never be larger than size in take()"
+            );
+
             let mut chunk = self.recv().await.err_tip(|| "During buf_channel::take()")?;
+            assert!(
+                self.partial.is_none(),
+                "Partial should have been consumed during the recv()"
+            );
             if chunk.is_empty() {
-                break; // EOF.
+                // Forward EOF to next recv() and return our current buffer.
+                self.partial = Some(Ok(chunk));
+                return Ok(output.freeze());
             }
 
             populate_partial_if_needed(output.len(), size, &mut chunk, &mut self.partial);
 
             output.put(chunk);
-
-            if output.len() >= size {
-                assert!(output.len() == size); // Length should never be larger than size here.
-                break;
-            }
         }
-        Ok(output.freeze())
     }
 }
 
--- a/nativelink-util/tests/buf_channel_test.rs
+++ b/nativelink-util/tests/buf_channel_test.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use nativelink_error::{make_err, Code, Error, ResultExt};
 use nativelink_util::buf_channel::make_buf_channel_pair;
 use tokio::try_join;
@@ -25,6 +25,7 @@ mod buf_channel_tests {
 
     const DATA1: &str = "foo";
     const DATA2: &str = "bar";
+    const DATA3: &str = "foobar1234";
 
     #[tokio::test]
     async fn smoke_test() -> Result<(), Error> {
@@ -219,6 +220,42 @@ mod buf_channel_tests {
             Result::<(), Error>::Ok(())
         };
         try_join!(tx_fut, rx_fut)?;
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn send_and_take_fuzz_test() -> Result<(), Error> {
+        const DATA3_END_POS: usize = DATA3.len() + 1;
+        for data_size in 1..DATA3_END_POS {
+            let data: Vec<u8> = DATA3.as_bytes()[0..data_size].to_vec();
+
+            for write_size in 1..DATA3_END_POS {
+                for read_size in 1..DATA3_END_POS {
+                    let tx_data = Bytes::from(data.clone());
+                    let expected_data = Bytes::from(data.clone());
+
+                    let (mut tx, mut rx) = make_buf_channel_pair();
+
+                    let tx_fut = async move {
+                        for i in (0..data_size).step_by(write_size) {
+                            tx.send(tx_data.slice(i..std::cmp::min(data_size, i + write_size)))
+                                .await?;
+                        }
+                        tx.send_eof().await?;
+                        Result::<(), Error>::Ok(())
+                    };
+                    let rx_fut = async move {
+                        let mut round_trip_data = BytesMut::new();
+                        for _ in (0..data_size).step_by(read_size) {
+                            round_trip_data.extend(rx.take(read_size).await?.iter());
+                        }
+                        assert_eq!(round_trip_data.freeze(), expected_data);
+                        Result::<(), Error>::Ok(())
+                    };
+                    try_join!(tx_fut, rx_fut)?;
+                }
+            }
+        }
         Ok(())
     }
 


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/buf_channel.rs` line 328 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NnzBlMZ7bpQeYNnCHM2:-NnzBlMZ7bpQeYNnCHM3:bpz7l18) ([raw file](https://github.com/TraceMachina/nativelink/blob/2b73750da1ed7beace5be095af88512264d3e7c3/nativelink-util/src/buf_channel.rs#L328)):*
> ```XML
>                 return Ok(first_chunk);
>             }
>             assert!(
> ```

nit: This assert is useless now.
___
*[`nativelink-util/src/buf_channel.rs` line 364 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NnzBH0uCNUUJ8NESJPm:-NnzBH0uCNUUJ8NESJPn:bow3tv0) ([raw file](https://github.com/TraceMachina/nativelink/blob/2b73750da1ed7beace5be095af88512264d3e7c3/nativelink-util/src/buf_channel.rs#L364)):*
> ```XML
>             output.put(chunk);
> 
>             if output.len() == size && self.partial.is_some() {
> ```

nit: For readability lets make this:
```rs
if self.partial.is_some() {
  assert!(output.len() == size, "If partial is set expected output length to be {size}");
  return Ok(output.freeze());
}
```
___
*[`nativelink-util/src/buf_channel.rs` line 364 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NnzCBK293zG_bRj2rpA:-NnzCBK293zG_bRj2rpB:b-zal54f) ([raw file](https://github.com/TraceMachina/nativelink/blob/2b73750da1ed7beace5be095af88512264d3e7c3/nativelink-util/src/buf_channel.rs#L364)):*
> ```XML
>             output.put(chunk);
> 
>             if output.len() == size && self.partial.is_some() {
> ```

nit: Move this (and the rest of the code below) to the line after `loop {`

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NnzCSTP6i9RtcUmQyqf:-NnzCSTP6i9RtcUmQyqg:b-ggbs7r) (no related file):*
Testing this is a bit tricky. I think we should do more of a fuzz test on this.

What I think we should do is make a test that does something like:
```
for data_size in 0..10 {
  data = random data of size "data_size";
  for read_size in 0..10 {
    for write_size in 0..10 {
      let (tx, rx) = BuffchannelPair();
      let send_fut = async move {
        .. .send() "data" in "write_size" chunks.
      };
      let recv_fut = async move {
        .. .take() in "read_size" chunks.
      };
      let (send_res, recv_res) = join!(send_fut, recv_fut);
      ... check results and ensure the receiver got all the data and matches "data".
    }
  }
}
```

___

steed924 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NnzCSTP6i9RtcUmQyqf:-NoCC53g6C7cjUyFAMO6:b-896fix) (no related file):*

Testing this is a bit tricky. I think we should do more of a fuzz test on this.

What I think we should do is make a test that does something like:
```
for data_size in 0..10 {
  data = random data of size "data_size";
  for read_size in 0..10 {
    for write_size in 0..10 {
      let (tx, rx) = BuffchannelPair();
      let send_fut = async move {
        .. .send() "data" in "write_size" chunks.
      };
      let recv_fut = async move {
        .. .take() in "read_size" chunks.
      };
      let (send_res, recv_res) = join!(send_fut, recv_fut);
      ... check results and ensure the receiver got all the data and matches "data".
    }
  }
}
```
</blockquote></details>

Done.

___

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), ubuntu-20.04, ubuntu-22.04

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCxSLD8ETLVc7sFQeZ:-NoCxSLD8ETLVc7sFQe_:b-mfwa7r) (no related file):*
Can you also verify if this new test fails without the other changes merged?

___
___
*[`nativelink-util/tests/buf_channel_test.rs` line 229 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCvTZ7BeNUJG1rSVH3:-NoCvTZ7BeNUJG1rSVH4:bz97r7o) ([raw file](https://github.com/TraceMachina/nativelink/blob/8eb895cde87d7318ab8b0225a6a5f09e2c113ac0/nativelink-util/tests/buf_channel_test.rs#L229)):*
> ```XML
>         for data_size in 1..11 {
>             // Create a random bytes of `data_size`.
>             let data: Vec<u8> = std::iter::repeat_with(|| rand::random::<u8>())
> ```

nit: Lets not make random, instead use a const array of size "11", then slice/copy it.
___
*[`nativelink-util/tests/buf_channel_test.rs` line 236 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCvhDx8JU71TgyBBfi:-NoCvhDx8JU71TgyBBfj:b3vx0k9) ([raw file](https://github.com/TraceMachina/nativelink/blob/8eb895cde87d7318ab8b0225a6a5f09e2c113ac0/nativelink-util/tests/buf_channel_test.rs#L236)):*
> ```XML
>                 for read_size in 1..11 {
>                     let tx_data = Bytes::from(data.clone());
>                     let rx_data = Bytes::from(data.clone());
> ```

nit: Rename this `expected_data`.
___
*[`nativelink-util/tests/buf_channel_test.rs` line 249 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCwp-t4m7zi7v4yHje:-NoCwp-t4m7zi7v4yHjf:b-czm301) ([raw file](https://github.com/TraceMachina/nativelink/blob/8eb895cde87d7318ab8b0225a6a5f09e2c113ac0/nativelink-util/tests/buf_channel_test.rs#L249)):*
> ```XML
>                     };
>                     let rx_fut = async move {
>                         let mut rx_mut_data = bytes::BytesMut::new();
> ```

nit: import `BytesMut` above, try not to use inline crate names.
___
*[`nativelink-util/tests/buf_channel_test.rs` line 249 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCx0HB0q6xPUQQvXhl:-NoCx0HB0q6xPUQQvXhm:b-w5atal) ([raw file](https://github.com/TraceMachina/nativelink/blob/8eb895cde87d7318ab8b0225a6a5f09e2c113ac0/nativelink-util/tests/buf_channel_test.rs#L249)):*
> ```XML
>                     };
>                     let rx_fut = async move {
>                         let mut rx_mut_data = bytes::BytesMut::new();
> ```

nit: Rename this `round_trip_data`

steed924 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), pre-commit-checks, publish-image, ubuntu-20.04, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoCxSLD8ETLVc7sFQeZ:-NoEKew12GnA_sudkCfh:bm9vobd) (no related file):*

Can you also verify if this new test fails without the other changes merged?
</blockquote></details>

The previous `buf_channel.rs` fails the newly written test - `send_and_take_test`.
It's clear that our current updates are fixing a bug.

___

allada (approved):
:lgtm:

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/606#-:-NoER7LeBIcjasN5UdXx:bnfp4nl)* status: 1 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

___
*[`nativelink-util/tests/buf_channel_test.rs` line 227 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoEPtySBZ0R7kNZbS0V:-NoEPtySBZ0R7kNZbS0W:b-xplcy2) ([raw file](https://github.com/TraceMachina/nativelink/blob/531e84c11f173ffa4ae080a6e4a80e57b90482bf/nativelink-util/tests/buf_channel_test.rs#L227)):*
> ```XML
> 
>     #[tokio::test]
>     async fn send_and_take_test() -> Result<(), Error> {
> ```

nit: Please add a reference to the ticket this test is a regression test against.
___
*[`nativelink-util/tests/buf_channel_test.rs` line 227 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoEQtBeBbEaDwnBzGVF:-NoEQtBeBbEaDwnBzGVG:b9878vw) ([raw file](https://github.com/TraceMachina/nativelink/blob/531e84c11f173ffa4ae080a6e4a80e57b90482bf/nativelink-util/tests/buf_channel_test.rs#L227)):*
> ```XML
> 
>     #[tokio::test]
>     async fn send_and_take_test() -> Result<(), Error> {
> ```

nit: rename this `send_and_take_fuzz_test()`
___
*[`nativelink-util/tests/buf_channel_test.rs` line 228 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/606#-NoEQaPECUdwxDQI5Ro6:-NoEQaPF7OmTBuxeTjpg:b-9ce9g1) ([raw file](https://github.com/TraceMachina/nativelink/blob/531e84c11f173ffa4ae080a6e4a80e57b90482bf/nativelink-util/tests/buf_channel_test.rs#L228)):*
> ```XML
>     #[tokio::test]
>     async fn send_and_take_test() -> Result<(), Error> {
>         for data_size in 1..11 {
> ```

nit: make `11` less magical and make it `const DATA3_END_POS: usize = DATA3.len() + 1;`

PR #496, opened 2023-12-16
Fix case where filesystem store future dropping causes issues

allada:
In a rare case where a file is uploaded, but while the rename() function is called the future is dropped the filesystem store can get into a bad state.

fixes: #495

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/496)

PR #613, opened 2024-01-17
Evict on touch failure

chrisstaite-menlo:
# Description

There's an unknown failure occuring under load where the filesystem store ends up with an entry in the temp path that doesn't exist.  I can't find that fault, but rather than persisting, it makes sense to avoid the failure persisting by removing it from the eviciting map in that case.

Therefore, this change introduces a change to the LenEntry such that touch() returns a bool.  If it returns false then it is removed from the EvictingMap immediately.

## Type of change

## How Has This Been Tested?

Running the code up on my cluster.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -266,7 +266,7 @@ impl LenEntry for FileEntryImpl {
     }
 
     #[inline]
-    async fn touch(&self) {
+    async fn touch(&self) -> bool {
         let result = self
             .get_file_path_locked(move |full_content_path| async move {
                 spawn_blocking(move || {
@@ -284,8 +284,10 @@ impl LenEntry for FileEntryImpl {
             })
             .await;
         if let Err(e) = result {
-            error!("{}", e);
+            error!("{e}");
+            return false;
         }
+        true
     }
 
     // unref() only triggers when an item is removed from the eviction_map. It is possible
--- a/nativelink-store/tests/filesystem_store_test.rs
+++ b/nativelink-store/tests/filesystem_store_test.rs
@@ -122,7 +122,7 @@ impl<Hooks: FileEntryHooks + 'static + Sync + Send> LenEntry for TestFileEntry<H
         self.inner.as_ref().unwrap().is_empty()
     }
 
-    async fn touch(&self) {
+    async fn touch(&self) -> bool {
         self.inner.as_ref().unwrap().touch().await
     }
 
@@ -1004,7 +1004,6 @@ mod filesystem_store_tests {
 
                 // Test that the empty digest file is created and contains an empty length.
                 if file_metadata.is_file() && file_metadata.len() == 0 {
-                    println!("{}", file_metadata.is_file());
                     return Ok(());
                 }
             }
@@ -1040,7 +1039,7 @@ mod filesystem_store_tests {
                     ..Default::default()
                 },
                 |_| sleep(Duration::ZERO),
-                |_, _| {
+                |from, to| {
                     // If someone locked our mutex, it means we need to pause, so we
                     // simply request a lock on the same mutex.
                     if RENAME_REQUEST_PAUSE_MUX.try_lock().is_err() {
@@ -1048,7 +1047,7 @@ mod filesystem_store_tests {
                         let _lock = RENAME_REQUEST_PAUSE_MUX.lock();
                         RENAME_IS_PAUSED.store(false, Ordering::Release);
                     }
-                    Ok(())
+                    std::fs::rename(from, to)
                 },
             )
             .await?,
@@ -1102,6 +1101,37 @@ mod filesystem_store_tests {
                 Ok(())
             })
             .await?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn deleted_file_removed_from_store() -> Result<(), Error> {
+        let digest = DigestInfo::try_new(HASH1, VALUE1.len())?;
+        let content_path = make_temp_path("content_path");
+        let temp_path = make_temp_path("temp_path");
+
+        let store = Box::pin(
+            FilesystemStore::<FileEntryImpl>::new_with_timeout_and_rename_fn(
+                &nativelink_config::stores::FilesystemStore {
+                    content_path: content_path.clone(),
+                    temp_path: temp_path.clone(),
+                    read_buffer_size: 1,
+                    ..Default::default()
+                },
+                |_| sleep(Duration::ZERO),
+                |from, to| std::fs::rename(from, to),
+            )
+            .await?,
+        );
+
+        store.as_ref().update_oneshot(digest, VALUE1.into()).await?;
+
+        let stored_file_path = OsString::from(format!("{}/{}-{}", content_path, digest.hash_str(), digest.size_bytes));
+        std::fs::remove_file(stored_file_path)?;
+
+        let digest_result = store.as_ref().has(digest).await.err_tip(|| "Failed to execute has")?;
+        assert!(digest_result.is_none());
 
         Ok(())
     }
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -20,7 +20,7 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use async_lock::Mutex;
 use async_trait::async_trait;
 use futures::stream::FuturesUnordered;
-use futures::{future, join, FutureExt, StreamExt};
+use futures::{future, join, StreamExt};
 use lru::LruCache;
 use nativelink_config::stores::EvictionPolicy;
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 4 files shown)


adam-singer (commented):
@chrisstaite-menlo there are some similarities in problem and solution at https://github.com/TraceMachina/nativelink/pull/601 , would you want to pull in that pr and see if it solves issue? I think there are bits that could be merged in terms of implementation here. What isn't obvious is how the unit tests aren't failing with this patch compared to #601     

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613#-:-NoO3c6xDmBWDU8Kp_v_:b-86zlfi)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoOFQvs4X0yTa2v_5qI:-NoOFQvs4X0yTa2v_5qJ:b56j5g6) (no related file):*
Is it possible the root cause is what is being addressed in:
https://github.com/TraceMachina/nativelink/pull/496

?

What you would be seeing is the file being added to the eviction map, the future then is dropped before `fs::rename` happens. This results in:
```rs
            encoded_file_path.path_type = PathType::Content;
            encoded_file_path.digest = final_digest;
```

never being triggered, so the entry in the evicting map is now pointing to the wrong file, and the file ends up being in temp, but not reachable.

___

chrisstaite-menlo:
*[`nativelink-util/src/evicting_map.rs` line 308 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoWSluj-Uvw8OHJJTUv:-NoWSluj-Uvw8OHJJTUw:bboasiv) ([raw file](https://github.com/TraceMachina/nativelink/blob/067eccd2f7bbf9fe77749ab86d2fc163e2992bae/nativelink-util/src/evicting_map.rs#L308)):*
> ```XML
>                         // Important to track the eviction size, otherwise if we
>                         // reach the maximum we end up eviciting everything!
>                         sum_store_size -= entry.data.len() as u64;
> ```

@allada @adam-singer In implementing this I spotted this fatal bug!

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/evicting_map.rs` line 338 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoW_Hyd71ysJb-xhYQ_:-NoW_Hyd71ysJb-xhYQa:b-to8piw) ([raw file](https://github.com/TraceMachina/nativelink/blob/554bd6732cb68115b130ed63c479e9f8ef7bf283/nativelink-util/src/evicting_map.rs#L338)):*
> ```XML
>             async move {
>                 for digest in remove_digests {
>                     // Do not use inner_remove as it calls evict_items, which
> ```

And this one too...

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 308 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoWSluj-Uvw8OHJJTUv:-Nosut1Y-xh_wcNDL1uI:b-cdvq37) ([raw file](https://github.com/TraceMachina/nativelink/blob/067eccd2f7bbf9fe77749ab86d2fc163e2992bae/nativelink-util/src/evicting_map.rs#L308)):*

@allada @adam-singer In implementing this I spotted this fatal bug!
</blockquote></details>

wow great catch! Guessing our tests don't have something around eviction for `max_bytes` / `max_count` settings. Could you include test coverage for that in fixing this bug? Should be able to get a few tests in that validate this as a bug and provide the tests as a protection against regression, `evicting_map_test.rs` and adjusting the  `EvictionPolicy` should be enough to provide/show/protect against it again

```
&EvictionPolicy {
                max_count: 0,
                max_seconds: 0,
                max_bytes: 0,
                evict_bytes: 0,
            },
```
___
*[`nativelink-util/src/evicting_map.rs` line 338 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoW_Hyd71ysJb-xhYQ_:-NosvxI74-v0hKdxqQxn:b-28jtts) ([raw file](https://github.com/TraceMachina/nativelink/blob/554bd6732cb68115b130ed63c479e9f8ef7bf283/nativelink-util/src/evicting_map.rs#L338)):*

And this one too...
</blockquote></details>

another great catch! Was hoping for the freebie on that call :/

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 308 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoWSluj-Uvw8OHJJTUv:-Np4AKYi3dY-GL0St1cM:by2n9do) ([raw file](https://github.com/TraceMachina/nativelink/blob/067eccd2f7bbf9fe77749ab86d2fc163e2992bae/nativelink-util/src/evicting_map.rs#L308)):*

wow great catch! Guessing our tests don't have something around eviction for `max_bytes` / `max_count` settings. Could you include test coverage for that in fixing this bug? Should be able to get a few tests in that validate this as a bug and provide the tests as a protection against regression, `evicting_map_test.rs` and adjusting the  `EvictionPolicy` should be enough to provide/show/protect against it again

```
&EvictionPolicy {
                max_count: 0,
                max_seconds: 0,
                max_bytes: 0,
                evict_bytes: 0,
            },
```
</blockquote></details>

I'm not actually sure if we can get to this state here now I think about it, because we would have evicted on insertion if we were over the size or length limits, right?  So perhaps this isn't as bad as I thought.  However, the lower call to evict_items for every removal by calling `remove_internal` will have definitely been significantly slower.

allada (approved):
:lgtm:

> Reviewed 3 of 4 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613#-:-Np5cISv3-P-4hK8AVh-:bnfp4nl)* status: 1 of 1 LGTMs obtained

___
*[`nativelink-util/src/evicting_map.rs` line 308 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoWSluj-Uvw8OHJJTUv:-Np5awXU4OQ0VpM1qRuV:b-i5c2vj) ([raw file](https://github.com/TraceMachina/nativelink/blob/067eccd2f7bbf9fe77749ab86d2fc163e2992bae/nativelink-util/src/evicting_map.rs#L308)):*

I'm not actually sure if we can get to this state here now I think about it, because we would have evicted on insertion if we were over the size or length limits, right?  So perhaps this isn't as bad as I thought.  However, the lower call to evict_items for every removal by calling `remove_internal` will have definitely been significantly slower.
</blockquote></details>

Yeah, evicting on insert was an original design decision. I think the code was safe the way it was, but doing what you have is technically safer.
___
*[`nativelink-util/src/evicting_map.rs` line 298 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/613#-Np5bPhW3jTumBEvXRm8:-Np5bPhW3jTumBEvXRm9:b-5nb5uw) ([raw file](https://github.com/TraceMachina/nativelink/blob/554bd6732cb68115b130ed63c479e9f8ef7bf283/nativelink-util/src/evicting_map.rs#L298)):*
> ```XML
>         let mut lru_len = state.lru.len();
>         let mut sum_store_size = state.sum_store_size;
>         let digest_entries: Vec<Option<T>> = digests
> ```

nit: Maybe call this `to_touch_or_remove` or something?

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613#-:-Np5uyZ_1joWrxYhWXmw:bnfp4nl)* status: 2 of 1 LGTMs obtained

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: 2 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/613#-NoOFQvs4X0yTa2v_5qI:-NpJt9Gu10RPGFTxO9c3:bfbo17u) (no related file):*

That is the root cause, but I think it would be prudent for the EvicitingMap to handle the case where the conditions under it change (such as an external factor deleting a file).  This would allow recovery without restarting the process.
</blockquote></details>

@allada just waiting on a resolution to this comment to merge.

___

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/613)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

Issue #610, opened 2024-01-17
Filesystem store breaking in error case


chrisstaite-menlo:
I've reported this before as occurring when the disk had run out of space, but this time it was caused by the GrpcStore erroring:

```
[...]

PR #616, opened 2024-01-17
GrpcStore retry first

chrisstaite-menlo:
# Description

We see the first request in a GrpcStore request failing but this bypasses the retry logic because it's already passed out of the retry function.

Read the first message from the upstream GrpcStore within the retry logic, only exit from the retry logic after the first message was recieved.

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/616)

### DIFF
--- a/nativelink-service/src/bytestream_server.rs
+++ b/nativelink-service/src/bytestream_server.rs
@@ -251,7 +251,7 @@ impl ByteStreamServer {
         let any_store = store.clone().inner_store(Some(digest)).as_any();
         let maybe_grpc_store = any_store.downcast_ref::<Arc<GrpcStore>>();
         if let Some(grpc_store) = maybe_grpc_store {
-            let stream = grpc_store.read(Request::new(read_request)).await?.into_inner();
+            let stream = grpc_store.read(Request::new(read_request)).await?;
             return Ok(Response::new(Box::pin(stream)));
         }
 
--- a/nativelink-store/src/grpc_store.rs
+++ b/nativelink-store/src/grpc_store.rs
@@ -20,7 +20,7 @@ use std::time::Duration;
 use async_trait::async_trait;
 use bytes::BytesMut;
 use futures::stream::{unfold, FuturesUnordered};
-use futures::{future, Future, Stream, TryStreamExt};
+use futures::{future, Future, Stream, StreamExt, TryStreamExt};
 use nativelink_error::{error_if, make_input_err, Error, ResultExt};
 use nativelink_proto::build::bazel::remote::execution::v2::action_cache_client::ActionCacheClient;
 use nativelink_proto::build::bazel::remote::execution::v2::content_addressable_storage_client::ContentAddressableStorageClient;
@@ -45,7 +45,7 @@ use rand::rngs::OsRng;
 use rand::Rng;
 use tokio::time::sleep;
 use tonic::transport::Channel;
-use tonic::{transport, IntoRequest, Request, Response, Streaming};
+use tonic::{transport, IntoRequest, Request, Response, Status, Streaming};
 use tracing::error;
 use uuid::Uuid;
 
@@ -63,6 +63,31 @@ pub struct GrpcStore {
     jitter_fn: Box<dyn Fn(Duration) -> Duration + Send + Sync>,
     retry: nativelink_config::stores::Retry,
     retrier: Retrier,
+}
+
+/// This provides a buffer for the first response from GrpcStore.read in order
+/// to allow the first read to occur within the retry loop.  That means that if
+/// the connection establishes fine, but reading the first byte of the file
+/// fails we have the ability to retry before returning to the caller.
+struct FirstStream {
+    /// Contains the first response from the stream (which could be an EOF,
+    /// hence the nested Option).  This should be populated on creation and
+    /// returned as the first result from the stream.  Subsequent reads from the
+    /// stream will use the encapsulated stream.
+    first_response: Option<Option<ReadResponse>>,
+    /// The stream to get responses from when first_response is None.
+    stream: Streaming<ReadResponse>,
+}
+
+impl Stream for FirstStream {
+    type Item = Result<ReadResponse, Status>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Option<Self::Item>> {
+        if let Some(first_response) = self.first_response.take() {
+            return std::task::Poll::Ready(first_response.map(Ok));
+        }
+        Pin::new(&mut self.stream).poll_next(cx)
+    }
 }
 
 impl GrpcStore {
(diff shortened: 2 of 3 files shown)


chrisstaite-menlo:
This was an attempt to fix #610 however, it doesn't appear to have made any difference.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/616)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

___
*[`nativelink-store/src/grpc_store.rs` line 68 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOAtDG4JEqLJ9qqfZI:-NoOAtDG4JEqLJ9qqfZJ:b-qcqw0u) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-store/src/grpc_store.rs#L68)):*
> ```XML
> }
> 
> struct FirstStream {
> ```

nit: Lets add some documentation above this on why we read the first stream item and how this helps.
___
*[`nativelink-store/src/grpc_store.rs` line 672 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOB3UG7qD1IuowQJvS:-NoOB3UG7qD1IuowQJvT:b-31n45l) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-store/src/grpc_store.rs#L672)):*
> ```XML
> 
>         loop {
>             let message = match future::poll_fn(|cx| Pin::new(&mut stream).poll_next(cx)).await {
> ```

nit: You should be able to import `futures::StreamExt` then call `stream.next().await`
___
*[`nativelink-util/src/buf_channel.rs` line 167 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOBT9r-2ICNZu2NbUm:-NoO903k-6y0wYTSmEEs:bwscjef) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-util/src/buf_channel.rs#L167)):*
> ```XML
>         let maybe_chunk = match self.partial.take() {
>             // Only return bytes if we haven't already reached EOF.
>             Some(result_bytes) => self.close_tx.is_some().then_some(result_bytes),
> ```

Oh interesting.

Lets reword this comment:
```
// `partial` is allowed to have empty bytes that represent EOF, but `self.rx.recv()` should never
// respond with empty bytes as EOF. If the `partial` is an empty bytes pass None to simulate
// the stream's version of EOF.
```
___
*[`nativelink-util/src/buf_channel.rs` line 167 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOEa026ZjGwg0nr_AL:-NoOEa026ZjGwg0nr_AM:biyrjdo) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-util/src/buf_channel.rs#L167)):*
> ```XML
>         let maybe_chunk = match self.partial.take() {
>             // Only return bytes if we haven't already reached EOF.
>             Some(result_bytes) => self.close_tx.is_some().then_some(result_bytes),
> ```

Do we really want to check `self.close_tx`? Shouldn't we just check if `result_bytes` is empty?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/616)* status: 0 of 1 LGTMs obtained, and pending CI: Remote / large-ubuntu-22.04

___
*[`nativelink-util/src/buf_channel.rs` line 167 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOBT9r-2ICNZu2NbUm:-NoW4gCwCpjVN8VeMQn9:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-util/src/buf_channel.rs#L167)):*

Oh interesting.

Lets reword this comment:
```
// `partial` is allowed to have empty bytes that represent EOF, but `self.rx.recv()` should never
// respond with empty bytes as EOF. If the `partial` is an empty bytes pass None to simulate
// the stream's version of EOF.
```
</blockquote></details>

Done.
___
*[`nativelink-util/src/buf_channel.rs` line 167 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/616#-NoOEa026ZjGwg0nr_AL:-NoW4gWm5per_lPrcxfH:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/68564f679f0c81a423798f1497bed8329922bc9c/nativelink-util/src/buf_channel.rs#L167)):*

Do we really want to check `self.close_tx`? Shouldn't we just check if `result_bytes` is empty?
</blockquote></details>

Done.

adam-singer (approved):
:lgtm:

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/616#-:-Not4bCX1mSql1COB-Pr:bnfp4nl)* status: 1 of 1 LGTMs obtained

___
*[`nativelink-store/src/grpc_store.rs` line 72 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/616#-Not3pu3C7fBsrZmytut:-Not3pu3C7fBsrZmytuu:b-xf44ib) ([raw file](https://github.com/TraceMachina/nativelink/blob/76d141ad485b9d79dbdea4f912cdd0843071a6f8/nativelink-store/src/grpc_store.rs#L72)):*
> ```XML
> /// the connection establishes fine, but reading the first byte of the file
> /// fails we have the ability to retry before returning to the caller.
> struct FirstStream {
> ```

nit: wondering out loud if this kind of trick would have been needed if we "warm up" clients, meaning before service is fully ready we do some no-op operations across clients so machinery / connections are established? 
 Does this kind of stream read failure happen regardless even if clients are warn/connected?

chrisstaite-menlo:
*[`nativelink-store/src/grpc_store.rs` line 72 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/616#-Not3pu3C7fBsrZmytut:-NouoJOJAywGm3XS0K-g:bqf9xiy) ([raw file](https://github.com/TraceMachina/nativelink/blob/76d141ad485b9d79dbdea4f912cdd0843071a6f8/nativelink-store/src/grpc_store.rs#L72)):*

nit: wondering out loud if this kind of trick would have been needed if we "warm up" clients, meaning before service is fully ready we do some no-op operations across clients so machinery / connections are established? 
 Does this kind of stream read failure happen regardless even if clients are warn/connected?
</blockquote></details>

I'm not really sure, I haven't had the issue since the file store fix has been put in place, but having said that I'm also running this code too, so maybe I have?

I am still having horrid issues with a GoAway frame being sent by the upstream store which then appears to drop all requests to that store.  I think we need to add a limit to the number of concurrent upstream GrpcStore requests we perform to avoid this.

allada (approved):
:lgtm: Sorry for delay.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/616#-:-Np5tA0P4iMDDOWt5-hf:ba5hwtw)* status: 2 of 1 LGTMs obtained

PR #633, opened 2024-01-25
Ignore fast store

chrisstaite-menlo:
# Description

If there's a value in the fast store, but it's been removed from the slow store and the slow store happens to require that value because it's being used for the execution, then this causes the execution to fail.  Add an option to make checks on the store go to the slow store every time to ensure it has the values.  Serving still comes from the fast store.

## Type of change

## How Has This Been Tested?

Tested on local cluster and added new test for the change.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633)

### DIFF
--- a/nativelink-store/src/fast_slow_store.rs
+++ b/nativelink-store/src/fast_slow_store.rs
@@ -111,26 +111,11 @@ impl Store for FastSlowStore {
         digests: &[DigestInfo],
         results: &mut [Option<usize>],
     ) -> Result<(), Error> {
-        self.pin_fast_store().has_with_results(digests, results).await?;
-        let missing_digests: Vec<DigestInfo> = results
-            .iter()
-            .zip(digests.iter())
-            .filter_map(|(result, digest)| result.map_or_else(|| Some(*digest), |_| None))
-            .collect();
-        if !missing_digests.is_empty() {
-            let mut slow_store_results = self.pin_slow_store().has_many(&missing_digests).await?.into_iter();
-            // We did not change our `result` order and the slow store's results are all the None
-            // entries in `results`. This means we can just iterate over the `results`, find the None
-            // items and put in whatever the next item in the results from `slow_store_results`.
-            for result in results.iter_mut() {
-                if result.is_none() {
-                    *result = slow_store_results
-                        .next()
-                        .err_tip(|| "slow_store_results out of sync with missing_digests")?;
-                }
-            }
-        }
-        Ok(())
+        // Only check the slow store because if it's not there, then something
+        // down stream might be unable to get it.  This should not affect
+        // workers as they only use get() and a CAS can use an
+        // ExistenceCacheStore to avoid the bottleneck.
+        self.pin_slow_store().has_with_results(digests, results).await
     }
 
     async fn update(
--- a/nativelink-store/tests/fast_slow_store_test.rs
+++ b/nativelink-store/tests/fast_slow_store_test.rs
@@ -355,4 +355,27 @@ mod fast_slow_store_tests {
         );
         get_res.merge(read_res)
     }
+
+    #[tokio::test]
+    async fn ignore_value_in_fast_store() -> Result<(), Error> {
+        let fast_store = Arc::new(MemoryStore::new(&nativelink_config::stores::MemoryStore::default()));
+        let slow_store = Arc::new(MemoryStore::new(&nativelink_config::stores::MemoryStore::default()));
+        let fast_slow_store = Arc::new(FastSlowStore::new(
+            &nativelink_config::stores::FastSlowStore {
+                fast: nativelink_config::stores::StoreConfig::memory(nativelink_config::stores::MemoryStore::default()),
+                slow: nativelink_config::stores::StoreConfig::memory(nativelink_config::stores::MemoryStore::default()),
+            },
+            fast_store.clone(),
+            slow_store,
+        ));
+        let digest = DigestInfo::try_new(VALID_HASH, 100).unwrap();
+        Pin::new(fast_store.as_ref())
+            .update_oneshot(digest, make_random_data(100).into())
+            .await?;
+        assert!(
+            Pin::new(fast_slow_store.as_ref()).has(digest).await?.is_none(),
+            "Expected data to not exist in store"
+        );
+        Ok(())
+    }
 }


aaronmondal (approved):
nit: Isn't it a bug if this feature is disabled? In what situation would we not have the fast store available? If there are two machines with different fast stores (e.g. different in-memory stores) and a common S3 slow store, do we want something like this?:

```
in fast?
  yes: stream from fast store to client
  no: in slow?
    yes: stream from slow store to client (But do we also want something like an implicit slow->fast upload?)
    no: build the artifact and upload to fast
``` 

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633#-:-Np1ZF9n4gL0eVXTU47P:b74969u)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

chrisstaite-menlo (commented):
Yeah, I wasn't sure if it was a bug or not, hence making it configurable.  For example, on a worker node, there's no reason to go and check with the slow store if you have everything and increase latency because if you've got it, then it's all good, you have everything to do the work requested.  However, for a proxy to the central storage it's an absolute must.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633#-:-Np3xKVl4bGLklP8cQm2:b-i1pfnc)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633)* status: 1 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 262 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/633#-Np5dYXo5O_bbh3zZg73:-Np5dYXo5O_bbh3zZg74:b-kkbm1c) ([raw file](https://github.com/TraceMachina/nativelink/blob/045e5c2557afbd0a212dc621e962088889e2026d/nativelink-config/src/stores.rs#L262)):*
> ```XML
>     /// will still come from the fast store if it exists.
>     #[serde(default)]
>     pub check_slow_store_only: bool,
> ```

Hmmm, I wounder if this should be always enabled now, and if people need the existence cache, they can use existence caching store as a layer?

chrisstaite-menlo:
*[`nativelink-config/src/stores.rs` line 262 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/633#-Np5dYXo5O_bbh3zZg73:-Np5e8cQ21OPfUr1JeHz:b-2u32mb) ([raw file](https://github.com/TraceMachina/nativelink/blob/045e5c2557afbd0a212dc621e962088889e2026d/nativelink-config/src/stores.rs#L262)):*

Hmmm, I wounder if this should be always enabled now, and if people need the existence cache, they can use existence caching store as a layer?
</blockquote></details>

Happy to go with whatever here, obviously it's much faster to look it up in the fast store if we don't care about the slow store (i.e. workers, filesystem slowstore).

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633)* status: 1 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 262 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/633#-Np5dYXo5O_bbh3zZg73:-Np5gM960_gtcCTYrqU4:b-fxz0zv) ([raw file](https://github.com/TraceMachina/nativelink/blob/045e5c2557afbd0a212dc621e962088889e2026d/nativelink-config/src/stores.rs#L262)):*

Happy to go with whatever here, obviously it's much faster to look it up in the fast store if we don't care about the slow store (i.e. workers, filesystem slowstore).
</blockquote></details>

From my understanding here's the breakdown of what components would ever call what:

Stores:
* ActionCache should be wrapped in `CompletenessCheckingStore` or else it can serve references to evicted items.
* CAS should be wrapped in `ExistenceCacheStore` and possibly `FastSlowStore`, for dramatic efficiency gains.

Scheduler:
* IIRC, scheduler doesn't actually need a store, unless it wants to extract details out about the job

Workers:
* Only ever calls `.get` & `.update` on CAS, never calls `.has`
* Always requires a `FastSlowStore(fast: FilesystemStore, slow: *)`, but it dynamically casts out the `FilesystemStore` at startup and talks to it directly (ie: bypass the Store interface).

I guess, for simplicity, lets remove this and we can add it back in if we ever find it is a bottleneck.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633)* status: 1 of 1 LGTMs obtained, and pending CI: Vercel, pre-commit-checks

___
*[`nativelink-config/src/stores.rs` line 262 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/633#-Np5dYXo5O_bbh3zZg73:-Np5vTce4tlWCYLywg84:b5c6qsd) ([raw file](https://github.com/TraceMachina/nativelink/blob/045e5c2557afbd0a212dc621e962088889e2026d/nativelink-config/src/stores.rs#L262)):*

From my understanding here's the breakdown of what components would ever call what:

Stores:
* ActionCache should be wrapped in `CompletenessCheckingStore` or else it can serve references to evicted items.
* CAS should be wrapped in `ExistenceCacheStore` and possibly `FastSlowStore`, for dramatic efficiency gains.

Scheduler:
* IIRC, scheduler doesn't actually need a store, unless it wants to extract details out about the job

Workers:
* Only ever calls `.get` & `.update` on CAS, never calls `.has`
* Always requires a `FastSlowStore(fast: FilesystemStore, slow: *)`, but it dynamically casts out the `FilesystemStore` at startup and talks to it directly (ie: bypass the Store interface).

I guess, for simplicity, lets remove this and we can add it back in if we ever find it is a bottleneck.
</blockquote></details>

Cool, I'll update to ignore the fast store then.

allada (approved):
:lgtm:

> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633#-:-Np66rJj5DvpWC4XiP3o:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

aaronmondal (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633#-:-Np6LUH-Ev5bQ-NJn78n:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

adam-singer (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/633)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

PR #634, opened 2024-01-25
Don't retry permanent failures

chrisstaite-menlo:
# Description

The retrier always retries on error.  However, some errors are permanent, for example NotFound.  This means that if a request is made to get something from the CAS (or GrpcStore AC) which is not there, it keeps trying even though it's still not there.  This defines the set of errors which shoudl be retried and fails otherwise.

## Type of change

## How Has This Been Tested?

Remote proxy getting data from the action cache now doesn't retry ad infinitum.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -548,6 +548,28 @@ pub struct GrpcStore {
     pub retry: Retry,
 }
 
+/// The possible error codes that might occur on an upstream request.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum ErrorCode {
+    Cancelled = 1,
+    Unknown = 2,
+    InvalidArgument = 3,
+    DeadlineExceeded = 4,
+    NotFound = 5,
+    AlreadyExists = 6,
+    PermissionDenied = 7,
+    ResourceExhausted = 8,
+    FailedPrecondition = 9,
+    Aborted = 10,
+    OutOfRange = 11,
+    Unimplemented = 12,
+    Internal = 13,
+    Unavailable = 14,
+    DataLoss = 15,
+    Unauthenticated = 16,
+    // Note: This list is duplicated from nativelink-error/lib.rs.
+}
+
 /// Retry configuration. This configuration is exponential and each iteration
 /// a jitter as a percentage is applied of the calculated delay. For example:
 /// ```rust,ignore
@@ -591,4 +613,18 @@ pub struct Retry {
     /// ```
     #[serde(default)]
     pub jitter: f32,
+
+    /// A list of error codes to retry on, if this is not set then the default
+    /// error codes to retry on are used.  These default codes are the most
+    /// likely to be non-permanent.
+    ///  - Unknown
+    ///  - Cancelled
+    ///  - DeadlineExceeded
+    ///  - ResourceExhausted
+    ///  - Aborted
+    ///  - Internal
+    ///  - Unavailable
+    ///  - DataLoss
+    #[serde(default)]
+    pub retry_on_errors: Option<Vec<ErrorCode>>,
 }
--- a/nativelink-error/src/lib.rs
+++ b/nativelink-error/src/lib.rs
@@ -312,6 +312,8 @@ pub enum Code {
     Unavailable = 14,
     DataLoss = 15,
     Unauthenticated = 16,
+    // NOTE: Additional codes must be added to stores.rs in ErrorCodes and also
+    // in both match statements in retry.rs.
 }
 
 impl From<i32> for Code {
--- a/nativelink-scheduler/src/grpc_scheduler.rs
+++ b/nativelink-scheduler/src/grpc_scheduler.rs
@@ -28,7 +28,7 @@ use nativelink_proto::build::bazel::remote::execution::v2::{
 };
 use nativelink_proto::google::longrunning::Operation;
 use nativelink_util::action_messages::{ActionInfo, ActionInfoHashKey, ActionState, DEFAULT_EXECUTION_PRIORITY};
-use nativelink_util::retry::{ExponentialBackoff, Retrier, RetryResult};
+use nativelink_util::retry::{Retrier, RetryResult};
 use nativelink_util::tls_utils;
 use parking_lot::Mutex;
 use rand::rngs::OsRng;
@@ -46,8 +46,6 @@ pub struct GrpcScheduler {
     capabilities_client: CapabilitiesClient<transport::Channel>,
     execution_client: ExecutionClient<transport::Channel>,
     platform_property_managers: Mutex<HashMap<String, Arc<PlatformPropertyManager>>>,
-    jitter_fn: Box<dyn Fn(Duration) -> Duration + Send + Sync>,
-    retry: nativelink_config::stores::Retry,
     retrier: Retrier,
 }
 
@@ -76,9 +74,11 @@ impl GrpcScheduler {
             capabilities_client: CapabilitiesClient::new(channel.clone()),
             execution_client: ExecutionClient::new(channel),
             platform_property_managers: Mutex::new(HashMap::new()),
-            jitter_fn,
-            retry: config.retry.clone(),
-            retrier: Retrier::new(Arc::new(|duration| Box::pin(sleep(duration)))),
+            retrier: Retrier::new(
+                Arc::new(|duration| Box::pin(sleep(duration))),
+                Arc::new(jitter_fn),
+                config.retry.to_owned(),
+            ),
         })
     }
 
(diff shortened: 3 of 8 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/634#-Np5mt-E7nl9S_6pFUbb):*
Actually, I'm not sure we should encode this strictly in the code, instead lets make this a config on every retry config.

Then make what you have the default.

The reason is because some S3 store implementations will return `NotFound` even though the item is there because it's eventual consistency. I is silly, but I've personally seen this and if you delay the read just a few micros it comes into existence.

I suggest make a `ErrorRetryMapping` struct in the config, but super simple, maybe just:
```rs
struct ErrorRetryMapping {
  not_found: bool,
  other: bool,
}
```

We can add more special cases to this later, and the default with be `{ not_found: false, other: true }`.

Thoughts?

chrisstaite-menlo:
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/634#-Np5mt-E7nl9S_6pFUbb):*

Actually, I'm not sure we should encode this strictly in the code, instead lets make this a config on every retry config.

Then make what you have the default.

The reason is because some S3 store implementations will return `NotFound` even though the item is there because it's eventual consistency. I is silly, but I've personally seen this and if you delay the read just a few micros it comes into existence.

I suggest make a `ErrorRetryMapping` struct in the config, but super simple, maybe just:
```rs
struct ErrorRetryMapping {
  not_found: bool,
  other: bool,
}
```

We can add more special cases to this later, and the default with be `{ not_found: false, other: true }`.

Thoughts?
</blockquote></details>

Sure.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 597 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLALTo5ncSb_I1xD34:-NpLkFDHA-L1fxnTlfAW:bteqvux) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-config/src/stores.rs#L597)):*

eeh, I've been trying to avoid setting anything that is not default to a non-default value in the config.

I'm actually thinking if maybe we should do this struct as an enum, then have it:

```rs
no_retry_errors: Option<HashSet<ErrorCode>>
```

The nice thing about doing it this way is `ErrorCode` can easily be used in other configs in the future (if needed).
</blockquote></details>

I'd like there to be a set of sensible codes that don't retry by default rather than requiring the user to specify them though, also requiring users to know error codes rather than nice strings seems a bit much.
___
*[`nativelink-error/src/lib.rs` line 315 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLBfbV4iUOXMf6jx0m:-NpLkSV39-DCOCRq2BIh:b-i6j14j) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-error/src/lib.rs#L315)):*

:-(
</blockquote></details>

Can't have exhaustive `match` outside of a module :'(

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Vercel, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), pre-commit-checks, publish-image, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable

___
*[`nativelink-config/src/stores.rs` line 597 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLALTo5ncSb_I1xD34:-NpLlXQ606zuo1zTf-Kb:b7vqv7z) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-config/src/stores.rs#L597)):*

I'd like there to be a set of sensible codes that don't retry by default rather than requiring the user to specify them though, also requiring users to know error codes rather than nice strings seems a bit much.
</blockquote></details>

Perhaps we could rename this to `do_not_retry_other` and then invert the boolean?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained, and pending CI: docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04)

___
*[`nativelink-util/src/retry.rs` line 97 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLmiKM3961qKVRzw3a:-NpLpsf_62hmolq-_y2C:b-bnway6) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-util/src/retry.rs#L97)):*

nit: not that we support other backoff functions yet, having backoff as a configuration would be a nice to have, non-blocker and maybe something to explore at a later point in time.  Exponential backoff is def most common/graceful in general scenarios
</blockquote></details>

Yes, that was one of the motivations for moving the configuration into `Retry` so it can be easily modified later without having to modify other parts of the system such as `GrpcStore` or `S3Store`.

allada (changes requested):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 597 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLALTo5ncSb_I1xD34:-NpNhFR-5xQVPWgX-JI1:bakp56f) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-config/src/stores.rs#L597)):*

Perhaps we could rename this to `do_not_retry_other` and then invert the boolean?
</blockquote></details>

Oh, I am saying we should default `None` to what you have defaulted now. I'm just afraid we are going to lock ourselves in as we add new implementations of stores that use this retry trait. For example, `RedisStore` might have a spurious bug where the credentials are only updated on failure (sounds dumb, but I'm sure you've seen dumber bugs), to which a user would just want to retry on `Unauthenticated` to get around this issue, but by default we should not do this.

Again, I don't want to block you, I just want to make sure we design our public api's with good thought.

adam-singer (approved):
> Reviewed 1 of 8 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 597 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/634#-NpLALTo5ncSb_I1xD34:-NpVbVcL1wZntO9QgEYM:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/7c8a40e3ffed5d0ba9a32d55335bbc6ef210ab03/nativelink-config/src/stores.rs#L597)):*

Oh, I am saying we should default `None` to what you have defaulted now. I'm just afraid we are going to lock ourselves in as we add new implementations of stores that use this retry trait. For example, `RedisStore` might have a spurious bug where the credentials are only updated on failure (sounds dumb, but I'm sure you've seen dumber bugs), to which a user would just want to retry on `Unauthenticated` to get around this issue, but by default we should not do this.

Again, I don't want to block you, I just want to make sure we design our public api's with good thought.
</blockquote></details>

Done.

allada (approved):
:lgtm:

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/634#-:-Npbgyvp0w6s-F-m2Tss:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #638, opened 2024-01-26
GrpcStore Write Retry

chrisstaite-menlo:
# Description

The current implementation of retry in GrpcStore is awkward and only allows retry up until the first call to the WriteRequestStreamWrapper, however this has a buffer of the first message in it.  Therefore, with a bit of refactoring we are able to retry up until the second message is requested by Tonic without any degredation in performance.  This has the added benefit of being able to refactor the interface to be a Stream.

## Type of change

## How Has This Been Tested?

Using proxy in my cluster.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)

### DIFF
--- a/nativelink-service/src/bytestream_server.rs
+++ b/nativelink-service/src/bytestream_server.rs
@@ -396,11 +396,11 @@ impl ByteStreamServer {
                     // by counting the number of bytes sent from the client. If they send
                     // less than the amount they said they were going to send and then
                     // close the stream, we know there's a problem.
-                    Ok(None) => return Err(make_input_err!("Client closed stream before sending all data")),
+                    None => return Err(make_input_err!("Client closed stream before sending all data")),
                     // Code path for client stream error. Probably client disconnect.
-                    Err(err) => return Err(err),
+                    Some(Err(err)) => return Err(err),
                     // Code path for received chunk of data.
-                    Ok(Some(write_request)) => write_request,
+                    Some(Ok(write_request)) => write_request,
                 };
 
                 if write_request.write_offset < 0 {
--- a/nativelink-service/tests/bytestream_server_test.rs
+++ b/nativelink-service/tests/bytestream_server_test.rs
@@ -235,7 +235,7 @@ pub mod write_tests {
             // Now disconnect our stream.
             drop(tx);
             let (result, _bs_server) = join_handle.await?;
-            assert!(result.is_ok(), "Expected success to be returned");
+            result?;
         }
         {
             // Check to make sure our store recorded the data properly.
--- a/nativelink-store/src/grpc_store.rs
+++ b/nativelink-store/src/grpc_store.rs
@@ -15,6 +15,7 @@
 use std::marker::Send;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -35,6 +36,7 @@ use nativelink_proto::google::bytestream::{
 };
 use nativelink_util::buf_channel::{DropCloserReadHalf, DropCloserWriteHalf};
 use nativelink_util::common::DigestInfo;
+use nativelink_util::resource_info::ResourceInfo;
 use nativelink_util::retry::{Retrier, RetryResult};
 use nativelink_util::store_trait::{Store, UploadSizeInfo};
 use nativelink_util::tls_utils;
@@ -44,7 +46,6 @@ use prost::Message;
 use rand::rngs::OsRng;
 use rand::Rng;
 use tokio::time::sleep;
-use tonic::transport::Channel;
 use tonic::{transport, IntoRequest, Request, Response, Status, Streaming};
 use tracing::error;
 use uuid::Uuid;
@@ -85,6 +86,132 @@ impl Stream for FirstStream {
             return std::task::Poll::Ready(first_response.map(Ok));
         }
         Pin::new(&mut self.stream).poll_next(cx)
+    }
+}
+
+/// This structure wraps all of the information required to perform a write
+/// request on the GrpcStore.  It stores the last message retrieved which allows
+/// the write to resume since the UUID allows upload resume at the server.
+struct WriteState<T, E>
+where
+    T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
+    E: Into<Error> + 'static,
+{
+    instance_name: String,
+    read_stream_error: Option<Error>,
+    read_stream: WriteRequestStreamWrapper<T, E>,
+    // Tonic doesn't appear to report an error until it has taken two messages,
+    // therefore we are required to buffer the last two messages.
+    cached_messages: [Option<WriteRequest>; 2],
+    // When resuming after an error, the previous messages are cloned into this
+    // queue upfront to allow them to be served back.
+    resume_queue: [Option<WriteRequest>; 2],
+    // An optimisation to avoid having to manage resume_queue when it's empty.
+    is_resumed: bool,
+}
+
+impl<T, E> WriteState<T, E>
+where
+    T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
+    E: Into<Error> + 'static,
+{
+    fn new(instance_name: String, read_stream: WriteRequestStreamWrapper<T, E>) -> Self {
+        Self {
+            instance_name,
+            read_stream_error: None,
+            read_stream,
+            cached_messages: [None, None],
+            resume_queue: [None, None],
+            is_resumed: false,
+        }
+    }
+
+    fn push_message(&mut self, message: WriteRequest) {
+        self.cached_messages.swap(0, 1);
+        self.cached_messages[0] = Some(message);
+    }
+
+    fn resumed_message(&mut self) -> Option<WriteRequest> {
+        if self.is_resumed {
+            // The resume_queue is a circular buffer, that we have to shift,
+            // since its only got two elements its a trivial swap.
+            self.resume_queue.swap(0, 1);
+            let message = self.resume_queue[0].take();
+            if message.is_none() {
+                self.is_resumed = false;
+            }
+            message
+        } else {
+            None
+        }
+    }
+
+    fn can_resume(&self) -> bool {
+        self.read_stream_error.is_none() && (self.cached_messages[0].is_some() || self.read_stream.is_first_msg())
+    }
+
+    fn resume(&mut self) {
+        self.resume_queue = self.cached_messages.clone();
+        self.is_resumed = true;
+    }
+}
+
+/// A wrapper around WriteState to allow it to be reclaimed from the underlying
+/// write call in the case of failure.
+struct WriteStateWrapper<T, E>
+where
+    T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
+    E: Into<Error> + 'static,
+{
+    shared_state: Arc<Mutex<WriteState<T, E>>>,
+}
+
+impl<T, E> Stream for WriteStateWrapper<T, E>
+where
+    T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
+    E: Into<Error> + 'static,
+{
+    type Item = WriteRequest;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // This should be an uncontended lock since write was called.
+        let mut local_state = self.shared_state.lock();
+        // If this is the first or second call after a failure and we have
+        // cached messages, then use the cached write requests.
+        let cached_message = local_state.resumed_message();
+        if cached_message.is_some() {
+            return Poll::Ready(cached_message);
+        }
+        // Read a new write request from the downstream.
+        let Poll::Ready(maybe_message) = Pin::new(&mut local_state.read_stream).poll_next(cx) else {
+            return Poll::Pending;
+        };
+        // Update the instance name in the write request and forward it on.
+        const IS_UPLOAD_TRUE: bool = true;
+        let result = match maybe_message {
+            Some(Ok(mut message)) => match ResourceInfo::new(&message.resource_name, IS_UPLOAD_TRUE) {
+                Ok(mut resource_name) => {
+                    if resource_name.instance_name != local_state.instance_name {
+                        resource_name.instance_name = &local_state.instance_name;
+                        message.resource_name = resource_name.to_string(IS_UPLOAD_TRUE);
+                    }
+                    // Cache the last request in case there is an error to allow
+                    // the upload to be resumed.
+                    local_state.push_message(message.clone());
+                    Some(message)
+                }
+                Err(err) => {
+                    error!("{err:?}");
+                    None
+                }
+            },
+            Some(Err(err)) => {
+                local_state.read_stream_error = Some(err);
+                None
+            }
+            None => None,
+        };
+        Poll::Ready(result)
     }
 }
 
@@ -237,16 +364,12 @@ impl GrpcStore {
     }
 
     fn get_read_request(&self, mut request: ReadRequest) -> Result<ReadRequest, Error> {
-        // `resource_name` pattern is: "{instance_name}/blobs/{hash}/{size}".
-        let first_slash_pos = request
-            .resource_name
-            .find('/')
-            .err_tip(|| "Resource name expected to follow pattern {instance_name}/blobs/{hash}/{size}")?;
-        request.resource_name = format!(
-            "{}/{}",
-            self.instance_name,
-            request.resource_name.get((first_slash_pos + 1)..).unwrap()
-        );
+        const IS_UPLOAD_FALSE: bool = false;
+        let mut resource_info = ResourceInfo::new(&request.resource_name, IS_UPLOAD_FALSE)?;
+        if resource_info.instance_name != self.instance_name {
+            resource_info.instance_name = &self.instance_name;
+            request.resource_name = resource_info.to_string(IS_UPLOAD_FALSE);
+        }
         Ok(request)
     }
 
@@ -295,81 +418,49 @@ impl GrpcStore {
             "CAS operation on AC store"
         );
 
-        struct LocalState<T, E>
-        where
-            T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
-            E: Into<Error> + 'static,
-        {
-            instance_name: String,
-            error: Mutex<Option<Error>>,
-            read_stream: Mutex<Option<WriteRequestStreamWrapper<T, E>>>,
-            client: ByteStreamClient<Channel>,
-        }
-
-        let local_state = Arc::new(LocalState {
-            instance_name: self.instance_name.clone(),
-            error: Mutex::new(None),
-            read_stream: Mutex::new(Some(stream)),
-            client: self.bytestream_client.clone(),
-        });
+        let local_state = Arc::new(Mutex::new(WriteState::new(self.instance_name.clone(), stream)));
 
         let result = self
             .retrier
-            .retry(
-                unfold(local_state, move |local_state| async move {
-                    let stream = unfold((None, local_state.clone()), move |(stream, local_state)| async {
-                        // Only consume the stream on the first request to read,
-                        // then pass it for future requests in the unfold.
-                        let mut stream = stream.or_else(|| local_state.read_stream.lock().take())?;
-                        let maybe_message = stream.next().await;
-                        if let Ok(maybe_message) = maybe_message {
-                            if let Some(mut message) = maybe_message {
-                                // `resource_name` pattern is: "{instance_name}/uploads/{uuid}/blobs/{hash}/{size}".
-                                let first_slash_pos = match message.resource_name.find('/') {
-                                    Some(pos) => pos,
-                                    None => {
-                                        error!("{}", "Resource name should follow pattern {instance_name}/uploads/{uuid}/blobs/{hash}/{size}");
-                                        return None;
-                                    }
-                                };
-                                message.resource_name = format!(
-                                    "{}/{}",
-                                    &local_state.instance_name,
-                                    message.resource_name.get((first_slash_pos + 1)..).unwrap()
-                                );
-                                return Some((message, (Some(stream), local_state)));
+            .retry(unfold(local_state, move |local_state| async move {
+                let mut client = self.bytestream_client.clone();
+                // The client write may occur on a separate thread and
+                // therefore in order to share the state with it we have to
+                // wrap it in a Mutex and retrieve it after the write
+                // has completed.  There is no way to get the value back
+                // from the client.
+                let result = client
+                    .write(WriteStateWrapper {
+                        shared_state: local_state.clone(),
+                    })
+                    .await;
+
+                // Get the state back from StateWrapper, this should be
+                // uncontended since write has returned.
+                let mut local_state_locked = local_state.lock();
+
+                let result = if let Some(err) = local_state_locked.read_stream_error.take() {
+                    // If there was an error with the stream, then don't
+                    // retry.
+                    RetryResult::Err(err)
+                } else {
+                    // On error determine whether it is possible to retry.
+                    match result.err_tip(|| "in GrpcStore::write") {
+                        Err(err) => {
+                            if local_state_locked.can_resume() {
+                                local_state_locked.resume();
+                                RetryResult::Retry(err)
+                            } else {
+                                RetryResult::Err(err.append("Retry is not possible"))
                             }
-                            return None;
                         }
-                        // TODO(allada) I'm sure there's a way to do this without a mutex, but rust can be super
-                        // picky with borrowing through a stream await.
-                        *local_state.error.lock() = Some(maybe_message.unwrap_err());
-                        None
-                    });
+                        Ok(response) => RetryResult::Ok(response),
+                    }
+                };
 
-                    let result = local_state.client.clone()
-                            .write(stream)
-                            .await
-                            .err_tip(|| "in GrpcStore::write");
-
-                    // If the stream has been consumed, don't retry, but
-                    // otherwise it's ok to try again.
-                    let result = if local_state.read_stream.lock().is_some() {
-                        result.map_or_else(RetryResult::Retry, RetryResult::Ok)
-                    } else {
-                        result.map_or_else(RetryResult::Err, RetryResult::Ok)
-                    };
-
-                    // If there was an error with the stream, then don't retry.
-                    let result = if let Some(err) = local_state.error.lock().take() {
-                        RetryResult::Err(err)
-                    } else {
-                        result
-                    };
-
-                    Some((result, local_state))
-                }),
-            )
+                drop(local_state_locked);
+                Some((result, local_state))
+            }))
             .await?;
         Ok(result)
     }
@@ -385,15 +476,12 @@ impl GrpcStore {
 
         let mut request = grpc_request.into_inner();
 
-        // `resource_name` pattern is: "{instance_name}/uploads/{uuid}/blobs/{hash}/{size}".
-        let first_slash_pos = request.resource_name.find('/').err_tip(|| {
-            "Resource name expected to follow pattern {instance_name}/uploads/{uuid}/blobs/{hash}/{size}"
-        })?;
-        request.resource_name = format!(
-            "{}/{}",
-            self.instance_name,
-            request.resource_name.get((first_slash_pos + 1)..).unwrap()
-        );
+        const IS_UPLOAD_TRUE: bool = true;
+        let mut request_info = ResourceInfo::new(&request.resource_name, IS_UPLOAD_TRUE)?;
+        if request_info.instance_name != self.instance_name {
+            request_info.instance_name = &self.instance_name;
+            request.resource_name = request_info.to_string(IS_UPLOAD_TRUE);
+        }
 
         self.perform_request(request, |request| async move {
             self.bytestream_client
--- a/nativelink-util/src/resource_info.rs
+++ b/nativelink-util/src/resource_info.rs
@@ -84,6 +84,7 @@ pub struct ResourceInfo<'a> {
     pub compressor: Option<&'a str>,
     pub digest_function: Option<&'a str>,
     pub hash: &'a str,
+    size: &'a str,
     pub expected_size: usize,
     pub optional_metadata: Option<&'a str>,
 }
@@ -128,6 +129,25 @@ impl<'a> ResourceInfo<'a> {
             output.instance_name = instance_name;
         }
         Ok(output)
+    }
+
+    pub fn to_string(&self, is_upload: bool) -> String {
+        [
+            Some(self.instance_name),
+            is_upload.then_some("uploads"),
+            self.uuid,
+            Some(self.compressor.map_or("blobs", |_| "compressed-blobs")),
+            self.compressor,
+            self.digest_function,
+            Some(self.hash),
+            Some(self.size),
+            self.optional_metadata,
+        ]
+        .into_iter()
+        .flatten()
+        .filter(|part| !part.is_empty())
+        .collect::<Vec<&str>>()
+        .join("/")
     }
 }
 
@@ -177,8 +197,9 @@ fn recursive_parse<'a>(
                     output.compressor = Some(part);
                     *bytes_processed += part.len() + SLASH_SIZE;
                     return Ok(state);
+                } else {
+                    return Err(make_input_err!("Expected compressor, got {part}"));
                 }
-                continue;
             }
             State::DigestFunction => {
                 state = State::Hash;
@@ -196,6 +217,7 @@ fn recursive_parse<'a>(
                 return Ok(State::Size);
             }
             State::Size => {
+                output.size = part;
                 output.expected_size = part
                     .parse::<usize>()
                     .map_err(|_| make_input_err!("Digest size_bytes was not convertible to usize. Got: {}", part))?;
--- a/nativelink-util/src/write_request_stream_wrapper.rs
+++ b/nativelink-util/src/write_request_stream_wrapper.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 The Native Link Authors. All rights reserved.
+// Copyright 2023-2024 The Native Link Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,8 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
 use futures::{Stream, StreamExt};
-use nativelink_error::{error_if, Error, ResultExt};
+use nativelink_error::{error_if, make_input_err, Error, ResultExt};
 use nativelink_proto::google::bytestream::WriteRequest;
 
 use crate::resource_info::ResourceInfo;
@@ -56,7 +59,6 @@ where
         let hash = resource_info.hash.to_string();
         let expected_size = resource_info.expected_size;
         let uuid = resource_info.uuid.map(|v| v.to_string());
-        let write_finished = first_msg.finish_write;
 
         Ok(WriteRequestStreamWrapper {
             instance_name,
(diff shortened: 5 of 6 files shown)


allada (commented):
+@blakehatch Could you take a first pass?

> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-Np5cofC6O3xvpb6hud3:brh2tb1)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

blakehatch:
> +@blakehatch Could you take a first pass?
> 
> > Reviewed 2 of 4 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-Np5cofC6O3xvpb6hud3:brh2tb1)_ status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

Sure!

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 127 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np5oD0N0H18g53hhtih:-Np5oD0N0H18g53hhtii:b-w360sx) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L127)):*
> ```XML
>                     maybe_message.err_tip(|| format!("Stream error at byte {}", self.bytes_received))
>                 }
>                 Poll::Ready(None) => Err(make_input_err!("Expected WriteRequest struct in stream")),
> ```

nit: `match Pin::new(&mut self.stream).poll\_next(cx) {

Poll::Pending => Poll::Pending,

Poll::Ready(maybe\_message) => maybe\_message?

.err\_tip(|| format!("Stream error at byte {}", self.bytes\_received)),

}`

Could be made more concise with ? to propogate the error up instead of handling error case separately in the match

chrisstaite-menlo:
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 127 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np5oD0N0H18g53hhtih:-Np5yfZqCk-AygB7uGOt:b-n37u6v) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L127)):*

nit: `match Pin::new(&mut self.stream).poll\_next(cx) {

Poll::Pending => Poll::Pending,

Poll::Ready(maybe\_message) => maybe\_message?

.err\_tip(|| format!("Stream error at byte {}", self.bytes\_received)),

}`

Could be made more concise with ? to propogate the error up instead of handling error case separately in the match
</blockquote></details>

It can't be because we would need to `return Poll::Ready(Some(Err(...))))` which I did for a while, but it was ugly.  This way we handle the error from `maybe_message` in the same way as the EOF error which standardises the code flow.

blakehatch (approved):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 116 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np5x2ZyCNBTp4gfRCUA:-Np5x2ZyCNBTp4gfRCUB:baljap9) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L116)):*
> ```XML
>         // subsequent message from the wrapped Stream.
>         let maybe_message = if self.message_count == 0 {
>             if let Some(first_msg) = self.first_msg.clone() {
> ```

Nit: let maybe\_message = if self.message\_count == 0 {

self.first\_msg.take().ok\_or\_else(|| make\_input\_err!("First message was lost in write stream wrapper"))

}

Could possibly be made more concise and a clone of the WriteRequest could be avoided by using take

chrisstaite (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 116 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np5x2ZyCNBTp4gfRCUA:-Np630GFAuH3IH6AvLq6:b-7u57qv) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L116)):*

Nit: let maybe\_message = if self.message\_count == 0 {

self.first\_msg.take().ok\_or\_else(|| make\_input\_err!("First message was lost in write stream wrapper"))

}

Could possibly be made more concise and a clone of the WriteRequest could be avoided by using take
</blockquote></details>

The whole point of this PR is that this is a clone() not a take() to allow a retry. This is not very inefficient since it contains a Bytes which is only a pointer.

adam-singer (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 296 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np87t8ODUgTPG8qUHHl:-Np87t8ODUgTPG8qUHHm:b28pgjh) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L296)):*
> ```XML
>         // A wrapper to allow the LocalState to be given to an attempt but
>         // retrieved if the attempt fails.
>         struct StateWrapper<T, E>
> ```

nit: Lets hoist this to the top of file, it's too complex (imo) to inline with the function.
___
*[`nativelink-store/src/grpc_store.rs` line 307 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8609E6abg4YsxGW1c:-Np8609E6abg4YsxGW1d:bcvvpun) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L307)):*
> ```XML
>         where
>             T: Stream<Item = Result<WriteRequest, E>> + Unpin + Send + 'static,
>             E: Into<Error> + 'static,
> ```

super nit: I'm 50% sure you don't need `'static` here.
___
*[`nativelink-store/src/grpc_store.rs` line 320 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np86rUe9IUOe4Ohtovo:-Np86rUe9IUOe4Ohtovp:b-yw9a43) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L320)):*
> ```XML
>                     Some(Ok(mut message)) => {
>                         // `resource_name` pattern is: "{instance_name}/uploads/{uuid}/blobs/{hash}/{size}".
>                         if let Some(first_slash_pos) = message.resource_name.find('/') {
> ```

nit: Can we instead use `ResourceInfo` to do this parsing?

I'm sure the reason you didn't want to use it before was because it cannot re-combine them (which I think we should implement anyway), secondly, You can't mutate easily, which you can probably change all those `&'a str`into `Cow<'a, str>` instead to get the same effect, but make it easier to mutate, then recombine.

A nice side-effect is that it will preserve things like `compressed-blobs` and `digest_function`.
___
*[`nativelink-store/src/grpc_store.rs` line 349 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np88elI8t_dzyU8whg2:-Np88elI8t_dzyU8whg3:bmbmhxe) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L349)):*
> ```XML
>                 unfold(local_state, move |local_state| async move {
>                     let mut client = local_state.lock().client.clone();
>                     // The client write may occur on a separate thread and
> ```

Actually, there's a trick you can use here...

If you want to be super optimized, we could make `shared_state` a raw value, and add a mutable reference to the output `&mut Option<SharedState>`, then make a custom destructor on `StateWrapper` that will move the SharedState to the parent's ref on destruction.

The outer function then just needs to assert it exists.

Technically faster, since it removes both the Arc and the Mutex, but a lot more complicated, so probably not worth the effort.

Plus IIRC, `parking_lot::Mutex` is a spinlock then a real lock or something, so in this use case it's only ever going to be atomic operations.
___
*[`nativelink-store/src/grpc_store.rs` line 366 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8BnEv6OxrPnWnkf9e:-Np8BnEv6OxrPnWnkf9f:b7vt9sj) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L366)):*
> ```XML
>                     // If the stream has been consumed, don't retry, but
>                     // otherwise it's ok to try again.
>                     let result = if result.is_err() && local_state_locked.read_stream.reset_for_retry() {
> ```

nit: Kinda complicated to read, more direct code, like:
```rs
match result {
  Err(err) => {
    if (local_state_locked.read_stream.is_retryable()) {
      local_state_locked.read_stream.reset();
      log::log!("GRPC message here stating it retried");
      RetryResult::Retry(())
    } else {
      RetryResult::Err(err.err_tip(|| "Retry was not possible in GrpcStore::write"))
    }
  }
  Ok(()) => RetryResult::Ok(()),
}
```
___
*[`nativelink-store/src/grpc_store.rs` line 369 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8DQGI9JycsJitFYlM:-Np8DQGI9JycsJitFYlN:bkwerx1) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L369)):*
> ```XML
>                         result
>                             .err_tip(|| "Retry is possible in GrpcStore::write")
>                             .map_or_else(RetryResult::Retry, RetryResult::Ok)
> ```

(related to above: This should never be `RetryResult::Ok` here.
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 1 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8DqpD33DYyBU6tOXO:-Np8DqpD33DYyBU6tOXP:b-z1r1sc) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L1)):*
> ```XML
> // Copyright 2024 The Native Link Authors. All rights reserved.
> ```

nit: 2023-2024
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 81 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8B4fAB7fRE4S-t96b:-Np8B4fAB7fRE4S-t96c:bkpenlz) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L81)):*
> ```XML
>     }
> 
>     pub fn reset_for_retry(&mut self) -> bool {
> ```

nit: Lets break this into a the mutable an immutable parts. I know it's silly, but I will just have better piece of mind someone wont screw this up in the future by having to split up.

ie:

`is_retryable()`
`reset()`
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 115 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8FDGL3PLS3wUg48P7:-Np8FDGL3PLS3wUg48P8:b-sewokc) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L115)):*
> ```XML
>         // Gets the next message, this is either the cached first or a
>         // subsequent message from the wrapped Stream.
>         let maybe_message = if self.message_count == 0 {
> ```

nit: I try to use `maybe_*` when referencing that it's an optional and `*_result`/`*_res` when referencing when it's a `Result`.

Here is a bit backwards.
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 137 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8FutAAxZaM6rTtFLV:-Np8FutAAxZaM6rTtFLW:b-7lhzfv) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L137)):*
> ```XML
>                 if self.message_count == 1 {
>                     // Upon a successful second message, we discard the first.
>                     self.first_msg.take();
> ```

nit: for more clarity:
```rs
self.first_msg = None;
```

(I like to use `.take()` for ownership transfers and assigns when not taking ownership).
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 140 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8Ggo13BkOX91tPMio:-Np8Ggo24OPCQTips0hI:bgclvsq) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L140)):*
> ```XML
>                 }
>                 self.write_finished = message.finish_write;
>                 self.bytes_received += message.data.len();
> ```

I'll trust your judgement on this, but is it going to be an issue with book-keeping this value if a stream is resumed? Specifically around how `first_msg` works (since first_msg` stays around until second message is requested).

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 320 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np86rUe9IUOe4Ohtovo:-NpJT-kw2-iBtGVmtkHj:b-yfgzaj) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L320)):*

nit: Can we instead use `ResourceInfo` to do this parsing?

I'm sure the reason you didn't want to use it before was because it cannot re-combine them (which I think we should implement anyway), secondly, You can't mutate easily, which you can probably change all those `&'a str`into `Cow<'a, str>` instead to get the same effect, but make it easier to mutate, then recombine.

A nice side-effect is that it will preserve things like `compressed-blobs` and `digest_function`.
</blockquote></details>

Yeah, I just copied the original code here, but there's no reason we can't improve it at the same time I suppose.
___
*[`nativelink-store/src/grpc_store.rs` line 349 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np88elI8t_dzyU8whg2:-NpJS4pECSvfrEU7j2p2:b-qa2mfk) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L349)):*

Actually, there's a trick you can use here...

If you want to be super optimized, we could make `shared_state` a raw value, and add a mutable reference to the output `&mut Option<SharedState>`, then make a custom destructor on `StateWrapper` that will move the SharedState to the parent's ref on destruction.

The outer function then just needs to assert it exists.

Technically faster, since it removes both the Arc and the Mutex, but a lot more complicated, so probably not worth the effort.

Plus IIRC, `parking_lot::Mutex` is a spinlock then a real lock or something, so in this use case it's only ever going to be atomic operations.
</blockquote></details>

I played with this for ages and it doesn't appear to be possible since the `client.write` requires that the structure passed to it is owned for `'static` and therefore we can't pass any instance reference to it.  This makes sense since there's no lifetime constraints on the `client.write` function call, so it's impossible to determine that the parameter hasn't been moved to another thread somewhere, hence the need for the `Mutex` and the lifetime could possibly live longer than the current scope, hence the need for the `Arc`.  I don't see any other way around this, even with tricks without `unsafe` code.
___
*[`nativelink-store/src/grpc_store.rs` line 369 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8DQGI9JycsJitFYlM:-NpJarKn14j4dzDU9yDr:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-store/src/grpc_store.rs#L369)):*

(related to above: This should never be `RetryResult::Ok` here.
</blockquote></details>

Done.
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 115 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8FDGL3PLS3wUg48P7:-NpJavvpBTXUxEWc4ICP:b-nwhns1) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L115)):*

nit: I try to use `maybe_*` when referencing that it's an optional and `*_result`/`*_res` when referencing when it's a `Result`.

Here is a bit backwards.
</blockquote></details>

I copied the original naming for the types here, but I'm not at all fixed to them.
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 140 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/638#-Np8Ggo13BkOX91tPMio:-NpJblv_DPXMrwsw-dfZ:bctlox1) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d24cf002af7e8cd21025a6f44444999f8455ea4/nativelink-util/src/write_request_stream_wrapper.rs#L140)):*

I'll trust your judgement on this, but is it going to be an issue with book-keeping this value if a stream is resumed? Specifically around how `first_msg` works (since first_msg` stays around until second message is requested).
</blockquote></details>

It really shouldn't be an issue so long as the invariant between `message_count` and `first_msg` is maintained within `reset`.  This is currently an unspoken invariant, so perhaps I'll document it in the struct.

allada (commented):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-util/src/resource_info.rs` line 88 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpL6s3K5YdMrAQlDgUr:-NpL6s3K5YdMrAQlDgUs:b-7ceqvo) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c7de247e3619b14443c9693651781180659614d/nativelink-util/src/resource_info.rs#L88)):*
> ```XML
>     pub digest_function: Option<&'a str>,
>     pub hash: &'a str,
>     size: &'a str,
> ```

nit: Any reason not to turn `expected_size` back into a string when needed?
___
*[`nativelink-util/src/resource_info.rs` line 144 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpL7lzI-BfZhNcc87BH:-NpL7lzI-BfZhNcc87BI:b-rlb2vf) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c7de247e3619b14443c9693651781180659614d/nativelink-util/src/resource_info.rs#L144)):*
> ```XML
>         [
>             Some(self.instance_name),
>             self.is_upload.then_some("uploads"),
> ```

nit: Instead of using the `ToString` trait, I suggest instead making this an impl with `.to_string(is_upload)`. This way it will allow the same object to be converted from a is_upload and then pass it to a download request or something if needed.
___
*[`nativelink-util/src/resource_info.rs` line 146 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpL799K5nnlZTA9zJ-R:-NpL799K5nnlZTA9zJ-S:b-ytb2l0) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c7de247e3619b14443c9693651781180659614d/nativelink-util/src/resource_info.rs#L146)):*
> ```XML
>             self.is_upload.then_some("uploads"),
>             self.uuid,
>             self.compressed.then_some("compressed-blobs").or(Some("blobs")),
> ```

nit: Any reason we can't use `self.compressor.map_or_else(|| "blobs", |_| "compressed-blobs")`?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-util/src/resource_info.rs` line 88 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpL6s3K5YdMrAQlDgUr:-NpLJnMV8fxxTFDAP-SD:bahjlnp) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c7de247e3619b14443c9693651781180659614d/nativelink-util/src/resource_info.rs#L88)):*

nit: Any reason not to turn `expected_size` back into a string when needed?
</blockquote></details>

It was really just to avoid the extra effort, but it doesn't matter either way really, it's probably not all that expensive.
___
*[`nativelink-util/src/resource_info.rs` line 146 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpL799K5nnlZTA9zJ-R:-NpLJUjLDdKmHJ-62YHv:b8xjhzh) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c7de247e3619b14443c9693651781180659614d/nativelink-util/src/resource_info.rs#L146)):*

nit: Any reason we can't use `self.compressor.map_or_else(|| "blobs", |_| "compressed-blobs")`?
</blockquote></details>

I originally did that, but we have a test that doesn't specify the compression format and I didn't check to see if that was standard or not, I should really.  I wondered if there was a default algorithm if compressed-blobs was used without an algorithm?

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

allada (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 135 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpNgTrT7UbbYdv4p2O-:-NpNgTrT7UbbYdv4p2O0:bo7e8xi) ([raw file](https://github.com/TraceMachina/nativelink/blob/a4259c5d96017d9ca6bab278571bdc86ff027685/nativelink-store/src/grpc_store.rs#L135)):*
> ```XML
>                 Ok(mut resource_name) => {
>                     resource_name.instance_name = &local_state.instance_name;
>                     message.resource_name = resource_name.to_string(true);
> ```

super nit: `.to_string(true /* is_upload */);`

chrisstaite-menlo (commented):
@blakehatch just waiting on your final sign-off here please

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpP5lgPE7qZeNMYU-9c:bt6nxwg)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

adam-singer (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained, and pending CI: macos-13 (waiting on @blakehatch)

chrisstaite-menlo (commented):
Have had another thought and completely changed this again.  I've basically reverted most changes in `WriteRequestStreamWrapper` and instead always buffer the last message in `GrpcStore::write` since `ByteStreamServer::write` can resume.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpTtAs_8yT6F_LXgXqk:bickcqz)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Bazel Dev / ubuntu-22.04, Vercel, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), pre-commit-checks, publish-image, ubuntu-20.04 / stable, ubuntu-22.04 (waiting on @blakehatch)

allada (changes requested):
> Reviewed 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 105 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVO73t9tOB0sS5NmOy:-NpVO73t9tOB0sS5NmOz:bnnzlys) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L105)):*
> ```XML
>     read_stream_error: Option<Error>,
>     read_stream: WriteRequestStreamWrapper<T, E>,
>     // Tonic doesn't appear to report an error until it has taken two messages,
> ```

curiously, is this some tonic buffer size config?

This does make sense though, since it'd fill one buffer up to completion, ship it off, but the previous message is still being processed by the receiver, then tonic gets an error on the 3rd message, so from the down stream component it look this.
___
*[`nativelink-store/src/grpc_store.rs` line 112 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVSiwZ4RVukR8-gb6P:-NpVSiwZ4RVukR8-gb6Q:b-ugln6s) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L112)):*
> ```XML
>     // message_before_last.  Will decrement and play back last_message next and
>     // will then decrement to 0 and start streaming again.
>     resume_count: u32,
> ```

Instead, can we do something like:

```rs
previous_messages: [Option<WriteRequest>, Option<WriteRequest>],
is_replaying: bool,
```

Then below something like:
```rs
if self.is_replaying {
  if self.previous_messages[0].is_some() {
    std::mem::swap(&mut self.previous_messages[0], &mut self.previous_messages[1]);
    return Poll::Ready(self.previous_messages[1].take());
  } else {
    self.is_replaying = false;
  }
}
```

It still is isn't pretty (the elegant way is to use a linked list, but that requires a heal allocation), but at least the book-keeping is much simpler and less "magic".
___
*[`nativelink-store/src/grpc_store.rs` line 189 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVPgcZ1qWCaB-K5io5:-NpVPgcZ1qWCaB-K5io6:bikyhro) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L189)):*
> ```XML
>                     // Cache the last request in case there is an error to allow
>                     // the upload to be resumed.
>                     local_state.message_before_last = local_state.last_message.take();
> ```

nit: Very petty, but does save a couple machine instructions, since it doesn't need to zero out `last_message`:
```rs
std::mem::swap(&mut local_state.message_before_last, &mut local_state.last_message);
```
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 114 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVKgy87T6NrfV-XkmX:-NpVKgy9DfvNQ1jh-wDi:bo0di7i) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-util/src/write_request_stream_wrapper.rs#L114)):*
> ```XML
>                     maybe_message.err_tip(|| format!("Stream error at byte {}", self.bytes_received))
>                 }
>                 Poll::Ready(None) => Err(make_input_err!("Expected WriteRequest struct in stream")),
> ```

more just general question... I wounder if this is actually a problem?

Like, if we don't read until the `None` in the stream I wounder if we can end up with it holding on to data?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 105 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVO73t9tOB0sS5NmOy:-NpVWKNn9pW8ilotCLD4:bq1i5o1) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L105)):*

curiously, is this some tonic buffer size config?

This does make sense though, since it'd fill one buffer up to completion, ship it off, but the previous message is still being processed by the receiver, then tonic gets an error on the 3rd message, so from the down stream component it look this.
</blockquote></details>

Not sure, haven't really looked much into how Tonic works.
___
*[`nativelink-store/src/grpc_store.rs` line 112 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVSiwZ4RVukR8-gb6P:-NpVWCSm8yKfh1QNjQy2:bllhlve) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L112)):*

Instead, can we do something like:

```rs
previous_messages: [Option<WriteRequest>, Option<WriteRequest>],
is_replaying: bool,
```

Then below something like:
```rs
if self.is_replaying {
  if self.previous_messages[0].is_some() {
    std::mem::swap(&mut self.previous_messages[0], &mut self.previous_messages[1]);
    return Poll::Ready(self.previous_messages[1].take());
  } else {
    self.is_replaying = false;
  }
}
```

It still is isn't pretty (the elegant way is to use a linked list, but that requires a heal allocation), but at least the book-keeping is much simpler and less "magic".
</blockquote></details>

That doesn't allow it to retry more than once, which if the connection is down it might take a couple of attempts to reconnect.
___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 114 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVKgy87T6NrfV-XkmX:-NpVWAByCy-yNNaIMgER:b-kk7zhq) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-util/src/write_request_stream_wrapper.rs#L114)):*

more just general question... I wounder if this is actually a problem?

Like, if we don't read until the `None` in the stream I wounder if we can end up with it holding on to data?
</blockquote></details>

It should be dropped though, right?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 189 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVPgcZ1qWCaB-K5io5:-NpVcG9Z3xUSbRkiarZk:b-fm1tgr) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L189)):*

nit: Very petty, but does save a couple machine instructions, since it doesn't need to zero out `last_message`:
```rs
std::mem::swap(&mut local_state.message_before_last, &mut local_state.last_message);
```
</blockquote></details>

Cannot borrow `local_state` as mutable twice...  I'm sure you must be able to do this though?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 189 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVPgcZ1qWCaB-K5io5:-NpVcrmB1mkqKnw7oJ8Q:b-irjf2u) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L189)):*

Cannot borrow `local_state` as mutable twice...  I'm sure you must be able to do this though?
</blockquote></details>

Helper method in the struct impl... that's a bit ugly, but I suppose it's actually better encapsulation.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained, and pending CI: Vercel, asan / ubuntu-22.04, integration-tests (20.04), pre-commit-checks (waiting on @blakehatch)

___
*[`nativelink-store/src/grpc_store.rs` line 112 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVSiwZ4RVukR8-gb6P:-NpZ80mD8JkB3l9rNtUP:b-y7m9ur) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-store/src/grpc_store.rs#L112)):*

That doesn't allow it to retry more than once, which if the connection is down it might take a couple of attempts to reconnect.
</blockquote></details>

I've done the `clone()` upfront when we call `resume()` and that's sorted it.

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained, and pending CI: macos-13 (waiting on @blakehatch)

___
*[`nativelink-util/src/write_request_stream_wrapper.rs` line 114 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/638#-NpVKgy87T6NrfV-XkmX:-Npbg7-i9eOLYPLkC2XH:b6z6yxh) ([raw file](https://github.com/TraceMachina/nativelink/blob/8d2bd643fa3cc22b16d6dd9c4b46fb25a5fb6b67/nativelink-util/src/write_request_stream_wrapper.rs#L114)):*

It should be dropped though, right?
</blockquote></details>

It's probably not a big deal, more of just a question of if we should assume streams are read until end on the happy path.

No action.

allada (approved):
:lgtm: Sorry about the delay, some stuff came up that I had to deal with.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpbgWZp7VUcXUQYeR0x:b-a7funw)* status: 0 of 1 LGTMs obtained, and pending CI: macos-13 (waiting on @blakehatch)

allada (approved):
@blakehatch , waiting on your stamp.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpeliGP5OlFbH6AZ7Nf:bctpggk)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

adam-singer (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: 0 of 1 LGTMs obtained (waiting on @blakehatch)

blakehatch (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpfJYDF5ZuEnfitA2wB:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

blakehatch (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/638#-:-NpfPtoV77Rfl1zNqcsX:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #656, opened 2024-02-03
Max concurrent GrpcStore streams

### LINKED ISSUE #602: GRPC store does not appear to timeout connections
If a connection is lost to a GRPC store it appears the connection is not terminated and restarted. We need to write a test and verify this claim.

chrisstaite-menlo:
# Description

Add a configuration option to limit the number of concurrent streams that are active with the upstream GrpcStore.

With this, we also introduce reconnection on error since Tonic doesn't do that.

Towards #602

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1707,6 +1707,7 @@ dependencies = [
 name = "nativelink-scheduler"
 version = "0.2.0"
 dependencies = [
+ "async-lock",
  "async-trait",
  "blake3",
  "futures",
--- a/nativelink-config/src/schedulers.rs
+++ b/nativelink-config/src/schedulers.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 The Native Link Authors. All rights reserved.
+// Copyright 2023-2024 The Native Link Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -134,6 +134,12 @@ pub struct GrpcScheduler {
     /// Retry configuration to use when a network request fails.
     #[serde(default)]
     pub retry: Retry,
+
+    /// Limit the number of simultaneous upstream requests to this many.  A
+    /// value of zero is treated as unlimited.  If the limit is reached the
+    /// request is queued.
+    #[serde(default)]
+    pub max_concurrent_requests: usize,
 }
 
 #[derive(Deserialize, Debug)]
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 The Native Link Authors. All rights reserved.
+// Copyright 2023-2024 The Native Link Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -546,6 +546,12 @@ pub struct GrpcStore {
     /// Retry configuration to use when a network request fails.
     #[serde(default)]
     pub retry: Retry,
+
+    /// Limit the number of simultaneous upstream requests to this many.  A
+    /// value of zero is treated as unlimited.  If the limit is reached the
+    /// request is queued.
+    #[serde(default)]
+    pub max_concurrent_requests: usize,
 }
 
 /// The possible error codes that might occur on an upstream request.
--- a/nativelink-scheduler/BUILD.bazel
+++ b/nativelink-scheduler/BUILD.bazel
@@ -30,6 +30,7 @@ rust_library(
         "//nativelink-proto",
         "//nativelink-store",
         "//nativelink-util",
+        "@crates//:async-lock",
         "@crates//:blake3",
         "@crates//:futures",
         "@crates//:hashbrown",
--- a/nativelink-scheduler/Cargo.toml
+++ b/nativelink-scheduler/Cargo.toml
@@ -13,6 +13,7 @@ nativelink-proto = { path = "../nativelink-proto" }
 #                    files somewhere else.
 nativelink-store = { path = "../nativelink-store" }
 
+async-lock = "3.2.0"
 async-trait = "0.1.74"
 blake3 = "1.5.0"
 prost = "0.12.3"
--- a/nativelink-scheduler/src/grpc_scheduler.rs
+++ b/nativelink-scheduler/src/grpc_scheduler.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 The Native Link Authors. All rights reserved.
+// Copyright 2023-2024 The Native Link Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ use nativelink_proto::build::bazel::remote::execution::v2::{
 };
 use nativelink_proto::google::longrunning::Operation;
 use nativelink_util::action_messages::{ActionInfo, ActionInfoHashKey, ActionState, DEFAULT_EXECUTION_PRIORITY};
+use nativelink_util::grpc_utils::ConnectionManager;
 use nativelink_util::retry::{Retrier, RetryResult};
 use nativelink_util::tls_utils;
 use parking_lot::Mutex;
@@ -36,17 +37,16 @@ use rand::Rng;
 use tokio::select;
 use tokio::sync::watch;
 use tokio::time::sleep;
-use tonic::{transport, Request, Streaming};
+use tonic::{Request, Streaming};
 use tracing::{error, info, warn};
 
 use crate::action_scheduler::ActionScheduler;
 use crate::platform_property_manager::PlatformPropertyManager;
 
 pub struct GrpcScheduler {
-    capabilities_client: CapabilitiesClient<transport::Channel>,
-    execution_client: ExecutionClient<transport::Channel>,
     platform_property_managers: Mutex<HashMap<String, Arc<PlatformPropertyManager>>>,
     retrier: Retrier,
+    connection_manager: ConnectionManager,
 }
 
 impl GrpcScheduler {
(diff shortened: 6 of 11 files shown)


allada (changes requested):
> Reviewed 6 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmMK_57PR40cMn7zB2:-NpmMK_57PR40cMn7zB3:b-33vb7t) (no related file):*
nit: It doesn't fix the issue, it is `towards` that issue.

___
___
*[`nativelink-config/src/schedulers.rs` line 139 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmKcKa25nIp80m5ZcY:-NpmKcKa25nIp80m5ZcZ:b-cs1p9m) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-config/src/schedulers.rs#L139)):*
> ```XML
> 
>     /// Limit the number of simultaneous upstream requests to this many.  A
>     /// value of zero is treated as unlimited.
> ```

nit: It might be good to document what happens if we exceed the limit (ie: do we backpressure/queue or drop).
___
*[`nativelink-config/src/stores.rs` line 551 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmKk2z5fqxIDJw4QoG:-NpmKk2z5fqxIDJw4QoH:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-config/src/stores.rs#L551)):*
> ```XML
> 
>     /// Limit the number of simultaneous upstream requests to this many.  A
>     /// value of zero is treated as unlimited.
> ```

ditto.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 76 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmLJ9m82TvX-2WqMxc:-NpmLJ9m82TvX-2WqMxd:b-khsjik) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-scheduler/src/grpc_scheduler.rs#L76)):*
> ```XML
> 
>     fn on_error(&self, err: &Error) {
>         if err.code != Code::Internal {
> ```

nit: Maybe `warn!()` the error?
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 76 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmLQAxAe-MirGfw2CA:-NpmLQAxAe-MirGfw2CB:b-fbscn7) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-scheduler/src/grpc_scheduler.rs#L76)):*
> ```XML
> 
>     fn on_error(&self, err: &Error) {
>         if err.code != Code::Internal {
> ```

Should we use some kind of map of codes to know if we should reconnect?
___
*[`nativelink-store/src/grpc_store.rs` line 223 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmM_F6AE6AI4RIp6tY:-NpmM_F6AE6AI4RIp6tZ:b-6lvrgu) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-store/src/grpc_store.rs#L223)):*
> ```XML
> /// This handles the permit for limiting concurrency, and also re-connecting the
> /// underlying channel on error.  It depends on users reporting all errors.
> struct Connection<'a> {
> ```

nit: This looks nearly identical to the one above in `grpc_scheduler`. Can we combine them w/ templates and put them into `grpc_utils` or something?

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-config/src/stores.rs` line 551 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmKk2z5fqxIDJw4QoG:-Npt_lZJE7HNOmrE7oYi:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-config/src/stores.rs#L551)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 76 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmLJ9m82TvX-2WqMxc:-Npt_nQp1xPPxfZLcue8:b-9y49j7) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-scheduler/src/grpc_scheduler.rs#L76)):*

nit: Maybe `warn!()` the error?
</blockquote></details>

That's the purpose of the upstream really.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 76 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NpmLQAxAe-MirGfw2CA:-Npt_ri140jp0OEfFzWH:b-njaurq) ([raw file](https://github.com/TraceMachina/nativelink/blob/02dd17b9f8eca16a65e5628de4e0a6cc2e020110/nativelink-scheduler/src/grpc_scheduler.rs#L76)):*

Should we use some kind of map of codes to know if we should reconnect?
</blockquote></details>

I don't think we need to, all of the transport errors are mapped to `Internal` within Tonic: https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L376

Notably, it states that these indicate an issue with the library: https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L282

allada (changes requested):
> Reviewed 4 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/grpc_utils.rs` line 91 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/656#-NptwbJaBPsgb4EA24oM:-NptwbJaBPsgb4EA24oN:b-jl2kyr) ([raw file](https://github.com/TraceMachina/nativelink/blob/ef4ba39fcacdaf675acb3f753e4d646069ea1adf/nativelink-util/src/grpc_utils.rs#L91)):*
> ```XML
>         // new connection now.
>         let mut channel_lock = self.parent.channel.lock();
>         // Channel already changed, don't do it again.
> ```

nit: Can we elaborate this a bit... it took me a while to understand this...
```
// This ensures only the first error reconnects and the other reconnect
// attempts use the Channel that the first one recreated.
```
___
*[`nativelink-util/src/grpc_utils.rs` line 92 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/656#-Npty7rH5ZUJ8Ux6LEpY:-Npty7rH5ZUJ8Ux6LEpZ:bsvx0eg) ([raw file](https://github.com/TraceMachina/nativelink/blob/ef4ba39fcacdaf675acb3f753e4d646069ea1adf/nativelink-util/src/grpc_utils.rs#L92)):*
> ```XML
>         let mut channel_lock = self.parent.channel.lock();
>         // Channel already changed, don't do it again.
>         if channel_lock.0 != self.channel_id {
> ```

I'm pretty sure you need to do this on all channels after they are updated:
```rs
self.channel_id = channel_lock.0
```

(sadly you'll need another mutex :-( )

Failing to do so will cause the second-reconnect attempt to result in `get_channel` always resulting in a bad channel.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/grpc_utils.rs` line 92 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/656#-Npty7rH5ZUJ8Ux6LEpY:-Nptz443AypWrJChEMSF:b-z3i1u) ([raw file](https://github.com/TraceMachina/nativelink/blob/ef4ba39fcacdaf675acb3f753e4d646069ea1adf/nativelink-util/src/grpc_utils.rs#L92)):*

I'm pretty sure you need to do this on all channels after they are updated:
```rs
self.channel_id = channel_lock.0
```

(sadly you'll need another mutex :-( )

Failing to do so will cause the second-reconnect attempt to result in `get_channel` always resulting in a bad channel.
</blockquote></details>

No, a `Connection` should only be used once.  I really wanted to make this easier by `impl tower::Service<http::HttpBody> for Connection` but that would involve adding a dependency to `tower`.

Essentially, we create a new `Channel` in the parent `ConnectionManager` on failure, but any existing `Connection` that's being used can complete (and perhaps error) by itself.  This is done because in the case of `GoAway` the other established streams on the `Channel` are able to complete, but we need to terminate after that graceful disconnect has finished.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained

___
*[`nativelink-util/src/grpc_utils.rs` line 92 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/656#-Npty7rH5ZUJ8Ux6LEpY:-NpuFRnwAkvWWzdvCQu8:bcitycy) ([raw file](https://github.com/TraceMachina/nativelink/blob/ef4ba39fcacdaf675acb3f753e4d646069ea1adf/nativelink-util/src/grpc_utils.rs#L92)):*

No, a `Connection` should only be used once.  I really wanted to make this easier by `impl tower::Service<http::HttpBody> for Connection` but that would involve adding a dependency to `tower`.

Essentially, we create a new `Channel` in the parent `ConnectionManager` on failure, but any existing `Connection` that's being used can complete (and perhaps error) by itself.  This is done because in the case of `GoAway` the other established streams on the `Channel` are able to complete, but we need to terminate after that graceful disconnect has finished.
</blockquote></details>

Can we add some of this to a comment? This seems like something that we will foot-gun ourselves on in the future if we don't document it.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, Vercel, asan / ubuntu-22.04, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04

___
*[`nativelink-util/src/grpc_utils.rs` line 92 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/656#-Npty7rH5ZUJ8Ux6LEpY:-NpuI10N2oCE1iDEHWtI:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ef4ba39fcacdaf675acb3f753e4d646069ea1adf/nativelink-util/src/grpc_utils.rs#L92)):*

Can we add some of this to a comment? This seems like something that we will foot-gun ourselves on in the future if we don't document it.
</blockquote></details>

Done.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/656)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #661, opened 2024-02-15
Account for block size in filesystem store for eviction purposes

### LINKED ISSUE #595: Filesystem store size should take disk block size into account
Current filesystem store takes only the size of the content into account, not the disk block size. It should compute the apx block size it'll use as a config and use it for size on disk eviction policies.

zbirenbaum:
# Description

Fixes #595 
Should prevent scenarios where fragmented files take up more space than is allowed by `max_bytes` and could therefore exceed the maximum space available

`file_size` has been renamed to `data_size` since the actual file size on the disk after accounting for block size is provided by `get_file_size()` now.

`FileEntry::len` will now return get_file_size() since it is used by the `eviction_map` to determine what to evict based on size

## Type of change

## How Has This Been Tested?

All affected existing tests have been updated to use a block size of 1 so that they do not  have a max_bytes limit smaller than block size (which will hang but wouldn't occur in production)

A new test has been added to ensure that the `get_file_size` calculation returns the correct result for different inputs 

## Checklist

- [x] Tests added/amended
- [x] PR is contained in a single commit, using `git amend`

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -242,6 +242,13 @@ pub struct FilesystemStore {
     /// value will cause items to never be removed from the store causing
     /// infinite memory usage.
     pub eviction_policy: Option<EvictionPolicy>,
+
+    /// The block size of the filesystem for the running machine
+    /// value is used to determine an entry's actual size on disk consumed
+    /// For a 4KB block size filesystem, a 1B file actually consumes 4KB
+    /// Default: 4096
+    #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
+    pub block_size: u64,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -41,6 +41,8 @@ use crate::cas_utils::is_zero_digest;
 
 // Default size to allocate memory of the buffer when reading files.
 const DEFAULT_BUFF_SIZE: usize = 32 * 1024;
+// Default block size of all major filesystems is 4KB
+const DEFAULT_BLOCK_SIZE: u64 = 4 * 1024;
 
 #[derive(Debug)]
 pub struct SharedContext {
@@ -119,17 +121,21 @@ fn to_full_path_from_digest(folder: &str, digest: &DigestInfo) -> OsString {
 #[async_trait]
 pub trait FileEntry: LenEntry + Send + Sync + Debug + 'static {
     /// Responsible for creating the underlying FileEntry.
-    fn create(file_size: u64, encoded_file_path: RwLock<EncodedFilePath>) -> Self;
+    fn create(data_size: u64, block_size: u64, encoded_file_path: RwLock<EncodedFilePath>) -> Self;
 
     /// Creates a (usually) temp file, opens it and returns the path to the temp file.
     async fn make_and_open_file(
+        block_size: u64,
         encoded_file_path: EncodedFilePath,
     ) -> Result<(Self, fs::ResumeableFileSlot<'static>, OsString), Error>
     where
         Self: Sized;
 
-    /// Returns the underlying reference to where the filesize is stored.
-    fn get_file_size(&mut self) -> &mut u64;
+    /// Returns the underlying reference to the size of the data in bytes
+    fn data_size_mut(&mut self) -> &mut u64;
+
+    /// Returns the actual size of the underlying file on the disk after accounting for filesystem block size.
+    fn size_on_disk(&self) -> u64;
 
     /// Gets the underlying EncodedfilePath.
     fn get_encoded_file_path(&self) -> &RwLock<EncodedFilePath>;
@@ -150,7 +156,8 @@ pub trait FileEntry: LenEntry + Send + Sync + Debug + 'static {
 }
 
 pub struct FileEntryImpl {
-    file_size: u64,
+    data_size: u64,
+    block_size: u64,
     encoded_file_path: RwLock<EncodedFilePath>,
 }
 
@@ -162,9 +169,10 @@ impl FileEntryImpl {
 
 #[async_trait]
 impl FileEntry for FileEntryImpl {
-    fn create(file_size: u64, encoded_file_path: RwLock<EncodedFilePath>) -> Self {
+    fn create(data_size: u64, block_size: u64, encoded_file_path: RwLock<EncodedFilePath>) -> Self {
         Self {
-            file_size,
+            data_size,
+            block_size,
             encoded_file_path,
         }
     }
@@ -173,6 +181,7 @@ impl FileEntry for FileEntryImpl {
     /// the cleanup of the file is handled without creating a FileEntry, which would
     /// try to cleanup the file as well during drop().
     async fn make_and_open_file(
+        block_size: u64,
         encoded_file_path: EncodedFilePath,
     ) -> Result<(FileEntryImpl, fs::ResumeableFileSlot<'static>, OsString), Error> {
         let temp_full_path = encoded_file_path.get_file_path();
@@ -192,6 +201,7 @@ impl FileEntry for FileEntryImpl {
         Ok((
             <FileEntryImpl as FileEntry>::create(
                 0, /* Unknown yet, we will fill it in later */
+                block_size,
                 RwLock::new(encoded_file_path),
             ),
             temp_file_result,
@@ -199,8 +209,12 @@ impl FileEntry for FileEntryImpl {
         ))
     }
 
-    fn get_file_size(&mut self) -> &mut u64 {
-        &mut self.file_size
+    fn data_size_mut(&mut self) -> &mut u64 {
+        &mut self.data_size
+    }
+
+    fn size_on_disk(&self) -> u64 {
+        self.data_size.div_ceil(self.block_size) * self.block_size
     }
 
     fn get_encoded_file_path(&self) -> &RwLock<EncodedFilePath> {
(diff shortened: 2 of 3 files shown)


allada (approved):
cc: @chrisstaite-menlo 


Looks great! Only a few nit-picky comments.

Thanks a lot :smile: 

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661#-:-NqfI3FKBonATx7gbqzK:bxyubjg)* status: 1 of 1 LGTMs obtained

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfHzjlBAJUSTDfY7Ot):*
nit: Please expand out the commit here.
___
*[`nativelink-store/src/filesystem_store.rs` line 134 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfFEj2-TS6FqTTtDW7:-NqfFEj2-TS6FqTTtDW8:b-wy01cx) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/src/filesystem_store.rs#L134)):*
> ```XML
>         Self: Sized;
> 
>     /// Returns the underlying reference to where the filesize is stored.
> ```

nit: This comment needs updated.
___
*[`nativelink-store/src/filesystem_store.rs` line 135 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfGF4k-k6iFvIqNWuS:-NqfGF4k-k6iFvIqNWuT:b35i9su) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/src/filesystem_store.rs#L135)):*
> ```XML
> 
>     /// Returns the underlying reference to where the filesize is stored.
>     fn get_data_size(&mut self) -> &mut u64;
> ```

nit: While we are here, can we rename this `.data_size_mut()`? (For consistency I'd also rename `get_file_size()` -> `file_size()`.

This will bring us slightly more inline with rust name convention.
___
*[`nativelink-store/src/filesystem_store.rs` line 137 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfFjoJ6RfmOpxJKzgu:-NqfFjoJ6RfmOpxJKzgv:b-6cmm9f) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/src/filesystem_store.rs#L137)):*
> ```XML
>     fn get_data_size(&mut self) -> &mut u64;
> 
>     /// Returns the actual size of the underlying file on the disk after accounting for filesystem block size
> ```

super nit: End in a period.
___
*[`nativelink-store/src/filesystem_store.rs` line 138 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfFbWlAntM7yImlXzV:-NqfFbWlAntM7yImlXzW:b9kp3x4) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/src/filesystem_store.rs#L138)):*
> ```XML
> 
>     /// Returns the actual size of the underlying file on the disk after accounting for filesystem block size
>     fn get_file_size(&self) -> u64;
> ```

nit: to reduce ambiguity here, I suggest `get_size_on_disk()`. It's probably confusing which one is which the way it's named.
___
*[`nativelink-store/tests/filesystem_store_test.rs` line 1151 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfHKkYA8WFjSqAwfXP:-NqfHKkYA8WFjSqAwfXQ:ba2z8pu) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/tests/filesystem_store_test.rs#L1151)):*
> ```XML
>     // Ensure that get_file_size() returns the correct number
>     // block_size if content length is < block_size
>     // ceil(content length / block_size) * block_size if content length is > block_size
> ```

nit: I think this comment is inverted and a bit confusing.
___
*[`nativelink-store/tests/filesystem_store_test.rs` line 1154 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfHR3U1sveiz9-Shzk:-NqfHR3U1sveiz9-Shzl:b-epsx82) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/tests/filesystem_store_test.rs#L1154)):*
> ```XML
>     #[tokio::test]
>     async fn get_file_size_uses_block_size() -> Result<(), Error> {
>         let short_digest = DigestInfo::try_new(HASH1, VALUE1.len())?;
> ```

nit: Can we just declare a value here with with the size that we are seeking in our test?

I really don't like talking about variable sized items when it's the thing we care about.

Maybe just declare `const VALUE_4K` and `const VALUE_1K` inline here and use them?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661)* status: 1 of 1 LGTMs obtained

zbirenbaum:
> _[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfHzjlBAJUSTDfY7Ot):_ nit: Please expand out the commit here.

I'm not sure how to do this, could you clarify what you mean by expand out the commit? I've been amending the main commit since that's what the guidelines recommended



> _[`nativelink-store/src/filesystem_store.rs` line 138 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfFbWlAntM7yImlXzV:-NqfFbWlAntM7yImlXzW:b9kp3x4) ([raw file](https://github.com/TraceMachina/nativelink/blob/c190eaa61665d72df584231caabde373c73c6b35/nativelink-store/src/filesystem_store.rs#L138)):_
> 
> > ```
> >     /// Returns the actual size of the underlying file on the disk after accounting for filesystem block size
> >     fn get_file_size(&self) -> u64;
> > ```
> 
> nit: to reduce ambiguity here, I suggest `get_size_on_disk()`. It's probably confusing which one is which the way it's named.
> 

Since another recommendation was to just make it `file_size` but I think this is clearer, I changed it to just `size_on_disk`

> > ```
> >     // Ensure that get_file_size() returns the correct number
> >     // block_size if content length is < block_size
> >     // ceil(content length / block_size) * block_size if content length is > block_size
> > ```
> 
> nit: I think this comment is inverted and a bit confusing.
> 

I think the comment is correct. If block size is 4KB and you create a file with 1B of contents, it will consume 4KB on disk. If you create a file with 4KB + 1B contents it will consume 8KB. I tested this on my laptop to confirm. Would including an example help?

allada (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661)* status: 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqjnQZH94Bi1Ef2zSgh:-NqjnQZH94Bi1Ef2zSgi:b-26cssz) (no related file):*
No action required, but I wanted to mention:

Ext4 filesystems don't actually use `ceil(data_size / block_size) * block_size` ... They will in fact inline data into the inode block if the data can fit (from what I remember reading)... This works similar to how short-string optimization works.

Again, no action, but I think it's important to know that our calculations are very "rough" and more of a "worst-case".

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/661#-NqfHzjlBAJUSTDfY7Ot):*

nit: Please expand out the commit here.
</blockquote></details>

Yeah, could you put in a short paragraph explaining what was done?

We have tools we use internally to roll this up into the release notes, and this is definitely something we want to include in the notes.

Format is as follows:
```
[Title]

[Short paragraph goes here]
```

I think we actually changed some stuff recently, so the commit message may not be the appropriate place... Can you do it here in the commit and in the github description? We require squash merge now due to the way github signs commits (we are still fiddling with some of the subtle details).
___
*[`nativelink-store/tests/filesystem_store_test.rs` line 1151 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/661#-Nqjm_q61E1q_314WOMH:-Nqjm_q61E1q_314WOMI:b-p4g4df) ([raw file](https://github.com/TraceMachina/nativelink/blob/429acb59f25d1786bf6d1b9b3154f438fbc489bd/nativelink-store/tests/filesystem_store_test.rs#L1151)):*
> ```XML
>     // Ensure that get_file_size() returns the correct number
>     // ceil(content length / block_size) * block_size
>     // Content length is 1B and block_size is 4B: size on disk = 4B
> ```

nit: Lets not talk in bytes for blocksize and just always talk in `4k` chunks. I understand what you are saying in the comment, but I had to re-read it like 3 times to realize the block size was a non-standard size.

Maybe elude to something more like:
```
(Assume block size of 4k)
1B data size = 4K size on disk
5K data size = 8K size on disk
```

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661)* status: 1 of 1 LGTMs obtained

zbirenbaum:
Sure, sorry about the confusion in the test case comment I updated it and the test case to use the default block size and generate a string of 1K and 5K length for the contents.

I added some extra information about the changes made in the commit description as well as how to restore the previous behavior

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/661#-:-Nqo-17y2qKkMXevy0jp:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #721, opened 2024-03-05
Update LICENSE

MarcusSorealheis:
barely

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/721)


MarcusSorealheis:
the precommit message here was very weird. It did not specify any error. We need to figure out why that is happening but now time to dig into it now. I'll use this branch to discuss precommit craziness more in the future.

PR #736, opened 2024-03-08
Use MiMalloc as global memory allocator

aaronmondal:
This should make allocation behavior more consistent across different builds and prevent memory fragmentation issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/736)

PR #738, opened 2024-03-08
Use jemalloc on non-windows to reduce memory fragmentation

### LINKED ISSUE #621: Memory store causes memory fragmentation
Here's some stats on real-world usage of Nativelink:
image

Since rust is not a garbage collected language it is not allowed to pack memory, so we need to deal with memory fragmentation ourselves. Normally programs don't need to think much about this but in this case we do.

### Details on how memory works:
Linux allocates memory in pages, often 4k. The memory manager will pack the memory for you by slotting the returned points from malloc usually from a pool of already allocated pages. This allows code to request (lets say) 100 bytes of data and that 100 bytes may share a page with other pointers in memory.

The down side is that small allocations that live for a long time might end up keeping the entire page (4k) allocated by the kernel. It is rare for an allocator to recycle memory pages that were full at one point; this means if a page was ever full, it will release the page back to the kernel once no data is allocated in the page instead of keep trying to pack new allocations into it.

### The problem
Looking at the graph above, you can see that roughly 50% of all of the CAS is ~128bytes or less. This means that the vast majority of items are super tiny and only use a little bit of memory. Since memory store frequently evicts items that have not been used recently from the store, it means that it's very common for an item to be put into it and shortly after to be evicted because the object is rarely used.

So, if a 4k page was allocated, then 32 x 128byte objects were loaded into it then an hour later 31 of the items were evicted, but 1 item is used very frequently, it means that we are now using 4k of memory for 128 bytes of data. Multiply this out and you can see how inefficient this can become.

### Possible solutions
1. A very simple solution to this problem is to simply have a timer or trigger that runs every once in a while and copies some of the data from the top of the LRU-Map (least recently used map) to a new place in memory then frees the old pointer. This is inefficient, but is extremely simple and will have very negligible impact on performance.
2. We could try to find a smarter Allocator that can handle this case better. This would be great if there's an allocator we can ask how compact our memory is or how packed our memory page is from a given pointer.
3. Similar to 1, We could store references to the data in a sorted order by age of insertion in addition to age of access (LRU), then periodically make full copies of data that has lived a long time in the LRU.

### Some things to keep in mind
Lots of items are very small, if 50% are 128bytes and a sha256/blake3 is 40 byte, this means just copying the hash + a pointer is about +37% for the lower 50%.

allada:
To help reduce/remove memory fragmentation caused by memory store we are switching our allocator to jemalloc. This should make our memory compaction significantly more efficient.

closes #621

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -38,7 +38,12 @@ rust_binary(
         "@crates//:tower",
         "@crates//:tracing",
         "@crates//:tracing-subscriber",
-    ],
+    ] + select({
+        "@platforms//os:windows": [],
+        "//conditions:default": [
+            "@crates//:tikv-jemallocator",
+        ],
+    }),
 )
 
 genrule(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1682,6 +1682,7 @@ dependencies = [
  "rustls-pemfile 2.1.1",
  "scopeguard",
  "serde_json5",
+ "tikv-jemallocator",
  "tokio",
  "tokio-rustls 0.25.0",
  "tonic 0.11.0",
@@ -2862,6 +2863,26 @@ checksum = "8b9ef9bad013ada3808854ceac7b46812a6465ba368859a37e2100283d2d719c"
 dependencies = [
  "cfg-if",
  "once_cell",
+]
+
+[[package]]
+name = "tikv-jemalloc-sys"
+version = "0.5.4+5.3.0-patched"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9402443cb8fd499b6f327e40565234ff34dbda27460c5b47db0db77443dd85d1"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
+name = "tikv-jemallocator"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "965fe0c26be5c56c94e38ba547249074803efd52adfb66de62107d95aab3eaca"
+dependencies = [
+ "libc",
+ "tikv-jemalloc-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,3 +52,6 @@ tonic = { version = "0.11.0", features = ["gzip", "tls"] }
 tower = "0.4.13"
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
+tikv-jemallocator = "0.5.4"
--- a/src/bin/nativelink.rs
+++ b/src/bin/nativelink.rs
@@ -66,6 +66,13 @@ use tower::util::ServiceExt;
 use tracing::{error, warn};
 use tracing_subscriber::filter::{EnvFilter, LevelFilter};
 
+// We use jemalloc because our program is frequently long-lived and memory fragmentation
+// happens often with memory store. Jemalloc nearly completely removes memory fragmentation
+// issues.
+#[cfg(not(target_env = "msvc"))]
+#[global_allocator]
+static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
 /// Note: This must be kept in sync with the documentation in `PrometheusConfig::path`.
 const DEFAULT_PROMETHEUS_METRICS_PATH: &str = "/metrics";
 


allada (commented):
+@aaronmondal 

fyi: @chrisstaite-menlo 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738#-:-NsUckNX6M5tTJLRsIXG:b9jwnn1)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, publish-image, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04 (waiting on @aaronmondal)

MarcusSorealheis (commented):
It's great that we have solved this here and I think it's a fantastic ongoing experiment for us. The `memory_store`will continue to surface bleeding edge issues. 

Now, more than ever I think we need to land the redis_store (not actually vendor specific) functionality. cc @blakehatch and @zbirenbaum 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738#-:-NsUfsnP5Na5AlDsrV17:b-c9rz5j)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, publish-image, ubuntu-20.04 / stable, ubuntu-22.04 / stable, windows-2022 / stable (waiting on @aaronmondal)

blakehatch:
> It's great that we have solved this here and I think it's a fantastic ongoing experiment for us. The `memory_store`will continue to surface bleeding edge issues.
> 
> Now, more than ever I think we need to land the redis_store (not actually vendor specific) functionality. cc @blakehatch and @zbirenbaum
> 
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738#-:-NsUfsnP5Na5AlDsrV17:b-c9rz5j)_ status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, publish-image, ubuntu-20.04 / stable, ubuntu-22.04 / stable, windows-2022 / stable (waiting on @aaronmondal)

Agreed I'll be dusting off the old redis store PR soon with a fresh rebase and clean up how the mocking is done. Will need to set up some integration tests with @aaronmondal and would love @zbirenbaum's help as well!

adam-singer (approved):
:lgtm:

We should try and collect some previous metrics on memory consumption and deploy this collecting same metrics to see a difference. jemalloc does allow for easier profiling and is a battle harden allocator, ymmv depending on what this is to solve. Longer term we should probably have some like https://www.polarsignals.com/blog/posts/2023/12/20/rust-memory-profiling integrated to pull a profile while system is running. 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738#-:-NsUncMK1uYEjMbm49RU:bwgo5y4)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), publish-image (waiting on @aaronmondal)

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), publish-image (waiting on @aaronmondal)

___
*[`Cargo.toml` line 57 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/738#-NsVXbtu-IoH2zqCIoYa:-NsVXbtu-IoH2zqCIoYb:b5eczdy) ([raw file](https://github.com/TraceMachina/nativelink/blob/82b5ad37db455daace7f717e8d5c5dc78ec2c5c0/Cargo.toml#L57)):*
> ```TOML
> 
> [target.'cfg(not(target_env = "msvc"))'.dependencies]
> tikv-jemallocator = "0.5.4"
> ```

It looks like we'd have to patch (or rewrite) the buildscript for the crate. The previous attempt to use MiMalloc #736 passed tests etc but didn't solve the defragmentation issue.

If I understand https://github.com/microsoft/mimalloc/issues/632 correctly it seems like it's possible to pull the defragmentation functionality from jemalloc into mimalloc with a patch like this: https://github.com/romange/helio/tree/master/patches

Maybe it's worth a shot to attempt using such a patched version of `mimalloc` rather than trying to fix the buildscript for `tikv-jemallocator`?

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/738)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), publish-image (waiting on @aaronmondal)

___
*[`Cargo.toml` line 57 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/738#-NsVXbtu-IoH2zqCIoYa:-NsVblLv-rnjjB0c-ujB:b-h42s6l) ([raw file](https://github.com/TraceMachina/nativelink/blob/82b5ad37db455daace7f717e8d5c5dc78ec2c5c0/Cargo.toml#L57)):*

It looks like we'd have to patch (or rewrite) the buildscript for the crate. The previous attempt to use MiMalloc #736 passed tests etc but didn't solve the defragmentation issue.

If I understand https://github.com/microsoft/mimalloc/issues/632 correctly it seems like it's possible to pull the defragmentation functionality from jemalloc into mimalloc with a patch like this: https://github.com/romange/helio/tree/master/patches

Maybe it's worth a shot to attempt using such a patched version of `mimalloc` rather than trying to fix the buildscript for `tikv-jemallocator`?
</blockquote></details>

Hmmm, maybe we should just make the copy in a different allocator then :-( .

PR #749, opened 2024-03-10
Use mimalloc as global memory allocator

aaronmondal:
This should make allocation behavior more consistent across different builds and reduce memory fragmentation issues.

The allocation behavior of the `nativelink` executable is now configurable with `MIMALLOC_*` environment variables as described here:

https://microsoft.github.io/mimalloc/environment.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -27,6 +27,7 @@ rust_binary(
         "@crates//:console-subscriber",
         "@crates//:futures",
         "@crates//:hyper",
+        "@crates//:mimalloc",
         "@crates//:parking_lot",
         "@crates//:prometheus-client",
         "@crates//:rustls-pemfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.10"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b79b82693f705137f8fb9b37871d99e4f9a7df12b917eed79c3d3954830a60b"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
  "once_cell",
@@ -192,7 +192,7 @@ dependencies = [
  "bytes",
  "fastrand",
  "hex",
- "http 0.2.11",
+ "http 0.2.12",
  "hyper",
  "ring",
  "time",
@@ -229,7 +229,7 @@ dependencies = [
  "aws-types",
  "bytes",
  "fastrand",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "percent-encoding",
  "pin-project-lite",
@@ -257,7 +257,7 @@ dependencies = [
  "aws-smithy-xml",
  "aws-types",
  "bytes",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "once_cell",
  "percent-encoding",
@@ -282,7 +282,7 @@ dependencies = [
  "aws-smithy-types",
  "aws-types",
  "bytes",
- "http 0.2.11",
+ "http 0.2.12",
  "once_cell",
  "regex-lite",
  "tracing",
@@ -304,7 +304,7 @@ dependencies = [
  "aws-smithy-types",
  "aws-types",
  "bytes",
- "http 0.2.11",
+ "http 0.2.12",
  "once_cell",
  "regex-lite",
  "tracing",
@@ -327,7 +327,7 @@ dependencies = [
  "aws-smithy-types",
  "aws-smithy-xml",
  "aws-types",
- "http 0.2.11",
+ "http 0.2.12",
  "once_cell",
  "regex-lite",
  "tracing",
@@ -349,8 +349,8 @@ dependencies = [
  "form_urlencoded",
  "hex",
  "hmac",
- "http 0.2.11",
- "http 1.0.0",
+ "http 0.2.12",
+ "http 1.1.0",
  "once_cell",
  "p256",
  "percent-encoding",
@@ -385,7 +385,7 @@ dependencies = [
  "crc32c",
  "crc32fast",
  "hex",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "md-5",
  "pin-project-lite",
@@ -417,7 +417,7 @@ dependencies = [
  "bytes",
  "bytes-utils",
  "futures-core",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "once_cell",
  "percent-encoding",
@@ -443,7 +443,7 @@ checksum = "882b95acc2ed531309fad4b6bc87486383bb89f644a4243b377a00895c701f00"
 dependencies = [
  "assert-json-diff",
  "aws-smithy-runtime-api",
- "http 0.2.11",
+ "http 0.2.12",
  "pretty_assertions",
  "regex-lite",
  "roxmltree",
@@ -475,7 +475,7 @@ dependencies = [
  "bytes",
  "fastrand",
  "h2",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "hyper",
  "hyper-rustls",
@@ -499,8 +499,8 @@ dependencies = [
  "aws-smithy-async",
  "aws-smithy-types",
  "bytes",
- "http 0.2.11",
- "http 1.0.0",
+ "http 0.2.12",
+ "http 1.1.0",
  "pin-project-lite",
  "tokio",
  "tracing",
@@ -517,7 +517,7 @@ dependencies = [
  "bytes",
  "bytes-utils",
  "futures-core",
- "http 0.2.11",
+ "http 0.2.12",
  "http-body",
  "itoa",
  "num-integer",
@@ -549,7 +549,7 @@ dependencies = [
  "aws-smithy-async",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
- "http 0.2.11",
+ "http 0.2.12",
  "rustc_version",
  "tracing",
 ]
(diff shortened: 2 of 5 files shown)


aaronmondal (commented):
+@allada +@adam-singer +@MarcusSorealheis +@zbirenbaum 

@allada and I talked at length about how to fix the fragmentation issues that @chrisstaite-menlo brought up. Initially we attempted to fix this with`jemalloc` in #738, but the additional build requirements associated with it (notably requiring `make` and `autotools`) made this infeasible.

As future improvements, there is an [additional patch](https://github.com/romange/helio/blob/master/patches/mimalloc-v2.0.9.patch) that we could add to potentially improve fragmentation behavior, and we can also go down the dragonflydb route of implementing active defragmentation directly against `mimalloc`'s API similar to this: https://github.com/dragonflydb/dragonfly/pull/523

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NseaIs99-d_1Ax6ldKk:b-ffdb8n)* status: 0 of 4 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, publish-image, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04 (waiting on @adam-singer, @allada, @MarcusSorealheis, and @zbirenbaum)

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-Nsftca7EvdFrwJj_yxn:bnfp4nl)* status: 1 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @adam-singer, @MarcusSorealheis, and @zbirenbaum)

___
*[`flake.nix` line 43 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/749#-NsftS_a7EODEBmNoNu_:-NsftS_a7EODEBmNoNua:biwbdqu) ([raw file](https://github.com/TraceMachina/nativelink/blob/a1b460db6e42c50d4ade3449363ef5d05e3111c1/flake.nix#L43)):*
> ```Nix
>         ...
>       }: let
>         stable-rust = pkgs.pkgsMusl.rust-bin.stable."1.76.0";
> ```

that's it?!?!?! This is all it took!?!!

I'm so confused on nix :stuck_out_tongue_winking_eye:

chrisstaite-menlo (commented):
Wow... Nice and easy swap.  Want me to run this and see how my memory performs?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NsgqflG5DMXoNpcZ3kl:b-3b0npd)* status: 1 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @adam-singer, @MarcusSorealheis, and @zbirenbaum)

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NsiF5Ws8ZCsN7IvdA1b:bnfp4nl)* status: 2 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @MarcusSorealheis and @zbirenbaum)

chrisstaite-menlo (commented):
My current build uses the Docker container with Bazel: https://github.com/TraceMachina/nativelink/blob/main/deployment-examples/docker-compose/Dockerfile is there an easy way to adapt that use use the build as you suggest?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NsiSB_KEqVlJl4qTIW6:b-x6ip9o)* status: 2 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @MarcusSorealheis and @zbirenbaum)

aaronmondal (commented):
@chrisstaite-menlo The build via the Dockerfile should work as well.

To give a bit more insight:

- The Bazel build uses whatever `libc` it finds on the system. Since the Dockerfile is based on an Ubuntu base image it'll default to linking `glibc` and its built-in `malloc`. This means you're currently using `glibc`'s implementation. After testing we found that this allocator doesn't free memory the way we want which causes the issue you described.
- At the moment the same is true for the "regular" Cargo build. Some CI workflows are currently failing because I'm trying to completely switch that to a static musl build to bring it closer to what the nix build does.
- The nix build already runs Cargo with a special configuration to statically link musl. The container image built from that is a minimal image with an overall closure size of ~28MB. The images we publish from that only contain that closure, i.e. a static nativelink executable and some certificates. In comparison, the Dockerfile-based image is ~4x larger at ~120MB.
- The nix/`musl` build doesn't show the same fragmentation behavior that we see with `glibc`. That is, *technically* it runs more stable and stays within expected memory bounds. However with `musl`'s default allocator this comes at the cost of a fairly significant slowdown (~2x IIRC).

In conclusion:

- You can test this with the Dockerfile as before. What you'll get is `nativelink` linked against Ubuntu's `glibc` but with `mimalloc` as allocator rather than `glibc`'s default allocator.
- The nix build would be nativelink+musl+mimalloc. I'd expect it to now behave fairly similar in performance to the nativelink+glibc+mimalloc builds.

For future work I'll try to get the Bazel build up to parity with the nix/cargo/musl build so that we have less variation and better portability between the builds.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NsiXt8I7xFQ6DbA6hTG:bl0hgg3)* status: 2 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @MarcusSorealheis and @zbirenbaum)

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749)* status: 2 of 4 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04 (waiting on @MarcusSorealheis and @zbirenbaum)

___
*[`flake.nix` line 43 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/749#-NsftS_a7EODEBmNoNu_:-Nsig-GR9X16ceCGF7RW:b-otnqit) ([raw file](https://github.com/TraceMachina/nativelink/blob/a1b460db6e42c50d4ade3449363ef5d05e3111c1/flake.nix#L43)):*

that's it?!?!?! This is all it took!?!!

I'm so confused on nix :stuck_out_tongue_winking_eye:
</blockquote></details>

There is a fair bit of nontrivial magic going on here. What we're doing is:

1. Take nixpkgs
2. Extend it with the rust-overlay
3. nixpkgs has cross-compilation toolchains built-in. That is, you can use things like `pkgs.pkgsStatic` to get a fully static nixpkgs, `pkgsMusl` for musl-based stuff and a large number of `pkgsCross.xxx` toolchains for crosscompilation: https://nix.dev/tutorials/cross-compilation.html
4. Take that cross-transitioned variant of the added rust toolchain
5. Wrap it in a specialized builder (`crane`) which invokes cargo and executes the build.
6. Re-export the cargo toolchain in the local devShell for development. This is still slighty broken in CI because even on musl-based systems some rust tools (e.g. rustdoc and clippy) are dynamically linked against `libgcc_s`. Shouldn't be too hard to work around though.

The original breakage with missing symbols came from the rust toolchains using a `glibc`-configured linker. The `pkgsMusl` change swaps that to a linker that uses Musl instead.

Technically, we could use `pkgsStatic` everywhere and things would almost always be guaranteed to be perfectly portable. The reason we're not doing this (yet) is because the static packageset isn't prebuilt by nixpkgs. So if we used that every user would have to wait for like 10 years for their local systems to build static GCC and LLVM toolchains lol

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-Nsju3Y7-AzYq3WQnsvH:bnfp4nl)* status: 2 of 4 LGTMs obtained (waiting on @MarcusSorealheis and @zbirenbaum)

aaronmondal (commented):
-@MarcusSorealheis -@zbirenbaum 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749#-:-NsjxI0X1sWza77X688x:b-6iup8o)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained

chrisstaite-menlo:
I've had this running from the docker-compose for 16 hours in prod now and memory usage is at 5GB and holding.  Hopefully sorted things out, will check again in a day or two.  CPU usage appears to be significantly reduced also which is weird, but may be just my sampling.

MarcusSorealheis:
Wow, that is fantastic news @chrisstaite-menlo

What operating system/version so that we can check another.

chrisstaite-menlo:
That's Ubuntu 20.04 with 4.15 kernel.  I'm due to upgrade all of the build machines to 22.04 soon, so will hopefully be able to test there also.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/749)* status: 2 of 2 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/749#-NtAU0JfD3DjvJnXu7BE:-NtAU0Jg4pwQG_E1mnBi:b36it7h) (no related file):*
@chrisstaite-menlo This is good, but not as good as I was hoping. When you have time, could you run nativelink with:
```
MIMALLOC_PAGE_RESET=1
```

I believe this is the flag that does most of the fragmentation reduction that we'll be setting, but we need to do more testing. As I mentioned before, we still need to tune mimalloc.

___

PR #778, opened 2024-03-19
Remove old actions with no listeners

### LINKED ISSUE #338: Actions with no listeners that are old should be removed from queue
If an action is old (set in config) and has no more listeners it should just be evicted from the queue, since no one is waiting for the action.

see: https://github.com/TraceMachina/turbo-cache/blob/f15146d0cf94e1a9159f14db775efe4a3e27355d/cas/scheduler/simple_scheduler.rs#L432

We should implement this using timeouts and upon timing out signal to the worker to kill the action.

zbirenbaum:
# Description

Implement scheduler side removal of actions with no listeners. Adds disconnect_timeout_s configuration field with default of 60s. If the client waiting on a given action is disconnected for longer than this duration without reconnecting the scheduler will stop tracking it. This does not remove it from the worker if the job has already been dispatched.

Fixes #338 

## Type of change

## How Has This Been Tested?

A test has been added to ensure that requested actions are removed if a client is found to be disconnected for `disconnect_timeout_s`

## Checklist
- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1803,6 +1803,7 @@ dependencies = [
  "prost",
  "prost-types",
  "rand",
+ "scopeguard",
  "tokio",
  "tokio-stream",
  "tonic 0.11.0",
--- a/nativelink-config/src/schedulers.rs
+++ b/nativelink-config/src/schedulers.rs
@@ -119,6 +119,12 @@ pub struct SimpleScheduler {
     /// The strategy used to assign workers jobs.
     #[serde(default)]
     pub allocation_strategy: WorkerAllocationStrategy,
+
+    /// Remove action from queue after this much time has elapsed without a listener
+    /// amount of time in seconds.
+    /// Default: 60 (seconds)
+    #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
+    pub disconnect_timeout_s: u64,
 }
 
 /// A scheduler that simply forwards requests to an upstream scheduler.  This
--- a/nativelink-scheduler/src/action_scheduler.rs
+++ b/nativelink-scheduler/src/action_scheduler.rs
@@ -47,6 +47,9 @@ pub trait ActionScheduler: Sync + Send + Unpin {
     /// Cleans up the cache of recently completed actions.
     async fn clean_recently_completed_actions(&self);
 
+    /// Inform the scheduler a client has disconnected
+    fn notify_client_disconnected(&self, unique_qualifier: &ActionInfoHashKey);
+
     /// Register the metrics for the action scheduler.
     fn register_metrics(self: Arc<Self>, _registry: &mut Registry) {}
 }
--- a/nativelink-scheduler/src/cache_lookup_scheduler.rs
+++ b/nativelink-scheduler/src/cache_lookup_scheduler.rs
@@ -151,6 +151,8 @@ impl CacheLookupScheduler {
 
 #[async_trait]
 impl ActionScheduler for CacheLookupScheduler {
+    fn notify_client_disconnected(&self, _unique_qualifier: &ActionInfoHashKey) {}
+
     async fn get_platform_property_manager(
         &self,
         instance_name: &str,
--- a/nativelink-scheduler/src/grpc_scheduler.rs
+++ b/nativelink-scheduler/src/grpc_scheduler.rs
@@ -152,6 +152,8 @@ impl GrpcScheduler {
 
 #[async_trait]
 impl ActionScheduler for GrpcScheduler {
+    fn notify_client_disconnected(&self, _unique_qualifier: &ActionInfoHashKey) {}
+
     async fn get_platform_property_manager(
         &self,
         instance_name: &str,
--- a/nativelink-scheduler/src/property_modifier_scheduler.rs
+++ b/nativelink-scheduler/src/property_modifier_scheduler.rs
@@ -47,6 +47,8 @@ impl PropertyModifierScheduler {
 
 #[async_trait]
 impl ActionScheduler for PropertyModifierScheduler {
+    fn notify_client_disconnected(&self, _unique_qualifier: &ActionInfoHashKey) {}
+
     async fn get_platform_property_manager(
         &self,
         instance_name: &str,
--- a/nativelink-scheduler/src/simple_scheduler.rs
+++ b/nativelink-scheduler/src/simple_scheduler.rs
@@ -18,10 +18,10 @@ use std::collections::BTreeMap;
 use std::hash::{Hash, Hasher};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::{Instant, SystemTime};
+use std::time::Instant;
 
 use async_trait::async_trait;
-use futures::Future;
+use futures::future::{BoxFuture, Future};
 use hashbrown::{HashMap, HashSet};
 use lru::LruCache;
 use nativelink_config::schedulers::WorkerAllocationStrategy;
@@ -38,7 +38,7 @@ use parking_lot::{Mutex, MutexGuard};
 use tokio::sync::{watch, Notify};
 use tokio::task::JoinHandle;
 use tokio::time::Duration;
-use tracing::{error, warn};
+use tracing::{error, info, warn};
 
 use crate::action_scheduler::ActionScheduler;
 use crate::platform_property_manager::PlatformPropertyManager;
@@ -56,6 +56,10 @@ const DEFAULT_RETAIN_COMPLETED_FOR_S: u64 = 60;
 /// Default times a job can retry before failing.
 /// If this changes, remember to change the documentation in the config.
 const DEFAULT_MAX_JOB_RETRIES: usize = 3;
+
+/// Default timeout for actions without any listeners.
+/// If this changes, remember to change the documentation in the config.
+const DEFAULT_DISCONNECT_TIMEOUT_S: u64 = 60;
 
 /// An action that is being awaited on and last known state.
 struct AwaitedAction {
@@ -173,7 +177,7 @@ impl Workers {
 }
 
 struct CompletedAction {
-    completed_time: SystemTime,
+    completed_instant: Instant,
     state: Arc<ActionState>,
 }
 
@@ -198,6 +202,41 @@ impl Borrow<ActionInfoHashKey> for CompletedAction {
     }
 }
 
+type NowFn = fn() -> Instant;
+type SleepFn = fn(Duration) -> BoxFuture<'static, ()>;
+
+/// Functions that may be injected for testing purposes, during standard control
+/// flows these are specified by the new function.
+pub struct Callbacks {
+    /// A function that gets the current time.
+    now_fn: NowFn,
+
+    /// A function that sleeps for a given Duration.
+    sleep_fn: SleepFn,
+}
+
+impl Callbacks {
+    pub fn new(now_fn: NowFn, sleep_fn: SleepFn) -> Self {
+        Self { now_fn, sleep_fn }
+    }
+
+    fn now(&self) -> Instant {
+        (self.now_fn)()
+    }
+
+    fn sleep(&self, duration: Duration) -> impl Future<Output = ()> {
+        (self.sleep_fn)(duration)
+    }
+}
+
+impl Default for Callbacks {
+    fn default() -> Self {
+        Callbacks {
+            now_fn: Instant::now,
+            sleep_fn: |duration| Box::pin(tokio::time::sleep(duration)),
+        }
+    }
+}
 struct SimpleSchedulerImpl {
     // BTreeMap uses `cmp` to do it's comparisons, this is a problem because we want to sort our
     // actions based on priority and insert timestamp but also want to find and join new actions
@@ -227,6 +266,9 @@ struct SimpleSchedulerImpl {
     /// Notify task<->worker matching engine that work needs to be done.
     tasks_or_workers_change_notify: Arc<Notify>,
     metrics: Arc<Metrics>,
+    /// How long the server will wait for a client to reconnect before removing the action from the queue.
+    disconnect_timeout_s: u64,
+    callbacks: Arc<Callbacks>,
 }
 
 impl SimpleSchedulerImpl {
@@ -313,11 +355,13 @@ impl SimpleSchedulerImpl {
     }
 
     fn clean_recently_completed_actions(&mut self) {
-        let expiry_time = SystemTime::now()
+        let expiry_time = self
+            .callbacks
+            .now()
             .checked_sub(self.retain_completed_for)
             .unwrap();
         self.recently_completed_actions
-            .retain(|action| action.completed_time > expiry_time);
+            .retain(|action| action.completed_instant > expiry_time);
     }
 
     fn find_recently_completed_action(
@@ -422,10 +466,21 @@ impl SimpleSchedulerImpl {
         Ok(())
     }
 
+    fn get_queued_action(&self, unique_qualifier: &ActionInfoHashKey) -> Option<&AwaitedAction> {
+        self.queued_actions_set
+            .get(unique_qualifier)
+            .and_then(|action_info| self.queued_actions.get(action_info))
+    }
+
+    fn get_active_action(&self, unique_qualifier: &ActionInfoHashKey) -> Option<&AwaitedAction> {
+        self.active_actions.get(unique_qualifier)
+    }
+
     // TODO(blaise.bruer) This is an O(n*m) (aka n^2) algorithm. In theory we can create a map
     // of capabilities of each worker and then try and match the actions to the worker using
     // the map lookup (ie. map reduce).
     fn do_try_match(&mut self) {
+        println!("do_try_match did run");
         // TODO(blaise.bruer) This is a bit difficult because of how rust's borrow checker gets in
         // the way. We need to conditionally remove items from the `queued_action`. Rust is working
         // to add `drain_filter`, which would in theory solve this problem, but because we need
@@ -619,7 +674,7 @@ impl SimpleSchedulerImpl {
 
         // Keep in case this is asked for soon.
         self.recently_completed_actions.insert(CompletedAction {
-            completed_time: SystemTime::now(),
+            completed_instant: self.callbacks.now(),
             state: running_action.current_state,
         });
 
@@ -647,7 +702,7 @@ impl SimpleScheduler {
     #[inline]
     #[must_use]
     pub fn new(scheduler_cfg: &nativelink_config::schedulers::SimpleScheduler) -> Self {
-        Self::new_with_callback(scheduler_cfg, || {
+        Self::new_with_callback(scheduler_cfg, Callbacks::default(), || {
             // The cost of running `do_try_match()` is very high, but constant
             // in relation to the number of changes that have happened. This means
             // that grabbing this lock to process `do_try_match()` should always
@@ -664,6 +719,7 @@ impl SimpleScheduler {
         F: Fn() -> Fut + Send + Sync + 'static,
     >(
         scheduler_cfg: &nativelink_config::schedulers::SimpleScheduler,
+        callbacks: Callbacks,
         on_matching_engine_run: F,
     ) -> Self {
         let platform_property_manager = Arc::new(PlatformPropertyManager::new(
@@ -688,6 +744,11 @@ impl SimpleScheduler {
             max_job_retries = DEFAULT_MAX_JOB_RETRIES;
         }
 
+        let mut disconnect_timeout_s = scheduler_cfg.disconnect_timeout_s;
+        if disconnect_timeout_s == 0 {
+            disconnect_timeout_s = DEFAULT_DISCONNECT_TIMEOUT_S;
+        }
+
         let tasks_or_workers_change_notify = Arc::new(Notify::new());
 
         let metrics = Arc::new(Metrics::default());
@@ -703,6 +764,8 @@ impl SimpleScheduler {
             max_job_retries,
             tasks_or_workers_change_notify: tasks_or_workers_change_notify.clone(),
             metrics: metrics.clone(),
+            disconnect_timeout_s,
+            callbacks: Arc::new(callbacks),
         }));
         let weak_inner = Arc::downgrade(&inner);
         Self {
@@ -777,6 +840,97 @@ impl SimpleScheduler {
 
 #[async_trait]
 impl ActionScheduler for SimpleScheduler {
+    fn notify_client_disconnected(&self, unique_qualifier: &ActionInfoHashKey) {
+        // TODO: Make this prettier.
+        // It's a bit tricky to comply with borrow checker
+        // but it should be possible to make this nicer.
+        let inner = self.get_inner_lock();
+        let Some(action) = inner
+            .get_queued_action(&unique_qualifier)
+            .or_else(|| inner.get_active_action(&unique_qualifier))
+        else {
+            warn!(
+                "Scheduler notified that client disconnected, but failed to find action {}",
+                unique_qualifier.digest.hash_str()
+            );
+            return;
+        };
+
+        if action.notify_channel.receiver_count() != 0 {
+            return;
+        }
+        let sleep_time = Duration::from_secs(inner.disconnect_timeout_s);
+        let callbacks = inner.callbacks.clone();
+        // Drop the mutex guard so we don't hold up access.
+        drop(inner);
+
+        let weak_inner = Arc::downgrade(&self.inner);
+
+        let unique_qualifier = unique_qualifier.clone();
+
+        // We create a spawn here which sleeps for disconnect_timeout_s
+        // and then checks to see if the listener count is still 0.
+        // If so, it moves to the removal stage, otherwise returns.
+        tokio::spawn(async move {
+            callbacks.sleep(sleep_time).await;
+            let Some(inner_mux) = weak_inner.upgrade() else {
+                return;
+            };
+            let mut inner = inner_mux.lock();
+            let Some(action) = inner
+                .queued_actions_set
+                .get(&unique_qualifier)
+                .and_then(|action_info| inner.queued_actions.get(action_info))
+            else {
+                if inner.active_actions.contains_key(&unique_qualifier) {
+                    warn!("Action was active and could not be killed");
+                } else {
+                    info!(
+                        "Action {} completed while client disconnected",
+                        unique_qualifier.digest.hash_str()
+                    );
+                }
+                return;
+            };
+
+            // If listener count is 0, remove the action from queued actions
+            // or active actions, whichever is applicable, and kill the action
+            // on the worker if it was active.
+            // If the listener count is not still 0, a client has reconencted
+            // and there is no more work to be done.
+            if action.notify_channel.receiver_count() != 0 {
+                info!(
+                    "Client reconnected before disconnect_timeout elsapsed for Action {}",
+                    unique_qualifier.digest.hash_str()
+                );
+                return;
+            }
+
+            warn!(
+                "Client disconnect timeout elapsed - Removing action with digest hash {}",
+                action.action_info.unique_qualifier.digest.hash_str()
+            );
+
+            println!("About to remove");
+            match inner.get_queued_action(&unique_qualifier) {
+                Some(_) => {
+                    // We can't use the action info from the above call due to borrow checker.
+                    let action_info = inner
+                        .queued_actions_set
+                        .get(&unique_qualifier)
+                        .unwrap()
+                        .clone();
+                    inner.queued_actions_set.remove(&action_info);
+                    inner.queued_actions.remove(&action_info);
+                }
+                None => {
+                    inner.active_actions.remove(&unique_qualifier);
+                    // TODO: Send kill on worker signal - PR: #842.
+                }
+            }
+        });
+    }
+
     async fn get_platform_property_manager(
         &self,
         _instance_name: &str,
@@ -1024,9 +1178,9 @@ impl MetricsComponent for SimpleScheduler {
 impl MetricsComponent for CompletedAction {
     fn gather_metrics(&self, c: &mut CollectorState) {
         c.publish(
-            "completed_timestamp",
-            &self.completed_time,
-            "The timestamp this action was completed",
+            "completed_instant_elapsed",
+            &self.completed_instant.elapsed(),
+            "The elapsed time since the action was completed",
         );
         c.publish(
             "current_state",
--- a/nativelink-scheduler/tests/simple_scheduler_test.rs
+++ b/nativelink-scheduler/tests/simple_scheduler_test.rs
@@ -31,7 +31,7 @@ use nativelink_proto::build::bazel::remote::execution::v2::{digest_function, Exe
 use nativelink_proto::com::github::trace_machina::nativelink::remote_execution::{
     update_for_worker, ConnectionResult, StartExecute, UpdateForWorker,
 };
-use nativelink_scheduler::simple_scheduler::SimpleScheduler;
+use nativelink_scheduler::simple_scheduler::{Callbacks, SimpleScheduler};
 use nativelink_scheduler::worker::{Worker, WorkerId};
 use nativelink_scheduler::worker_scheduler::WorkerScheduler;
 use nativelink_util::common::DigestInfo;
@@ -95,6 +95,8 @@ async fn setup_action(
 
 #[cfg(test)]
 mod scheduler_tests {
+    use std::time::Instant;
+
     use pretty_assertions::assert_eq;
 
     use super::*; // Must be declared in every module.
@@ -107,6 +109,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -160,6 +163,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -224,6 +228,7 @@ mod scheduler_tests {
                 worker_timeout_s: WORKER_TIMEOUT_S,
                 ..Default::default()
             },
+            Callbacks::default(),
             || async move {},
         );
         let action_digest1 = DigestInfo::new([99u8; 32], 512);
@@ -367,6 +372,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -442,6 +448,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -524,6 +531,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -625,6 +633,7 @@ mod scheduler_tests {
         const WORKER_ID: WorkerId = WorkerId(0x0010_0010);
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -666,6 +675,7 @@ mod scheduler_tests {
                 worker_timeout_s: WORKER_TIMEOUT_S,
                 ..Default::default()
             },
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -765,6 +775,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -868,6 +879,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -971,6 +983,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -1066,6 +1079,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -1195,6 +1209,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest1 = DigestInfo::new([11u8; 32], 512);
@@ -1339,6 +1354,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest1 = DigestInfo::new([11u8; 32], 512);
@@ -1394,6 +1410,7 @@ mod scheduler_tests {
                 max_job_retries: 2,
                 ..Default::default()
             },
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
@@ -1523,6 +1540,7 @@ mod scheduler_tests {
         // DropChecker.
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             move || {
                 // This will ensure dropping happens if this function is ever dropped.
                 let _drop_checker = drop_checker.clone();
@@ -1548,6 +1566,7 @@ mod scheduler_tests {
 
         let scheduler = SimpleScheduler::new_with_callback(
             &nativelink_config::schedulers::SimpleScheduler::default(),
+            Callbacks::default(),
             || async move {},
         );
         let action_digest = DigestInfo::new([99u8; 32], 512);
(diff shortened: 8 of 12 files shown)


zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Vercel, pre-commit-checks

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtNs-vl1PJB6EvVwWlS:-NtNs-vmCLvOeFbwGFIJ:b-n7m3m2) ([raw file](https://github.com/TraceMachina/nativelink/blob/4b1c4de69dd76ad88107a7c6c9a87d7160351658/nativelink-scheduler/src/simple_scheduler.rs#L77)):*
> ```XML
> 
>     /// Updated on every client connect and periodically while it has listeners.
>     last_update_timestamp: Mutex<u64>
> ```

Mutex added due to borrow checker difficulties with modifying this field from `active_actions`.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 80 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtNraCY5GpqXQlG4hOw:-NtNraCY5GpqXQlG4hOx:b8jmgo4) ([raw file](https://github.com/TraceMachina/nativelink/blob/4b1c4de69dd76ad88107a7c6c9a87d7160351658/nativelink-scheduler/src/simple_scheduler.rs#L80)):*
> ```XML
> }
> 
> impl AwaitedAction {
> ```

These helpers are added to avoid in-lining long variable references and mutex locking but might not be necessary

adam-singer (changes requested):
Should we be using [struct.Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) over [struct.SystemTime](https://doc.rust-lang.org/std/time/struct.SystemTime.html) in this case?

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778#-:-NtWG5Ga99uu6EpccnFx:b-vnzc15)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 475 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWE3012nhQMnPFxd54:-NtWE3012nhQMnPFxd55:b5wahir) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L475)):*
> ```XML
>             // add update to queued action update timestamp here
>             let action = self.queued_actions.get_mut(&action_info).unwrap();
>             let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
> ```

Would it make a difference if we captured system time now before the loop?
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 548 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWDJOt84qynHGY-3Hb:-NtWDJOt84qynHGY-3Hc:b2f56mh) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L548)):*
> ```XML
>         for running_action in running_actions {
>             if running_action.action.notify_channel.receiver_count() > 0 {
>                 running_action.set_last_update_timestamp(SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs());
> ```

grab now before the else/if and having it recalled

```
let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
```

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 471 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwQbDM0-6DJULUOcga:-NtwQbDM0-6DJULUOcgb:bwytru4) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L471)):*
> ```XML
>         // unstable feature [see: https://github.com/rust-lang/rust/issues/70530]).
>         let action_infos: Vec<Arc<ActionInfo>> =
>             self.queued_actions.keys().rev().cloned().collect();
> ```

Make this `let queued_actions = self.queued_actions.iter().rev().cloned().collect()`. This will allow you to then use:
```rs
for (action_info, action) in queued_actions {
```

And you won't need to deal with unwrap and such.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRK51CvxkFx7T_lYg:-NtwRK51CvxkFx7T_lYh:b-r1srxo) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*
> ```XML
>             let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
>             if action.notify_channel.receiver_count() > 0 {
>                 action.set_last_update_timestamp(now)
> ```

nit: semi-colon.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-NtwT5As4BIfE0gOh9Ma:b-qcxair) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*
> ```XML
>             let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
>             if action.notify_channel.receiver_count() > 0 {
>                 action.set_last_update_timestamp(now)
> ```

I really feel that we should be able to detect when clients disconnect and set it then.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 480 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRgY9AJG2AKt3gQtU:-NtwRgY9AJG2AKt3gQtV:b-34xdy4) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L480)):*
> ```XML
>             } else if action.get_last_update_timestamp() + self.disconnect_timeout_s < now {
>                 self.queued_actions_set.remove(&action_info);
>                 self.queued_actions.remove(&action_info);
> ```

nit: Add a warn message here saying it is being removed.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 545 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRvmt7WRRdTN4d0MZ:-NtwRvmt7WRRdTN4d0M_:bnw0q68) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L545)):*
> ```XML
> 
>         let mut remove_actions = Vec::new();
>         let running_actions = &mut self.active_actions.values().collect::<Vec<_>>();
> ```

I'm kinda confused on how this ever worked because I believe this is grabbing a reference to a temporary.

In any case you should be beable to just use:

```rs
for running_action in self.active_Actions.values() {
```
since it is not modifying it in the loop.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1634 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSX396ZgxFqnSkeMF:-NtwSX396ZgxFqnSkeMG:b-7kooul) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1634)):*
> ```XML
>         .await?;
> 
>         // Drop our receiver
> ```

nit: period.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1641 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSZQb44xc9y-QHx21:-NtwSZQb44xc9y-QHx22:b-7kooul) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1641)):*
> ```XML
>         tokio::task::yield_now().await;
> 
>         // Sleep for longer than disconnect_timeout_s
> ```

nit: period.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1642 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSoCaDUnpIslmIm58:-NtwSoCaDUnpIslmIm59:bqymbch) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1642)):*
> ```XML
> 
>         // Sleep for longer than disconnect_timeout_s
>         let _ = sleep(Duration::from_secs(DISCONNECT_TIMEOUT_S + 1)).await;
> ```

We should never call native `sleep()` in tests. It's generally considered bad practice. In this case you probably want to hijack the time function inside the scheduler.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1652 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSaK30p7ocNIAEjJV:-NtwSaK30p7ocNIAEjJW:b-7kooul) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1652)):*
> ```XML
>         }
> 
>         // Setup a second action so matching engine is scheduled to rerun
> ```

nit: period.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1665 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSdUSA6u9acwM1urv:-NtwSdUSA6u9acwM1urw:b-7kooul) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1665)):*
> ```XML
>         tokio::task::yield_now().await;
> 
>         // Check to make sure that the action was removed
> ```

nit: period.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1669 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSfmb9BKbbODj8vv8:-NtwSfmb9BKbbODj8vv9:bcpg7mj) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1669)):*
> ```XML
>             .find_existing_action(&unique_qualifier)
>             .await
>             .is_none(),);
> ```

nit: missing message.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-NtwTBf-DF1Q5hn_pZEH:bfp0zmq) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*
> ```XML
> 
>     /// Updated on every client connect and periodically while it has listeners.
>     last_update_timestamp: Mutex<u64>
> ```

I'll help you with this when this PR is closer to being done.

However, `Mutex<(any primitive)>` is silly, since `Atomic{Primitive}` is always better (I can't think of any reason it wouldn't be faster/better, since it's non-blocking and sometimes has hardware support).

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Publish image, Vercel, docker-compose-compiles-nativelink (20.04), pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, vale, and 7 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-NufJQlm1viKI47VBBLJ:byi55r4) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*

I'll help you with this when this PR is closer to being done.

However, `Mutex<(any primitive)>` is silly, since `Atomic{Primitive}` is always better (I can't think of any reason it wouldn't be faster/better, since it's non-blocking and sometimes has hardware support).
</blockquote></details>

I changed it to be AtomicU64 and it works great
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 471 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwQbDM0-6DJULUOcga:-NufJ_3nDCdHt5o9UtjJ:b1zp18x) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L471)):*

Make this `let queued_actions = self.queued_actions.iter().rev().cloned().collect()`. This will allow you to then use:
```rs
for (action_info, action) in queued_actions {
```

And you won't need to deal with unwrap and such.
</blockquote></details>

`let queued_actions = self.queued_actions.iter().rev().cloned().collect()` gives a type error. I tried a bunch of different things to get it to work but couldn't figure it out.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 475 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWE3012nhQMnPFxd54:-NufK13X5_dvJOVI0VNz:b-kcw19m) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L475)):*

Would it make a difference if we captured system time now before the loop?
</blockquote></details>

Do_try_match is the slowest part of the scheduler. If we capture it before the loop and it takes a substantial amount of time to complete, it might fail to time out some actions until the next pass.

Whether or not that it acceptable is something I'm not sure about
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRK51CvxkFx7T_lYg:-NufKrtN7ECg_MT0E8EY:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

nit: semi-colon.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-NufKu6aB3w0Iph4q9q5:beq11pr) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

I really feel that we should be able to detect when clients disconnect and set it then.
</blockquote></details>

I think we discussed this previously and decided that putting this in do_try_match was preferable to creating a new spawn which iterates to check and sleeps. Is there another method of checking when the client drops?
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 480 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRgY9AJG2AKt3gQtU:-NufT9jtE6Gq8xPrtKeo:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L480)):*

nit: Add a warn message here saying it is being removed.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 545 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwRvmt7WRRdTN4d0MZ:-NufTCOsE2TP7YLgD-87:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L545)):*

I'm kinda confused on how this ever worked because I believe this is grabbing a reference to a temporary.

In any case you should be beable to just use:

```rs
for running_action in self.active_Actions.values() {
```
since it is not modifying it in the loop.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 548 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWDJOt84qynHGY-3Hb:-NufTjYh-t4kM1gUsvjk:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L548)):*

grab now before the else/if and having it recalled

```
let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
```
</blockquote></details>

Done.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1634 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSX396ZgxFqnSkeMF:-NufTn-vA5qXk4pZZBmi:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1634)):*

nit: period.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1641 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSZQb44xc9y-QHx21:-NufU0RS76yOJ1uGiDmj:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1641)):*

nit: period.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1642 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSoCaDUnpIslmIm58:-NufU1eg52VXUCFx-gUh:b-szepm7) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1642)):*

We should never call native `sleep()` in tests. It's generally considered bad practice. In this case you probably want to hijack the time function inside the scheduler.
</blockquote></details>

I changed it so that it calls a function which sets the `last_update_timestamp` instead of sleeping.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1652 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSaK30p7ocNIAEjJV:-NufUGBX2YLq7zW_ajTh:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1652)):*

nit: period.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1665 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSdUSA6u9acwM1urv:-NufUHuI3JY4vRJDKJ5s:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1665)):*

nit: period.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1669 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSfmb9BKbbODj8vv8:-NufUImc4qBu9lrE1-zh:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1669)):*

nit: missing message.
</blockquote></details>

Done.

adam-singer (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Remote / large-ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, windows-2022 / stable, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04, and 7 discussions need to be resolved

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and 9 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-NujA-SlCCF4GSMHUhCi:bkg771q) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*

I changed it to be AtomicU64 and it works great
</blockquote></details>

Why `Arc`?
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 471 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwQbDM0-6DJULUOcga:-NujAsX823qQjkravtS7:b3zpf6w) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L471)):*

`let queued_actions = self.queued_actions.iter().rev().cloned().collect()` gives a type error. I tried a bunch of different things to get it to work but couldn't figure it out.
</blockquote></details>

Did you instruct `.collect()` what the output container type was? Remember `.collect()` does not know what the type of the output will be, it just implements a trait that many containers implement. In this case you can use something like:
```rs
let queued_actions: Vec<(ActionInfo, Action)> = self.queued_actions.iter().rev().cloned().collect()
```
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 475 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWE3012nhQMnPFxd54:-NujBPuc25ZgoM78_X-h:btzztll) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L475)):*

Do_try_match is the slowest part of the scheduler. If we capture it before the loop and it takes a substantial amount of time to complete, it might fail to time out some actions until the next pass.

Whether or not that it acceptable is something I'm not sure about
</blockquote></details>

I would not over optimize that. if `.do_try_match` is taking more than 1 second we are in serious trouble.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-NujCDhG8UVUeQfgDUvx:b-fpsjam) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

I think we discussed this previously and decided that putting this in do_try_match was preferable to creating a new spawn which iterates to check and sleeps. Is there another method of checking when the client drops?
</blockquote></details>

Yeah, the cleanup happens here, but flagging when clients disconnect can/should happen when the event happens.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 60 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujA7v64WZOMIAyoREC:-NujA7v64WZOMIAyoRED:b-7kooul) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L60)):*
> ```XML
> const DEFAULT_MAX_JOB_RETRIES: usize = 3;
> 
> /// Default timeout for actions without any listeners
> ```

nit: period.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 363 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujAORA6T-Ikyx5dfys:-NujAORA6T-Ikyx5dfyt:b-2yo8ud) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L363)):*
> ```XML
>     }
> 
>     fn set_action_last_update_for_test(
> ```

nit: Inline this below, we want to limit the number of these floating around. These functions are generally allowed to be more ugly, since they are not supposed to be public facing.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 473 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujBfEpDEscU1lAJMCF:-NujBfEpDEscU1lAJMCG:b-6c1ew7) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L473)):*
> ```XML
>             // add update to queued action update timestamp here
>             let action = self.queued_actions.get_mut(&action_info).unwrap();
>             let now = SystemTime::now()
> ```

Using `SystemTime` inlined like this will make testing really difficult. I suggest making a constructor that you can pass in the time function.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1642 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSoCaDUnpIslmIm58:-NujCVTL65A8jGmSQb8j:bswpotf) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1642)):*

I changed it so that it calls a function which sets the `last_update_timestamp` instead of sleeping.
</blockquote></details>

But we are back to the original problem since we are relying on wall time because we are now using `SystemTime`.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and 9 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-NujpXRtD2QdzyyHDzmQ:b-dvjgkw) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*

Why `Arc`?
</blockquote></details>

I wasn't sure where we landed on when to update this. I can take it out for now but it will need to be an Arc if we are updating this from another thread that detects when the user disconnects.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 471 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwQbDM0-6DJULUOcga:-NujrdUb3UWpaxhswX71:b-svbjke) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L471)):*

Did you instruct `.collect()` what the output container type was? Remember `.collect()` does not know what the type of the output will be, it just implements a trait that many containers implement. In this case you can use something like:
```rs
let queued_actions: Vec<(ActionInfo, Action)> = self.queued_actions.iter().rev().cloned().collect()
```
</blockquote></details>

I did, I tried all the following and more but none of them worked:
`let queued_actions: Vec<(ActionInfo, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&ActionInfo, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(ActionInfo, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&ActionInfo, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(Arc<ActionInfo>, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(Arc<ActionInfo>, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();
`let queued_actions: Vec<(&Arc<ActionInfo>, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&Arc<ActionInfo>, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`

Here's an example of the full error:
 │     the method `collect` exists for struct `Cloned<Rev<Iter<'_, Arc<ActionInfo>, AwaitedAction>>>`, but its trait bounds were not satisfied rustc (E0599) [456, 106]
 │      the full type name has been written to '/Users/zach/Development/nativelink/target/debug/deps/nativelink_scheduler-1259087a177e1a06.long-type-1362048726443463511.txt' 
 │      the following trait bounds were not satisfied: 
 │      `<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>> as std::iter::Iterator>::Item = &_` 
 │      which is required by `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      which is required by `&mut std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │     the method `collect` exists for struct `Cloned<Rev<Iter<'_, Arc<ActionInfo>, AwaitedAction>>>`, but its trait bounds were not satisfied rustc (E0599) [456, 106]
 │      the full type name has been written to '/Users/zach/Development/nativelink/target/debug/deps/nativelink_scheduler-d0f178a39f61bd0a.long-type-2575323896278722321.txt' 
 │      the following trait bounds were not satisfied: 
 │      `<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>> as std::iter::Iterator>::Item = &_` 
 │      which is required by `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      which is required by `&mut std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │     the method call chain might not have had the expected associated types rustc (E0271) [456, 84]
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-Nujqmtp0OXxWfZrf0Ua:b-ume55j) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

Yeah, the cleanup happens here, but flagging when clients disconnect can/should happen when the event happens.
</blockquote></details>

How/where would we detect that the client disconnect has occurred? I was under the impression the only way to do this was to have a spawn which loops over the actions to check their receiver count.

I thought with disconnects it isn't possible to know until you request or try to send something to the other side
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1642 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSoCaDUnpIslmIm58:-NujuDg89_BiyA9D5T4C:b-6y7mom) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1642)):*

But we are back to the original problem since we are relying on wall time because we are now using `SystemTime`.
</blockquote></details>

I'm a bit confused on a couple things here.

Where is the time function set/used in the scheduler? I see SystemTime used other places inside of it but not a `now_fn` like `RunningActionsManagerImpl` has.

I don't mind changing it to use that structure, but I just want to make sure that's something we want to include in this PR.

I'm also very confused about why we are relying on wall time here, the actual time doesn't matter at all unless I'm missing something huge.

The comparison done in the code to determine whether or not to time out is:
```
// If this is true the action is marked as old
running_action.get_last_update_timestamp() + self.disconnect_timeout_s < now
```
where:
`running_action.get_last_update_timestamp()` returns some uint
`now` is the current time since unix epoch in seconds
`self.disconnect_timeout_s` is a value like `1`

if we set `running_action.get_last_update_timestamp()` to be `SystemTime::now() - (self.disconnect_timeout_s+1)` then even if no time at all elapses between the first call to `SystemTime` in the test and the second one in `do_try_match` we end up with:

`now - (disconnect_timeout_s-1) + disconnect_timeout_s < now` = `-1 < 0` which will always return true

If the thought is that something like daylight savings or some system setting change could mess up the system time between the two calls to this test, we could just set the last_update_timestamp to 0 which would make it

`January 1st, 1970 + disconnect_timeout_s < now`

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Bazel Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), pre-commit-checks, ubuntu-20.04 / stable, vale, and 8 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 363 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujAORA6T-Ikyx5dfys:-Nuk-Cb-5CmI-hOUSaWN:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L363)):*

nit: Inline this below, we want to limit the number of these floating around. These functions are generally allowed to be more ugly, since they are not supposed to be public facing.
</blockquote></details>

Done.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04, and 8 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-NukNmvx-PcpCYGA_zIs:bjb6wwy) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*

I wasn't sure where we landed on when to update this. I can take it out for now but it will need to be an Arc if we are updating this from another thread that detects when the user disconnects.
</blockquote></details>

If we do that we will change the type. In this case we are not even sure how the future will end up, so keep it as simple as possible for now (no Arc).
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-NukQ2pKAtJupjdickNn:b-6j35ub) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

How/where would we detect that the client disconnect has occurred? I was under the impression the only way to do this was to have a spawn which loops over the actions to check their receiver count.

I thought with disconnects it isn't possible to know until you request or try to send something to the other side
</blockquote></details>

Use `scope_guard` to detect when the stream is dropped in `execution_server::to_execute_stream()` that then calls a function on the scheduler to let the scheduler know that a client disconnected. The function would then take check if it's the last one subscribed and then the time.

Btw, this will make testing significantly easier with these public functions.

adam-singer (commented):
> Reviewed 10 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and 12 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 837 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/778#-Nv5RkpqDoNP9iXYl_YF:-Nv5RkpqDoNP9iXYl_YG:b1xfh1h) ([raw file](https://github.com/TraceMachina/nativelink/blob/624eda2e29d31b33e2bf4989278d7528d0b1ef97/nativelink-scheduler/src/simple_scheduler.rs#L837)):*
> ```XML
> #[async_trait]
> impl ActionScheduler for SimpleScheduler {
>     fn notify_client_disconnected(&self, unique_qualifier: ActionInfoHashKey) {
> ```

When short returning in some of the `let/else` expressions, should we `debug`/`warn` log, it seems in some of the cases it might be useful to know that we are hitting what seems unexpected state or racing states?
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 837 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/778#-Nv5SEZe9c4_cN5zO7xL:-Nv5SEZe9c4_cN5zO7xM:b-pswmea) ([raw file](https://github.com/TraceMachina/nativelink/blob/624eda2e29d31b33e2bf4989278d7528d0b1ef97/nativelink-scheduler/src/simple_scheduler.rs#L837)):*
> ```XML
> #[async_trait]
> impl ActionScheduler for SimpleScheduler {
>     fn notify_client_disconnected(&self, unique_qualifier: ActionInfoHashKey) {
> ```

We should be incrementing counters on different states here, not sure if a counter on `notify_client_disconnected` is useful (could be to match if we count successful removals from queues), we should counter on `inner.queued_actions*` removals and `inner.active_actions.remove`.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 859 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/778#-Nv5RJ6SEMCfq8gjijMw:-Nv5RJ6SEMCfq8gjijMx:b-lqdy88) ([raw file](https://github.com/TraceMachina/nativelink/blob/624eda2e29d31b33e2bf4989278d7528d0b1ef97/nativelink-scheduler/src/simple_scheduler.rs#L859)):*
> ```XML
>         let weak_inner = Arc::downgrade(&self.inner);
> 
>         tokio::spawn(async move {
> ```

I think we should explain (comments) why we are spawning at this point and what the next stage of removing from queues means. 
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 899 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/778#-Nv5R0TVEpoDDEIJlUt3:-Nv5R0TVEpoDDEIJlUt4:b3r6c82) ([raw file](https://github.com/TraceMachina/nativelink/blob/624eda2e29d31b33e2bf4989278d7528d0b1ef97/nativelink-scheduler/src/simple_scheduler.rs#L899)):*
> ```XML
>                         return;
>                     };
>                     // TODO: Send kill on worker signal - PR: #842.
> ```

Couldn't we just call `inner.active_actions.remove(&unique_qualifier);` since the match would fall through and return?

allada (commented):
Yes, I think `Instant` is the proper approach. It also solves a bit of other issues with importing `SysemTime`.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778#-:-Nv7hPMxEnugiX0BExxE:be2sqxz)* status: 0 of 1 LGTMs obtained, and 12 discussions need to be resolved

zbirenbaum (commented):
> Dismissed @adam-singer and @allada from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 0 of 1 LGTMs obtained, and 7 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 77 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwTBf-DF1Q5hn_pZEG:-Nv8YvJ5Cjc_Ds0cV1tu:b-4kirhb) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L77)):*

If we do that we will change the type. In this case we are not even sure how the future will end up, so keep it as simple as possible for now (no Arc).
</blockquote></details>

Done. The var no longer exists so dismissing this.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 471 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwQbDM0-6DJULUOcga:-Nv8Z42S67A4JtalS-_m:bk5pxgz) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L471)):*

I did, I tried all the following and more but none of them worked:
`let queued_actions: Vec<(ActionInfo, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&ActionInfo, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(ActionInfo, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&ActionInfo, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(Arc<ActionInfo>, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(Arc<ActionInfo>, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();
`let queued_actions: Vec<(&Arc<ActionInfo>, AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`
`let queued_actions: Vec<(&Arc<ActionInfo>, &AwaitedAction)> = self.queued_actions.iter().rev().cloned().collect();`

Here's an example of the full error:
 │     the method `collect` exists for struct `Cloned<Rev<Iter<'_, Arc<ActionInfo>, AwaitedAction>>>`, but its trait bounds were not satisfied rustc (E0599) [456, 106]
 │      the full type name has been written to '/Users/zach/Development/nativelink/target/debug/deps/nativelink_scheduler-1259087a177e1a06.long-type-1362048726443463511.txt' 
 │      the following trait bounds were not satisfied: 
 │      `<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>> as std::iter::Iterator>::Item = &_` 
 │      which is required by `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      which is required by `&mut std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │     the method `collect` exists for struct `Cloned<Rev<Iter<'_, Arc<ActionInfo>, AwaitedAction>>>`, but its trait bounds were not satisfied rustc (E0599) [456, 106]
 │      the full type name has been written to '/Users/zach/Development/nativelink/target/debug/deps/nativelink_scheduler-d0f178a39f61bd0a.long-type-2575323896278722321.txt' 
 │      the following trait bounds were not satisfied: 
 │      `<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>> as std::iter::Iterator>::Item = &_` 
 │      which is required by `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      `std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │      which is required by `&mut std::iter::Cloned<std::iter::Rev<std::collections::btree_map::Iter<'_, std::sync::Arc<nativelink_util::action_messages::ActionInfo>, simple_scheduler::AwaitedAction>>>: std::iter::Iterator` 
 │     the method call chain might not have had the expected associated types rustc (E0271) [456, 84]
</blockquote></details>

Done. `do_try_match` no longer has any edits and leaving this how it works on main so dismissing this.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 475 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWE3012nhQMnPFxd54:-Nv8ZfDV3eYgmeP06FY2:bk5pxgz) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L475)):*

I would not over optimize that. if `.do_try_match` is taking more than 1 second we are in serious trouble.
</blockquote></details>

Done. `do_try_match` no longer has any edits and leaving this how it works on main so dismissing this.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 477 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwT5As4BIfE0gOh9M_:-NuzH7L55NmF-EBkrxu5:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L477)):*

Use `scope_guard` to detect when the stream is dropped in `execution_server::to_execute_stream()` that then calls a function on the scheduler to let the scheduler know that a client disconnected. The function would then take check if it's the last one subscribed and then the time.

Btw, this will make testing significantly easier with these public functions.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 548 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtWDJOt84qynHGY-3Hb:-Nv8_CsXCN7HGUnJK3PF:bk5pxgz) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/src/simple_scheduler.rs#L548)):*

Done.
</blockquote></details>

Done. `do_try_match` no longer has any edits and leaving this how it works on main so dismissing this.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 60 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujA7v64WZOMIAyoREC:-Nv8YuH0ASTjLjX6Lrnd:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L60)):*

nit: period.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 473 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NujBfEpDEscU1lAJMCF:-Nv8_-VwDJBdw2Rn7Dwv:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/2e287e5dba0cc8c19d5363e1b8fe350ed80000bc/nativelink-scheduler/src/simple_scheduler.rs#L473)):*

Using `SystemTime` inlined like this will make testing really difficult. I suggest making a constructor that you can pass in the time function.
</blockquote></details>

Done.

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778#-:-Nv9JHRB47RmjRt-evcJ:bnfp4nl)* status: 1 of 1 LGTMs obtained, and 4 discussions need to be resolved

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 1 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 485 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NvXZ2Lo7xvc8TxgHc0h:-NvXZ2Lo7xvc8TxgHc0i:b-jefw7h) ([raw file](https://github.com/TraceMachina/nativelink/blob/825aa7943a1705359a18e0bed5b96eeb7da2140b/nativelink-scheduler/src/simple_scheduler.rs#L485)):*
> ```XML
>             self.queued_actions.keys().rev().cloned().collect();
>         for action_info in action_infos {
>             // add update to queued action update timestamp here
> ```

don't forget this comment.
___
*[`nativelink-scheduler/tests/simple_scheduler_test.rs` line 1642 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/778#-NtwSoCaDUnpIslmIm58:-NvXZNTyAmbtdXAvxKkK:bb4krlc) ([raw file](https://github.com/TraceMachina/nativelink/blob/ecadfb0ee7bb2062ad0023d8307572a3df44010f/nativelink-scheduler/tests/simple_scheduler_test.rs#L1642)):*

I'm a bit confused on a couple things here.

Where is the time function set/used in the scheduler? I see SystemTime used other places inside of it but not a `now_fn` like `RunningActionsManagerImpl` has.

I don't mind changing it to use that structure, but I just want to make sure that's something we want to include in this PR.

I'm also very confused about why we are relying on wall time here, the actual time doesn't matter at all unless I'm missing something huge.

The comparison done in the code to determine whether or not to time out is:
```
// If this is true the action is marked as old
running_action.get_last_update_timestamp() + self.disconnect_timeout_s < now
```
where:
`running_action.get_last_update_timestamp()` returns some uint
`now` is the current time since unix epoch in seconds
`self.disconnect_timeout_s` is a value like `1`

if we set `running_action.get_last_update_timestamp()` to be `SystemTime::now() - (self.disconnect_timeout_s+1)` then even if no time at all elapses between the first call to `SystemTime` in the test and the second one in `do_try_match` we end up with:

`now - (disconnect_timeout_s-1) + disconnect_timeout_s < now` = `-1 < 0` which will always return true

If the thought is that something like daylight savings or some system setting change could mess up the system time between the two calls to this test, we could just set the last_update_timestamp to 0 which would make it

`January 1st, 1970 + disconnect_timeout_s < now`
</blockquote></details>

If we move everything to `Instant` it will likely resolve all of these issues.

adam-singer (commented):
> Reviewed 6 of 8 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/778)* status: 1 of 1 LGTMs obtained, and 3 discussions need to be resolved

PR #793, opened 2024-03-25
Publish nativelink-worker image for C++

aaronmondal:
This should simplify external setups as it's no longer a requirement to build worker images if they work with a "standard" C++ toolchain.

Since the toolchain images are quite somewhat more complex than the minimal NativeLink image the pipelines now publish trivy scan results.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/793)


aaronmondal:
Not ready yet.

PR #806, opened 2024-03-28
Connection Manager Rewrite

chrisstaite-menlo:
# Description

The connection manager written in grpc_utils made incorrect assumptions about how the tonic and tower implementations were written and is not suitable for maintaining multiple connections and ensuring stability.

Completely rewrite this to manage the tonic::Channel for each tonic::Endpoint itself to make a simpler external API and ensure that connection errors are handled correctly. This is performed by using a single worker loop that manages all of the connections and wrapping each connection to inform state to the worker.

## Type of change

## How Has This Been Tested?

Run up on my US proxy service which is now capable of completing a fresh clean build rather than stalling out on concurrent uploads.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)


aaronmondal (approved):
:lgtm:

But I feel like I must be overlooking something lol. Deferring to @allada :sweat_smile: 

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806#-:-Nu6MFp62r36P4uxRT7i:b-bwa0ym)* status: 1 of 1 LGTMs obtained, and 1 discussions need to be resolved

___
*[`nativelink-util/src/grpc_utils.rs` line 94 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-Nu6JiT_CAF_5eWATRfC:-Nu6JiT_CAF_5eWATRfD:bw4dxah) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L94)):*
> ```XML
>     /// Connected channels that are waiting for poll_ready for the Service.
>     /// This has a key of the endpoint index and iteration to the Channel value.
>     pending_channels: HashMap<ChannelIdentifier, Channel>,
> ```

nit: I believe it's safe to use `ahash::HashMap` instead of the `std::collections::HashMap` here: https://github.com/tkaitchuck/ahash

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 26 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGZPSH9-V1zAv84h29:-NuGZPSH9-V1zAv84h2A:b-yn7yjk) (no related file):*
It feels like we are doing a lot of book keeping with `connecting_channels`, `available_channels`, `pending_channels`, `connecting_channels`, and   `waiting_connections`. I feel like there's a pattern that could have this auto-managed for us.

Maybe we can use a `Semaphore` or `tokio::sync::watch` to lease out connections & channels and let `drop()` put them back into the available pool.

___
___
*[`nativelink-config/src/schedulers.rs` line 145 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFH9W1o3oa_xWbXkS:-NuGFH9W1o3oa_xWbXkT:bn9ccgs) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-config/src/schedulers.rs#L145)):*
> ```XML
> 
>     /// The number of connections to make to each specified endpoint to balance
>     /// the load over multiple TCP connections.  Zero is always changed to 1.
> ```

nit: Suggest also saying `Default: 1`
___
*[`nativelink-config/src/stores.rs` line 565 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFLBJAnvXcBWJxq__:-NuGFLBJAnvXcBWJxq_a:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-config/src/stores.rs#L565)):*
> ```XML
> 
>     /// The number of connections to make to each specified endpoint to balance
>     /// the load over multiple TCP connections.  Zero is always changed to 1.
> ```

ditto.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 168 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFcal-K0OVUioWEGi:-NuGFcal-K0OVUioWEGj:b2snv1h) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-scheduler/src/grpc_scheduler.rs#L168)):*
> ```XML
>         self.perform_request(instance_name, |instance_name| async move {
>             // Not in the cache, lookup the capabilities with the upstream.
>             let channel = self.connection_manager.connection().await?;
> ```

nit: Can we `.err_tip()` here for debug tracing?
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 221 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFkR12L_oADI5f1Yk:-NuGFkR12L_oADI5f1Yl:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-scheduler/src/grpc_scheduler.rs#L221)):*
> ```XML
>         let result_stream = self
>             .perform_request(request, |request| async move {
>                 let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 241 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFlTy44f_txi-K-NS:-NuGFlTy44f_txi-K-NT:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-scheduler/src/grpc_scheduler.rs#L241)):*
> ```XML
>         let result_stream = self
>             .perform_request(request, |request| async move {
>                 let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 149 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFoSf9MuBgUgt56QL:-NuGFoSf9MuBgUgt56QM:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L149)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 170 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFpXO7W5CStXuMW-X:-NuGFpXO7W5CStXuMW-Y:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L170)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 191 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFq0gAQyKBmbW9Fub:-NuGFq0gAQyKBmbW9Fuc:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L191)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 212 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFqYB245JmnGjSnXF:-NuGFqYB245JmnGjSnXG:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L212)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 235 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFsf7EazL0VRoEtab:-NuGFsf7EazL0VRoEtac:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L235)):*
> ```XML
>         request: ReadRequest,
>     ) -> Result<impl Stream<Item = Result<ReadResponse, Status>>, Error> {
>         let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 349 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGKEt-pPPcaugThNT:-NuGGKEt-pPPcaugThNU:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L349)):*
> ```XML
> 
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 365 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGKxA3rvCdK9aOg0H:-NuGGKxA3rvCdK9aOg0I:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L365)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-store/src/grpc_store.rs` line 381 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGLYI3VSJI7Msd-qg:-NuGGLYI3VSJI7Msd-qh:b-rqxvo7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L381)):*
> ```XML
>         request.instance_name = self.instance_name.clone();
>         self.perform_request(request, |request| async move {
>             let channel = self.connection_manager.connection().await?;
> ```

ditto.
___
*[`nativelink-util/src/grpc_utils.rs` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGIk5n6olN8rXP6Z2O:-NuGIk5n6olN8rXP6Z2P:b-rsq06q) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L1)):*
> ```XML
> // Copyright 2024 The NativeLink Authors. All rights reserved.
> ```

It looks like this file is now all about `GrpcConnectionManager`. We should probably rename this file.
___
*[`nativelink-util/src/grpc_utils.rs` line 32 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGH0zV4TKcuwayCjMB:-NuGH0zW7wj7GceQsw2_:b-qpcsdb) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L32)):*
> ```XML
> pub struct ConnectionManager {
>     // The channel to request connections from the worker.
>     worker_tx: UnboundedSender<WorkerRequest>,
> ```

It feels like it would be a good idea to use a bounded channel instead so backpressure can be created. I'm always afraid of unbounded channels for this reason. If we did I suggest don't make it configurable just hard code a high limit. Thoughts?
___
*[`nativelink-util/src/grpc_utils.rs` line 36 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGHaMw3vTkmDl3yKiP:-NuGHaMx3HRcpsGkdl05:b7hzvef) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L36)):*
> ```XML
> 
> /// The requests that can be made from the ConnectionManager to the
> /// ConnectionManagerWorker such as requesting
> ```

nit: Incomplete comment.
___
*[`nativelink-util/src/grpc_utils.rs` line 47 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGJE5z8OFX6ckB8luC:-NuGJE5z8OFX6ckB8luD:b-xa4e0p) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L47)):*
> ```XML
>     endpoint_index: usize,
>     /// A unique identifier for this particular connection to the Endpoint.
>     iteration: usize,
> ```

nit: Lets rename this `connection_index`. It took me a while to figure out what this was.
___
*[`nativelink-util/src/grpc_utils.rs` line 82 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGJUva65FlBD-VMYne:-NuGJUva65FlBD-VMYnf:b-119e9j) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L82)):*
> ```XML
>     /// The endpoints to establish Channels and the current iteration of
>     /// connection to that endpoint.
>     endpoints: Vec<(usize, Endpoint)>,
> ```

nit: Lets alias this `usize` to something like `EndpointIndex` (Maybe do the same for above `usize`s).
___
*[`nativelink-util/src/grpc_utils.rs` line 87 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGKDno9mK_O_UL4ZTY:-NuGKDno9mK_O_UL4ZTZ:baxkk99) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L87)):*
> ```XML
>     /// If the number of connections are restricted, then the number of
>     /// connections that are currently allowed to be made.
>     available_connections: Option<usize>,
> ```

nit: Lets make this a `usize` and make it `usize::MAX` if zero.

Or maybe better yet, always enforce some sane limit. I personally like enforcing limits on uses by default and let them put higher/lower values when they discover limitations.
___
*[`nativelink-util/src/grpc_utils.rs` line 94 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-Nu6JiT_CAF_5eWATRfC:-NuEmylDAWoPIIIcGFxn:b6fzoyd) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L94)):*

nit: I believe it's safe to use `ahash::HashMap` instead of the `std::collections::HashMap` here: https://github.com/tkaitchuck/ahash
</blockquote></details>

We use `hashbrown::HashMap` for non-std hashmaps.
___
*[`nativelink-util/src/grpc_utils.rs` line 103 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGLKn66ZAQY01d9--q:-NuGLKn66ZAQY01d9--r:baerh6a) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L103)):*
> ```XML
>     /// set of Endpoints.  This will restrict the number of concurrent requests
>     /// assuming that the user of this connection manager uses the connection
>     /// only once and reports all errors.
> ```

What does the last sentence mean here? I'm not sure I understand. Shouldn't this be wrapped to prevent a user from abusing it?
___
*[`nativelink-util/src/grpc_utils.rs` line 121 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGLySsAENPHCsmXNL_:-NuGLySsAENPHCsmXNLa:b-8zta83) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L121)):*
> ```XML
>             waiting_connections: VecDeque::new(),
>         };
>         let connections_per_endpoint = if connections_per_endpoint > 0 {
> ```

nit: For readability, just make `connections_per_endpoint` mut and modify it in the if-statement.
___
*[`nativelink-util/src/grpc_utils.rs` line 148 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGPnP5ACE7L5DnslLv:-NuGPnP5ACE7L5DnslLw:b-tsc7s2) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L148)):*
> ```XML
> 
> impl ConnectionManagerWorker {
>     async fn execute(
> ```

nit: suggest better name. Maybe `service_requests`?
___
*[`nativelink-util/src/grpc_utils.rs` line 187 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGMMpREJdKhWg3JQNA:-NuGMMpREJdKhWg3JQNB:bhps4zo) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L187)):*
> ```XML
>                 _ = self.connect_next() => {}
>             }
>         }
> ```

nit: Add `// Unreachable.` after the loop.
___
*[`nativelink-util/src/grpc_utils.rs` line 356 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGSC3VAjvoUZ3jUPmH:-NuGSC3VAjvoUZ3jUPmI:bds6pk3) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L356)):*
> ```XML
>     identifier: ChannelIdentifier,
>     /// The wrapped future that actually does the work.
>     inner: channel::ResponseFuture,
> ```

nit: Generally it's a good idea to put `inner` objects first in the list so the optimizer can combine/inline functions that have the same index offsets.

chrisstaite (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 26 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGZPSH9-V1zAv84h29:-NuISnSr-EKiqjZTjJmr:b-xu7y81) (no related file):*

It feels like we are doing a lot of book keeping with `connecting_channels`, `available_channels`, `pending_channels`, `connecting_channels`, and   `waiting_connections`. I feel like there's a pattern that could have this auto-managed for us.

Maybe we can use a `Semaphore` or `tokio::sync::watch` to lease out connections & channels and let `drop()` put them back into the available pool.
</blockquote></details>

The concept is heavily based off the implementation in tower, which has to perform all this management. I did start off trying to use the tower implementation, but it didn't give access to the errors and we'd end up with the same compromise as before. I don't think that a watch or a Semaphore would help simplify this at all.

I also considered a state machine, but we actually want different management of the Channel, for example connecting needs to be a monitored future, available needs to be a queue, pending could be simply managed my the Connection, but then we'd need a separate list of errors which can happen asynchronously as Channels are cloned to reuse a connection multiple times. The waiting connections could probably be handled by a Semaphore, but I feel it's much simpler to reason with reconnection attempts.

I designed this to be totally lock free and simply mutable on a single worker with a simple channel input and output. I welcome library suggestions, but I haven't seen anything that fits any parts of the use case yet.

___
___
*[`nativelink-util/src/grpc_utils.rs` line 187 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGMMpREJdKhWg3JQNA:-NuIUuHqA5E6WP7glZGa:bvjm0a5) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L187)):*

nit: Add `// Unreachable.` after the loop.
</blockquote></details>

It is reachable when the ConnectionManager is dropped. There *could* be cleanup work afterwards, none is needed though.
___
*[`nativelink-util/src/grpc_utils.rs` line 356 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGSC3VAjvoUZ3jUPmH:-NuIV4DV3RP6tzNbl0Ar:b-mdvnjq) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L356)):*

nit: Generally it's a good idea to put `inner` objects first in the list so the optimizer can combine/inline functions that have the same index offsets.
</blockquote></details>

Oh, nice.. hadn't thought of that

chrisstaite (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 26 discussions need to be resolved

___
*[`nativelink-util/src/grpc_utils.rs` line 32 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGH0zV4TKcuwayCjMB:-NuIVCIp69-Cj4u5Tiy6:by6hgyl) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L32)):*

It feels like it would be a good idea to use a bounded channel instead so backpressure can be created. I'm always afraid of unbounded channels for this reason. If we did I suggest don't make it configurable just hard code a high limit. Thoughts?
</blockquote></details>

Yeah, I'd thought that too. I already planned to swap it out, just haven't yet.
___
*[`nativelink-util/src/grpc_utils.rs` line 36 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGHaMw3vTkmDl3yKiP:-NuIVIBM54aOQHizgwA3:brdhx24) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L36)):*

nit: Incomplete comment.
</blockquote></details>

Yeah, apparently I drifted off...

chrisstaite:

Could you link which implementation it's based on?

I'm going to look to see if maybe we could upstream some changes to reduce our burden of maintenance. My initial search didn't yield much.
</blockquote></details>

https://docs.rs/tower/latest/src/tower/balance/p2c/service.rs.html#32-45

adam-singer (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 26 discussions need to be resolved

___
*[`nativelink-util/src/grpc_utils.rs` line 194 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdZAI0bQ9kr7NGTsH:-NuPdZAI0bQ9kr7NGTsI:b-i42q80) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L194)):*
> ```XML
>             // Sleep for a really long time, this will be cancelled by the
>             // select in execute.
>             tokio::time::sleep(Duration::from_secs(100000)).await;
> ```

Which thread is this running in and how many possible threads could hit this sleep? 
___
*[`nativelink-util/src/grpc_utils.rs` line 222 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdqSN6z2PNW5hLFw-:-NuPdqSN6z2PNW5hLFw0:b-840mr6) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L222)):*
> ```XML
>             }
>             self.connecting_channels.push(Box::pin(async move {
>                 if is_backoff {
> ```

Can we utilize the retry.rs backoff utility code vs one off sleeping?

chrisstaite (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 26 discussions need to be resolved

___
*[`nativelink-util/src/grpc_utils.rs` line 194 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdZAI0bQ9kr7NGTsH:-NuPhjKq7N91Cq5rNXab:b47atv7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L194)):*

Which thread is this running in and how many possible threads could hit this sleep?
</blockquote></details>

It's not running in any thread, it's a tokio sleep. Essentially, we want to suspend this future forever such that it always returns Pending. It will get cancelled by the select that called it and the function re-entered at which time, connecting_channels can be reassessed.
___
*[`nativelink-util/src/grpc_utils.rs` line 222 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdqSN6z2PNW5hLFw-:-NuPiBae852V02q16Ge4:b-g7ad0h) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L222)):*

Can we utilize the retry.rs backoff utility code vs one off sleeping?
</blockquote></details>

I considered it, but that's a lot of config for reconnecting, also retry.rs wraps a function... Maybe we could do it for the whole future...

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 18 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 565 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFLBJAnvXcBWJxq__:-NuSlVYa62JfLRt-CVTd:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-config/src/stores.rs#L565)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 221 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFkR12L_oADI5f1Yk:-NuSlWvp6VS0jMDCRqon:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-scheduler/src/grpc_scheduler.rs#L221)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/grpc_scheduler.rs` line 241 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFlTy44f_txi-K-NS:-NuSlXHVBJFq_aWe2mrF:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-scheduler/src/grpc_scheduler.rs#L241)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 149 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFoSf9MuBgUgt56QL:-NuSlXz80v13zakMXpgP:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L149)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 170 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFpXO7W5CStXuMW-X:-NuSlYOu4M5e7cxdi6f0:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L170)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 191 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFq0gAQyKBmbW9Fub:-NuSlYmcAOePT--xizUy:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L191)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 212 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFqYB245JmnGjSnXF:-NuSlZ8HDsJPNPfbUWzi:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L212)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 235 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGFsf7EazL0VRoEtab:-NuSlZYZ9CAAFTXZhi9L:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L235)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 349 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGKEt-pPPcaugThNT:-NuSl_xr5t6FLYkwJHWb:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L349)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 365 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGKxA3rvCdK9aOg0H:-NuSlaHP2Af-1DueJPqV:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L365)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-store/src/grpc_store.rs` line 381 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGGLYI3VSJI7Msd-qg:-NuSlaiCDWKaqkjRGCGa:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-store/src/grpc_store.rs#L381)):*

ditto.
</blockquote></details>

Done.
___
*[`nativelink-util/src/grpc_utils.rs` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGIk5n6olN8rXP6Z2O:-NuSlbRmAxRoKOcL-Ymc:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L1)):*

It looks like this file is now all about `GrpcConnectionManager`. We should probably rename this file.
</blockquote></details>

Done.
___
*[`nativelink-util/src/grpc_utils.rs` line 32 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGH0zV4TKcuwayCjMB:-NuSoTu94rilqfD-pWXh:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L32)):*
<details><summary><i>Previously, chrisstaite (Chris) wrote…</i></summary><blockquote>

Yeah, I'd thought that too. I already planned to swap it out, just haven't yet.
</blockquote></details>

Done.
___
*[`nativelink-util/src/grpc_utils.rs` line 94 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-Nu6JiT_CAF_5eWATRfC:-NuSpLDoAf1vUV_951es:b2pk6pf) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L94)):*

We use `hashbrown::HashMap` for non-std hashmaps.
</blockquote></details>

I think I'll leave this as-is since the std HashMap is now hashbrown anyway.
___
*[`nativelink-util/src/grpc_utils.rs` line 103 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGLKn66ZAQY01d9--q:-NuSpgbR84DgOerypk9v:b-4qskxj) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L103)):*

What does the last sentence mean here? I'm not sure I understand. Shouldn't this be wrapped to prevent a user from abusing it?
</blockquote></details>

Yeah, this is a lay-over from the old implementation that required user input, I've re-written this for the new implementation.
___
*[`nativelink-util/src/grpc_utils.rs` line 222 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdqSN6z2PNW5hLFw-:-NuSqFW47N4yZqNBqqTK:b-ducmlu) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L222)):*
<details><summary><i>Previously, chrisstaite (Chris) wrote…</i></summary><blockquote>

I considered it, but that's a lot of config for reconnecting, also retry.rs wraps a function... Maybe we could do it for the whole future...
</blockquote></details>

There's a few issues with this, but I think I've overcome them.  Firstly, the `Retrier` requires the result to be an `Error`, so I wrapped it to provide the `ConnectionIdentifier` context in the case of error.  Second, the `Retrier` does not have a concept of infinite retries, so in the case the it runs out of attempts, it simply re-enters.  The annoying point is that we have to make a clone of the `Retrier` because of the lifetime of `self`, `connecting_channels` will be dropped before the `retrier` and therefore it should be safe, but I can't convince the checker without introducing a lifetime to the worker.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Vercel, docker-compose-compiles-nativelink (20.04), pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, vale, and 17 discussions need to be resolved

___
*[`nativelink-util/src/grpc_utils.rs` line 194 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuPdZAI0bQ9kr7NGTsH:-NuTGn709d1msVX2nM49:b-k8amd5) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L194)):*
<details><summary><i>Previously, chrisstaite (Chris) wrote…</i></summary><blockquote>

It's not running in any thread, it's a tokio sleep. Essentially, we want to suspend this future forever such that it always returns Pending. It will get cancelled by the select that called it and the function re-entered at which time, connecting_channels can be reassessed.
</blockquote></details>

Moved to `futures::future::pending::<()>().await` instead.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, zig-cc ubuntu-20.04, zig-cc ubuntu-22.04, and 17 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuGZPSH9-V1zAv84h29:-NuTRWv65WYM-jkdQMay:bvzp8s4) (no related file):*
<details><summary><i>Previously, chrisstaite (Chris) wrote…</i></summary><blockquote>

https://docs.rs/tower/latest/src/tower/balance/p2c/service.rs.html#32-45
</blockquote></details>

Moved `pending_channels` to passing the channel to the `Connection` on creation and then passing it back on `Connected` or `Dropped` using `.take()`.  The down side is that we don't have visibility of the connection if a transport error happens on one stream first while it's pending in another `Connection`, however the error should occur again in the other `Connection` and resolve itself, so I don't see it as an issue.

___

chrisstaite-menlo (commented):
Have deployed this out to our fleet and it appears to be functioning well.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806#-:-NuUO7UZ4weh7Q7v0Rxj:bcmidk2)* status: 1 of 1 LGTMs obtained, and 17 discussions need to be resolved

adam-singer (commented):
> Reviewed 5 of 8 files at r3, 2 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Bazel Dev / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Vercel, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), pre-commit-checks, ubuntu-20.04 / stable, and 18 discussions need to be resolved

___
*[`nativelink-util/src/connection_manager.rs` line 442 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuUXhMM6lMoMrEqAxZl:-NuUXhMM6lMoMrEqAxZm:bg2xomb) ([raw file](https://github.com/TraceMachina/nativelink/blob/62cf43dd8f1132470a365e87c101cac1612d56a4/nativelink-util/src/connection_manager.rs#L442)):*
> ```XML
>                 }
>                 Err(err) => {
>                     debug!("Error while creating connection on channel: {err:?}");
> ```

I think this should be `warn!` or `error!`, is it expected to be noisy?

chrisstaite-menlo:
*[`nativelink-util/src/connection_manager.rs` line 442 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuUXhMM6lMoMrEqAxZl:-NuUYpbc55If1_48P_Ag:b-i5kr6c) ([raw file](https://github.com/TraceMachina/nativelink/blob/62cf43dd8f1132470a365e87c101cac1612d56a4/nativelink-util/src/connection_manager.rs#L442)):*

I think this should be `warn!` or `error!`, is it expected to be noisy?
</blockquote></details>

Because a `Channel` is cloned, when it terminates all clones will report the error and thus this will be very noisy.  Instead, we log it when it reconnects up in `connect_endpoint` when it's resolved down to a single reconnection, the issue is that the error message has been lost by that context, so it may be helpful to have this at debug level at times.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and pending CI: docker-compose-compiles-nativelink (20.04), integration-tests (20.04), integration-tests (22.04), and 15 discussions need to be resolved

chrisstaite:

Moved `pending_channels` to passing the channel to the `Connection` on creation and then passing it back on `Connected` or `Dropped` using `.take()`.  The down side is that we don't have visibility of the connection if a transport error happens on one stream first while it's pending in another `Connection`, however the error should occur again in the other `Connection` and resolve itself, so I don't see it as an issue.
</blockquote></details>

So, it looks like some of this can be implemented by:
```
let (channel, tx) = tonic::Channel::balance_channel(1024);
for (index, endpoint) in endpoints.enumerate() {
    for connection in 0..connections {
        let _ = tx.try_send(Change::Insert(index * connections + connection, endpoint));
    }
}
```
Which will make multiple connections to each `Endpoint`.

Then we have three issues left:
 - Limit the number of connections (counting the number of `Channel`s that are in use)
 - Don't start a new channel while an existing stream is being set up (moving back to `available` after `poll_ready` completes)
 - Reconnect on transport error during the service execution (looks like that is an issue here: https://github.com/hyperium/tonic/blob/eeb3268f71ae5d1107c937392389db63d8f721fb/tonic/src/transport/service/reconnect.rs#L215 in that it simply forwards errors unless they happened during the connection stage)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806)* status: 1 of 1 LGTMs obtained, and 1 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/806#-NuvNoZf8FR6NQNQ6BJk:-NuvNoZf8FR6NQNQ6BJl:b-hv3kf0) (no related file):*
FYI, sorry for the delay on this. I'll make sure we get everything done for this this week. I subconsciously was avoiding giving an lgtm  until after we cut the `0.3.0` release. It should be out in the next day or two and then I'm happy to land this.

Again sorry, these level of refactors are just super scary to me :-( 

___
___
*[`nativelink-util/src/grpc_utils.rs` line 94 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/806#-Nu6JiT_CAF_5eWATRfC:-NuU_nO5CGeHo_CAQVip:bcoppek) ([raw file](https://github.com/TraceMachina/nativelink/blob/5850758dd64cd27f7d073ec1b656afffc11c1c2f/nativelink-util/src/grpc_utils.rs#L94)):*

I think I'll leave this as-is since the std HashMap is now hashbrown anyway.
</blockquote></details>

:exploding_head:

allada (approved):
:lgtm:

Sorry about the delay. Now that 0.3.0 is out, I'm happy to land this.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806#-:-Nv7pImR4UnVq1upaWz6:b3uo8y2)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

chrisstaite-menlo:
Just waiting on @adam-singer

adam-singer (approved):
:lgtm:

> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/806#-:-Nv9MPDZ47SnBM57mpGv:bnfp4nl)* status: <span>:shipit:</span> complete! 3 of 1 LGTMs obtained

PR #893, opened 2024-04-27
update bazel version on the docs

### LINKED ISSUE #892: Quickstart docs are outdated
Bazel version is outdated.
my 2 cents will be to concentrate the documentation in one place because the readme and the quickstark will keep drifting apart

chinchaun:
# Description
Updated the Bazel version to run nativelink

Fixes #892 

## Type of change

- [ ] ~Bug fix (non-breaking change which fixes an issue)~
- [ ] ~New feature (non-breaking change which adds functionality)~
- [ ] ~Breaking change (fix or feature that would cause existing functionality to
  not work as expected)~
- [ ] ~This change requires a documentation update~

## How Has This Been Tested?

## Checklist

- [ ] ~Updated documentation if needed~
- [ ] ~ Tests added/amended~

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893)

### DIFF
--- a/nativelink-docs/docs/Quickstart.mdx
+++ b/nativelink-docs/docs/Quickstart.mdx
@@ -85,13 +85,13 @@ apt update
 apt install apt-transport-https curl gnupg
 
 # Download the Bazel binary for ARM Linux.
-curl -LO https://github.com/bazelbuild/bazel/releases/download/6.4.0/bazel-6.4.0-linux-arm64
+curl -LO https://github.com/bazelbuild/bazel/releases/download/7.0.2/bazel-7.0.2-linux-arm64
 
 # Make the binary executable.
-chmod +x bazel-6.4.0-linux-arm64
+chmod +x bazel-7.0.2-linux-arm64
 
 # Move the binary to a directory in your PATH.
-mv bazel-6.4.0-linux-arm64 /usr/local/bin/bazel
+mv bazel-7.0.2-linux-arm64 /usr/local/bin/bazel
 
 # Verify the installation.
 bazel --version
@@ -115,13 +115,13 @@ apt update
 apt install apt-transport-https curl gnupg
 
 # Download the Bazel binary for x86 Linux.
-curl -LO https://github.com/bazelbuild/bazel/releases/download/6.4.0/bazel-6.4.0-linux-x86_64
+curl -LO https://github.com/bazelbuild/bazel/releases/download/7.0.2/bazel-7.0.2-linux-x86_64
 
 # Make the binary executable.
-chmod +x bazel-6.4.0-linux-x86_64
+chmod +x bazel-7.0.2-linux-x86_64
 
 # Move the binary to a directory in your PATH.
-mv bazel-6.4.0-linux-x86_64 /usr/local/bin/bazel
+mv bazel-7.0.2-linux-x86_64 /usr/local/bin/bazel
 
 # Verify the installation.
 bazel --version
@@ -174,7 +174,7 @@ and `worker`.
 
 **Build requirements:**
 
-* Bazel 6.4.0+
+* Bazel 7.0.2
 * A recent C++ toolchain with LLD as linker
 
 :::tip


aaronmondal (changes requested):
Thanks for the PR :heart:

I believe to avoid drift in the future @blakehatch is working on auto-syncing this file with the `README.md`. Maybe it would make sense to blindly copy-paste the current readme to this file and adjust it to work in the `mdx`?

cc @nfarah86 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893#-:-NwZtKSZEbYmFRJKLRBe:b-xdpoeh)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZrDJH33Tjs8cCQfnu):*
nit: Capitalized `Update`.
___
*[`nativelink-docs/docs/Quickstart.mdx` line 88 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZrK6R7QQcbIsWJFZV:-NwZrK6R7QQcbIsWJFZW:b-csjyfi) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L88)):*
> ```MDX
> 
> # Download the Bazel binary for ARM Linux.
> curl -LO https://github.com/bazelbuild/bazel/releases/download/7.1.1/bazel-7.1.1-linux-arm64
> ```

Oh damn this section is indeed very outdated.

Since we have a `.bazelversion` in the root of the repository which currently uses `7.0.2`, I'm not sure whether a build with `7.1.1` would work.

Maybe we could make this more generic by suggesting to install [`bazelisk`](https://github.com/bazelbuild/bazelisk) instead:

E.g. something like:

```bash
go install \
  github.com/bazelbuild/bazelisk@c7c6c19799ff408c48bdce6b7557217ad0050b17
```
___
*[`nativelink-docs/docs/Quickstart.mdx` line 177 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZt2gs8dLB0HkWMzqf:-NwZt2gt9UsZyFkvRGzf:b69d7tr) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L177)):*
> ```MDX
> **Build requirements:**
> 
> * Bazel 7.0.0+
> ```

This is `7.0.2` at the moment (without `+`, as we only support exactly one Bazel version at the moment)

chinchaun (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-docs/docs/Quickstart.mdx` line 177 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZt2gs8dLB0HkWMzqf:-Nw_GnoI5QciiA1P91Pl:b-n17fyd) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L177)):*

This is `7.0.2` at the moment (without `+`, as we only support exactly one Bazel version at the moment)
</blockquote></details>

I use bazel-7.1.1-linux-x86_64 inside docker in Mac to run the quickstart if that is helpful.
FYI I got this error in the logs, should I open a bug or is related on the new version of bazel?



_Code snippet:_
```MDX
 2024-04-23T14:39:27.860558Z ERROR nativelink_worker::running_actions_manager: Command returned exit code 127 :  /usr/bin/env: 'python3': No such file or directory

    at nativelink-worker/src/running_actions_manager.rs:1108 on tokio-runtime-worker ThreadId(13)

  2024-04-23T14:39:28.771583Z ERROR nativelink: Failed running service : hyper::Error(Io, Kind(NotConnected))
    at src/bin/nativelink.rs:737 on tokio-runtime-worker ThreadId(9)

  2024-04-23T14:40:08.428907Z ERROR nativelink_worker::running_actions_manager: Command returned exit code 127 :  /usr/bin/env: 'python3': No such file or directory

    at nativelink-worker/src/running_actions_manager.rs:1108 on tokio-runtime-worker ThreadId(17)

  2024-04-23T14:40:08.638775Z ERROR nativelink: Failed running service : hyper::Error(Io, Kind(NotConnected))
    at src/bin/nativelink.rs:737 on tokio-runtime-worker ThreadId(12)

  2024-04-23T14:40:08.639122Z ERROR nativelink: Failed running service : hyper::Error(Io, Kind(NotConnected))
    at src/bin/nativelink.rs:737 on tokio-runtime-worker ThreadId(12)
```

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-docs/docs/Quickstart.mdx` line 177 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZt2gs8dLB0HkWMzqf:-Nw_T8VB6Jltu2lxw3XH:b-7z8xxn) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L177)):*
<details><summary><i>Previously, chinchaun (Jean Delannoy) wrote…</i></summary><blockquote>

I use bazel-7.1.1-linux-x86_64 inside docker in Mac to run the quickstart if that is helpful.
FYI I got this error in the logs, should I open a bug or is related on the new version of bazel?
</blockquote></details>

Could you try invoking the command with `--keep_going`? It could be that it's just the python test that fails because of path mismatch between linux and mac.

But it could also be related to Bazel-7.1.1. I believe there were a bunch of changes around mac and sandboxing recently, so that could also be causing it.

chinchaun (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-docs/docs/Quickstart.mdx` line 88 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZrK6R7QQcbIsWJFZV:-Nw_Ii-l7boevospAbhA:b-8zfbag) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L88)):*

Oh damn this section is indeed very outdated.

Since we have a `.bazelversion` in the root of the repository which currently uses `7.0.2`, I'm not sure whether a build with `7.1.1` would work.

Maybe we could make this more generic by suggesting to install [`bazelisk`](https://github.com/bazelbuild/bazelisk) instead:

E.g. something like:

```bash
go install \
  github.com/bazelbuild/bazelisk@c7c6c19799ff408c48bdce6b7557217ad0050b17
```
</blockquote></details>

That will create the need to install go and bazelisk in your system and that can lead to some friction in trying the quickstart for the first time, what are your thoughts on that?

aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-docs/docs/Quickstart.mdx` line 88 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZrK6R7QQcbIsWJFZV:-NwfgVf_6NOIUFECMzK1:b-tnrn4e) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L88)):*
<details><summary><i>Previously, chinchaun (Jean Delannoy) wrote…</i></summary><blockquote>

That will create the need to install go and bazelisk in your system and that can lead to some friction in trying the quickstart for the first time, what are your thoughts on that?
</blockquote></details>

Hmm makes sense. There are many ways to install bazelisk, but for someone new to Bazel it could be confusing why one should use another tool (and create aliases) to actually invoke bazel.

The more I think about this the more it seems to me like the only variant that is guaranteed to work would be to recommend installing nix. Our nix environment is very robust, continuously tested in CI and required for the more advanced examples (like the K8s examples) anyways.

For a long time it was a bit complicated to set Nix up, but the new Rust-based installer makes this a simple copy-paste operation which will automatically get the exact Bazel version we need and prepare the user for any additional examples:

```
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install

nix develop
```

This will also give the user `cargo` and the `native` command (and cloud CLIs like `native`).

This won't make the Bazel build work initially, but I can add exceptions for MacOS to exclude the python targets which we know would fail (more ideally I'd just fix the tests on Mac, but no idea how much effort that would take).

Ref: https://github.com/NixOS/experimental-nix-installer

chinchaun:
> Please rebase this PR on latest. It seems to have pulled in a lot of changes that don't seem correct.

Yes, I made a mistake with the rebasing, sorry about that I'm going to take a look in a bit

aaronmondal (changes requested):
xref https://github.com/TraceMachina/nativelink/pull/912

> Reviewed 208 of 208 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/893#-:-NxpAJmf8pzQzZ1XTzE-:b-2hhy3d)* status: 0 of 1 LGTMs obtained, and 2 discussions need to be resolved

___
*[`nativelink-docs/docs/Quickstart.mdx` line 88 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZrK6R7QQcbIsWJFZV:-Nxp2y7aANMza765CEQG:b3wbts0) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L88)):*
<details><summary><i>Previously, chinchaun (Jean Delannoy) wrote…</i></summary><blockquote>

ok, if you think this is the best approach, we can go this route, maybe we can do it in another PR and start drafting the docs with Nix?
</blockquote></details>

Yeah I think this is best. I understand that it can be very confusing how we handle builds. Let me try to clarify:

- The only build that makes it into production images is Cargo wrapped in nix, but not via the nix shell but via the `nix build .#nativelink` target. This target is fully reproducible and uses a musl toolchain. This build produces a static binary with very consistent behavior on all systems. This executable is also the one that ends up in the published nativelink images (and in fact these images only contain that static binary)

The `nix build .#nativelink` command only works on MacOS and Linux and should be the preferred method for anything you want to run locally.

The reason we don't use it during development is because this build is very slow and doesn't support incremental rebuilds.

- The bazel build technically only exists to make development more convenient and because it makes for a nice dev experience to have sanitizer targets (`--config=asan`) and to run clippy/rustfmt checks on `bazel test`. These builds are not guaranteed to be  reproducible though, e.g. they leak the host's glibc into the build. So they're good for development and dogfooding purposes but not ideal for production deployments.

- When you run nix develop you'll get access to the local remote execution builds which can be enabled via `--config=lre`. This is (to the best of my knowledge) a novel attempt at making the Bazel build truly reproducible. The issue with these builds is that because they link against glibc2.39 from nixpkgs that the executable (and often the build itself) with the `lre` config don't work unless you use a rolling release distro like Gentoo or NixOS (and maybe the latest ubuntu?). So we can't really recommend those for a "generic" setup.

- IIRC the Bazel build doesn't work on MacOS when in the nix devenv :sweat_smile:, so cargo seems like the better choice there (it's fairly high prio to get nix+MacOS+bazel working, but turned out to be a bit trickier than anticipated)

- There are also manual `cargo build` builds. Both in the `nix develop` shell and without it. These use the same cargo setup as the `nix build` executables but again have the potential to leak host information into the build. Convenient for development but not suitable for production builds.

- The general philosophy is to have a code layout that is compatible with Bazel, Cargo and Buck2. We're not running Buck2 builds yet, but we do plan to do this at some point and have a "native" non-nix-wrapped version and a nix-wrapped version. Like the manual Bazel and Cargo builds this will be mostly for dogfooding purposes.
___
*[`nativelink-docs/docs/Quickstart.mdx` line 177 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/893#-NwZt2gs8dLB0HkWMzqf:-Nxp8SnxA2LpOI-Tw7R2:bip50o2) ([raw file](https://github.com/TraceMachina/nativelink/blob/71063e712143173838cab1b7dfab733393c6a7f9/nativelink-docs/docs/Quickstart.mdx#L177)):*
<details><summary><i>Previously, chinchaun (Jean Delannoy) wrote…</i></summary><blockquote>

I tried with Bazel 7.1.1 and 7.0.2 and I got the same error
</blockquote></details>

Oh I totally misread this log initially. It's complaining that it can't find `python3` in the environment that you're executing into. It looks like this is caused by some script that uses a shebang like this:

```bash
#!/usr/bin/env python3
```

If the `env` that you're calling doesn't have `python3` in its PATH it causes the error you're seeing. Some things that might help fix this:

1. Modify the correct PATH to include a `python3`.
2. Change the shebang to point to a python interpreter directly (eg `#!/usr/bin/python3`)
3. Disable the failing target (via `select` or commenting it out). If you're testing this with the nativelink tests, I believe the only target that requires python is the update_protos test: https://github.com/TraceMachina/nativelink/blob/8a632953b86395088e4ab8c1e160a650739549b7/nativelink-proto/BUILD.bazel#L95-L105

MarcusSorealheis:
@chinchaun Can you please wrap this up?

aaronmondal:
Superceeded by:

- #1127

PR #911, opened 2024-05-13
Conclude bzlmod migration

aaronmondal:
This removes the obsolete WORKSPACE file.

Starting with this commit, the MODULE.bazel.lock file is checked in to the repository to lock down the Bazel-side requirements.

Migrates from the Nix-packaged Bazel to a Nix-packaged bazelisk which fetches the correct `.bazelversion`. This means that we're no longer dependent on the Bazel from nixpkgs and can update this dependency more freely.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/911)


aaronmondal:
Needs more testing.

Issue #931, opened 2024-05-26
Move EvictingMap to it's own crate


allada:
`EvictingMap` is pretty abstract and does not have any required hard dependencies on the rest of NativeLink. In addition this kind of map is likely useful to other people. Due to this, we should remove the remaining dependencies and publish it as an official crate.

PR #932, opened 2024-05-26
Refactor EvictingMap so it does not use DigestInfo

allada:
In prep to move to an abstract key instead of DigestInfo, this change cleans up EvictingMap and removes dependencies on DigestInfo.

towards: #931

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932)

### DIFF
--- a/nativelink-store/src/existence_cache_store.rs
+++ b/nativelink-store/src/existence_cache_store.rs
@@ -43,7 +43,7 @@ impl LenEntry for ExistanceItem {
 
 pub struct ExistenceCacheStore {
     inner_store: Arc<dyn Store>,
-    existence_cache: EvictingMap<ExistanceItem, SystemTime>,
+    existence_cache: EvictingMap<DigestInfo, ExistanceItem, SystemTime>,
 }
 
 impl ExistenceCacheStore {
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -401,13 +401,13 @@ pub fn digest_from_filename(file_name: &str) -> Result<DigestInfo, Error> {
 const SIMULTANEOUS_METADATA_READS: usize = 200;
 
 async fn add_files_to_cache<Fe: FileEntry>(
-    evicting_map: &EvictingMap<Arc<Fe>, SystemTime>,
+    evicting_map: &EvictingMap<DigestInfo, Arc<Fe>, SystemTime>,
     anchor_time: &SystemTime,
     shared_context: &Arc<SharedContext>,
     block_size: u64,
 ) -> Result<(), Error> {
     async fn process_entry<Fe: FileEntry>(
-        evicting_map: &EvictingMap<Arc<Fe>, SystemTime>,
+        evicting_map: &EvictingMap<DigestInfo, Arc<Fe>, SystemTime>,
         file_name: &str,
         atime: SystemTime,
         data_size: u64,
@@ -519,7 +519,7 @@ async fn prune_temp_path(temp_path: &str) -> Result<(), Error> {
 
 pub struct FilesystemStore<Fe: FileEntry = FileEntryImpl> {
     shared_context: Arc<SharedContext>,
-    evicting_map: Arc<EvictingMap<Arc<Fe>, SystemTime>>,
+    evicting_map: Arc<EvictingMap<DigestInfo, Arc<Fe>, SystemTime>>,
     block_size: u64,
     read_buffer_size: usize,
     sleep_fn: fn(Duration) -> Sleep,
--- a/nativelink-store/src/memory_store.rs
+++ b/nativelink-store/src/memory_store.rs
@@ -51,7 +51,7 @@ impl LenEntry for BytesWrapper {
 }
 
 pub struct MemoryStore {
-    evicting_map: EvictingMap<BytesWrapper, SystemTime>,
+    evicting_map: EvictingMap<DigestInfo, BytesWrapper, SystemTime>,
 }
 
 impl MemoryStore {
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -12,26 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Borrow;
+use std::cmp::Eq;
 use std::fmt::Debug;
+use std::hash::Hash;
 use std::ops::DerefMut;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use async_lock::Mutex;
 use async_trait::async_trait;
-use futures::stream::FuturesUnordered;
-use futures::{future, join, StreamExt};
 use lru::LruCache;
 use nativelink_config::stores::EvictionPolicy;
 use serde::{Deserialize, Serialize};
 use tracing::{event, Level};
 
-use crate::common::DigestInfo;
 use crate::metrics_utils::{CollectorState, Counter, CounterWithTime, MetricsComponent};
 
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
-pub struct SerializedLRU {
-    pub data: Vec<(DigestInfo, i32)>,
+pub struct SerializedLRU<K> {
+    pub data: Vec<(K, i32)>,
     pub anchor_time: u64,
 }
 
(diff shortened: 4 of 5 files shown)


allada (commented):
+@adam-singer  or +@zbirenbaum 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932#-:-NyqdUls6TmYmhDUP86C:b7hsfod)* status: 0 of 2 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Bazel Dev / ubuntu-22.04, Publish image, Vercel, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, vale (waiting on @adam-singer and @zbirenbaum)

zbirenbaum (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932)* status: 0 of 2 LGTMs obtained, and 3 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-util/src/evicting_map.rs` line 194 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-NyqnOwT9hoUxLzDc8Bt:-NyqnOwT9hoUxLzDc8Bu:b72ayax) ([raw file](https://github.com/TraceMachina/nativelink/blob/b7b58a7af3378d14780970f39e918e9d64131777/nativelink-util/src/evicting_map.rs#L194)):*
> ```XML
>     }
> 
>     pub async fn build_lru_index(&self) -> SerializedLRU {
> ```

It doesn't look like there were any references to these functions in the codebase so removing instead of porting is probably the right move, but it was a bit confusing at first glance. I think it might be better to do two PRs to keep this diff clean and have a place to provide context on what build_lru_index and restore_lru were initially for if it's at all relevant. I can quickly approve the removal one first and we can PR the rest on top of it.
___
*[`nativelink-util/src/evicting_map.rs` line 253 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-NyqzbP37ilYL4giKNEe:-NyqzbP4DuEMoeKCQj1O:b-t9zixi) ([raw file](https://github.com/TraceMachina/nativelink/blob/b47c9fee23abad54b83a71d03e743253d38eaace/nativelink-util/src/evicting_map.rs#L253)):*
> ```XML
> 
>     /// Return the size of a `key`, if not found `None` is returned.
>     pub async fn size_for_key(&self, key: &impl Borrow<K>) -> Option<usize> {
> ```

nit: you don't need the `impl Borrow<K>` or `k.borrow()` here
```
    pub async fn size_for_key(&self, key: &K) -> Option<usize> {
        let mut results = [None];
        self.sizes_for_keys([key], &mut results[..]).await;
        results[0]
    }
```
___
*[`nativelink-util/src/evicting_map.rs` line 276 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-Nyqu4sL9l_0wSHy_SX2:-Nyqu4sL9l_0wSHy_SX3:b-v4ec15) ([raw file](https://github.com/TraceMachina/nativelink/blob/b47c9fee23abad54b83a71d03e743253d38eaace/nativelink-util/src/evicting_map.rs#L276)):*
> ```XML
>                     // Since we are not inserting anythign we don't need to evict based
>                     // on the size of the store.
>                     let should_evict = self.should_evict(lru_len, entry, 0, u64::MAX);
> ```

I went down a rabbit hole refactoring this to be a bit clearer and ideally delegate each entry to another function but it's out of the scope of this PR and possibly not worthwhile. I do think this should be commented stating why getting the size for the keys implicitly evicts and that it uses the key rather than just peeking it.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932)* status: 0 of 2 LGTMs obtained, and 3 discussions need to be resolved (waiting on @adam-singer and @zbirenbaum)

___
*[`nativelink-util/src/evicting_map.rs` line 194 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-NyqnOwT9hoUxLzDc8Bt:-NyvdvO-7w33UMFne9qq:b-svsnli) ([raw file](https://github.com/TraceMachina/nativelink/blob/b7b58a7af3378d14780970f39e918e9d64131777/nativelink-util/src/evicting_map.rs#L194)):*

It doesn't look like there were any references to these functions in the codebase so removing instead of porting is probably the right move, but it was a bit confusing at first glance. I think it might be better to do two PRs to keep this diff clean and have a place to provide context on what build_lru_index and restore_lru were initially for if it's at all relevant. I can quickly approve the removal one first and we can PR the rest on top of it.
</blockquote></details>

Normally I would agree, but in this case it's a refactor anyway, and as long as we are not introducing features with modifications it should generally be considered ok.

If these functions were used somewhere in the code base it would not have been able to compile or a test would have failed (if it was referenced outside code base).

As a rule of thumb, deleting code is generally considered ok. I don't see how it makes code review more difficult, since it's just removal without modifications related to it.
___
*[`nativelink-util/src/evicting_map.rs` line 253 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-NyqzbP37ilYL4giKNEe:-NyveYnhCt3-FKg5L-6s:b-ljlq17) ([raw file](https://github.com/TraceMachina/nativelink/blob/b47c9fee23abad54b83a71d03e743253d38eaace/nativelink-util/src/evicting_map.rs#L253)):*

nit: you don't need the `impl Borrow<K>` or `k.borrow()` here
```
    pub async fn size_for_key(&self, key: &K) -> Option<usize> {
        let mut results = [None];
        self.sizes_for_keys([key], &mut results[..]).await;
        results[0]
    }
```
</blockquote></details>

I'm using `Borrow`, so the caller doesn't need to actually need to be a `K` and instead just needs a way to reference into `K`. For example, we might implement:
```rs
impl Borrow<DigestInfo> for ActionInfo {
  fn borrow(a: &ActionInfo) -> &DigestInfo {
    a.digest
  }
}
```

It's following the same api that `HashMap` is following:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932)* status: 0 of 2 LGTMs obtained, and 3 discussions need to be resolved (waiting on @adam-singer and @zbirenbaum)

___
*[`nativelink-util/src/evicting_map.rs` line 276 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-Nyqu4sL9l_0wSHy_SX2:-NyvgMqi6z1dEkjaguTM:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/b47c9fee23abad54b83a71d03e743253d38eaace/nativelink-util/src/evicting_map.rs#L276)):*

I went down a rabbit hole refactoring this to be a bit clearer and ideally delegate each entry to another function but it's out of the scope of this PR and possibly not worthwhile. I do think this should be commented stating why getting the size for the keys implicitly evicts and that it uses the key rather than just peeking it.
</blockquote></details>

Done.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932)* status: 0 of 2 LGTMs obtained, and 1 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-util/src/evicting_map.rs` line 253 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/932#-NyqzbP37ilYL4giKNEe:-NywUYiH5PZNAXt0hRn9:b-ibmgos) ([raw file](https://github.com/TraceMachina/nativelink/blob/b47c9fee23abad54b83a71d03e743253d38eaace/nativelink-util/src/evicting_map.rs#L253)):*

I'm using `Borrow`, so the caller doesn't need to actually need to be a `K` and instead just needs a way to reference into `K`. For example, we might implement:
```rs
impl Borrow<DigestInfo> for ActionInfo {
  fn borrow(a: &ActionInfo) -> &DigestInfo {
    a.digest
  }
}
```

It's following the same api that `HashMap` is following:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get
</blockquote></details>

Got it, in that case it looks good

zbirenbaum (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932#-:-NywUkpu-GwiDYPAbKfb:bg7igkv)* status: 1 of 2 LGTMs obtained, and 1 discussions need to be resolved (waiting on @adam-singer)

zbirenbaum (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932#-:-NywUomC-cMWC7D2yC2N:bnfp4nl)* status: 1 of 2 LGTMs obtained, and 1 discussions need to be resolved (waiting on @adam-singer)

allada (commented):
-@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/932#-:-NywV1AHEHlHLEbgqCeV:b-sd1e3m)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #937, opened 2024-05-29
Add Operation State API

zbirenbaum:
# Description

Introduces nativelink-state and operation_state_manager API

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/937)

PR #927, opened 2024-05-22
Include UUID in ActionState

zbirenbaum:
# Description

Changes the field used for identifying which action an ActionState corresponds to include a uuid as well as the ActionInfoHashKey. All existing functionality is kept the same by making use of the nested ActionInfoHashKey contained within the Id. These changes will provide the basis for all usage of Id in followup changes to the scheduler.  This breaks compatibility for forwarding an operation from one remote execution system to another that does not use our operation name format (ie: very unlikely, but possible).

Please delete options that aren't relevant.
- [x] Refactor
- [x] Breaking Change 
## How Has This Been Tested?

All existing tests have been modified to remain functionally equivalent.

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/927)

PR #941, opened 2024-05-29
Redis wrapper types

zbirenbaum:
PR for early feedback. Main issues here are:
1. We could just throw a Serialize, Deserialize on all these structs instead of going through this effort and having to maintain them in two places.
2. PlatformProperties is going to take some work to get to dynamically serialize properly. We ideally want to be able to get it in the form of: `property_name: { value_type: [Exact|Minimum|...], value: [String|u64] }` but that's deceptively difficult to go to and from. Its also impossible to do from the proto's since the way things are currently done the platform property manager resolves those types dynamically.

Relies on #927  #937

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/941)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,6 +699,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
 
 [[package]]
+name = "bitmap"
+version = "3.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e97e87564167fab42092c423e491e0fbb77b04ae02b5a78559a1396c052e01e9"
+
+[[package]]
 name = "blake3"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


adam-singer (commented):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/941)* status: 0 of 1 LGTMs obtained, and pending CI: Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Publish image, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, ubuntu-20.04 / stable, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved

___
*[`nativelink-util/src/action_messages.rs` line 107 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/941#-Nz4wkcR7zX8wCJmQ8fN:-Nz4wkcR7zX8wCJmQ8fO:bil00wj) ([raw file](https://github.com/TraceMachina/nativelink/blob/f6620ecd805741a553ed35939aa5af823d2f50d9/nativelink-util/src/action_messages.rs#L107)):*
> ```XML
>     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
>         f.write_fmt(format_args!(
>             "{}:{}",
> ```

Would it be useful to also display the hash code?

zbirenbaum:
Outdated and alternative solution was merged. Closing

PR #981, opened 2024-06-07
Add a config option to prefix keys in Redis stores

### LINKED ISSUE #977: Redis store should support prefixing
We should have a way to prefix redis stores so when keys are inserted all keys get some prefix based on a config.

caass:
# Description

Add a configuration option that allows for setting an optional prefix to all the keys used with a Redis store. I added some documentation to the `RedisStore` configuration struct, but I'm not sure if there's docs I should update elsewhere.

Fixes #977

## Type of change

## How Has This Been Tested?

I added duplicates of existing tests of the Redis store to use prefixes

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -879,6 +879,15 @@ pub struct RedisStore {
     /// Default: 10
     #[serde(default)]
     pub connection_timeout_s: u64,
+
+    /// An optional prefix to prepend to all keys in this store.
+    ///
+    /// Setting this value can make it convenient to query or
+    /// organize your data according to the shared prefix.
+    ///
+    /// Default: (Empty String / No Prefix)
+    #[serde(default)]
+    pub key_prefix: String,
 }
 
 /// Retry configuration. This configuration is exponential and each iteration
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -28,7 +28,7 @@ use nativelink_util::health_utils::{HealthRegistryBuilder, HealthStatus, HealthS
 use nativelink_util::metrics_utils::{Collector, CollectorState, MetricsComponent, Registry};
 use nativelink_util::store_trait::{StoreDriver, StoreKey, UploadSizeInfo};
 use redis::aio::{ConnectionLike, ConnectionManager};
-use redis::AsyncCommands;
+use redis::{AsyncCommands, ToRedisArgs};
 
 use crate::cas_utils::is_zero_digest;
 
@@ -44,6 +44,11 @@ pub enum LazyConnection<T: ConnectionLike + Unpin + Clone + Send + Sync> {
 pub struct RedisStore<T: ConnectionLike + Unpin + Clone + Send + Sync = ConnectionManager> {
     lazy_conn: ArcCell<LazyConnection<T>>,
     temp_name_generator_fn: fn() -> String,
+
+    /// A common prefix to append to all keys before they are sent to Redis.
+    ///
+    /// See [`RedisStore::key_prefix`](`nativelink_config::stores::RedisStore::key_prefix`).
+    key_prefix: String,
 }
 
 impl RedisStore {
@@ -77,9 +82,10 @@ impl RedisStore {
 
         let lazy_conn = LazyConnection::Future(conn_fut);
 
-        Ok(RedisStore::new_with_conn_and_name_generator(
+        Ok(RedisStore::new_with_conn_and_name_generator_and_prefix(
             lazy_conn,
             || uuid::Uuid::new_v4().to_string(),
+            config.key_prefix.clone(),
         ))
     }
 }
@@ -89,9 +95,22 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync> RedisStore<T> {
         lazy_conn: LazyConnection<T>,
         temp_name_generator_fn: fn() -> String,
     ) -> RedisStore<T> {
+        RedisStore::new_with_conn_and_name_generator_and_prefix(
+            lazy_conn,
+            temp_name_generator_fn,
+            String::new(),
+        )
+    }
+
+    pub fn new_with_conn_and_name_generator_and_prefix(
+        lazy_conn: LazyConnection<T>,
+        temp_name_generator_fn: fn() -> String,
+        key_prefix: String,
+    ) -> RedisStore<T> {
         RedisStore {
             lazy_conn: ArcCell::new(Arc::new(lazy_conn)),
             temp_name_generator_fn,
+            key_prefix,
         }
     }
 
@@ -103,6 +122,30 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync> RedisStore<T> {
         self.lazy_conn
             .set(Arc::new(LazyConnection::Connection(result.clone())));
         result
+    }
+
+    /// Encode a [`StoreKey`] so it can be sent to Redis.
+    fn encode_key(&self, key: StoreKey) -> impl ToRedisArgs {
+        // TODO(caass): Once https://github.com/redis-rs/redis-rs/pull/1219 makes it into a release,
+        // this can be changed to
+        // ```rust
+        // if self.key_prefix.is_empty() {
+        //   key.as_str()
+        // } else {
+        //   let mut encoded_key = String::with_capacity(self.key_prefix.len() + key_body.len());
+        //   encoded_key.push_str(&self.key_prefix);
+        //   encoded_key.push_str(&key_body);
+        //   Cow::Owned(encoded_key)
+        // }
+        //```
+        // to avoid an allocation
+        let key_body = key.as_str();
+
+        let mut encoded_key = String::with_capacity(self.key_prefix.len() + key_body.len());
+        encoded_key.push_str(&self.key_prefix);
+        encoded_key.push_str(&key_body);
+
+        encoded_key
     }
 }
 
@@ -128,7 +171,7 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync + 'static> StoreDriver for 
                 zero_digest_indexes.push(index);
             }
 
-            pipe.strlen(key.as_str().as_ref());
+            pipe.strlen(self.encode_key(key.borrow()));
         });
 
         let digest_sizes = pipe
@@ -182,7 +225,7 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync + 'static> StoreDriver for 
                         return Ok(());
                     }
                     if force_recv {
-                        conn.append(key.as_str().as_ref(), &chunk[..])
+                        conn.append(self.encode_key(key.borrow()), &chunk[..])
                             .await
                             .map_err(from_redis_err)
                             .err_tip(|| "In RedisStore::update() single chunk")?;
@@ -209,7 +252,7 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync + 'static> StoreDriver for 
 
         pipe.cmd("RENAME")
             .arg(temp_key.get_or_init(make_temp_name))
-            .arg(key.as_str().as_ref());
+            .arg(self.encode_key(key));
         pipe.query_async(&mut conn)
             .await
             .map_err(from_redis_err)
@@ -237,7 +280,7 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync + 'static> StoreDriver for 
         let mut conn = self.get_conn().await?;
         if length == Some(0) {
             let exists = conn
-                .exists::<_, bool>(key.as_str().as_ref())
+                .exists::<_, bool>(self.encode_key(key.borrow()))
                 .await
                 .map_err(from_redis_err)
                 .err_tip(|| "In RedisStore::get_part::zero_exists")?;
@@ -264,7 +307,7 @@ impl<T: ConnectionLike + Unpin + Clone + Send + Sync + 'static> StoreDriver for 
             let current_end =
                 std::cmp::min(current_start.saturating_add(READ_CHUNK_SIZE), end_position) - 1;
             let chunk = conn
-                .getrange::<_, Bytes>(key.as_str().as_ref(), current_start, current_end)
+                .getrange::<_, Bytes>(self.encode_key(key.borrow()), current_start, current_end)
                 .await
                 .map_err(from_redis_err)
                 .err_tip(|| "In RedisStore::get_part::getrange")?;
--- a/nativelink-store/tests/redis_store_test.rs
+++ b/nativelink-store/tests/redis_store_test.rs
@@ -126,6 +126,55 @@ async fn upload_and_get_data() -> Result<(), Error> {
 }
 
 #[nativelink_test]
+async fn upload_and_get_data_with_prefix() -> Result<(), Error> {
+    let data = Bytes::from_static(b"14");
+    let prefix = "TEST_PREFIX-";
+
+    let digest = DigestInfo::try_new(VALID_HASH1, 2)?;
+    let packed_hash_hex = format!("{prefix}{}-{}", digest.hash_str(), digest.size_bytes);
+
+    let chunk_data = "14";
+
+    let redis_connection = MockRedisConnectionBuilder::new()
+        .pipe(&[("APPEND", &[TEMP_UUID, chunk_data], Ok(&[redis::Value::Nil]))])
+        .cmd("APPEND", &[&packed_hash_hex, ""], Ok(""))
+        .pipe(&[(
+            "RENAME",
+            &[TEMP_UUID, &packed_hash_hex],
+            Ok(&[redis::Value::Nil]),
+        )])
+        .pipe(&[(
+            "STRLEN",
+            &[&packed_hash_hex],
+            Ok(&[redis::Value::Bulk(vec![redis::Value::Int(2)])]),
+        )])
+        .cmd("GETRANGE", &[&packed_hash_hex, "0", "1"], Ok("14"))
+        .build();
+
+    let store = RedisStore::new_with_conn_and_name_generator_and_prefix(
+        LazyConnection::Connection(Ok(redis_connection)),
+        mock_uuid_generator,
+        prefix.to_string(),
+    );
+
+    store.update_oneshot(digest, data.clone()).await?;
+
+    let result = store.has(digest).await?;
+    assert!(
+        result.is_some(),
+        "Expected redis store to have hash: {VALID_HASH1}",
+    );
+
+    let result = store
+        .get_part_unchunked(digest, 0, Some(data.clone().len()))
+        .await?;
+
+    assert_eq!(result, data, "Expected redis store to have updated value",);
+
+    Ok(())
+}
+
+#[nativelink_test]
 async fn upload_empty_data() -> Result<(), Error> {
     let data = Bytes::from_static(b"");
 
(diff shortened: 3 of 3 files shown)


caass (commented):
+@allada 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981#-:-NzkPd6q8CaBITyGSKZb:bwi8nah)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, and 3 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-config/src/stores.rs` line 889 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkPmhDEZMFJRpUHPEh:-NzkPmhDEZMFJRpUHPEi:b132kl5) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-config/src/stores.rs#L889)):*
> ```XML
>     ///
>     /// Default: `""`
>     pub key_prefix: Option<String>,
> ```

FYI I thought about making this just a `String` with `#[serde(default)]`, but later in the implementation it wound up adding an allocation in cases where there wasn't a prefix.
___
*[`nativelink-store/src/redis_store.rs` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkQDGE5zPe049w2N1H:-NzkQDGE5zPe049w2N1I:b-fzwqs9) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L42)):*
> ```XML
> /// e.g. attach a prefix to every key. Instead, you can use [`RedisStore::key`] to create a
> /// `RedisStoreKey`.
> struct RedisStoreKey<'a, 's> {
> ```

I decided to write a custom newtype here because I felt that it was an appropriate level of abstraction.

In one extreme, we just manually call what became the implementation of `ToRedisArgs` every time we pass a key to Redis. I felt that was unmaintainable -- what if we add other configuration options later that also impact how keys are encoded? It would be pretty nightmarish to manually ensure every time we talked to Redis we ran the logic for encoding the keys properly.

On the other hand, we could define a whole layer on top of the Redis library that enforces at compile time that we only use a processed key (i.e. `RedisStoreKey`) instead of `StoreKey.as_str().as_ref()`, but that felt pretty brittle in the other direction -- what if we add more functionality to the `StoreDriver` trait? Then we'd need to update our intermediate layer as well.

Here, we do define a newtype to use, and describe with doc comments how it's supposed to be used. The way to construct a `RedisStoreKey` is to use `RedisStore::key` (or the `RedisStoreKey { prefix: Some("prefix"), body: Cow::Borrowed("body") }` syntax, which we could prevent by moving this into a child module...), and since it implements `ToRedisArgs` you can pass that directly to calls to Redis.
___
*[`nativelink-store/src/redis_store.rs` line 52 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkTwEj9U-SL6LmHY7B:-NzkTwEj9U-SL6LmHY7C:b8iqird) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L52)):*
> ```XML
>         W: ?Sized + redis::RedisWrite,
>     {
>         let body = self.body.as_bytes();
> ```

In the `Cow::Owned` case, this doesn't actually call `String::as_bytes` -- is calls `String::as_ref().as_bytes`, which delegates to `str::as_bytes`. Ultimately, these are all zero-cost abstractions that come out in the wash, but I thought it was worth mentioning.
___
*[`nativelink-store/src/redis_store.rs` line 52 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUKTv8p83NT7o9iQ0:-NzkUKTv8p83NT7o9iQ1:b-yy3iyq) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L52)):*
> ```XML
>         W: ?Sized + redis::RedisWrite,
>     {
>         let body = self.body.as_bytes();
> ```

Nit: This doesn't need to be its own declaration, we can inline it.
___
*[`nativelink-store/src/redis_store.rs` line 131 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUT0g983nDiEFSdyC:-NzkUT0g983nDiEFSdyD:bgz6cme) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L131)):*
> ```XML
>             lazy_conn,
>             || uuid::Uuid::new_v4().to_string(),
>             config.key_prefix.clone(),
> ```

The alternative to cloning here is to make `RedisStore` dependent on the lifetime of `Config`, which technically should be possible (as a config is likely to live for the rest of the program), but since this isn't in the hot path and only happens when a `RedisStore` is constructed I figure this is better for ergonomics.
___
*[`nativelink-store/src/redis_store.rs` line 160 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUkmaCyYbOxKR99KN:-NzkUkmaCyYbOxKR99KO:b-5hzny1) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L160)):*
> ```XML
> 
>     /// Wrap a [`StoreKey`] in a [`RedisStoreKey`] which can be sent to Redis.
>     fn key<'a, 's, 'k: 's>(&'a self, key: &'k StoreKey<'s>) -> RedisStoreKey<'a, 's> {
> ```

Holy lifetimes...these all need to be explicit so the compiler can work out what's happening, which is that:

1. The key prefix is guaranteed to be valid for the lifetime of the borrow of `self` (the `RedisStore`)
2. The key body is guaranteed to be valid for the lifetime of the `StoreKey` (since it's referencing the `str` slice inside it)
3. The borrow of the `StoreKey` is guaranteed to be valid for at least the lifetime of the `StoreKey` (by the rules of the borrow checker -- since if the lifetime of `StoreKey` could be shorter than its own borrow, we'd get unsoundness).
___
*[`nativelink-store/src/redis_store.rs` line 160 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkVw9t6c6ss74xaHuW:-NzkVw9t6c6ss74xaHuX:b-3zkd7g) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L160)):*
> ```XML
> 
>     /// Wrap a [`StoreKey`] in a [`RedisStoreKey`] which can be sent to Redis.
>     fn key<'a, 's, 'k: 's>(&'a self, key: &'k StoreKey<'s>) -> RedisStoreKey<'a, 's> {
> ```

Maybe this method should be named something else? `encode_key`?

caass (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, and 2 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 52 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUKTv8p83NT7o9iQ0:-NzkWXmNBEaILxaSsPd6:bb6un1j) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L52)):*

Nit: This doesn't need to be its own declaration, we can inline it.
</blockquote></details>

Done

allada (commented):
Overall looks good :-)  

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981#-:-NzkkRZk2KyluEBCXyDt:b471irb)* status: 0 of 1 LGTMs obtained, and 5 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 889 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkPmhDEZMFJRpUHPEh:-NzkibFo1e05gCNT2Gt3:b-ruh5f6) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-config/src/stores.rs#L889)):*

FYI I thought about making this just a `String` with `#[serde(default)]`, but later in the implementation it wound up adding an allocation in cases where there wasn't a prefix.
</blockquote></details>

It shouldn't. Empty strings don't have any data on the heap.

See: https://doc.rust-lang.org/std/string/struct.String.html#method.new
___
*[`nativelink-config/src/stores.rs` line 888 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkiPzI1o7kNsnxMIYY:-NzkiPzI1o7kNsnxMIYZ:b-bp22v2) ([raw file](https://github.com/TraceMachina/nativelink/blob/9aa5782d3018edb33548911be603bb22d3a95f13/nativelink-config/src/stores.rs#L888)):*
> ```XML
>     /// organize your data according to the shared prefix.
>     ///
>     /// Default: `""`
> ```

nit: `Default: (Empty String / No prefix)`
___
*[`nativelink-store/src/redis_store.rs` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkQDGE5zPe049w2N1H:-NzkjdTI62yu7IlWiRMN:bgsd0k4) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L42)):*

I decided to write a custom newtype here because I felt that it was an appropriate level of abstraction.

In one extreme, we just manually call what became the implementation of `ToRedisArgs` every time we pass a key to Redis. I felt that was unmaintainable -- what if we add other configuration options later that also impact how keys are encoded? It would be pretty nightmarish to manually ensure every time we talked to Redis we ran the logic for encoding the keys properly.

On the other hand, we could define a whole layer on top of the Redis library that enforces at compile time that we only use a processed key (i.e. `RedisStoreKey`) instead of `StoreKey.as_str().as_ref()`, but that felt pretty brittle in the other direction -- what if we add more functionality to the `StoreDriver` trait? Then we'd need to update our intermediate layer as well.

Here, we do define a newtype to use, and describe with doc comments how it's supposed to be used. The way to construct a `RedisStoreKey` is to use `RedisStore::key` (or the `RedisStoreKey { prefix: Some("prefix"), body: Cow::Borrowed("body") }` syntax, which we could prevent by moving this into a child module...), and since it implements `ToRedisArgs` you can pass that directly to calls to Redis.
</blockquote></details>

To be honest, I think this is one step too far for readability. This is adding enough complexity that we should just allocate a new string for every key if that's what it requires. I mean, we are already doing an allocation in `BytesMut`, so I don't see a reason to do a string allocation instead.
___
*[`nativelink-store/src/redis_store.rs` line 131 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUT0g983nDiEFSdyC:-Nzkj3UF-rLG0rE2LiaC:bepe7v) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L131)):*

The alternative to cloning here is to make `RedisStore` dependent on the lifetime of `Config`, which technically should be possible (as a config is likely to live for the rest of the program), but since this isn't in the hot path and only happens when a `RedisStore` is constructed I figure this is better for ergonomics.
</blockquote></details>

No, it's a one time cost. It's fine to copy.
___
*[`nativelink-store/src/redis_store.rs` line 160 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUkmaCyYbOxKR99KN:-NzkjDLyBCEeYtYM9IsX:b-afbmvr) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L160)):*

Holy lifetimes...these all need to be explicit so the compiler can work out what's happening, which is that:

1. The key prefix is guaranteed to be valid for the lifetime of the borrow of `self` (the `RedisStore`)
2. The key body is guaranteed to be valid for the lifetime of the `StoreKey` (since it's referencing the `str` slice inside it)
3. The borrow of the `StoreKey` is guaranteed to be valid for at least the lifetime of the `StoreKey` (by the rules of the borrow checker -- since if the lifetime of `StoreKey` could be shorter than its own borrow, we'd get unsoundness).
</blockquote></details>

Use `key: StoreKey<'_>`

`StoreKey` only owns data if it is required. The caller would use `.borrow()` if they don't want to pass ownership of the heap.
___
*[`nativelink-store/src/redis_store.rs` line 92 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkixeF66V94c6mMjVU:-NzkixeF66V94c6mMjVV:b2n4mm4) ([raw file](https://github.com/TraceMachina/nativelink/blob/9aa5782d3018edb33548911be603bb22d3a95f13/nativelink-store/src/redis_store.rs#L92)):*
> ```XML
>     ///
>     /// See [`RedisStore::key_prefix`](`nativelink_config::stores::RedisStore::key_prefix`).
>     key_prefix: Option<String>,
> ```

nit: Use `String` w/o `Option`.
___
*[`nativelink-store/tests/redis_store_test.rs` line 300 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkkLim28nSWUx4ymfl:-NzkkLim28nSWUx4ymfm:baz91un) ([raw file](https://github.com/TraceMachina/nativelink/blob/9aa5782d3018edb33548911be603bb22d3a95f13/nativelink-store/tests/redis_store_test.rs#L300)):*
> ```XML
> 
> #[nativelink_test]
> async fn test_uploading_large_data_with_prefix() -> Result<(), Error> {
> ```

nit: This test is a bit too much. We are testing the important stuff with the above tests.

caass (commented):
@allada Ok, I'll push some changes and re-request review.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981#-:-NzmzVCV5U2otel27ERA:b-99wadb)* status: 0 of 1 LGTMs obtained, and 2 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-config/src/stores.rs` line 889 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkPmhDEZMFJRpUHPEh:-NzmyVLtDGfaAezaBB73:b-gmvpce) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-config/src/stores.rs#L889)):*

It shouldn't. Empty strings don't have any data on the heap.

See: https://doc.rust-lang.org/std/string/struct.String.html#method.new
</blockquote></details>

I misspoke, what I meant is that later in `RedisStoreKey::write_redis_args` we would basically try to copy an empty buffer (i.e. `String.as_bytes()`), which could be avoided if we knew up from the string was empty.
___
*[`nativelink-store/src/redis_store.rs` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkQDGE5zPe049w2N1H:-Nzmz-nG4ZhQ4TrsmaRY:b-h17zeg) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L42)):*

To be honest, I think this is one step too far for readability. This is adding enough complexity that we should just allocate a new string for every key if that's what it requires. I mean, we are already doing an allocation in `BytesMut`, so I don't see a reason to do a string allocation instead.
</blockquote></details>

Got it :+1: I think also an easier way to read this would be to do like `if string.is_empty() { /* just encode the key body */ } else { encode the prefix }`.
___
*[`nativelink-store/src/redis_store.rs` line 131 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkUT0g983nDiEFSdyC:-NzmzI_68Rkw1tbNj8V1:b-8sabn4) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L131)):*

No, it's a one time cost. It's fine to copy.
</blockquote></details>

Yea that's what I figured :+1:
___
*[`nativelink-store/tests/redis_store_test.rs` line 300 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkkLim28nSWUx4ymfl:-NzmzSKl4wSO5EDy1eKW:bb6mt9b) ([raw file](https://github.com/TraceMachina/nativelink/blob/9aa5782d3018edb33548911be603bb22d3a95f13/nativelink-store/tests/redis_store_test.rs#L300)):*

nit: This test is a bit too much. We are testing the important stuff with the above tests.
</blockquote></details>

:+1:

allada (approved):
:lgtm:

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/981#-:-NznL71tA6i9PGtEdH3X:bnfp4nl)* status: 1 of 1 LGTMs obtained, and pending CI: pre-commit-checks

___
*[`nativelink-store/src/redis_store.rs` line 52 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkTwEj9U-SL6LmHY7B:-NznJzfZ1BbmRAAI3dRG:b155y4k) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L52)):*

In the `Cow::Owned` case, this doesn't actually call `String::as_bytes` -- is calls `String::as_ref().as_bytes`, which delegates to `str::as_bytes`. Ultimately, these are all zero-cost abstractions that come out in the wash, but I thought it was worth mentioning.
</blockquote></details>

Yeah, maybe there's a little more stack used, but it should hit L1 cache each time (if the optimizer doesn't optimize it away).
___
*[`nativelink-store/src/redis_store.rs` line 160 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/981#-NzkVw9t6c6ss74xaHuW:-NznKmEPDbh6bkrF-DjE:b-w7jzgw) ([raw file](https://github.com/TraceMachina/nativelink/blob/71f8260e4b89b4498aa3bb2df7678f8da9b51374/nativelink-store/src/redis_store.rs#L160)):*

Maybe this method should be named something else? `encode_key`?
</blockquote></details>

I try to be more explicit, like: `redis_args_for_key`. But not strict on this here.

PR #988, opened 2024-06-11
Add memory store optimized subscription API

### LINKED ISSUE #987: Implement optimized version of subscribe for MemoryStore
Now that the subscribe API is implemented, we should add an optimized variation for memory store.

allada:
This will make memory store much more efficient for cases where subscribe is used.

closes #987

- - -
This change is [Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988)

### DIFF
--- a/nativelink-scheduler/tests/cache_lookup_scheduler_test.rs
+++ b/nativelink-scheduler/tests/cache_lookup_scheduler_test.rs
@@ -50,9 +50,9 @@ struct TestContext {
 
 fn make_cache_scheduler() -> Result<TestContext, Error> {
     let mock_scheduler = Arc::new(MockActionScheduler::new());
-    let ac_store = Store::new(Arc::new(MemoryStore::new(
+    let ac_store = Store::new(MemoryStore::new(
         &nativelink_config::stores::MemoryStore::default(),
-    )));
+    ));
     let cache_scheduler = CacheLookupScheduler::new(ac_store.clone(), mock_scheduler.clone())?;
     Ok(TestContext {
         mock_scheduler,
--- a/nativelink-store/src/default_store_factory.rs
+++ b/nativelink-store/src/default_store_factory.rs
@@ -50,7 +50,7 @@ pub fn store_factory<'a>(
 ) -> Pin<FutureMaybeStore<'a>> {
     Box::pin(async move {
         let store: Arc<dyn StoreDriver> = match backend {
-            StoreConfig::memory(config) => Arc::new(MemoryStore::new(config)),
+            StoreConfig::memory(config) => MemoryStore::new(config),
             StoreConfig::experimental_s3_store(config) => Arc::new(S3Store::new(config).await?),
             StoreConfig::redis_store(config) => Arc::new(RedisStore::new(config)?),
             StoreConfig::verify(config) => Arc::new(VerifyStore::new(
--- a/nativelink-store/src/memory_store.rs
+++ b/nativelink-store/src/memory_store.rs
@@ -12,19 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::fmt::Debug;
 use std::pin::Pin;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::time::SystemTime;
 
 use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
-use nativelink_error::{Code, Error, ResultExt};
+use nativelink_error::{make_err, Code, Error, ResultExt};
 use nativelink_util::buf_channel::{DropCloserReadHalf, DropCloserWriteHalf};
 use nativelink_util::evicting_map::{EvictingMap, LenEntry};
 use nativelink_util::health_utils::{default_health_status_indicator, HealthStatusIndicator};
 use nativelink_util::metrics_utils::{Collector, CollectorState, MetricsComponent, Registry};
-use nativelink_util::store_trait::{StoreDriver, StoreKey, UploadSizeInfo};
+use nativelink_util::store_trait::{
+    StoreDriver, StoreKey, StoreOptimizations, StoreSubscription, StoreSubscriptionItem,
+    UploadSizeInfo,
+};
+use parking_lot::{RwLock, RwLockUpgradableReadGuard};
+use tokio::sync::watch;
+use tracing::{event, Level};
 
 use crate::cas_utils::is_zero_digest;
 
@@ -49,17 +56,91 @@ impl LenEntry for BytesWrapper {
     }
 }
 
+/// A subscription to a key in the MemoryStore.
+struct MemoryStoreSubscription {
+    store: Weak<MemoryStore>,
+    key: StoreKey<'static>,
+    receiver: Option<watch::Receiver<Result<Arc<dyn StoreSubscriptionItem>, Error>>>,
+}
+
+#[async_trait]
+impl StoreSubscription for MemoryStoreSubscription {
+    fn peek(&self) -> Result<Arc<dyn StoreSubscriptionItem>, Error> {
+        self.receiver.as_ref().unwrap().borrow().clone()
+    }
+
+    async fn changed(&mut self) -> Result<Arc<dyn StoreSubscriptionItem>, Error> {
+        self.receiver
+            .as_mut()
+            .unwrap()
+            .changed()
+            .await
+            .map_err(|e| {
+                make_err!(
+                    Code::Internal,
+                    "Sender dropped in DefaultStoreSubscription::changed - {e:?}"
+                )
+            })?;
+        self.receiver.as_ref().unwrap().borrow().clone()
+    }
+}
+
+/// When the subscription is dropped, we need to remove the subscription from the store
+/// to prevent memory leaks.
+impl Drop for MemoryStoreSubscription {
+    fn drop(&mut self) {
+        // Make sure we manually drop receiver first.
+        self.receiver = None;
+        let Some(store) = self.store.upgrade() else {
+            return;
+        };
+        store.remove_dropped_subscription(self.key.borrow().into_owned());
+    }
+}
+
+struct MemoryStoreSubscriptionItem {
+    store: Weak<MemoryStore>,
+    key: StoreKey<'static>,
+}
+
+#[async_trait]
+impl StoreSubscriptionItem for MemoryStoreSubscriptionItem {
+    async fn get_key<'a>(&'a self) -> Result<StoreKey<'a>, Error> {
+        Ok(self.key.borrow())
+    }
+
+    async fn get_part(
+        &self,
+        writer: &mut DropCloserWriteHalf,
+        offset: usize,
+        length: Option<usize>,
+    ) -> Result<(), Error> {
+        let store = self
+            .store
+            .upgrade()
+            .err_tip(|| "Store dropped in MemoryStoreSubscriptionItem::get_part")?;
+        Pin::new(store.as_ref())
+            .get_part(self.key.borrow(), writer, offset, length)
+            .await
+    }
+}
+
+type SubscriptionSender = watch::Sender<Result<Arc<dyn StoreSubscriptionItem>, Error>>;
 pub struct MemoryStore {
+    weak_self: Weak<Self>,
     evicting_map: EvictingMap<StoreKey<'static>, BytesWrapper, SystemTime>,
+    subscriptions: RwLock<HashMap<StoreKey<'static>, SubscriptionSender>>,
 }
 
 impl MemoryStore {
-    pub fn new(config: &nativelink_config::stores::MemoryStore) -> Self {
+    pub fn new(config: &nativelink_config::stores::MemoryStore) -> Arc<Self> {
         let empty_policy = nativelink_config::stores::EvictionPolicy::default();
         let eviction_policy = config.eviction_policy.as_ref().unwrap_or(&empty_policy);
-        MemoryStore {
+        Arc::new_cyclic(|weak_self| MemoryStore {
+            weak_self: weak_self.clone(),
             evicting_map: EvictingMap::new(eviction_policy, SystemTime::now()),
-        }
+            subscriptions: RwLock::new(HashMap::new()),
+        })
     }
 
     /// Returns the number of key-value pairs that are currently in the the cache.
@@ -70,6 +151,21 @@ impl MemoryStore {
 
     pub async fn remove_entry(&self, key: StoreKey<'_>) -> bool {
         self.evicting_map.remove(&key.into_owned()).await
+    }
+
+    /// Tells the store that a subscription has been dropped and gives an opportunity to clean up.
+    fn remove_dropped_subscription(&self, key: StoreKey<'static>) {
+        let mut subscriptions = self.subscriptions.write();
+        let Some(sender) = subscriptions.get(&key) else {
+            event!(
+                Level::ERROR,
+                "Subscription dropped for key '{key:?}' but not found in subscriptions"
+            );
+            return;
+        };
+        if sender.receiver_count() == 0 {
+            subscriptions.remove(&key);
+        }
     }
 }
 
@@ -114,8 +210,15 @@ impl StoreDriver for MemoryStore {
         };
 
         self.evicting_map
-            .insert(key.into_owned(), BytesWrapper(final_buffer))
+            .insert(key.borrow().into_owned(), BytesWrapper(final_buffer))
             .await;
+        {
+            // Notify all subscribers that the key has been modified.
+            let subscription_lock = self.subscriptions.read();
+            if let Some(sender) = subscription_lock.get(&key.into_owned()) {
+                sender.send_modify(|_| { /* We just need to flag it modified. */ });
+            }
+        }
         Ok(())
     }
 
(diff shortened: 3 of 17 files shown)


allada (commented):
+@caass 
+@adam-singer 
fyi: @zbirenbaum 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/988#-:-O-7noW_3GnmoYMN9n67:bl42blb)* status: 0 of 2 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable (waiting on @adam-singer and @caass)

adam-singer (approved):
:lgtm:

> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/988#-:-O-8b1bKDp5FqqLKXy5c:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @caass)

adam-singer (approved):
should use documentation around the new structs introduced 

> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988#-:-O-8b5JHD-EmFT-XmbYq:b-13ovck)* status: 1 of 2 LGTMs obtained (waiting on @caass)

caass (approved):
This looks pretty good, I have a few minor concerns but overall the big question is: do we want to add a dependency on [dashmap](https://docs.rs/dashmap/latest/dashmap/) in order to make the code here more readable?

> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988#-:-O-COdCV0DBQrAle1B3O:b3fxkco)* status: 1 of 2 LGTMs obtained, and pending CI: Installation / macos-13, Publish image, Publish nativelink-worker-lre-cc, macos-13, and 4 discussions need to be resolved (waiting on @caass)

___
*[`nativelink-store/src/memory_store.rs` line 63 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLYBNEukfpAuD5QTr:-O-CLYBNEukfpAuD5QTs:bljfa7n) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L63)):*
> ```XML
>     store: Weak<MemoryStore>,
>     key: StoreKey<'static>,
>     receiver: Option<watch::Receiver<Result<Arc<dyn StoreSubscriptionItem>, Error>>>,
> ```

Nit: can you use the `SubscriptionSender` alias here? And declare it before this struct?
___
*[`nativelink-store/src/memory_store.rs` line 97 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLihZ9F09gOLLF_G9:-O-CLihZ9F09gOLLF_GA:bmag53u) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L97)):*
> ```XML
>             return;
>         };
>         store.subscription_dropped(self.key.borrow().into_owned());
> ```

Minor: I think this is more clear as an `if let` since the logic is "if we are the only handle, notify the store we're dropped". Like basically inverting this to

```rust
if let Some(store) = self.store.upgrade() {
    store.subscription_dropped(self.key.borrow().into_owned());
}
```



_Code quote:_
```XML
        let Some(store) = self.store.upgrade() else {
            return;
        };
        store.subscription_dropped(self.key.borrow().into_owned());
```
___
*[`nativelink-store/src/memory_store.rs` line 132 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CMSZs5kqj8lyZpOi-:-O-CMSZs5kqj8lyZpOi0:b2gre0c) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L132)):*
> ```XML
>     weak_self: Weak<Self>,
>     evicting_map: EvictingMap<StoreKey<'static>, BytesWrapper, SystemTime>,
>     subscriptions: RwLock<HashMap<StoreKey<'static>, SubscriptionSender>>,
> ```

FYI this looks a lot like we want to use [dashmap](https://docs.rs/dashmap/latest/dashmap/), which is basically an auto-sharding replacement for `RwLock<HashMap>` (which leads to lower lock contention). If we're ok adding a dependency, it'd make things simpler on our side since the locking logic would be all in that crate.
___
*[`nativelink-store/src/memory_store.rs` line 157 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CMtoADa3DAjLxIDgH:-O-CMtoADa3DAjLxIDgI:b72chrd) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L157)):*
> ```XML
> 
>     /// Tells the store that a subscription has been dropped and gives an opportunity to clean up.
>     fn subscription_dropped(&self, key: StoreKey<'static>) {
> ```

Nit: should this be more imperative? Like `drop_subscription` instead of `subscription_dropped`? I guess the more technically correct name would be like. `remove_dropped_subscription` or `notify_of_dropped_subscription`. But `subscription_dropped` sounds like it should be a `bool` or a `fn(&self) -> bool` to me.
___
*[`nativelink-store/src/memory_store.rs` line 213 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CNISvBG9KrGhKzsnK:-O-CNISvBG9KrGhKzsnL:b13ajrx) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L213)):*
> ```XML
> 
>         self.evicting_map
>             .insert(key.borrow().into_owned(), BytesWrapper(final_buffer))
> ```

Unrelated to the current PR but I feel like this pattern of `key.borrow()` and `key.borrow().into_owned()` everywhere could probably benefit from using some traits to turn it into a proper smart pointer. Like it really wants to be a version of `Arc`, right? But like an `Arc<enum KeyKind>` basically?

I dunno, I guess it's fine for now but it generates a lot of noise around the codebase.
___
*[`nativelink-store/src/memory_store.rs` line 259 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CNlwg3FLo0GcdznNi:-O-CNlwg3FLo0GcdznNj:b-pxsx5c) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L259)):*
> ```XML
> 
>     fn optimized_for(&self, optimization: StoreOptimizations) -> bool {
>         optimization == StoreOptimizations::SubscribeChanges
> ```

Nit: This doesn't matter at this stage but later on we'll probably want to use `matches!(optimization, StoreOptimizations::SubscribeChanges | StoreOptimizations::SomeOtherOptimization)` so it could make sense to use that syntax now.
___
*[`nativelink-store/src/memory_store.rs` line 269 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-COJQpDNjgsbJdvJwm:-O-COJQpDNjgsbJdvJwn:b-r2za18) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L269)):*
> ```XML
>             rx.mark_changed();
>             return Box::new(MemoryStoreSubscription {
>                 store: self.weak_self.clone(),
> ```

I guess it's correct to have this api where the struct stores a weak pointer to it's own allocation but it's weird to me to not use something like `Arc::clone(self).downgrade()`

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988)* status: 1 of 2 LGTMs obtained, and 1 discussions need to be resolved (waiting on @caass)

___
*[`nativelink-store/src/memory_store.rs` line 63 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLYBNEukfpAuD5QTr:-O-EYJSl1gmYPZ3NtTmv:b-92gzsa) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L63)):*

Nit: can you use the `SubscriptionSender` alias here? And declare it before this struct?
</blockquote></details>

These are different types:
```
watch::Sender<Result<Arc<dyn StoreSubscriptionItem>, Error>>;

vs

watch::Receiver<Result<Arc<dyn StoreSubscriptionItem>, Error>>
```
___
*[`nativelink-store/src/memory_store.rs` line 97 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLihZ9F09gOLLF_G9:-O-EV6dd4h4wU0jyvb5r:b-21tjsb) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L97)):*

Minor: I think this is more clear as an `if let` since the logic is "if we are the only handle, notify the store we're dropped". Like basically inverting this to

```rust
if let Some(store) = self.store.upgrade() {
    store.subscription_dropped(self.key.borrow().into_owned());
}
```
</blockquote></details>

I generally prefer the return-early-return-often style. In rust it's not as big of a deal, but it does reduce indentation in exchange for more lines.
___
*[`nativelink-store/src/memory_store.rs` line 132 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CMSZs5kqj8lyZpOi-:-O-EVZi6ClVQeB-GmMD2:b-66qh7t) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L132)):*

FYI this looks a lot like we want to use [dashmap](https://docs.rs/dashmap/latest/dashmap/), which is basically an auto-sharding replacement for `RwLock<HashMap>` (which leads to lower lock contention). If we're ok adding a dependency, it'd make things simpler on our side since the locking logic would be all in that crate.
</blockquote></details>

I'm not opposed to this, but instead lets do it more wide-spread to replace all cases of this.

My only concern is doing a security check/audit on them. They are big enough that I'm not worried, but our dependencies are where our security concerns live.
___
*[`nativelink-store/src/memory_store.rs` line 157 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CMtoADa3DAjLxIDgH:-O-EVvO12AUfugOXPBlR:bdaf8v3) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L157)):*

Nit: should this be more imperative? Like `drop_subscription` instead of `subscription_dropped`? I guess the more technically correct name would be like. `remove_dropped_subscription` or `notify_of_dropped_subscription`. But `subscription_dropped` sounds like it should be a `bool` or a `fn(&self) -> bool` to me.
</blockquote></details>

I could make it `on_subscription_dropped`, because that's what it actually is. Done.
___
*[`nativelink-store/src/memory_store.rs` line 213 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CNISvBG9KrGhKzsnK:-O-EWbQr7ykjBegB7TNT:bikgcxk) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L213)):*

Unrelated to the current PR but I feel like this pattern of `key.borrow()` and `key.borrow().into_owned()` everywhere could probably benefit from using some traits to turn it into a proper smart pointer. Like it really wants to be a version of `Arc`, right? But like an `Arc<enum KeyKind>` basically?

I dunno, I guess it's fine for now but it generates a lot of noise around the codebase.
</blockquote></details>

`Arc` is always heap allocated, which is why we use it sparingly. I know it looks weird, but the nice thing is now that all the areas are marked, it's easy to regex-replace all occurrences if we need.

The big thing I was trying to avoid was `.clone()` because it forces a copy, when most of the time `.borrow()` is enough. I also considered making it `fn to_owned(&self) -> StoreKey<'static>`, but opted out of this because there are some cases that you could move the heap allocated string instead of copying it. Yes annoying, but at the end it's not as bad as it sounds, since it's easy to change later.
___
*[`nativelink-store/src/memory_store.rs` line 259 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CNlwg3FLo0GcdznNi:-O-EXJlsAtXNOfeRXDKr:bi44uqs) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L259)):*

Nit: This doesn't matter at this stage but later on we'll probably want to use `matches!(optimization, StoreOptimizations::SubscribeChanges | StoreOptimizations::SomeOtherOptimization)` so it could make sense to use that syntax now.
</blockquote></details>

Yeah, we do this in other stores.
___
*[`nativelink-store/src/memory_store.rs` line 269 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-COJQpDNjgsbJdvJwm:-O-EXZP9BUAmDzC8WVbi:b-1pq0rh) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L269)):*

I guess it's correct to have this api where the struct stores a weak pointer to it's own allocation but it's weird to me to not use something like `Arc::clone(self).downgrade()`
</blockquote></details>

Yes, but doing this actually causes 2 additional needless counter changes. I think what you mean is `Arc::downgrade(self)`

In any case we have a weak ref, so might as well use it.

allada (commented):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988)* status: 1 of 2 LGTMs obtained, and 1 discussions need to be resolved (waiting on @caass)

caass (approved):
:lgtm:

> Reviewed 16 of 17 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/988#-:-O-GRaR7932Aj5zMTFH3:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained

___
*[`nativelink-store/src/memory_store.rs` line 63 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLYBNEukfpAuD5QTr:-O-GQkfc5jWePwA9Y2be:b-ncebi6) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L63)):*

These are different types:
```
watch::Sender<Result<Arc<dyn StoreSubscriptionItem>, Error>>;

vs

watch::Receiver<Result<Arc<dyn StoreSubscriptionItem>, Error>>
```
</blockquote></details>

Whoops lol
___
*[`nativelink-store/src/memory_store.rs` line 97 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CLihZ9F09gOLLF_G9:-O-GQooaA2zM-G4LlALN:b-flevv9) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L97)):*

I generally prefer the return-early-return-often style. In rust it's not as big of a deal, but it does reduce indentation in exchange for more lines.
</blockquote></details>

Yea I understand, I just thought in this instance it was more clear. Either way, though, it's not a big deal.
___
*[`nativelink-store/src/memory_store.rs` line 132 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CMSZs5kqj8lyZpOi-:-O-GR9e08lySbpptWS62:b-dxcbts) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L132)):*

I'm not opposed to this, but instead lets do it more wide-spread to replace all cases of this.

My only concern is doing a security check/audit on them. They are big enough that I'm not worried, but our dependencies are where our security concerns live.
</blockquote></details>

Makes sense.
___
*[`nativelink-store/src/memory_store.rs` line 213 at r2](https://reviewable.io/reviews/nativelink/nativelink/988#-O-CNISvBG9KrGhKzsnK:-O-GRQyn-ZOgfDrBLAZx:b-qqwet0) ([raw file](https://github.com/nativelink/nativelink/blob/994877cabd4be2e670e9403c6994b842c1bfbe9c/nativelink-store/src/memory_store.rs#L213)):*

`Arc` is always heap allocated, which is why we use it sparingly. I know it looks weird, but the nice thing is now that all the areas are marked, it's easy to regex-replace all occurrences if we need.

The big thing I was trying to avoid was `.clone()` because it forces a copy, when most of the time `.borrow()` is enough. I also considered making it `fn to_owned(&self) -> StoreKey<'static>`, but opted out of this because there are some cases that you could move the heap allocated string instead of copying it. Yes annoying, but at the end it's not as bad as it sounds, since it's easy to change later.
</blockquote></details>

Totally

PR #993, opened 2024-06-12
Implementation of WorkerStateManager

adam-singer:
# Description

Implement WorkerStateManager for simple scheduler
    
Implementation of `WorkerStateManager` where the beginnings of being able to
tuck mutations behind a trait for `StateManager`. `StateManager` now wraps
the `StateManagerImpl` for inner state structure `StateManagerImpl`.
All mutations should be done on a lock on `inner` in the future.
Moving implementation code of `update_action` into `WorkerStateManager` trait.

Fixes https://github.com/nativelink/nativelink/issues/994

## Type of change

- [x] Refactor

## How Has This Been Tested?

Unit tests

## Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/993)

### DIFF
--- a/nativelink-scheduler/src/operation_state_manager.rs
+++ b/nativelink-scheduler/src/operation_state_manager.rs
@@ -119,7 +119,7 @@ pub trait WorkerStateManager {
     /// did not change with a modified timestamp in order to prevent
     /// the operation from being considered stale and being rescheduled.
     async fn update_operation(
-        &self,
+        &mut self,
         operation_id: OperationId,
         worker_id: WorkerId,
         action_stage: Result<ActionStage, Error>,
--- a/nativelink-scheduler/src/scheduler_state/metrics.rs
+++ b/nativelink-scheduler/src/scheduler_state/metrics.rs
@@ -19,27 +19,58 @@ pub(crate) struct Metrics {
     pub(crate) add_action_joined_running_action: CounterWithTime,
     pub(crate) add_action_joined_queued_action: CounterWithTime,
     pub(crate) add_action_new_action_created: CounterWithTime,
+    pub(crate) update_action_missing_action_result: CounterWithTime,
+    pub(crate) update_action_from_wrong_worker: CounterWithTime,
+    pub(crate) update_action_no_more_listeners: CounterWithTime,
+    pub(crate) workers_evicted: CounterWithTime,
+    pub(crate) workers_evicted_with_running_action: CounterWithTime,
+    pub(crate) retry_action: CounterWithTime,
+    pub(crate) retry_action_max_attempts_reached: CounterWithTime,
+    pub(crate) retry_action_no_more_listeners: CounterWithTime,
+    pub(crate) retry_action_but_action_missing: CounterWithTime,
 }
 
 impl Metrics {
     pub fn gather_metrics(&self, c: &mut CollectorState) {
-        c.publish_with_labels(
-            "add_action",
-            &self.add_action_joined_running_action,
-            "Stats about add_action().",
-            vec![("result".into(), "joined_running_action".into())],
-        );
-        c.publish_with_labels(
-            "add_action",
-            &self.add_action_joined_queued_action,
-            "Stats about add_action().",
-            vec![("result".into(), "joined_queued_action".into())],
-        );
-        c.publish_with_labels(
-            "add_action",
-            &self.add_action_new_action_created,
-            "Stats about add_action().",
-            vec![("result".into(), "new_action_created".into())],
-        );
+        {
+            c.publish_with_labels(
+                "add_action",
+                &self.add_action_joined_running_action,
+                "Stats about add_action().",
+                vec![("result".into(), "joined_running_action".into())],
+            );
+            c.publish_with_labels(
+                "add_action",
+                &self.add_action_joined_queued_action,
+                "Stats about add_action().",
+                vec![("result".into(), "joined_queued_action".into())],
+            );
+            c.publish_with_labels(
+                "add_action",
+                &self.add_action_new_action_created,
+                "Stats about add_action().",
+                vec![("result".into(), "new_action_created".into())],
+            );
+        }
+        {
+            c.publish_with_labels(
+                "update_action_errors",
+                &self.update_action_missing_action_result,
+                "Stats about errors when worker sends update_action() to scheduler. These errors are not complete, just the most common.",
+                vec![("result".into(), "missing_action_result".into())],
+            );
+            c.publish_with_labels(
+                "update_action_errors",
+                &self.update_action_from_wrong_worker,
+                "Stats about errors when worker sends update_action() to scheduler. These errors are not complete, just the most common.",
+                vec![("result".into(), "from_wrong_worker".into())],
+            );
+            c.publish_with_labels(
+                "update_action_errors",
+                &self.update_action_no_more_listeners,
+                "Stats about errors when worker sends update_action() to scheduler. These errors are not complete, just the most common.",
+                vec![("result".into(), "no_more_listeners".into())],
+            );
+        }
     }
 }
--- a/nativelink-scheduler/src/scheduler_state/state_manager.rs
+++ b/nativelink-scheduler/src/scheduler_state/state_manager.rs
@@ -15,22 +15,29 @@
 use std::cmp;
 use std::collections::BTreeMap;
 use std::sync::Arc;
+use std::time::SystemTime;
 
 use async_trait::async_trait;
 use futures::stream;
 use hashbrown::{HashMap, HashSet};
-use nativelink_error::{Error, ResultExt};
-use nativelink_util::action_messages::{ActionInfo, ActionStage, ActionState, OperationId};
+use nativelink_error::{make_err, make_input_err, Code, Error, ResultExt};
+use nativelink_util::action_messages::{
+    ActionInfo, ActionResult, ActionStage, ActionState, ExecutionMetadata, OperationId, WorkerId,
+};
 use tokio::sync::watch;
+use tokio::sync::watch::error::SendError;
+use tracing::{event, Level};
 
 use crate::operation_state_manager::{
     ActionStateResult, ActionStateResultStream, ClientStateManager, OperationFilter,
+    WorkerStateManager,
 };
 use crate::scheduler_state::awaited_action::AwaitedAction;
 use crate::scheduler_state::client_action_state_result::ClientActionStateResult;
 use crate::scheduler_state::completed_action::CompletedAction;
 use crate::scheduler_state::metrics::Metrics;
 use crate::scheduler_state::workers::Workers;
+use crate::worker::WorkerUpdate;
 
 #[repr(transparent)]
 pub(crate) struct StateManager {
@@ -45,6 +52,7 @@ impl StateManager {
         active_actions: HashMap<Arc<ActionInfo>, AwaitedAction>,
         recently_completed_actions: HashSet<CompletedAction>,
         metrics: Arc<Metrics>,
+        max_job_retries: usize,
     ) -> Self {
         Self {
             inner: StateManagerImpl {
@@ -54,6 +62,7 @@ impl StateManager {
                 active_actions,
                 recently_completed_actions,
                 metrics,
+                max_job_retries,
             },
         }
     }
@@ -102,6 +111,110 @@ pub(crate) struct StateManagerImpl {
     pub(crate) recently_completed_actions: HashSet<CompletedAction>,
 
     pub(crate) metrics: Arc<Metrics>,
+
+    /// Default times a job can retry before failing.
+    pub(crate) max_job_retries: usize,
+}
+
+impl StateManager {
+    /// Modifies the `stage` of `current_state` within `AwaitedAction`. Sends notification channel
+    /// the new state.
+    ///
+    ///
+    /// # Discussion
+    ///
+    /// The use of `Arc::make_mut` is potentially dangerous because it clones the data and
+    /// invalidates all weak references to it. However, in this context, it is considered
+    /// safe because the data is going to be re-sent back out. The primary reason for using
+    /// `Arc` is to reduce the number of copies, not to enforce read-only access. This approach
+    /// ensures that all downstream components receive the same pointer. If an update occurs
+    /// while another thread is operating on the data, it is acceptable, since the other thread
+    /// will receive another update with the new version.
+    ///
+    fn mutate_stage(
+        awaited_action: &mut AwaitedAction,
+        action_stage: ActionStage,
+    ) -> Result<(), SendError<Arc<ActionState>>> {
+        Arc::make_mut(&mut awaited_action.current_state).stage = action_stage;
+        awaited_action
+            .notify_channel
+            .send(awaited_action.current_state.clone())
+    }
+
+    /// Modifies the `priority` of `action_info` within `ActionInfo`.
+    ///
+    fn mutate_priority(action_info: &mut Arc<ActionInfo>, priority: i32) {
+        Arc::make_mut(action_info).priority = priority;
+    }
+
+    /// Evicts the worker from the pool and puts items back into the queue if anything was being executed on it.
+    fn immediate_evict_worker(&mut self, worker_id: &WorkerId, err: Error) {
+        if let Some(mut worker) = self.inner.workers.remove_worker(worker_id) {
+            self.inner.metrics.workers_evicted.inc();
+            // We don't care if we fail to send message to worker, this is only a best attempt.
+            let _ = worker.notify_update(WorkerUpdate::Disconnect);
+            for action_info in worker.running_action_infos.drain() {
+                self.inner.metrics.workers_evicted_with_running_action.inc();
+                self.retry_action(&action_info, worker_id, err.clone());
+            }
+        }
+    }
+
+    fn retry_action(&mut self, action_info: &Arc<ActionInfo>, worker_id: &WorkerId, err: Error) {
+        match self.inner.active_actions.remove(action_info) {
+            Some(running_action) => {
+                let mut awaited_action: AwaitedAction = running_action;
+                let send_result = if awaited_action.attempts >= self.inner.max_job_retries {
+                    self.inner.metrics.retry_action_max_attempts_reached.inc();
+                    let action_stage_completed = ActionStage::Completed(ActionResult {
+                        execution_metadata: ExecutionMetadata {
+                            worker: format!("{worker_id}"),
+                            ..ExecutionMetadata::default()
+                        },
+                        error: Some(err.merge(make_err!(
+                            Code::Internal,
+                            "Job cancelled because it attempted to execute too many times and failed"
+                        ))),
+                        ..ActionResult::default()
+                    });
+                    StateManager::mutate_stage(&mut awaited_action, action_stage_completed)
+                    // Do not put the action back in the queue here, as this action attempted to run too many
+                    // times.
+                } else {
+                    self.inner.metrics.retry_action.inc();
+                    let send_result =
+                        StateManager::mutate_stage(&mut awaited_action, ActionStage::Queued);
+                    self.inner.queued_actions_set.insert(action_info.clone());
+                    self.inner
+                        .queued_actions
+                        .insert(action_info.clone(), awaited_action);
+                    send_result
+                };
+
+                if send_result.is_err() {
+                    self.inner.metrics.retry_action_no_more_listeners.inc();
+                    // Don't remove this task, instead we keep them around for a bit just in case
+                    // the client disconnected and will reconnect and ask for same job to be executed
+                    // again.
+                    event!(
+                        Level::WARN,
+                        ?action_info,
+                        ?worker_id,
+                        "Action has no more listeners during evict_worker()"
+                    );
+                }
+            }
+            None => {
+                self.inner.metrics.retry_action_but_action_missing.inc();
+                event!(
+                    Level::ERROR,
+                    ?action_info,
+                    ?worker_id,
+                    "Worker stated it was running an action, but it was not in the active_actions"
+                );
+            }
+        }
+    }
 }
 
 #[async_trait]
@@ -132,7 +245,7 @@ impl ClientStateManager for StateManager {
 
             // In the event our task is higher priority than the one already scheduled, increase
             // the priority of the scheduled one.
-            Arc::make_mut(&mut arc_action_info).priority = new_priority;
+            StateManager::mutate_priority(&mut arc_action_info, new_priority);
 
             let result = Arc::new(ClientActionStateResult::new(
                 queued_action.notify_channel.subscribe(),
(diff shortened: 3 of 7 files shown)


zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Installation / macos-13, Publish image, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, windows-2022 / stable, and 1 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/operation_state_manager.rs` line 122 at r1](https://reviewable.io/reviews/nativelink/nativelink/993#-O-DHSOh7K9w9AvLyJMa:-O-DHSOh7K9w9AvLyJMb:b-ubk2kn) ([raw file](https://github.com/nativelink/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/operation_state_manager.rs#L122)):*
> ```XML
>     /// the operation from being considered stale and being rescheduled.
>     async fn update_operation(
>         &mut self,
> ```

Lets find a way to work around whatever is making this mut necessary by locking inside if necessary.

I'm guessing that the retry logic requiring mutable access is what is causing this to be necessary? IIRC that unfortunately goes pretty deep but it may be time to address it before this is landed.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Installation / macos-13, Publish image, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-22.04, windows-2022 / stable, and 2 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 222 at r1](https://reviewable.io/reviews/nativelink/nativelink/993#-O-DIUg5-fnWKnRBx6mi:-O-DIUg5-fnWKnRBx6mj:b-ibds2v) ([raw file](https://github.com/nativelink/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/scheduler_state/state_manager.rs#L222)):*
> ```XML
>         }
> 
>         Arc::make_mut(&mut running_action.current_state).stage = action_stage;
> ```

I know this line came from the original code but `Arc::make_mut` is pretty dangerous since it clones the data and invalidates all weak references to it. I think the changes necessary to allow `update_action` (and ideally `retry_action`) to not require a mutable reference to self would also allow us to modify this in a safer way.

allada (commented):
> Reviewed 4 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 8 discussions need to be resolved (waiting on @adam-singer)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EfEuM0o4taKIQaLp-):*
nit: Extend this commit message.
___
*[`nativelink-scheduler/src/operation_state_manager.rs` line 122 at r1](https://reviewable.io/reviews/nativelink/nativelink/993#-O-DHSOh7K9w9AvLyJMa:-O-EfIiHBoyigG5uQ5_2:bxsb0st) ([raw file](https://github.com/nativelink/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/operation_state_manager.rs#L122)):*

Lets find a way to work around whatever is making this mut necessary by locking inside if necessary.

I'm guessing that the retry logic requiring mutable access is what is causing this to be necessary? IIRC that unfortunately goes pretty deep but it may be time to address it before this is landed.
</blockquote></details>

Adam and I talked about this. We decided to make it mutable for now and change it later because of some other issues he was having.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 468 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EfmRAEr3bMnTNHS8c:-O-EfmRAEr3bMnTNHS8d:b-1fwlr7) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/simple_scheduler.rs#L468)):*
> ```XML
>         &mut self,
>         worker_id: &WorkerId,
>         action_info_hash_key: &ActionInfoHashKey,
> ```

nit: Can we change this to take an owned version of this now?

It's generally a good idea to allow the caller to decide if it wants to copy it or move it.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 222 at r1](https://reviewable.io/reviews/nativelink/nativelink/993#-O-DIUg5-fnWKnRBx6mi:-O-Ei0Jc04OnxoyYaok2:bf2d3rb) ([raw file](https://github.com/nativelink/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/scheduler_state/state_manager.rs#L222)):*

I know this line came from the original code but `Arc::make_mut` is pretty dangerous since it clones the data and invalidates all weak references to it. I think the changes necessary to allow `update_action` (and ideally `retry_action`) to not require a mutable reference to self would also allow us to modify this in a safer way.
</blockquote></details>

I do agree, but It's actually safe to do because we are going to re-send the data back out. The only reason it's an `Arc` is to reduce the number of copies. It's not really used to make it read-only. Doing it this way makes it so all down-stream components get the same pointer. If an update happens while another thread is operating on the data, it's ok, since the other thread is going to get another update with the new version anyway.

It might be worth hiding this logic behind a function and documenting it though.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 89 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-Egb-d-O4Gi_EUx3tf:-O-Egb-d-O4Gi_EUx3tg:b-wyyzbq) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L89)):*
> ```XML
>             // We don't care if we fail to send message to worker, this is only a best attempt.
>             let _ = worker.notify_update(WorkerUpdate::Disconnect);
>             // We create a temporary Vec to avoid doubt about a possible code
> ```

nit: This looks like a stale comment.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 98 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-Egek1DIfLAuhJHY52:-O-Egek1DIfLAuhJHY53:b3shytz) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L98)):*
> ```XML
>         // TODO(adams): call this after eviction.
>         // Note: Calling this many time is very cheap, it'll only trigger `do_try_match` once.
>         //self.tasks_or_workers_change_notify.notify_one();
> ```

nit: don't forget.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O-EhD0g4EuU2ZTM0urI:b-9oafvt) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*
> ```XML
>         action_stage: Result<ActionStage, Error>,
>         let action_stage = action_stage?;
> ```

This looks weird, we take a result, but require it to be `Ok`?
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 175 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EhRhf3DoehbAFDIef:-O-EhRhf3DoehbAFDIeg:b-ftsvdo) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L175)):*
> ```XML
>             event!(
>                 Level::ERROR,
>                 ?action_info_hash_key, /*?action_info_hash_key,*/
> ```

nit: cleanup.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/nativelink/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 8 discussions need to be resolved

___
*[`nativelink-scheduler/src/operation_state_manager.rs` line 122 at r1](https://reviewable.io/reviews/nativelink/nativelink/993#-O-DHSOh7K9w9AvLyJMa:-O-DmaF61s9m14zXfqlP:btvi4h0) ([raw file](https://github.com/nativelink/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/operation_state_manager.rs#L122)):*

Adam and I talked about this. We decided to make it mutable for now and change it later because of some other issues he was having.
</blockquote></details>

This came around to locking, right now locking is in place, moving the lock inside will be easier once of the stateful code is out of the simple scheduler.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 468 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EfmRAEr3bMnTNHS8c:-O-F6-1MBC-QBlje8ErA:bemrh2l) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/simple_scheduler.rs#L468)):*

nit: Can we change this to take an owned version of this now?

It's generally a good idea to allow the caller to decide if it wants to copy it or move it.
</blockquote></details>

Owned in this case is we want to pass ` action_info_hash_key: ActionInfoHashKey` instead of ` action_info_hash_key: &ActionInfoHashKey` ?
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/nativelink/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O-F6AUjBQMcBVNhUEbM:bh3yq6u) ([raw file](https://github.com/nativelink/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*

This looks weird, we take a result, but require it to be `Ok`?
</blockquote></details>

If it isn't `Ok` what should we do given the function returns `()` or `Error`, I assumed by doing `action_stage?` that would propagate the failure back out. If we don't have an `action_stage` what else would we do? I forget do we have a different code path in the worker we need to support in this definition?

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 7 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/operation_state_manager.rs` line 122 at r1](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-DHSOh7K9w9AvLyJMa:-O-bsQrO2lsEl_E08t_Y:btn461w) ([raw file](https://github.com/tracemachina/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/operation_state_manager.rs#L122)):*

This came around to locking, right now locking is in place, moving the lock inside will be easier once of the stateful code is out of the simple scheduler.
</blockquote></details>

Got it
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 222 at r1](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-DIUg5-fnWKnRBx6mi:-O-bscpG1lvYKRhz_Ab5:b-v6nsfu) ([raw file](https://github.com/tracemachina/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/scheduler_state/state_manager.rs#L222)):*

I do agree, but It's actually safe to do because we are going to re-send the data back out. The only reason it's an `Arc` is to reduce the number of copies. It's not really used to make it read-only. Doing it this way makes it so all down-stream components get the same pointer. If an update happens while another thread is operating on the data, it's ok, since the other thread is going to get another update with the new version anyway.

It might be worth hiding this logic behind a function and documenting it though.
</blockquote></details>

I agree with hiding it behind a documented function being a good course of action. That can probably wait for a followup though.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 1 discussions need to be resolved (waiting on @adam-singer)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EfEuM0o4taKIQaLp-):*

nit: Extend this commit message.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 468 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EfmRAEr3bMnTNHS8c:-O-d_JOu2eMZ4VwSbsd8:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/simple_scheduler.rs#L468)):*

Owned in this case is we want to pass ` action_info_hash_key: ActionInfoHashKey` instead of ` action_info_hash_key: &ActionInfoHashKey` ?
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 222 at r1](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-DIUg5-fnWKnRBx6mi:-O-d_UAX4ADqlyQMym7e:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/aa3e1bf7592ba91b1d9860658bf29cbc74fa1c14/nativelink-scheduler/src/scheduler_state/state_manager.rs#L222)):*

I agree with hiding it behind a documented function being a good course of action. That can probably wait for a followup though.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 89 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-Egb-d-O4Gi_EUx3tf:-O-dTo7TBtxAT7NCg-rO:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L89)):*

nit: This looks like a stale comment.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 98 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-Egek1DIfLAuhJHY52:-O-d_L5wDDIME4hxgONr:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L98)):*

nit: don't forget.
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O-de8q85Win0hR4Z3wX:b-sh3idu) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*

If it isn't `Ok` what should we do given the function returns `()` or `Error`, I assumed by doing `action_stage?` that would propagate the failure back out. If we don't have an `action_stage` what else would we do? I forget do we have a different code path in the worker we need to support in this definition?
</blockquote></details>

cc: @allada 
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 175 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EhRhf3DoehbAFDIef:-O-ddu298aBLJJE6S6Cb:bb6un1j) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L175)):*

nit: cleanup.
</blockquote></details>

Done

allada (approved):
:lgtm:

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993#-:-O-xG2HoEq8ke10psIsG:bnfp4nl)* status: 0 of 1 LGTMs obtained, and pending CI: docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), windows-2022 / stable, and 2 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O-xFk2w3xJ5FiIbZUS4:b-xztf7j) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*

cc: @allada
</blockquote></details>

Shouldn't the operation be updated to reflect the Err case? The way it's done here it does not propagate the error to the client.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 133 at r3](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-xFQYg8ynrp7G8PgeD:-O-xFQYg8ynrp7G8PgeE:b-dy22d6) ([raw file](https://github.com/tracemachina/nativelink/blob/dd3c137d3d69f180f7c498a66cc0ec8fc6591f2f/nativelink-scheduler/src/scheduler_state/state_manager.rs#L133)):*
> ```XML
>     ///
>     fn mutate_stage(awaited_action: &mut AwaitedAction, action_stage: ActionStage) {
>         Arc::make_mut(&mut awaited_action.current_state).stage = action_stage;
> ```

If we mutate the stage, we probably should also do the notification to any listeners here too?

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), windows-2022 / stable, and 2 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O07_Gh_37-AbBdZtDbD:b-4008oa) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*

Shouldn't the operation be updated to reflect the Err case? The way it's done here it does not propagate the error to the client.
</blockquote></details>

I'm confused by this statement, the error case here is `action_stage` came in as an error, using `action_stage?` afaict means it would propagate that same error back out
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 133 at r3](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-xFQYg8ynrp7G8PgeD:-O07_eTA8Vw0aRrKTQSi:b5cojhl) ([raw file](https://github.com/tracemachina/nativelink/blob/dd3c137d3d69f180f7c498a66cc0ec8fc6591f2f/nativelink-scheduler/src/scheduler_state/state_manager.rs#L133)):*

If we mutate the stage, we probably should also do the notification to any listeners here too?
</blockquote></details>

When `StateManager::mutate_stage` is called the following code has handled doing `notify_channel.send()`. Guess you are saying thats related and we should fold that code into this method

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, and 2 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 133 at r3](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-xFQYg8ynrp7G8PgeD:-O07dQKK3ZlseGKRB-XB:b-896fix) ([raw file](https://github.com/tracemachina/nativelink/blob/dd3c137d3d69f180f7c498a66cc0ec8fc6591f2f/nativelink-scheduler/src/scheduler_state/state_manager.rs#L133)):*

When `StateManager::mutate_stage` is called the following code has handled doing `notify_channel.send()`. Guess you are saying thats related and we should fold that code into this method
</blockquote></details>

Done.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable, and 2 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 169 at r2](https://reviewable.io/reviews/tracemachina/nativelink/993#-O-EhD0g4EuU2ZTM0urH:-O0AOT6y0NDIf54YvlLZ:b-qyg96b) ([raw file](https://github.com/tracemachina/nativelink/blob/11fc58812969b17c762cbf7beabfbc0c8ea64f22/nativelink-scheduler/src/scheduler_state/state_manager.rs#L169)):*

I'm confused by this statement, the error case here is `action_stage` came in as an error, using `action_stage?` afaict means it would propagate that same error back out
</blockquote></details>

Taken note in NL-320, will follow up on that refactor after this lands

adam-singer (commented):
> Reviewed 8 of 9 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained, and 2 discussions need to be resolved (waiting on @adam-singer)

allada (approved):
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993)* status: 0 of 1 LGTMs obtained (waiting on @adam-singer)

adam-singer (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/993#-:-O0AdHenDtRZzSm54qvF:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained

PR #1027, opened 2024-06-19
Add optional pub sub publisher to store writes

blakehatch:
# Description

This optional publisher for stores writes is a cleaner, more flexible, and simpler alternative to using keyspace notifications.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -879,6 +879,17 @@ pub struct RedisStore {
     #[serde(default)]
     pub connection_timeout_s: u64,
 
+    /// An optional and experimental Redis channel to publish write events to.
+    ///
+    /// If set, every time a write operation is made to a Redis node
+    /// then an event will be published to a Redis channel with the given name.
+    /// If unset, the writes will still be made,
+    /// but the write events will not be published.
+    ///
+    /// Default: (Empty String / No Channel)
+    #[serde(default)]
+    pub experimental_pub_sub_channel: Option<String>,
+
     /// An optional prefix to prepend to all keys in this store.
     ///
     /// Setting this value can make it convenient to query or
--- a/nativelink-service/src/bep_server.rs
+++ b/nativelink-service/src/bep_server.rs
@@ -65,6 +65,8 @@ impl BepServer {
             .as_ref()
             .err_tip(|| "Expected stream_id to be set")?;
 
+        let sequence_number = build_event.sequence_number;
+
         let mut buf = BytesMut::new();
         request
             .encode(&mut buf)
@@ -73,8 +75,8 @@ impl BepServer {
         self.store
             .update_oneshot(
                 StoreKey::Str(Cow::Owned(format!(
-                    "{}-{}-{}",
-                    stream_id.build_id, stream_id.invocation_id, stream_id.component
+                    "LifecycleEvent-{}-{}-{}",
+                    &stream_id.build_id, &stream_id.invocation_id, sequence_number,
                 ))),
                 buf.freeze(),
             )
@@ -104,6 +106,7 @@ impl BepServer {
             let sequence_number = ordered_build_event.sequence_number;
 
             let mut buf = BytesMut::new();
+
             request
                 .encode(&mut buf)
                 .err_tip(|| "Could not encode PublishBuildToolEventStreamRequest proto")?;
@@ -111,8 +114,8 @@ impl BepServer {
             store
                 .update_oneshot(
                     StoreKey::Str(Cow::Owned(format!(
-                        "{}-{}-{}",
-                        stream_id.build_id, stream_id.invocation_id, stream_id.component
+                        "BuildToolEventStream-{}-{}-{}",
+                        &stream_id.build_id, &stream_id.invocation_id, sequence_number,
                     ))),
                     buf.freeze(),
                 )
--- a/nativelink-service/tests/bep_server_test.rs
+++ b/nativelink-service/tests/bep_server_test.rs
@@ -93,15 +93,11 @@ async fn publish_lifecycle_event_test() -> Result<(), Box<dyn std::error::Error>
         invocation_id: "some-invocation-id".to_string(),
         component: BuildComponent::Controller as i32,
     };
-    let store_key = StoreKey::Str(Cow::Owned(format!(
-        "{}-{}-{}",
-        stream_id.build_id, stream_id.invocation_id, stream_id.component
-    )));
 
     let request = PublishLifecycleEventRequest {
         service_level: ServiceLevel::Interactive as i32,
         build_event: Some(OrderedBuildEvent {
-            stream_id: Some(stream_id),
+            stream_id: Some(stream_id.clone()),
             sequence_number: 1,
             event: Some(BuildEvent {
                 event_time: Some(Timestamp::date(1999, 1, 6)?),
@@ -118,6 +114,15 @@ async fn publish_lifecycle_event_test() -> Result<(), Box<dyn std::error::Error>
         project_id: "some-project-id".to_string(),
         check_preceding_lifecycle_events_present: false,
     };
+
+    let sequence_number = request.clone().build_event.unwrap().sequence_number;
+
+    let store_key = StoreKey::Str(Cow::Owned(format!(
+        "LifecycleEvent-{}-{}-{}",
+        stream_id.clone().build_id,
+        stream_id.clone().invocation_id,
+        sequence_number
+    )));
 
     bep_server
         .publish_lifecycle_event(Request::new(request.clone()))
@@ -165,7 +170,7 @@ async fn publish_build_tool_event_stream_test() -> Result<(), Box<dyn std::error
     }
     .await?;
 
-    let (requests, store_key) = {
+    let (requests, store_keys) = {
         // Construct some requests to send off and a store key to retrieve them with.
         let stream_id = StreamId {
             build_id: "some-build-id".to_string(),
@@ -271,11 +276,22 @@ async fn publish_build_tool_event_stream_test() -> Result<(), Box<dyn std::error
         ];
 
         (
-            requests,
-            StoreKey::Str(Cow::Owned(format!(
-                "{}-{}-{}",
-                stream_id.build_id, stream_id.invocation_id, stream_id.component
-            ))),
+            requests.clone(),
+            requests
+                .iter()
+                .map(|request| {
+                    StoreKey::Str(Cow::Owned(format!(
+                        "BuildToolEventStream-{}-{}-{}",
+                        stream_id.build_id,
+                        stream_id.invocation_id,
+                        request
+                            .ordered_build_event
+                            .as_ref()
+                            .unwrap()
+                            .sequence_number
+                    )))
+                })
+                .collect::<Vec<_>>(),
         )
     };
 
@@ -296,6 +312,7 @@ async fn publish_build_tool_event_stream_test() -> Result<(), Box<dyn std::error
 
             // First, check if the response matches what we expect.
             assert_eq!(response.sequence_number, sequence_number);
+
             assert_eq!(
                 response.stream_id,
                 request
@@ -307,7 +324,12 @@ async fn publish_build_tool_event_stream_test() -> Result<(), Box<dyn std::error
             // Second, check if the message was forwarded correctly.
             let (mut writer, mut reader) = make_buf_channel_pair();
             bep_store
-                .get_part(store_key.borrow(), &mut writer, 0, None)
+                .get_part(
+                    store_keys[sequence_number as usize - 1].clone(),
+                    &mut writer,
+                    0,
+                    None,
+                )
                 .await?;
             let encoded_request = reader.recv().await?;
 
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -245,6 +245,7 @@ pub struct RedisStore<C: ConnectionLike + Clone = ConnectionKind> {
 
     /// A function used to generate names for temporary keys.
     temp_name_generator_fn: fn() -> String,
+    pub_sub_channel: Option<String>,
 
     /// A common prefix to append to all keys before they are sent to Redis.
     ///
@@ -287,6 +288,7 @@ impl RedisStore<ConnectionKind> {
             RedisStore::new_with_conn_and_name_generator_and_prefix(
                 connection,
                 || uuid::Uuid::new_v4().to_string(),
+                config.experimental_pub_sub_channel.clone(),
                 config.key_prefix.clone(),
             ),
         ))
@@ -302,6 +304,7 @@ impl<C: ConnectionLike + Clone + Send + 'static> RedisStore<C> {
         RedisStore::new_with_conn_and_name_generator_and_prefix(
             connection,
             temp_name_generator_fn,
+            None,
             String::new(),
         )
     }
@@ -310,11 +313,13 @@ impl<C: ConnectionLike + Clone + Send + 'static> RedisStore<C> {
     pub fn new_with_conn_and_name_generator_and_prefix(
         connection: BackgroundConnection<C>,
         temp_name_generator_fn: fn() -> String,
+        pub_sub_channel: Option<String>,
         key_prefix: String,
     ) -> Self {
         RedisStore {
             connection,
             temp_name_generator_fn,
+            pub_sub_channel,
             key_prefix,
         }
     }
@@ -460,6 +465,7 @@ where
                             .map_err(from_redis_err)
                             .err_tip(|| "In RedisStore::update() single chunk")?;
                     }
+
                     break 'outer;
                 }
 
@@ -477,16 +483,25 @@ where
                 .await
                 .map_err(from_redis_err)
                 .err_tip(|| "In RedisStore::update::query_async")?;
+
             pipe.clear();
         }
 
         pipe.cmd("RENAME")
             .arg(temp_key.get_or_init(make_temp_name))
-            .arg(final_key);
+            .arg(&final_key);
+
         pipe.query_async(&mut conn)
             .await
             .map_err(from_redis_err)
             .err_tip(|| "In RedisStore::update")?;
+
+        if let Some(pub_sub_channel) = &self.pub_sub_channel {
+            conn.publish(pub_sub_channel, &final_key)
+                .await
+                .map_err(from_redis_err)
+                .err_tip(|| "Failed to publish temp key value to configured channel")?
+        }
 
         Ok(())
     }
--- a/nativelink-store/tests/redis_store_test.rs
+++ b/nativelink-store/tests/redis_store_test.rs
@@ -157,6 +157,7 @@ async fn upload_and_get_data_with_prefix() -> Result<(), Error> {
     let store = MockRedisStore::new_with_conn_and_name_generator_and_prefix(
         BackgroundConnection::with_initializer(async move { Ok::<_, Error>(redis_connection) }),
         mock_uuid_generator,
+        None,
         prefix.to_string(),
     );
 
@@ -213,6 +214,7 @@ async fn upload_empty_data_with_prefix() -> Result<(), Error> {
     let store = MockRedisStore::new_with_conn_and_name_generator_and_prefix(
         BackgroundConnection::with_initializer(async move { Ok::<_, Error>(redis_connection) }),
         mock_uuid_generator,
+        None,
         prefix.to_string(),
     );
 


caass (changes requested):
Awesome :) I love it !! I had like one blocking concern and mostly a few minor nits.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-:-O-lQBDu1ZmgXkEuMCxL:b-qgw5fi)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 4 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPyi60juDTRy4-e6j):*
Minor: Can you make this more descriptive? Like at least mention that it's for the Redis store.
___
*[`nativelink-config/src/stores.rs` line 881 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPWZz0PrIS1rZlt6x:-O-lPWZz0PrIS1rZlt6y:b-qt0i1u) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L881)):*
> ```XML
>     /// Default: 10
>     #[serde(default)]
>     pub connection_timeout_s: u64,
> ```

FYI unrelated to this current PR, these should be like. `Duration`s deserialized with `serde_with` or something. Whatever.
___
*[`nativelink-config/src/stores.rs` line 883 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lOvjl9NWD6u7-OVyo:-O-lOvjl9NWD6u7-OVyp:bz9cqa1) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L883)):*
> ```XML
>     pub connection_timeout_s: u64,
> 
>     /// An optional channel to publish writes to redis.
> ```

Nit: Can you clarify this a little bit? Maybe like 

```rust
/// An optional Redis channel to publish write events to.
///
/// If set, every time a write operation is made to a Redis node then an event will be published to a Redis channel with the given name.
/// If unset, the writes will still be made, but the write events will not be published.
///
/// Default: (Empty String / No Channel)
```
___
*[`nativelink-store/src/redis_store.rs` line 97 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lN55u5zgVKN7VH6Gx:-O-lN55u5zgVKN7VH6Gy:b-dxsvrr) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L97)):*
> ```XML
> 
> impl<T: ConnectionLike + Unpin + Clone + Send + Sync> RedisStore<T> {
>     pub fn new_with_conn_and_name_generator(
> ```

Minor: Can you rename this (and i think there's another method we should actually delete probably? Like we should probably have one method that's like `new(config: &Config)` and one that's `with_conn_and_pub_sub_channel_and_temp_name_generator_fn`
___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-lNLeOEDZWFflq9_sw:b-s6l39x) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*
> ```XML
>                 .await
>                 .map_err(from_redis_err)
>                 .err_tip(|| "Failed to publish temp key value to configured channel")?;
> ```

Try:

```rust
if let Some(ref channel) = self.pub_sub_channel {
// more code...
}
```

Also: for the call to `publish`, you shouldn't use the `Debug` representation -- use the `Display` representation instead. Check this out: https://doc.rust-lang.org/rust-by-example/hello/print.html.

Also also: it seems bizarre to me that the way to publish here is to basically upload the key, upload the value, and then also in a separate transaction upload the value _again_. Is that really what we have to do? Can't we just like. Upload the value once? I don't know the answer to this it just seems like we should be able to execute those all at once. Maybe there's a write-and-publish operation?



_Code quote:_
```XML
        if !self.pub_sub_channel.is_none() {
            let temp_key_value = conn.get(temp_key.get_or_init(make_temp_name)).await?;

            conn.publish(&self.pub_sub_channel, format!("{:?} {:?}", key, temp_key_value))
                .await
                .map_err(from_redis_err)
                .err_tip(|| "Failed to publish temp key value to configured channel")?;
```

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 4 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPyi60juDTRy4-e6j):*

Minor: Can you make this more descriptive? Like at least mention that it's for the Redis store.
</blockquote></details>

Oh yes that was a mistype this is not a store trait feature haha
___
*[`nativelink-store/src/redis_store.rs` line 97 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lN55u5zgVKN7VH6Gx:-O-lReGj8gn1NQzU_BPo:b15cxm9) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L97)):*

Minor: Can you rename this (and i think there's another method we should actually delete probably? Like we should probably have one method that's like `new(config: &Config)` and one that's `with_conn_and_pub_sub_channel_and_temp_name_generator_fn`
</blockquote></details>

This is a weird holdover from how the mocking has to be done to make the UUID generation deterministic you have to pass in a name_generator. I'm not opposed to changing it to be cleaner if possible but it would probably be best suited for a separate PR.
___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-lR-LO8Jvq_5YPGvLk:bvwhm9o) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

Try:

```rust
if let Some(ref channel) = self.pub_sub_channel {
// more code...
}
```

Also: for the call to `publish`, you shouldn't use the `Debug` representation -- use the `Display` representation instead. Check this out: https://doc.rust-lang.org/rust-by-example/hello/print.html.

Also also: it seems bizarre to me that the way to publish here is to basically upload the key, upload the value, and then also in a separate transaction upload the value _again_. Is that really what we have to do? Can't we just like. Upload the value once? I don't know the answer to this it just seems like we should be able to execute those all at once. Maybe there's a write-and-publish operation?
</blockquote></details>

The issue with this is that we don't call SET to put all the data up we incrementally build up the data with APPEND and then at the end swap out the temporary key. So essentially I have to wait for it to be built up and then when the key swap happens I then grab the full value to publish. Shouldn't be a huge deal performance-wise since this is an optional feature but I agree it would be nice to do both at the same time.

adam-singer (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 4 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 881 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPWZz0PrIS1rZlt6x:-O-mCzZe203QWycZW15Q:bgn5084) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L881)):*

FYI unrelated to this current PR, these should be like. `Duration`s deserialized with `serde_with` or something. Whatever.
</blockquote></details>

Good point, lets move this discussion out of this PR. There have been prior discussions around friendly formats for number formats that has landed. We can bring up `Duration` for replacing integers offline, I assume we would also get a friendly parsing scheme when making that move.

caass (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 0 of 1 LGTMs obtained, and pending CI: pre-commit-checks, and 3 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPyi60juDTRy4-e6j):*

Oh yes that was a mistype this is not a store trait feature haha
</blockquote></details>

:)
___
*[`nativelink-config/src/stores.rs` line 881 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lPWZz0PrIS1rZlt6x:-O-mMrpyByvGkiTSFLST:b-q4fy4h) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L881)):*

Good point, lets move this discussion out of this PR. There have been prior discussions around friendly formats for number formats that has landed. We can bring up `Duration` for replacing integers offline, I assume we would also get a friendly parsing scheme when making that move.
</blockquote></details>

Totally, I just saw it on my reviewable and I was like "aaaghhh!!" but yes not a blocker here :)
___
*[`nativelink-store/src/redis_store.rs` line 97 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lN55u5zgVKN7VH6Gx:-O-mN3io55YOKvQEQfEG:bb74ce2) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L97)):*

This is a weird holdover from how the mocking has to be done to make the UUID generation deterministic you have to pass in a name_generator. I'm not opposed to changing it to be cleaner if possible but it would probably be best suited for a separate PR.
</blockquote></details>

Sure
___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-mN8SU8UDRDiEaKZIX:b-kw3xw0) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

The issue with this is that we don't call SET to put all the data up we incrementally build up the data with APPEND and then at the end swap out the temporary key. So essentially I have to wait for it to be built up and then when the key swap happens I then grab the full value to publish. Shouldn't be a huge deal performance-wise since this is an optional feature but I agree it would be nice to do both at the same time.
</blockquote></details>

I think you can probably make it part of the same pipeline? During my work to implement cluster mode, I did some research around like. How transactions and stuff works in redis -- pipelining allows you to submit multiple commands over the same TCP request, which would make me feel a lot better. That's where we have `redis::pipe()` and then add commands to the `Pipe`, before finally sending it over the connection.

Also, one time when I tried to do this I got some borrow checker problems because of moving the pipe in and out of async blocks -- so let me know if that's the case.

caass (changes requested):
Nice work rebasing, sorry I made about a billion changes here. @blakehatch let's maybe pair to remove the hangup I have over pipelining the `PUBLISH` command?

> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-:-O-weMId2F0LKbrRqGl1:bqbxfac)* status: 0 of 1 LGTMs obtained, and 1 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 883 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lOvjl9NWD6u7-OVyo:-O-wc3xA7eBlc4kKHOQa:bb6ywmx) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L883)):*

Nit: Can you clarify this a little bit? Maybe like 

```rust
/// An optional Redis channel to publish write events to.
///
/// If set, every time a write operation is made to a Redis node then an event will be published to a Redis channel with the given name.
/// If unset, the writes will still be made, but the write events will not be published.
///
/// Default: (Empty String / No Channel)
```
</blockquote></details>

LGTM
___
*[`nativelink-service/tests/bep_server_test.rs` line 97 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-wcH1gDToXSYjw1pOq:-O-wcH1gDToXSYjw1pOr:by11y7x) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/tests/bep_server_test.rs#L97)):*
> ```XML
>     };
>     let store_key = StoreKey::Str(Cow::Owned(format!(
>         "{}-{}-{}-L",
> ```

What's up with the `-L` here and `-B` elsewhere?
___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-wcfsk1ixqSRxAGFij:b-z1fa49) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

@caass Just to make sure I'm understanding this correctly, are the two uploads you are referring to the steps where:
1. value is appended to temp-key
2. value in temp-key is assigned to actual key through rename

The sync transaction blocks where you can get a value and then perform some operation on it might be a good fit here, but pipelining alone won't allow you to do a get + set gotten value, and I think redis-rs has some issues with transaction blocks + async...

@blakehatch If you have to wait have to wait for the key swap to publish, why not do publication after the rename? Right now this code looks very much like it's publishing the temp value:
```
            conn.publish(&self.pub_sub_channel, format!("{:?} {:?}", key, temp_key_value))
                .await
                .map_err(from_redis_err)
                .err_tip(|| "Failed to publish temp key value to configured channel")?;
```

It might be worthwhile to put some more documentation for the logic here and maybe do the publish step after rename so that it's clearer we are publishing the final value
</blockquote></details>

@zbirenbaum No,I'm talking about how we call `conn.publish()` instead of `pipe.cmd("PUBLISH")`. My gut feeling is that redis doesn't mind making the `PUBLISH` command part of the same transaction? @blakehatch did that come up at all for you?

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 0 of 1 LGTMs obtained, and 8 discussions need to be resolved

___
*[`nativelink-config/examples/basic_cas.json` line 37 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xP35BBeIqtwblJgzs:-O-xP35BBeIqtwblJgzt:b-7qri16) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-config/examples/basic_cas.json#L37)):*
> ```JSON
>       }
>     },
>     "BEP_STORE": {
> ```

nit: Don't put this as a default. This is the baseline of running NL without any dependencies.
___
*[`nativelink-config/src/stores.rs` line 891 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPNYD2z8LqMfi_52v:-O-xPNYE9rjSuyGfyV0c:bubh35s) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-config/src/stores.rs#L891)):*
> ```XML
>     /// Default: (Empty String / No Channel)
>     #[serde(default)]
>     pub pub_sub_channel: Option<String>,
> ```

nit: mark this `experimental_`
___
*[`nativelink-service/src/bep_server.rs` line 76 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPRVW6GA2h35RiPR4:-O-xPRVW6GA2h35RiPR5:b-kjj35i) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L76)):*
> ```XML
>             .update_oneshot(
>                 StoreKey::Str(Cow::Owned(format!(
>                     "{}-{}-{}-L",
> ```

What is `L`?
___
*[`nativelink-service/src/bep_server.rs` line 112 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPhbWDiHPukF9hD9D:-O-xPhbWDiHPukF9hD9E:b-l7msuo) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L112)):*
> ```XML
>                 .err_tip(|| "Could not encode PublishBuildToolEventStreamRequest proto")?;
> 
>             if !buf.is_empty() {
> ```

nit: Shouldn't we publish empty protos?
___
*[`nativelink-service/src/bep_server.rs` line 116 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xP_IwD9fmaJX0D28V:-O-xP_IwD9fmaJX0D28W:b-n9d8ww) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L116)):*
> ```XML
>                     .update_oneshot(
>                         StoreKey::Str(Cow::Owned(format!(
>                             "{}-{}-{}-B",
> ```

what is `B`?
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQ0Up50lxw4D3yPe5:-O-xQ0Up50lxw4D3yPe6:bn6mij4) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*
> ```XML
> 
>                 if self.pub_sub_channel.is_some() {
>                     conn.publish(&self.pub_sub_channel, format!("{:?} {:?}", key, &chunk[..]))
> ```

nit: You probably don't want to publish a `Some<String>`. You probably should unwrap it in the `if-statement`.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQCXqAK2kACVCELcn:-O-xQCXqAK2kACVCELco:b-x3stnw) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*
> ```XML
> 
>                 if self.pub_sub_channel.is_some() {
>                     conn.publish(&self.pub_sub_channel, format!("{:?} {:?}", key, &chunk[..]))
> ```

I don't think we should publish the data. We should publish the key and let the receiver lookup the data when it's received.

Yes this does mean they might get an old version, but we currently don't need to support receiving every event. We only currently care about most recent event.

blakehatch (commented):
Thanks! Appreciate all the thoughts :)

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-:-O-rVAf-94IDNdMUwhze:b-bqcnqs)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 8 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 883 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lOvjl9NWD6u7-OVyo:-O-rTnfJ26i1Ix6ARFnN:bri454t) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-config/src/stores.rs#L883)):*

Nit: Can you clarify this a little bit? Maybe like 

```rust
/// An optional Redis channel to publish write events to.
///
/// If set, every time a write operation is made to a Redis node then an event will be published to a Redis channel with the given name.
/// If unset, the writes will still be made, but the write events will not be published.
///
/// Default: (Empty String / No Channel)
```
</blockquote></details>

Clarified, also added that it is experimental since it is mostly a nice to have for BEP.
___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-rUBzfAQqDGShAu42k:b-l40e64) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

@caass Just to make sure I'm understanding this correctly, are the two uploads you are referring to the steps where:
1. value is appended to temp-key
2. value in temp-key is assigned to actual key through rename

The sync transaction blocks where you can get a value and then perform some operation on it might be a good fit here, but pipelining alone won't allow you to do a get + set gotten value, and I think redis-rs has some issues with transaction blocks + async...

@blakehatch If you have to wait have to wait for the key swap to publish, why not do publication after the rename? Right now this code looks very much like it's publishing the temp value:
```
            conn.publish(&self.pub_sub_channel, format!("{:?} {:?}", key, temp_key_value))
                .await
                .map_err(from_redis_err)
                .err_tip(|| "Failed to publish temp key value to configured channel")?;
```

It might be worthwhile to put some more documentation for the logic here and maybe do the publish step after rename so that it's clearer we are publishing the final value
</blockquote></details>

I've been testing it to just dump the bytes currently in the chunk over to the receiving end and I haven't had any issues. I've been having some issues with grabbing the value at the time of the swap and is a performance hit as it requires an entire additional redis operation instead of leveraging the data in the chunk being passed around. 

As this is experimental and for BEP which benefits from performance over mild stability increases here (occasionally maybe getting a partial message). I'd rather use this current setup esp as I've never seen a BEP Event message not fit in the chunk.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQCXqAK2kACVCELcn:-O-xRvZK-7QyUr_-DkOy:b-z7ymvn) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*

I don't think we should publish the data. We should publish the key and let the receiver lookup the data when it's received.

Yes this does mean they might get an old version, but we currently don't need to support receiving every event. We only currently care about most recent event.
</blockquote></details>

I actually ended up needing every event hence why I made this feature for myself.

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 8 discussions need to be resolved

___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-xSJ-B51VBgWi3zxIN:b-94p7sd) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

I've been testing it to just dump the bytes currently in the chunk over to the receiving end and I haven't had any issues. I've been having some issues with grabbing the value at the time of the swap and is a performance hit as it requires an entire additional redis operation instead of leveraging the data in the chunk being passed around. 

As this is experimental and for BEP which benefits from performance over mild stability increases here (occasionally maybe getting a partial message). I'd rather use this current setup esp as I've never seen a BEP Event message not fit in the chunk.
</blockquote></details>

Swapped to pipe, it functions equivalently

caass (approved):
Ofc :) :lgtm:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-:-O-xV2Vc98zSKGcXYD7Q:b-phceu4)* status: 1 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, and 7 discussions need to be resolved

___
*[`nativelink-store/src/redis_store.rs` line 266 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-lNLeN9IvF3hLJIdqo:-O-xVFJe9e_u4-awgGp2:bb7ot1r) ([raw file](https://github.com/TraceMachina/nativelink/blob/610a446b0ca06991526a6d673d65ff095a1c3e91/nativelink-store/src/redis_store.rs#L266)):*

Swapped to pipe, it functions equivalently
</blockquote></details>

swag

allada (changes requested):
> Reviewed 1 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 1 of 1 LGTMs obtained, and 4 discussions need to be resolved

___
*[`nativelink-service/src/bep_server.rs` line 76 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPRVW6GA2h35RiPR4:-O09pCciAM0c7XiG4BfC:boxexl6) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L76)):*

Indicates that the protobuf being stored in a LifeCycle object and not a BuildEvent object. They were also overwriting each other so redis was never saving the last build object it was being overwritten by the less granular and useful lifecycle object.

Also indicates how to deserialize them instead of having to use try catch style fallbacks for deserializing the two options.
</blockquote></details>

I see, in that case we should make a function that names this, so the code becomes self-explanatory.
___
*[`nativelink-service/src/bep_server.rs` line 112 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPhbWDiHPukF9hD9D:-O09pPLiCLi6pu7L0Y2m:b6r66bj) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L112)):*

I don't really see why it would be useful but I can understand not putting any filtering at this stage. Tbh tho anything ingesting it only cares about a small subset of these which definitely does not include empty protos. If there's some debugging reason this would be helpful tho I can understand that.
</blockquote></details>

I would not put this branch in.

protobufs are weird, an empty proto actually can have data. For example a proto of:
```
{
  id: 0,
  found_items: [],
  prefix: "",
}
```

will be an empty proto, but this proto might have meaning.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQ0Up50lxw4D3yPe5:-O09prCn8sO5BOgh0ZTH:bggnoez) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*

It's not a Some<String>, it's an Option<String> and is guaranteed to be something because of the conditional it's in.
</blockquote></details>

I'm saying:
```rs
if let Some(pub_sub_channel) = &self.pub_sub_channel {
  pipe.publish(pub_sub_channel, format!("{:?} {:?}", key, &chunk[..]));
}
```

Otherwise you are publishing the optional and not the actual value.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQCXqAK2kACVCELcn:-O09qGJaEjegnZ_37UEV:b-rq8o4z) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*

I can maybe switch it to this later but this is far simpler and again is optional. Would make getting V1 of BEP over the line far easier
</blockquote></details>

In that case, we should make a new key for every item.

example: `{build_id}-{event_id}-{index}`

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 1 of 1 LGTMs obtained, and pending CI: Publish image, Vercel, pre-commit-checks, and 4 discussions need to be resolved

___
*[`nativelink-service/src/bep_server.rs` line 76 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPRVW6GA2h35RiPR4:-O09u6gx-zI6WyuL9DpH:b-gp71r2) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L76)):*

I see, in that case we should make a function that names this, so the code becomes self-explanatory.
</blockquote></details>

I can write -Lifecycle or something instead of -L. It's already contained in a function that clearly indicates which of the two is being sent it's just unclear that -L is indicating that.
___
*[`nativelink-service/src/bep_server.rs` line 112 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xPhbWDiHPukF9hD9D:-O09uy3IEz2Vrsa9DJOc:bk2fmmr) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-service/src/bep_server.rs#L112)):*

I would not put this branch in.

protobufs are weird, an empty proto actually can have data. For example a proto of:
```
{
  id: 0,
  found_items: [],
  prefix: "",
}
```

will be an empty proto, but this proto might have meaning.
</blockquote></details>

Makes sense and shouldn't cause too much strain I haven't seen any empty protobufs get sent in my testing anyways.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQCXqAK2kACVCELcn:-O09stJ6E88eiuRxpVmb:b-erlf6t) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*

In that case, we should make a new key for every item.

example: `{build_id}-{event_id}-{index}`
</blockquote></details>

Ok I'll switch to this cuz it would be nice to be able to reconstruct the full history if the Microservice ever goes down and we just get forget with no fire.
___
*[`nativelink-store/src/redis_store.rs` line 478 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O-xQ0Up50lxw4D3yPe5:-O09tNuY4LZXm6ROKC5P:b-l1aqyy) ([raw file](https://github.com/TraceMachina/nativelink/blob/cc767ae84c29fc11c1fd5bbcaef6c6a7ce9f5019/nativelink-store/src/redis_store.rs#L478)):*

I'm saying:
```rs
if let Some(pub_sub_channel) = &self.pub_sub_channel {
  pipe.publish(pub_sub_channel, format!("{:?} {:?}", key, &chunk[..]));
}
```

Otherwise you are publishing the optional and not the actual value.
</blockquote></details>

Makes sense and still working fine same as before I didn't accidentally adjust to the option on the other end haha. Done.

allada (changes requested):
> Reviewed 2 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 1 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 5 discussions need to be resolved

___
*[`nativelink-service/src/bep_server.rs` line 60 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0ArSaC840z23hTrw79:-O0ArSaC840z23hTrw7A:b7xmde7) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/src/bep_server.rs#L60)):*
> ```XML
>         invocation_id: &'a str,
>         index: i64,
>     ) -> StoreKey<'a> {
> ```

nit: The contract is actually:
```rs
fn create_store_key(
        event_type: &str,
        build_id: &str,
        invocation_id: &str,
        index: i64,
    ) -> StoreKey<'static>
```
___
*[`nativelink-service/src/bep_server.rs` line 61 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0Arm5dAtz740LRavfW:-O0Arm5dAtz740LRavfX:b-2bllq7) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/src/bep_server.rs#L61)):*
> ```XML
>         index: i64,
>     ) -> StoreKey<'a> {
>         StoreKey::Str(Cow::Owned(format!(
> ```

nit: lol, now I would inline this, since you are not suffixing it with `-L` and it's self-explanatory now if inlined.
___
*[`nativelink-service/tests/bep_server_test.rs` line 199 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0AsI4WB4w-SF0jWUjS:-O0AsI4WB4w-SF0jWUjT:bwxyk3q) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/tests/bep_server_test.rs#L199)):*
> ```XML
>                 ordered_build_event: Some(OrderedBuildEvent {
>                     stream_id: Some(stream_id.clone()),
>                     sequence_number: 1,
> ```

This looks wrong. Why is it not passing the sequence number any more?
___
*[`nativelink-service/tests/bep_server_test.rs` line 298 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0AsWpN3NrjKAl9FzZi:-O0AsWpN3NrjKAl9FzZj:b-3aiopj) ([raw file](https://github.com/TraceMachina/nativelink/blob/47a25b87e2c9159fcf9d93fd28e62e59e5684f65/nativelink-service/tests/bep_server_test.rs#L298)):*
> ```XML
> 
>             // First, check if the response matches what we expect.
>             assert_eq!(response.sequence_number, sequence_number);
> ```

nit: Why is this changed?
___
*[`nativelink-store/src/redis_store.rs` line 500 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0Asg2qBNnQJFp2qHGo:-O0Asg2qBNnQJFp2qHGp:b-hb2gok) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-store/src/redis_store.rs#L500)):*
> ```XML
> 
>         if let Some(pub_sub_channel) = &self.pub_sub_channel {
>             conn.publish(pub_sub_channel, format!("{final_key}"))
> ```

nit: try `final_key.into()` or `&final_key` (or see if it'll take it without conversion.

blakehatch (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027)* status: 1 of 1 LGTMs obtained, and pending CI: Installation / ubuntu-22.04, Publish image, Vercel, asan / ubuntu-22.04, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, vale, and 2 discussions need to be resolved

___
*[`nativelink-service/src/bep_server.rs` line 60 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0ArSaC840z23hTrw79:-O0B1FmQ1h-dc7-sXRPS:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/src/bep_server.rs#L60)):*

nit: The contract is actually:
```rs
fn create_store_key(
        event_type: &str,
        build_id: &str,
        invocation_id: &str,
        index: i64,
    ) -> StoreKey<'static>
```
</blockquote></details>

Done.
___
*[`nativelink-service/src/bep_server.rs` line 61 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0Arm5dAtz740LRavfW:-O0B47d_4eR6u9Ej2MT3:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/src/bep_server.rs#L61)):*

nit: lol, now I would inline this, since you are not suffixing it with `-L` and it's self-explanatory now if inlined.
</blockquote></details>

Done.
___
*[`nativelink-service/tests/bep_server_test.rs` line 199 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0AsI4WB4w-SF0jWUjS:-O0B48MPE1i0wor8l1BL:bp3aow6) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-service/tests/bep_server_test.rs#L199)):*

This looks wrong. Why is it not passing the sequence number any more?
</blockquote></details>

Reverted.
___
*[`nativelink-service/tests/bep_server_test.rs` line 298 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0AsWpN3NrjKAl9FzZi:-O0B45zUEOmoVuJI3zZ3:bp3aow6) ([raw file](https://github.com/TraceMachina/nativelink/blob/47a25b87e2c9159fcf9d93fd28e62e59e5684f65/nativelink-service/tests/bep_server_test.rs#L298)):*

nit: Why is this changed?
</blockquote></details>

Reverted.
___
*[`nativelink-store/src/redis_store.rs` line 500 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-O0Asg2qBNnQJFp2qHGo:-O0B0vFK1bdE6zpWw9cZ:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/704379ef554993739967237966eed30f166fec5d/nativelink-store/src/redis_store.rs#L500)):*

nit: try `final_key.into()` or `&final_key` (or see if it'll take it without conversion.
</blockquote></details>

Done.

allada (approved):
:lgtm:

> Reviewed 1 of 7 files at r3, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1027#-:-O0BHORP4a_w6d1-Je_T:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 1 LGTMs obtained

PR #437, opened 2023-12-01
Rename ExistanceStore to ExistanceCacheStore

allada:
Minor changes to ExistanceStore to be more consistent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/native-link/437)

PR #1057, opened 2024-06-26
Resolved Bytes vtable

aleksdmladenovic:
# Description
I exposed the drop function of Bytes for resolving the memory fragmentation issues because Bytes are often very small (~50% of the data is less than 500 bytes). I invoked the drop function by using std::mem::drop().

[#437](https://github.com/tokio-rs/bytes/issues/437)

## Type of change

## How Has This Been Tested?

- [x] Tests added/amended

*****
For testing, I added one test script for evicting items when there is not enough space.

**Script**: cargo test --test memory_store_test insert_drop_test -- --nocapture

**Output**: 
- Printed **Evicting item** when the items should be evicted because there is no enough space.
- Printed **BytesWrapper dropped** and the Bytes data when the Bytes are dropped.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1057)


allada (commented):
This won't work for what we are trying to do. Hyper/Tonic/S3 takes a `Bytes` when sending http requests and we don't want to force a copy every time someone reads something. We need the drop to be called when Hyper/Tonic/S3 is done with it, not when we are done with it.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1057#-:-O0K44pX3WBVoko-Y4-9:bk1y1pn)* status: 0 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), license/cla, macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

Issue #359, opened 2023-10-31
[Design] Support shared database API for schedulers


allada:
We should support a transactional database API for schedulers. By doing this we could run multiple schedulers and have each worker talk to any scheduler. The schedulers would then be able to synchronize through the transactional database.

Right now everything is in memory and uses mutex for synchronization. If we supported row-level-locking via an API it would unlock a lot of possibilities.

PR #1069, opened 2024-06-29
[Refactor] Move worker notification in SimpleScheduler under Workers

allada:
Moves the logic on when the matching enginge trigger gets run to
under the workers struct where easy. This splits the logic of
when a task is changed and matching engine needs to run and when
a task gets run and the matching engine needs to be run.

towards: #359

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1069)

### DIFF
--- a/nativelink-scheduler/src/action_scheduler.rs
+++ b/nativelink-scheduler/src/action_scheduler.rs
@@ -42,7 +42,7 @@ pub trait ActionScheduler: Sync + Send + Unpin {
     async fn find_existing_action(
         &self,
         unique_qualifier: &ActionInfoHashKey,
-    ) -> Option<watch::Receiver<Arc<ActionState>>>;
+    ) -> Result<Option<watch::Receiver<Arc<ActionState>>>, Error>;
 
     /// Cleans up the cache of recently completed actions.
     async fn clean_recently_completed_actions(&self);
--- a/nativelink-scheduler/src/cache_lookup_scheduler.rs
+++ b/nativelink-scheduler/src/cache_lookup_scheduler.rs
@@ -222,11 +222,11 @@ impl ActionScheduler for CacheLookupScheduler {
     async fn find_existing_action(
         &self,
         unique_qualifier: &ActionInfoHashKey,
-    ) -> Option<watch::Receiver<Arc<ActionState>>> {
+    ) -> Result<Option<watch::Receiver<Arc<ActionState>>>, Error> {
         {
             let cache_check_actions = self.cache_check_actions.lock();
             if let Some(rx) = subscribe_to_existing_action(&cache_check_actions, unique_qualifier) {
-                return Some(rx);
+                return Ok(Some(rx));
             }
         }
         // Cache skipped may be in the upstream scheduler.
--- a/nativelink-scheduler/src/grpc_scheduler.rs
+++ b/nativelink-scheduler/src/grpc_scheduler.rs
@@ -260,7 +260,7 @@ impl ActionScheduler for GrpcScheduler {
     async fn find_existing_action(
         &self,
         unique_qualifier: &ActionInfoHashKey,
-    ) -> Option<watch::Receiver<Arc<ActionState>>> {
+    ) -> Result<Option<watch::Receiver<Arc<ActionState>>>, Error> {
         let request = WaitExecutionRequest {
             name: unique_qualifier.action_name(),
         };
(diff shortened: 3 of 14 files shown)


allada (commented):
+@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1069#-:-O0WVZ7I-M95JBheTPMD:b-bn1n1c)* status: 0 of 1 LGTMs obtained, and pending CI: Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable (waiting on @adam-singer)

zbirenbaum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1069)* status: 0 of 1 LGTMs obtained, and 2 discussions need to be resolved (waiting on @adam-singer)

___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 389 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1069#-O0XEkhNDnHib-y-BzP1:-O0XEkhNDnHib-y-BzP2:b-xgz5eo) ([raw file](https://github.com/TraceMachina/nativelink/blob/34d93b7d1cda7c0f4a430364b7588002ab612494/nativelink-scheduler/src/simple_scheduler.rs#L389)):*
> ```XML
> 
>                     let operation_id = state.id.clone();
>                     let ret = <StateManager as MatchingEngineStateManager>::update_operation(
> ```

If `maybe_worker_id` is `None` we shouldn't update the action at all. We should only mark it as `Executing` when we know for certain that there is a worker to run it on.

If the ID is none we failed to find a worker that meets the action criteria, which is a totally valid case. just create a `INFO` event and continue. That should clean up all the logic below substantially.

This is a bit simpler and should be functionally equivalent:
```
let maybe_worker_id = self
    .workers
    .find_worker_for_action(&action_info.platform_properties);

let Some(worker_id) = maybe_worker_id else {
    event!(
        Level::INFO,
        "Failed to find worker for action: {}",
        action_info.unique_qualifier.action_name()
    );
    continue;
};

// At this point we know WorkerId is Some and can mark it as Executing.
let operation_id = state.id.clone();
let ret = <StateManager as MatchingEngineStateManager>::update_operation(
    &self.state_manager,
    operation_id.clone(),
    maybe_worker_id,
    Ok(ActionStage::Executing),
)
.await;

if let Err(e) = ret {
    let max_job_retries = self.max_job_retries;
    let metrics = self.metrics.clone();
    // TODO(allada) This is to get around rust borrow checker with double mut borrows
    // of a mutex lock. Once the scheduler is fully moved to state manager this can be
    // removed.
    let state_manager = self.state_manager.clone();
    let mut inner_state = state_manager.inner.lock().await;
    SimpleSchedulerImpl::immediate_evict_worker(
        &mut inner_state,
        &mut self.workers,
        max_job_retries,
        &metrics,
        &worker_id,
        e.clone(),
    );
    event!(
        Level::ERROR,
        ?e,
        "update operation failed for {}",
        operation_id
    );
    continue;
}

// Once we get here we know that the operation update was successful so we notify the worker.
let run_action_result = self
    .worker_notify_run_action(worker_id, action_info.clone())
    .await;

if let Err(err) = run_action_result {
    event!(
        Level::ERROR,
        ?err,
        ?worker_id,
        ?action_info,
        "failed to run worker_notify_run_action in SimpleSchedulerImpl::do_try_match"
    );
}

```
___
*[`nativelink-scheduler/src/simple_scheduler.rs` line 427 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1069#-O0XJcL1BQIuOEPqVrxW:-O0XJcL1BQIuOEPqVrxX:b-cws3h9) ([raw file](https://github.com/TraceMachina/nativelink/blob/34d93b7d1cda7c0f4a430364b7588002ab612494/nativelink-scheduler/src/simple_scheduler.rs#L427)):*
> ```XML
>                             .await
>                         {
>                             event!(
> ```

Is it really ok to just emit an event here? Wouldn't the operation be stuck  if we update the scheduler-side state to say it is executing on a worker before we actually succeed in assigning it to the worker?

We should do a second update here to fix the mismatch
```
let revert_res = MatchingEngineStateManager>::update_operation(
    &self.state_manager,
    operation_id.clone(),
    None,
    Ok(ActionStage::Queued),
)
.await;
```

It's pretty unlikely revert_res will fail here, but if it does that's a pretty bad error I think.

adam-singer (approved):
:lgtm:

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1069#-:-O0XXB6e7SFRsV617wpL:bnfp4nl)* status: 1 of 1 LGTMs obtained, and 2 discussions need to be resolved

PR #1089, opened 2024-07-06
QA fixes for scheduler-v2

zbirenbaum:
# Description
Introduces various code quality improvements to scheduler-v2 which allow bazel test to pass.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)

### DIFF
--- a/nativelink-proto/genproto/com.github.trace_machina.nativelink.remote_execution.pb.rs
+++ b/nativelink-proto/genproto/com.github.trace_machina.nativelink.remote_execution.pb.rs
@@ -71,6 +71,7 @@ pub struct ExecuteResult {
     /// that case the server SHOULD infer the digest function using the
     /// length of the action digest hash and the digest functions announced
     /// in the server's capabilities.
+    /// TODO!!! THIS IS UNUSED AFAIK.
     #[prost(
         enumeration = "super::super::super::super::super::build::bazel::remote::execution::v2::digest_function::Value",
         tag = "7"
@@ -159,7 +160,7 @@ pub struct StartExecute {
     pub execute_request: ::core::option::Option<
         super::super::super::super::super::build::bazel::remote::execution::v2::ExecuteRequest,
     >,
-    /// Id of the operation.
+    /// / Id of the operation.
     #[prost(string, tag = "4")]
     pub operation_id: ::prost::alloc::string::String,
     /// / The time at which the command was added to the queue to allow population
--- a/nativelink-scheduler/BUILD.bazel
+++ b/nativelink-scheduler/BUILD.bazel
@@ -55,6 +55,7 @@ rust_library(
         "@crates//:scopeguard",
         "@crates//:serde",
         "@crates//:serde_json",
+        "@crates//:static_assertions",
         "@crates//:tokio",
         "@crates//:tokio-stream",
         "@crates//:tonic",
--- a/nativelink-scheduler/src/cache_lookup_scheduler.rs
+++ b/nativelink-scheduler/src/cache_lookup_scheduler.rs
@@ -44,6 +44,10 @@ use tracing::{event, Level};
 use crate::action_scheduler::ActionScheduler;
 use crate::platform_property_manager::PlatformPropertyManager;
 
+/// A future containing the resolved `ClientOperationId` once it is figured out.
+/// This future may be cloned and will always yield the same value once resolved.
+type ClientOperationIdFuture = SharedFuture<oneshot::Receiver<ClientOperationId>>;
+
 /// Actions that are having their cache checked or failed cache lookup and are
 /// being forwarded upstream.  Missing the skip_cache_check actions which are
 /// forwarded directly.
@@ -97,10 +101,7 @@ async fn get_action_from_store(
 fn subscribe_to_existing_action(
     cache_check_actions: &MutexGuard<CheckActions>,
     unique_qualifier: &ActionInfoHashKey,
-) -> Option<(
-    SharedFuture<oneshot::Receiver<ClientOperationId>>,
-    watch::Receiver<Arc<ActionState>>,
-)> {
+) -> Option<(ClientOperationIdFuture, watch::Receiver<Arc<ActionState>>)> {
     cache_check_actions
         .get(unique_qualifier)
         .map(|(client_operation_id_rx, rx)| {
--- a/nativelink-scheduler/src/scheduler_state/awaited_action.rs
+++ b/nativelink-scheduler/src/scheduler_state/awaited_action.rs
@@ -153,7 +153,7 @@ impl AwaitedAction {
     /// will hold a lock preventing anyone else from reading or modifying the
     /// sort key until the result is dropped.
     #[must_use]
-    pub fn set_priority<'a>(&'a self, new_priority: i32) -> Option<SortInfoLock<'a>> {
+    pub fn set_priority(&self, new_priority: i32) -> Option<SortInfoLock> {
         let sort_info_lock = self.sort_info.upgradable_read();
         if sort_info_lock.priority == new_priority {
             return None;
@@ -173,7 +173,7 @@ impl AwaitedAction {
     }
 
     /// Gets the sort info of the action.
-    pub fn get_sort_info<'a>(&'a self) -> SortInfoLock<'a> {
+    pub fn get_sort_info(&self) -> SortInfoLock {
         let sort_info = self.sort_info.read();
         SortInfoLock {
             previous_sort_key: sort_info.sort_key,
@@ -199,7 +199,7 @@ impl AwaitedAction {
 
     /// Gets the worker id that is currently processing this action.
     pub fn get_worker_id(&self) -> Option<WorkerId> {
-        self.worker_id.read().clone()
+        *self.worker_id.read()
     }
 
     /// Sets the worker id that is currently processing this action.
@@ -244,6 +244,7 @@ impl AwaitedAction {
 pub struct AwaitedActionSortKey(u128);
 
 impl AwaitedActionSortKey {
+    #[rustfmt::skip]
     const fn new(priority: i32, insert_timestamp: u64, hash: [u8; 4]) -> Self {
         // Shift `new_priority` so [`i32::MIN`] is represented by zero.
         // This makes it so any nagative values are positive, but
@@ -251,9 +252,10 @@ impl AwaitedActionSortKey {
         const MIN_I32: i64 = (i32::MIN as i64).abs();
         let priority = ((priority as i64 + MIN_I32) as u32).to_be_bytes();
 
+        // Invert our timestamp so the larger the timestamp the lower the number.
+        // This makes timestamp descending order instead of ascending.
         let timestamp = (insert_timestamp ^ u64::MAX).to_be_bytes();
 
-        #[cfg_attr(rustfmt, rustfmt_skip)]
         AwaitedActionSortKey(u128::from_be_bytes([
             priority[0], priority[1], priority[2], priority[3],
             timestamp[0], timestamp[1], timestamp[2], timestamp[3],
@@ -286,7 +288,7 @@ const_assert_eq!(
     // Note: `6543210fedcba987` are the inverted bits of `9abcdef012345678`.
     // This effectively inverts the priority to now have the highest priority
     // be the lowest timestamps.
-    AwaitedActionSortKey(0x92345678_6543210fedcba987_9abcdef0).0
+    AwaitedActionSortKey(0x9234_5678_6543_210f_edcb_a987_9abc_def0).0
 );
 // Ensure the priority is used as the sort key first.
 const_assert!(
@@ -354,11 +356,9 @@ impl MetricsComponent for AwaitedAction {
         );
         c.publish(
             "worker_id",
-            &format!(
-                "{}",
-                self.get_worker_id()
-                    .map_or(String::new(), |v| v.to_string())
-            ),
+            &self
+                .get_worker_id()
+                .map_or(String::new(), |v| v.to_string()),
             "The current worker processing the action (if any).",
         );
     }
--- a/nativelink-scheduler/src/scheduler_state/client_action_state_result.rs
+++ b/nativelink-scheduler/src/scheduler_state/client_action_state_result.rs
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{borrow::Cow, sync::Arc};
+use std::borrow::Cow;
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use nativelink_error::Error;
--- a/nativelink-scheduler/src/scheduler_state/matching_engine_action_state_result.rs
+++ b/nativelink-scheduler/src/scheduler_state/matching_engine_action_state_result.rs
@@ -12,16 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{borrow::Cow, sync::Arc};
+use std::borrow::Cow;
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use nativelink_error::Error;
 use nativelink_util::action_messages::{ActionInfo, ActionState};
 use tokio::sync::watch;
 
-use crate::operation_state_manager::ActionStateResult;
-
 use super::awaited_action::AwaitedAction;
+use crate::operation_state_manager::ActionStateResult;
 
 pub struct MatchingEngineActionStateResult {
     awaited_action: Arc<AwaitedAction>,
--- a/nativelink-scheduler/src/scheduler_state/state_manager.rs
+++ b/nativelink-scheduler/src/scheduler_state/state_manager.rs
@@ -21,7 +21,7 @@ use async_lock::Mutex;
 use async_trait::async_trait;
 use futures::stream::{self, unfold};
 use hashbrown::HashMap;
-use nativelink_error::{make_err, Code, Error};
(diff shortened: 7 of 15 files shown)


zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and pending CI: vale, and 3 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 258 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18sdN00gt_CFGpxTTD:-O18sdN00gt_CFGpxTTE:bjt952r) ([raw file](https://github.com/TraceMachina/nativelink/blob/7e1d597c929b85a8ae1d2e9328c2b91772672326/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L258)):*
> ```XML
>         let timestamp = (insert_timestamp ^ u64::MAX).to_be_bytes();
> 
>         // TODO: I can't find a way to make clippy not complain about this that doesn't introduce another complaint.
> ```

Using  `#[cfg_attr(rustfmt, rustfmt_skip)]` gives a clippy deprecation error to use `#[rustfmt::skip]` but using `#[rustfmt::skip]` on just this line instead of the function gives a different error regarding experimental feature usage.
___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 283 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18sUTC8clkLLgYBFb6:-O18sUTC8clkLLgYBFb7:bpwoln2) ([raw file](https://github.com/TraceMachina/nativelink/blob/7e1d597c929b85a8ae1d2e9328c2b91772672326/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L283)):*
> ```XML
> // Ensure the size of the sort key is the same as a `u64`.
> assert_eq_size!(AwaitedActionSortKey, u128);
> ```

I'm not sure what to do about this whole code section tbh.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 114 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18sLiSCd3yWBXNLfgA:-O18sLiSCd3yWBXNLfgB:b62eg9l) ([raw file](https://github.com/TraceMachina/nativelink/blob/7e1d597c929b85a8ae1d2e9328c2b91772672326/nativelink-scheduler/src/scheduler_state/state_manager.rs#L114)):*
> ```XML
>     }
> 
>     #[allow(clippy::mutable_key_type)]
> ```

Can we find some way to not use mutable keys?

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 5 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 284 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uANv8M6Mgk1o3uhS:-O18uANv8M6Mgk1o3uhT:bxlp265) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L284)):*
> ```XML
> assert_eq_size!(AwaitedActionSortKey, u128);
> 
> // TODO: Shouldn't these be tests? Why are we inlining asserts here.
> ```

I'm not sure what to do about this whole section.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 259 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uHpYDtS00TSQDE_1:-O18uHpYDtS00TSQDE_2:b-c2tdka) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/state_manager.rs#L259)):*
> ```XML
>     }
> 
>     // TODO: ActionInfo isn't an error. The compiler wants to box this
> ```

Remove comment once implementation is given the goahead.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 300 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uVjz2wBaY6eQlTgg:-O18uVjz2wBaY6eQlTgh:blq72dc) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/state_manager.rs#L300)):*
> ```XML
>                     });
>             let Some(mut sorted_awaited_action) = maybe_sorted_awaited_action else {
>                 // TODO: Either use event on all of the above error here, but both is overkill.
> ```

Address this comment.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 8 discussions need to be resolved

___
*[`nativelink-scheduler/BUILD.bazel` line 58 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18vGP54htyyn6tF0g0:-O18vGP54htyyn6tF0g1:b-jn7l7s) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/BUILD.bazel#L58)):*
> ```Starlark
>         "@crates//:serde",
>         "@crates//:serde_json",
>         "@crates//:static_assertions",
> ```

@jhpratt said that we can use `assert!()` now instead as of recent versions of rust... maybe we should try?
___
*[`nativelink-scheduler/src/cache_lookup_scheduler.rs` line 47 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18vZZrCoIP-qXTdoAT:-O18vZZrCoIP-qXTdoAU:bsybmos) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/cache_lookup_scheduler.rs#L47)):*
> ```XML
> use crate::platform_property_manager::PlatformPropertyManager;
> 
> // TODO: Document this type. Why is it only used in CacheLookupScheduler for subscription.
> ```

```rs
/// A future containing the resolved `ClientOperationId` once it is figured out.
/// This future may be cloned and will always yield the same value once resolved.
type SubscriptionFuture = SharedFuture<oneshot::Receiver<ClientOperationId>>;
```
___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 284 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uANv8M6Mgk1o3uhS:-O18wK8XBcr7WeS5rPux:b-42rtk9) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L284)):*

I'm not sure what to do about this whole section.
</blockquote></details>

We could make them tests, but since this entire struct is `const`, we can just make the compiler test the functions for us.

No need to build a binary for these, when the compiler can test it.
___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 365 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18wea70aJ673KZa23p:-O18wea70aJ673KZa23q:b-zi2enl) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L365)):*
> ```XML
>                 .get_worker_id()
>                 .map_or(String::new(), |v| v.to_string())
>                 .to_string(),
> ```

nit: `.to_string()` probably not needed here.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 8 discussions need to be resolved

___
*[`nativelink-scheduler/src/cache_lookup_scheduler.rs` line 47 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18vZZrCoIP-qXTdoAT:-O18xu0CAgPXI7Lr9gkz:b-28mc0t) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/cache_lookup_scheduler.rs#L47)):*

```rs
/// A future containing the resolved `ClientOperationId` once it is figured out.
/// This future may be cloned and will always yield the same value once resolved.
type SubscriptionFuture = SharedFuture<oneshot::Receiver<ClientOperationId>>;
```
</blockquote></details>

Actually:
```rs
/// A future containing the resolved `ClientOperationId` once it is figured out.
/// This future may be cloned and will always yield the same value once resolved.
type ClientOperationIdFuture = SharedFuture<oneshot::Receiver<ClientOperationId>>;
```

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and pending CI: vale, and 5 discussions need to be resolved

___
*[`nativelink-scheduler/BUILD.bazel` line 58 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18vGP54htyyn6tF0g0:-O19GcOxDeFfZEaqdeVP:b2ncqjo) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/BUILD.bazel#L58)):*

@jhpratt said that we can use `assert!()` now instead as of recent versions of rust... maybe we should try?
</blockquote></details>

I tried this but couldn't get any of it to compile using the `assert!` equivalents. 
___
*[`nativelink-scheduler/src/cache_lookup_scheduler.rs` line 47 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18vZZrCoIP-qXTdoAT:-O19GqthDe-LhwlNsALR:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/cache_lookup_scheduler.rs#L47)):*

Actually:
```rs
/// A future containing the resolved `ClientOperationId` once it is figured out.
/// This future may be cloned and will always yield the same value once resolved.
type ClientOperationIdFuture = SharedFuture<oneshot::Receiver<ClientOperationId>>;
```
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 114 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18sLiSCd3yWBXNLfgA:-O19GSPS4UZowqFbC1gn:b-1l2ycy) ([raw file](https://github.com/TraceMachina/nativelink/blob/7e1d597c929b85a8ae1d2e9328c2b91772672326/nativelink-scheduler/src/scheduler_state/state_manager.rs#L114)):*

Can we find some way to not use mutable keys?
</blockquote></details>

Discussed offline. Added to top level so only one ignore is necessary.

allada (commented):
> Reviewed 11 of 15 files at r1, 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 7 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 255 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1CnkPWE823eIg9DY0E:-O1CnkPWE823eIg9DY0F:bqpvhvq) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L255)):*
> ```XML
>         let priority = ((priority as i64 + MIN_I32) as u32).to_be_bytes();
> 
>         let timestamp = (insert_timestamp ^ u64::MAX).to_be_bytes();
> ```

nit: While here can we say:
```rs
// Invert our timestamp so the larger the timestamp the lower the number.
// This makes timestamp descending order instead of ascending.
```
___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 283 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1Co2gy628qjHbib90d:-O1Co2gy628qjHbib90e:bdabcfu) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L283)):*
> ```XML
> 
> // TODO: Shouldn't these be tests? Why are we inlining asserts here.
> // Also there are a ton of complaints about every part of this and it's not maintainable code.
> ```

Then just fix it. Don't write a comment in the code complaining unless there's a clear path on how to fix it.
___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 292 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1CoCL50ailkmbjDc0C:-O1CoCL50ailkmbjDc0D:b-tcnmc7) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L292)):*
> ```XML
>     // This effectively inverts the priority to now have the highest priority
>     // be the lowest timestamps.
>     AwaitedActionSortKey(0x9234_5678_6543_210f_edcb_a987_9abc_def0).0
> ```

nit: Is this required? The underscores on the original was done that way to show where the different variables live:
```
0x92345678_6543210fedcba987_9abcdef0
-priority-_----timestamp---_--hash--
```
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 259 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uHpYDtS00TSQDE_1:-O1CpSwzAHzxg8FgL109:b-tgd7y2) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/state_manager.rs#L259)):*

Remove comment once implementation is given the goahead.
</blockquote></details>

Looking over the code, it looks like we should instead be passing in `unique_qualifier: &ActionInfoHashKey` and `priority: i32`

This will solve both problems if we rewrite this.

We might also want to make the return:
`Result<Option<Arc<ClientActionStateResult>>, Error>`
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 300 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O18uVjz2wBaY6eQlTgg:-O1CpsI3A0C2Aa5R1Y67:bsvykzk) ([raw file](https://github.com/TraceMachina/nativelink/blob/a72ea49324e3a1353ccf3679edd72711c2eeab9c/nativelink-scheduler/src/scheduler_state/state_manager.rs#L300)):*

Address this comment.
</blockquote></details>

If this happens it's kind of a big deal. If these get out of sync we have major problems, and worried about not being loud enough.

I would say we event and return an error here.
___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 445 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1CqTDHDHMPAFkjIf2h:-O1CqTDHDHMPAFkjIf2i:b-g61o5f) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/state_manager.rs#L445)):*
> ```XML
>             let inner = state.inner.lock().await;
> 
>             #[allow(clippy::mutable_key_type)]
> ```

needed still?

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 3 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/awaited_action.rs` line 283 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1Co2gy628qjHbib90d:-O1Dl58s1jLKZnDWyWCD:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/awaited_action.rs#L283)):*

Then just fix it. Don't write a comment in the code complaining unless there's a clear path on how to fix it.
</blockquote></details>

Done.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1089)* status: 0 of 1 LGTMs obtained, and 2 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 445 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1089#-O1CqTDHDHMPAFkjIf2h:-O1Do8-520L0vjylgF6b:b2nhswy) ([raw file](https://github.com/TraceMachina/nativelink/blob/a5c750a6c5354bb0a2a409999b88d8a8e26474e4/nativelink-scheduler/src/scheduler_state/state_manager.rs#L445)):*

needed still?
</blockquote></details>

Yes

zbirenbaum:
Merged in #1092

PR #1121, opened 2024-07-09
Fix sleep await duration hang

adam-singer:
While testing it was observed that `state_manager::make_client_keepalive_spawn` would enter into the `spawn!` but hang even after sleep duration passed. Some digging into https://github.com/rust-lang/rust/issues/57391 && https://doc.rust-lang.org/src/core/time.rs.html#110 shows that using `const Duration` might not be stable in all cases.

After applying this change awaking from sleep does work. Still suspect other issues around creating high number of sleeping threads for client connections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1121)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1121)* status: 0 of 1 LGTMs obtained, and 1 discussions need to be resolved

___
*[`nativelink-scheduler/src/scheduler_state/state_manager.rs` line 905 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1121#-O1N-uK29ueKiwuPGe9A:-O1N-uK33TykLLhYLOI-:bvhz3ho) ([raw file](https://github.com/TraceMachina/nativelink/blob/577785ded45e3c4cda1de4f26b6ad9cc2ea86496/nativelink-scheduler/src/scheduler_state/state_manager.rs#L905)):*
> ```XML
>     spawn!("client_action_state_result_keepalive", async move {
>         loop {
>             tokio::time::sleep(Duration::from_secs(KEEPALIVE_DURATION_SECS)).await;
> ```

nit: Can we instead just move this to a variable on `inner_weak`? We are going to need to do this anyway when we move to use configs.

PR #1122, opened 2024-07-09
Add new basic_cas that resembles production setup

blakehatch:
# Description
This variant of basic_cas is based on a production setup combined with the convience of using basic_cas locally as an all-in-one CAS, AC, and Worker,

## Type of change

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1122)

### DIFF
--- a/docs/src/content/docs/guides/prod-configurations.mdx
+++ b/docs/src/content/docs/guides/prod-configurations.mdx
@@ -1,0 +1,253 @@
+---
+title: "Production Configurations"
+description: "NativeLink production configs"
+pagefind: true
+---
+
+# Recommended Basic CAS
+
+This configuration is based on a production setup used by NativeLink. The main
+difference is that in a production environment, the CAS, AC, and Worker are
+usually separated. However, this version combines all three, similar to the
+basic CAS.
+
+Most production configurations use Redis and S3. However, this configuration
+is intended to serve as a bridge between the basic CAS, which can be used out
+of the box with the local filesystem and memory as stores, and a full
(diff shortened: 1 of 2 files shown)


allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1122)* status: 0 of 1 LGTMs obtained, and 1 discussions need to be resolved

___
*[`nativelink-config/examples/basic_cas_recommended.json` line 39 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1122#-O1MzCq46Zo0IFGBPpbx:-O1MzCq56T7Oft21dQJS:btf5nf) ([raw file](https://github.com/TraceMachina/nativelink/blob/5e55028b5c5e1aa6a833f23e9e975376489814fa/nativelink-config/examples/basic_cas_recommended.json#L39)):*
> ```JSON
>       }
>     },
>     "CAS_FAST_SLOW_STORE": {
> ```

nit: I don't think this will work. If used as a worker store, it needs to have the fast store be a `filesystem` and the slow store can be whatever.

What you actually want is to split this into a few different stores and use different ones in different places.

The CAS should use `fast_slow { fast -> size_partitioning { lower -> memory, upper -> noop }, slow -> filesystem }`

The worker should use: `fast_slow { fast -> filesystem, slow -> noop }` (but put comment that `noop` should be replaced with a remote storage store (S3, GRPC, Redis, exc...).

The filesystem stores both are referencing should be the same store.

PR #1180, opened 2024-07-19
Double protect output stream of verify store

allada:
VerifyStore's job is to protect corrupted data from being received. It is possible for the down-stream component to try and accept data before receiving the EOF. To keep this from being a problem, we have VeifyStore do the checks when conditions are met before sending the last chunk of data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1180)

### DIFF
--- a/nativelink-store/src/verify_store.rs
+++ b/nativelink-store/src/verify_store.rs
@@ -55,7 +55,7 @@ impl VerifyStore {
         &self,
         mut tx: DropCloserWriteHalf,
         mut rx: DropCloserReadHalf,
-        size_info: UploadSizeInfo,
+        maybe_expected_digest_size: Option<u64>,
         original_hash: [u8; 32],
         mut maybe_hasher: Option<&mut D>,
     ) -> Result<(), Error> {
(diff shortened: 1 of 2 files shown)


aaronmondal (approved):
:lgtm:

nit: Reconsider whether this is really something that should be enabled by default. With the S3 store it only affected a part of the system where we knew that this was an issue.

In contrast, since the VerifyStore is often used in conjunction with the MemoryStore it might be in a hotter path and the loop here might be called a lot more.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1180#-:-O2C-Aj6Ezy_a0WQUDvr:bozv635)* status: 1 of 1 LGTMs obtained, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Installation / macos-13, Remote / large-ubuntu-22.04, macos-13, and 1 discussions need to be resolved

___
*[`nativelink-store/src/verify_store.rs` line 100 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1180#-O2Bz3eE1WwatyqvuEbc:-O2Bz3eE1WwatyqvuEbd:b1vozxo) ([raw file](https://github.com/TraceMachina/nativelink/blob/bb893e4dcbb247329552f6db31af111515b45b1f/nativelink-store/src/verify_store.rs#L100)):*
> ```XML
>             }
> 
>             if done {
> ```

nit: The branching with the `done` and `chunk.is_empty()` is a bit confusing and I'm not sure whether this loop is easy enough for the compiler to unroll. Instead, there might be a more "inline-y" way to map out these branches to increase the potential for the compiler to specialize the branches.

allada (commented):
This should not be causing any performance regression. No additional buffering is taking place, this is just changing the logic to finish hasing before sending it instead of on next loop cycle.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1180#-:-O2CKdp6D5diZdMtklnq:bv0fo63)* status: 1 of 1 LGTMs obtained, and 1 discussions need to be resolved

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1180)* status: 1 of 1 LGTMs obtained, and 1 discussions need to be resolved

___
*[`nativelink-store/src/verify_store.rs` line 100 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1180#-O2Bz3eE1WwatyqvuEbc:-O2CLfca4gSJWD3QMDei:b-qtqa8k) ([raw file](https://github.com/TraceMachina/nativelink/blob/bb893e4dcbb247329552f6db31af111515b45b1f/nativelink-store/src/verify_store.rs#L100)):*

nit: The branching with the `done` and `chunk.is_empty()` is a bit confusing and I'm not sure whether this loop is easy enough for the compiler to unroll. Instead, there might be a more "inline-y" way to map out these branches to increase the potential for the compiler to specialize the branches.
</blockquote></details>

To be honest, I looked over this a few times and could not come up with an easier way to express it :-(

PR #1182, opened 2024-07-20
WIP: RedisAwaitedActionDb

zbirenbaum:
Commit for early feedback and design planning.

# Description

Notable partial or full TODO items include:
- Handling drop events
- Serialize/Deserialize impl for ActionState
- Make client actually accessible from RedisStore
- Error handling
- Return sorted AwaitedActions with ChunkStream 

## Type of change

Please delete options that aren't relevant.

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Please also list any relevant details for your test configuration

## Checklist

- [ ] Updated documentation if needed
- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1182)

PR #1206, opened 2024-07-27
Convert AwaitedAction to and from raw bytes

zbirenbaum:
Makes AwaitedAction Serialize/Deserialize and adds ability to convert to
and from raw bytes. This will be used to communicate operation data to
and external data storage such as Redis through the store API.

Towards #1182

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1782,6 +1782,8 @@ dependencies = [
  "prost",
  "rand",
  "scopeguard",
+ "serde",
+ "serde_json",
  "static_assertions",
  "tokio",
  "tokio-stream",
--- a/nativelink-scheduler/BUILD.bazel
+++ b/nativelink-scheduler/BUILD.bazel
@@ -43,6 +43,8 @@ rust_library(
         "@crates//:parking_lot",
         "@crates//:rand",
         "@crates//:scopeguard",
+        "@crates//:serde",
+        "@crates//:serde_json",
         "@crates//:static_assertions",
         "@crates//:tokio",
         "@crates//:tokio-stream",
--- a/nativelink-scheduler/Cargo.toml
+++ b/nativelink-scheduler/Cargo.toml
@@ -27,6 +27,8 @@ tokio = "1.38.0"
 tokio-stream = { version = "0.1.15", default-features = false }
 tonic = { version = "0.12.0", default-features = false }
 tracing = { version = "0.1.40", default-features = false }
+serde = { version = "1.0.203", features = ["rc"] }
+serde_json = "1.0.120"
 static_assertions = "1.1.0"
 
 [dev-dependencies]
--- a/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
+++ b/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
@@ -15,18 +15,20 @@
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use nativelink_error::{make_input_err, Error, ResultExt};
 use nativelink_metric::{
     MetricFieldData, MetricKind, MetricPublishKnownKindData, MetricsComponent,
 };
 use nativelink_util::action_messages::{
     ActionInfo, ActionStage, ActionState, OperationId, WorkerId,
 };
+use serde::{Deserialize, Serialize};
 use static_assertions::{assert_eq_size, const_assert, const_assert_eq};
 
 /// The version of the awaited action.
 /// This number will always increment by one each time
 /// the action is updated.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 struct AwaitedActionVersion(u64);
 
 impl MetricsComponent for AwaitedActionVersion {
@@ -40,7 +42,7 @@ impl MetricsComponent for AwaitedActionVersion {
 }
 
 /// An action that is being awaited on and last known state.
-#[derive(Debug, Clone, MetricsComponent)]
+#[derive(Debug, Clone, MetricsComponent, Serialize, Deserialize)]
 pub struct AwaitedAction {
     /// The current version of the action.
     #[metric(help = "The version of the AwaitedAction")]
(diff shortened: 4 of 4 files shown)


allada (changes requested):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206)* status: 0 of 1 LGTMs obtained, and 1 of 4 files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-scheduler/src/awaited_action_db/awaited_action.rs` line 152 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1206#-O2ml_Ge77Y1FmdcIxIS:-O2ml_Ge77Y1FmdcIxIT:b-wilz5e) ([raw file](https://github.com/TraceMachina/nativelink/blob/f975d3b96974d02d28ba4f087464ece3f716bbb2/nativelink-scheduler/src/awaited_action_db/awaited_action.rs#L152)):*
> ```XML
> }
> 
> impl TryInto<Vec<u8>> for AwaitedAction {
> ```

Can we use protos or json strings instead?

the reason is because `bincode` is not multi-language compatible and allowing users to read the state out of redis directly (for scaling cloud instances) is a real use case.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206)* status: 0 of 1 LGTMs obtained, and 0 of 4 files reviewed, and pending CI: pre-commit-checks, and 1 discussions need to be resolved

___
*[`nativelink-scheduler/src/awaited_action_db/awaited_action.rs` line 152 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1206#-O2ml_Ge77Y1FmdcIxIS:-O2q3Ehe2kYkzbYDcn8D:b-rfdzqo) ([raw file](https://github.com/TraceMachina/nativelink/blob/f975d3b96974d02d28ba4f087464ece3f716bbb2/nativelink-scheduler/src/awaited_action_db/awaited_action.rs#L152)):*

Can we use protos or json strings instead?

the reason is because `bincode` is not multi-language compatible and allowing users to read the state out of redis directly (for scaling cloud instances) is a real use case.
</blockquote></details>

Great point, updated it to use serde_json.

As a side note I did some digging and found this really useful tool which compared the performance of various encoders/decoders and serde_json is a really poor performer in the grand scheme of things.

https://david.kolo.ski/rust_serialization_benchmark

Prost is better but would lock us out of most or possibly all redis-specific advantages for performance.

I don't think this is something to prioritize atm though, serde_json should be fine for now, and this is abstracted to just produce/consume bytes so easily updatable later.

allada (approved):
:lgtm:

> Reviewed 1 of 4 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206#-:-O2w7zCY5hYZKtYzC8gi:bnfp4nl)* status: 1 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-scheduler/src/awaited_action_db/awaited_action.rs` line 152 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1206#-O2ml_Ge77Y1FmdcIxIS:-O2w7lF6ASYVwE-wElzX:b-nkk3oh) ([raw file](https://github.com/TraceMachina/nativelink/blob/f975d3b96974d02d28ba4f087464ece3f716bbb2/nativelink-scheduler/src/awaited_action_db/awaited_action.rs#L152)):*

Great point, updated it to use serde_json.

As a side note I did some digging and found this really useful tool which compared the performance of various encoders/decoders and serde_json is a really poor performer in the grand scheme of things.

https://david.kolo.ski/rust_serialization_benchmark

Prost is better but would lock us out of most or possibly all redis-specific advantages for performance.

I don't think this is something to prioritize atm though, serde_json should be fine for now, and this is abstracted to just produce/consume bytes so easily updatable later.
</blockquote></details>

Yeah, performance is not something that is critical here, since we are going to be able to paralyze this. We could also make this a config option at some point.

zbirenbaum (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206)* status: 1 of 1 LGTMs obtained, and 2 of 4 files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Cloudflare Pages, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (20.04), docker-compose-compiles-nativelink (22.04), integration-tests (20.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable

___
*[`nativelink-scheduler/src/awaited_action_db/awaited_action.rs` line 152 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1206#-O2ml_Ge77Y1FmdcIxIS:-O2w9aLD4NL0P20FCrdI:b-hfghm4) ([raw file](https://github.com/TraceMachina/nativelink/blob/f975d3b96974d02d28ba4f087464ece3f716bbb2/nativelink-scheduler/src/awaited_action_db/awaited_action.rs#L152)):*

Yeah, performance is not something that is critical here, since we are going to be able to paralyze this. We could also make this a config option at some point.
</blockquote></details>

Sounds good

zbirenbaum (commented):
> Reviewed 1 of 4 files at r2, 1 of 3 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1206)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

Issue #1082, opened 2024-07-05
Add documentation tests to CI


aaronmondal:
To prevent issues like #1081 in the future.

While it's a bit tedious to create entire deployment pipelines, we should have at least run doc builds as a CI test. Preview builds already run the Typescript linting infrastructure and invoke the generator scripts on all inputs used which should cover a reasonably large surface area of potential issues.

Evaan2001:
I'm working on this issue right now; PR should hopefully be up tomorrow, July 28th.

Evaan2001:
Pull request is up – github.com/TraceMachina/nativelink/pull/1217. I'm failing a CI test. I edited the run_integration_tests.sh script and one of my comments seems to cause the failure. As this is my first time editing a CI pipeline, some feedback/direction would be great!

PR #1224, opened 2024-07-30
Add "docs builds" Into CI Tests

Evaan2001:
# Description

This commit tackles Issue #1082, where the goal is to run "docs builds" as a CI test to ensure documentation gets built properly.

I saw that .github/workflows/main.yml has a CI job / GithubAction called "integration-tests" that executes the commands contained in run_integration_tests.sh, a shell script. This shell script has the line "set -euo pipefail" which will cause the execution of this script to break whenever a command has a non-Zero exit code. 

I added the command "docs builds" to run_integration_tests.sh. In case "docs builds" fails, it will exit with a non-zero exit code and cause run_integration_tests.sh to terminate abruptly which would in turn fail the "integration-tests" job in main.yml. This addition should ensure that "docs builds" always runs before merging a PR.

Fixes # 1082

## Type of change

## How Has This Been Tested?

Need some help here – I tested "docs builds" by temporarily creating files with invalid rust syntax to examine error codes. I've never worked on editing CI pipelines and was unsure about what further tests to conduct. I'm unsure if GitHub Actions will find the "docs" command, which builds off Rust's cargo doc. Any suggestions would be helpful :)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1224)

### DIFF
--- a/run_integration_tests.sh
+++ b/run_integration_tests.sh
@@ -95,6 +95,10 @@ DID_FAIL=0
 export NATIVELINK_DIR="$CACHE_DIR/nativelink"
 mkdir -p "$NATIVELINK_DIR"
 
+# Running "docs builds" as CI to ensure documentation doesn't break after merging
+echo "Running \"docs builds\" to ensure documentation doesn't break"
+docs builds
+
 for pattern in "${TEST_PATTERNS[@]}"; do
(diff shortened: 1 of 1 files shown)


aaronmondal:
The goal of #1082 is to add documentation tests to CI. The way to solve this issue is to take the test infrastructure from the docs directory and run it in our GHA workflows. I'll close this for now as it touches a part of the code that is unrelated to the docs.

PR #1226, opened 2024-08-01
Add Int as StoreKey type

zbirenbaum:
# Description
Adding Int as StoreKey type enables using index based bounds on list calls. This is useful when communicating with some backends such as redis, and can enable further optimizations in the future.

## Type of change

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1226)

### DIFF
--- a/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
+++ b/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
@@ -176,6 +176,12 @@ impl TryFrom<&[u8]> for AwaitedAction {
 #[repr(transparent)]
 pub struct AwaitedActionSortKey(u64);
 
+impl AwaitedActionSortKey {
+    pub fn as_u64(&self) -> u64 {
+        self.0
+    }
+}
+
 impl MetricsComponent for AwaitedActionSortKey {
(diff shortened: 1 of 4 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1226)* status: 0 of 1 LGTMs obtained, and 0 of 4 files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-store/src/shard_store.rs` line 130 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1226#-O3FXKqaCYpd-vmYL0by:-O3FXKqaCYpd-vmYL0bz:bg65zqe) ([raw file](https://github.com/TraceMachina/nativelink/blob/35d8701dc1da2c93b8718fa42a4c14f21477bb38/nativelink-store/src/shard_store.rs#L130)):*
> ```XML
>                 (key_u64 >> 32) as u32 // We only need the top 32 bits.
>             }
>             StoreKey::Int(i) => {
> ```

After the other discussion, I don't think we need to do this. We can use a lexical index for this.

PR #1227, opened 2024-08-01
Remove unnecessary sync trait bounds.

zbirenbaum:
# Description
Removes unnecessary sync bounds from functions inside of the
AwaitedActionDb trait. These bounds interfere with implementations while
attempting to communicate to with the redis store.

towards #1182

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1227)

PR #1228, opened 2024-08-01
awaited action db

zbirenbaum:
# Description
WIP branch for feedback.

includes: #1227 #1226
succeeds: #1182 
## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1228)


adam-singer (changes requested):
> Reviewed 9 of 9 files at r1, 3 of 5 files at r2, 2 of 5 files at r3, 6 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1228)* status: 0 of 1 LGTMs obtained, and 18 of 19 files reviewed, and pending CI: pre-commit-checks, and 13 discussions need to be resolved

___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 48 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3EvBVP8zctzrELCozA:-O3EvBVQ7UhfAO-aV2I-:b-u9jhh0) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L48)):*
> ```XML
> };
> 
> /// Duration to wait before sending client keep alive messages.
> ```

Why is this talking about duration and keep alive messages? 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 74 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3EvibW5Pp0ayFc4Yhx:-O3EvibW5Pp0ayFc4Yhy:bbg2myb) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L74)):*
> ```XML
> }
> 
> impl RedisAwaitedActionDbImpl {
> ```

should leave a comment here about the state of still being in memory until follow up reviews 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 147 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3Evh633rQs2XOT4Cy_:-O3Evh633rQs2XOT4Cya:btfvlc4) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L147)):*
> ```XML
>         }
> 
>         // .map(|sorted_awaited_action| async {
> ```

dead code
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 184 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F0UzL16D5gxSrzBT2:-O3F0UzL16D5gxSrzBT3:b-f515k2) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L184)):*
> ```XML
>             }
>             loop {
>                 let Ok(msg) = handle_next(stream.next().await) else {
> ```

I could be lost here, I see that `handle_next` returns an `Err` object and there is a `else/return` statement here, does the return hoist the returned `Err` from `handle_next` as the returned value? It must if this code compiles

tioli: I do wonder for unwrapping the `Option` there should be a more concise way of representing it, the `match` statement isn't terrible, but usually  when you see `Some(Ok(v)) => { Ok(v) },` in functional programming means a map error or map empty of some sort could of been sequenced differently via some combinators.

```
fn handle_next(next: Option<Result<Message, BroadcastStreamRecvError>>) -> Result<Message, Error> {
    next.ok_or_else(|| make_err!(Code::Internal, "RedisAwaitedActionDb::subscription_listener subscription update stream was closed"))?
        .map_err(|e| make_err!(Code::Internal, "{}", e.to_string()))
}
``` 

or 

```
let closed_stream_msg = "RedisAwaitedActionDb::subscription_listener subscription update stream was closed"
let Ok(msg) = stream.next().await
    .ok_or_else(|| make_err!(Code::Internal, closed_stream_msg))?
    .map_err(|e| make_err!(Code::Internal, "{}", e.to_string())) else {
                    event!(Level::ERROR, closed_stream_msg);
                    return

};
```
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 194 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F2p4v4FAakVF11h_4:-O3F2p4v4FAakVF11h_5:b-xwzulx) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L194)):*
> ```XML
>                     Some(inner_mutex) => {
>                         let mut inner = inner_mutex.lock().await;
>                         let state: AwaitedAction = AwaitedAction::try_from(bytes).unwrap();
> ```

How do these two `unwrap()` calls bbuble out as failures? Do we guard against the conversion or let the program halt?
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 239 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F3L9p1Hrl9eZMWBJl:-O3F3L9p1Hrl9eZMWBJm:bp2nimx) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L239)):*
> ```XML
>         let mut inner = self.inner.lock().await;
>         let Some(tx) = inner.get_operation_sender(operation_id).await else {
>             return Ok(None)
> ```

How abnormal would it be to return `None`, is it something we should log a message for? 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 249 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F6p0h1o1mQ9QAJzFc:-O3F6p0h1o1mQ9QAJzFd:b-s5s4r9) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L249)):*
> ```XML
>     async fn get_operation_id_by_client_id(&self, client_id: &ClientOperationId) -> Result<Option<OperationId>, Error> {
>         let result: Result<OperationId, Error> = {
>             let key = format!("cid:{client_id}");
> ```

tioli: Should be able to return early instead of additional usage of a `Result` enum

```
    let key = format!("cid:{client_id}");
    let bytes = self.store.get_part_unchunked(&key, 0, None).await?;
    let s = match from_utf8(&bytes) {
        Ok(s) => s,
        Err(e) => return Err(make_input_err!("Decoding bytes failed with error: {e}")),
    };
    match OperationId::try_from(s) {
        Ok(operation_id) => Ok(Some(operation_id)),
        Err(e) => match e.code {
            Code::NotFound => Ok(None),
            _ => Err(e),
        },
    }
``` 

There is a very similar logic of this code in `get_operation_id_by_hash_key` probably could be simplified and then make the rest of it generic private function. 

I think it would be ideal to have some information in the error message that helps identify where it orginates from or unique information about `ClientOperationId` or `ActionUniqueQualifier` in the error message.
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 289 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F7z-IBQ5iWXE9mRMb:-O3F7z-IBQ5iWXE9mRMc:b-1zv8d1) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L289)):*
> ```XML
>     }
> 
>     async fn get_awaited_action_by_operation_id(&self, operation_id: &OperationId) -> Result<Option<AwaitedAction>, Error> {
> ```

tioli: Factor out result object

```
    let key = format!("oid:{operation_id}");
    match self.store.get_part_unchunked(&key, 0, None).await {
        Ok(bytes) => match AwaitedAction::try_from(bytes.as_slice()) {
            Ok(action) => Ok(Some(action)),
            Err(e) => Err(e),
        },
        Err(e) => match e.code {
            Code::NotFound => Ok(None),
            _ => Err(e),
        },
    }
```
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 314 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3F9x377Xe98PAms3u-:-O3F9x377Xe98PAms3u0:b-40aoup) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L314)):*
> ```XML
>         client_operation_id: &ClientOperationId,
>     ) -> Result<Option<RedisOperationSubscriber>, Error> {
>         match self.get_operation_id_by_client_id(client_operation_id).await {
> ```

tioli: Might be overly dense but I tend to like combinators, given my newness in rust I assume these do what they would in other languages with combinators

```
        let i: Result<Option<RedisOperationSubscriber>, Error> =
            self.get_operation_id_by_client_id(client_operation_id).await
                .flat_map(|opt| async { 
                    opt.map(|operation_id| self.subscribe_to_operation(&operation_id)) 
                }).collect().await;
        i
```

Usually then I see the "trinary state monad" (just made that up), where its some ADT (`Either` kind `Ok`/`Err`) that contains an `Option` and needs to be transformed on `Ok` but kept the same for non-Ok, usually there is some simplified combinator sequence for it.. usually, sometimes its much clearer for readability if not used, tioli.
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 382 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3FBjun095pR9Wl9tE8:-O3FBjun095pR9Wl9tE9:bgkcw4x) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L382)):*
> ```XML
>                 let operation_id = OperationId::new(action_info.unique_qualifier.clone());
>                 let mut inner = self.inner.lock().await;
>                 // let sub = self.subscribe_to_operation(&operation_id).await;
> ```

nit: dead code 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 405 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3FDW606FoRiOBYdDcc:-O3FDW606FoRiOBYdDcd:b-hefn2g) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L405)):*
> ```XML
>                 self.store.update_oneshot(action_hash_key.as_str(), operation_id.to_string().into()).await?;
> 
>                 // If we are going in or out of Queued State, need to add/remove the sort key from the index
> ```

nit: seems like todo for follow up review? 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 437 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3EzDhRCRCUW9x3d-h0:-O3EzDhRCRCUW9x3d-h1:b-8mg9u6) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L437)):*
> ```XML
>             // but that doesn't exist in the std or any libraries we use.
>             if desc {
>                 for result in iterator.rev() {
> ```

If the returned type is `DoubleEndedIterator` in RedisAwaitedActionDb::get_range_of_actions() do we still need to call `rev()`? Wouldn't that mean we are reversing an iterator that is already feeding from the end or is there some magic with the interfaces where being the underlying a `DoubleEndedIterator` the call to `rev()` is no-op? 
___
*[`nativelink-scheduler/src/redis_awaited_action_db.rs` line 489 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1228#-O3FDddJ1DVnrGC5Qxz7:-O3FDddJ1DVnrGC5Qxz8:bj8jm7) ([raw file](https://github.com/TraceMachina/nativelink/blob/4dbf382fb84d5a568af3f04e52aadcd3547ac063/nativelink-scheduler/src/redis_awaited_action_db.rs#L489)):*
> ```XML
> 
> //
> // impl RedisActionStream {
> ```

nit: dead code

zbirenbaum:
Succeeded by #1249

PR #1214, opened 2024-07-29
Remove ClientOperationId and move all to OperationId

allada:
This is in prep to support generic layering of workers & schedulers. We can no longer know if the source is a client or just an up-stream-component, so we need to fall back to just using OperationId.

towards #1213

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1214)

PR #1245, opened 2024-08-07
Update variables to reflect OperationId changes

zbirenbaum:
# Description
Minor cosmetic fix to reflect changes in #1214

## Type of change

- [x] Code QA

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1245)

### DIFF
--- a/nativelink-scheduler/src/awaited_action_db/mod.rs
+++ b/nativelink-scheduler/src/awaited_action_db/mod.rs
@@ -77,10 +77,10 @@ pub trait AwaitedActionSubscriber: Send + Sync + Sized + 'static {
 pub trait AwaitedActionDb: Send + Sync + MetricsComponent + 'static {
     type Subscriber: AwaitedActionSubscriber;
 
-    /// Get the AwaitedAction by the client operation id.
+    /// Get the AwaitedAction by the operation id.
     fn get_awaited_action_by_id(
         &self,
-        client_operation_id: &OperationId,
+        operation_id: &OperationId,
     ) -> impl Future<Output = Result<Option<Self::Subscriber>, Error>> + Send;
 
     /// Get all AwaitedActions. This call should be avoided as much as possible.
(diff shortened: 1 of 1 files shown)


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1245)* status: 0 of 1 LGTMs obtained, and 0 of 1 files reviewed, and 3 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/1245#-O3hSCpu7di3YsLOyBdS:-O3hSCpu7di3YsLOyBdT:b91igex) (no related file):*
We need to keep the logic of `ClientOperationId`, but the api surface can only know about `OperationId`, since this module is the only module that deals with both sides of the client<->worker relationship it needs to keep the logic separate.

___
___
*[`nativelink-scheduler/src/awaited_action_db/mod.rs` line 83 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1245#-O3hRqd5Cy08_BStiWCb:-O3hRqd5Cy08_BStiWCc:b-drshe6) ([raw file](https://github.com/TraceMachina/nativelink/blob/33f09cbd1b2833956ffb268f786a7c035f375dae/nativelink-scheduler/src/awaited_action_db/mod.rs#L83)):*
> ```XML
>     fn get_awaited_action_by_id(
>         &self,
>         client_operation_id: &OperationId,
> ```

This is for the client_operation_id, not the worker_operation_id.
___
*[`nativelink-scheduler/src/awaited_action_db/mod.rs` line 116 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1245#-O3hS38rEJMEylpjPMMT:-O3hS38rEJMEylpjPMMU:b-k3admz) ([raw file](https://github.com/TraceMachina/nativelink/blob/33f09cbd1b2833956ffb268f786a7c035f375dae/nativelink-scheduler/src/awaited_action_db/mod.rs#L116)):*
> ```XML
>     fn add_action(
>         &self,
>         client_operation_id: OperationId,
> ```

This is for the client side of the operation id.

Issue #1247, opened 2024-08-08
[Bug] Please provide a `sleep_impl` on the config, or disable timeouts.


adam-singer:
[ghcr.io/tracemachina/nativelink:nqy6kcs06frck1hbgxy02gqp6vsm9dxf](http://ghcr.io/tracemachina/nativelink:nqy6kcs06frck1hbgxy02gqp6vsm9dxf) has been reported to throw the following exceptions

```
Defaulted container "cas" out of: cas, filter
{"timestamp":"2024-08-07T21:50:11.885249Z","level":"DEBUG","fields":{"message":"loading default partitions"},"target":"aws_sdk_sts::endpoint_lib"}
thread 'main' panicked at /REDACTED/aws-config-1.5.4/src/web_identity_token.rs:239:25:
Invalid client configuration: An async sleep implementation is required for retry to work. Please provide a `sleep_impl` on the config, or disable timeouts.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The reason for this failure is `aws-smithy-runtime` was no longer receiving the feature flag `rt-tokio`

PR #1248, opened 2024-08-08
[Bug] Add rt-tokio feature to aws-sdk-s3

adam-singer:
# Description

Feature `rt-tokio` had been transitively dependended on, durning refactors of `Cargo.toml` files it was accidentally dropped.

Fixes https://github.com/TraceMachina/nativelink/issues/1247

## Type of change

## How Has This Been Tested?

Will be deployed to staging systems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1248)

### DIFF
--- a/nativelink-store/Cargo.toml
+++ b/nativelink-store/Cargo.toml
@@ -14,7 +14,9 @@ async-trait = "0.1.80"
 aws-config = { version = "1.5.4", default-features = false, features = [
   "rustls",
 ] }
-aws-sdk-s3 = { version = "1.41.0", default-features = false }
+aws-sdk-s3 = { version = "1.41.0", features = [
+  "rt-tokio",
+], default-features = false }
 aws-smithy-runtime = { version = "1.6.2" }
 bincode = "1.3.3"
 blake3 = { version = "1.5.2", default-features = false }


MarcusSorealheis:
Thank you for fixing this one. It goes without saying that this will be in our point release.

aaronmondal (approved):
It would be nice to have some sort of regression test for this (or at least an issue that tracks implementation of a regression test).

I'll create a point release after this is merged.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1248#-:-O3mc3ljD8IyjfHpKzQH:b9c4gjb)* status: 1 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-store/Cargo.toml` line 17 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1248#-O3mbpBRC-dyGDw78l0T:-O3mbpBRC-dyGDw78l0U:b-ibtetx) ([raw file](https://github.com/TraceMachina/nativelink/blob/511e2a549dc384a3e712ffd1bc5ede76d0a9e940/nativelink-store/Cargo.toml#L17)):*
> ```TOML
>   "rustls",
> ] }
> aws-sdk-s3 = { version = "1.41.0", features = [
> ```

nit: Is there a way we could add a regression test or something? Maybe just a feature-check in the source for the S3 store so that builds fail early if this accidentally removed again in the future?

allada (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1248#-:-O3mfLai3--v_zgzbP5x:bnfp4nl)* status: 2 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-store/Cargo.toml` line 17 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1248#-O3mbpBRC-dyGDw78l0T:-O3meodt8prpCQ4mcxAQ:boyy4q6) ([raw file](https://github.com/TraceMachina/nativelink/blob/511e2a549dc384a3e712ffd1bc5ede76d0a9e940/nativelink-store/Cargo.toml#L17)):*

nit: Is there a way we could add a regression test or something? Maybe just a feature-check in the source for the S3 store so that builds fail early if this accidentally removed again in the future?
</blockquote></details>

This should have been caught in CI, the reason it didnt was because `aws-sdk-s3` has to be defined in `dev-dependencies` (below) as well as here, but only got enabled below.

Writing a regression test for this is not going to be easy, because it will require a full integration test.

adam-singer (commented):
The test could of been something that asserts what the defaults provided from the library, as @allada it was a feature enabled in dev deps so we would still have missed it. We could.. put an assert in the main code upon init to halt if the credential provider util isn't returning expected sleep functions, don't think its the right place, integration tests would be better. 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1248#-:-O3ml1UF13l0fd1l-ZEK:b3vxb6n)* status: 2 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

PR #1250, opened 2024-08-08
Attach NativeLink version and commit hash

### LINKED ISSUE #1044: All client side errors should have the NativeLink version attached to it
To make it easier to understand what version people are using when there's a problem, we should attach a hash that represents what version or commit of NL they are using.

cc: @aaronmondal for how we might be able to stamp this by default but not invalidate cache?

aleksdmladenovic:
# Description

NativeLink version and commit hash are attached as runtime environment variables for Cargo and Bazel. For Cargo, `build.rs` file is used for embedding version and commit hash as environment varaibles. For Bazel, `.bazelrc` file is used for the same purporse. Wrote `update_bazelrc.sh` script for adding environment variables to `.bazelrc` file.

Bazel didn't support `build.rs` method and it needs to be configured in `.bazelrc` file.
I tried to make `.bazelrc` file to load the version information and git commit hash inside it but it didn't support retrieving shell script results. So shell script file came into solution for that.

I'm not 100% sure if my solution is correct but be ready to tolerate the current solution.

Fixes #1044 

## Type of change

## Checklist

- [ ] Tests added/amended


aleksdmladenovic:
+@MarcusSorealheis, @allada, @aaronmondal

aaronmondal (changes requested):
Thanks a lot for working on this. I'm very sorry but I don't think we can merge this PR or this feature in general.

There are IMO too many potential problems that come from tracking a version hash in the executable. Users can already derive the commit hash of a container (it's written into the container manifest of all images). Adding it to the executable would require elaborate binary signing and verification setups. It would also break our CI efficiency.

This is not true for the "regular" version like `0.5.0` though. This is already available via `nativelink --version` so there should be an easy way to get that version without any external environment variables. This version is much slower moving and would be safe to print.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1250#-:-O3mTY9Y-i3u7ZwTkP4C:b-mvsr6y)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, and 2 discussions need to be resolved

___
*[`.bazelrc` line 88 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1250#-O3mTHO0D_82UX0Es2hU:-O3mTHO0D_82UX0Es2hV:b-vsxphz) ([raw file](https://github.com/TraceMachina/nativelink/blob/293651c274cc70f965940a720455c2d7a73d20d1/.bazelrc#L88)):*
> ```
> 
> # Generated by update_bazelrc.sh
> build --action_env=NATIVELINK_APP_VERSION=0.5.0
> ```

Since `nativelink --version` already prints this information should already be present in the executable.
___
*[`.bazelrc` line 89 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1250#-O3mOnizBZ9sw7bTGs7c:-O3mOnizBZ9sw7bTGs7d:bgfp94l) ([raw file](https://github.com/TraceMachina/nativelink/blob/293651c274cc70f965940a720455c2d7a73d20d1/.bazelrc#L89)):*
> ```
> # Generated by update_bazelrc.sh
> build --action_env=NATIVELINK_APP_VERSION=0.5.0
> build --action_env=NATIVELINK_GIT_COMMIT_HASH=c3f648ecaad4861983bce1a5dc67781685bd1e80
> ```

This approach doesn't properly track the commit. If you look at the hash of this commit it'll be different from the one here. If you update it and amend the change the git commit hash will change again.

So essentially it's not possible to textually pin this hash against the "self" commit.

Tracking the commit hash also breaks caching as every build will have a different hash. Even on README changes we'd have to rebuild the executable. At the moment we get around this by only tracking the commit hash in the published container images which can reuse nativelink builds from previous runs.

MarcusSorealheis:
We can re-open after we have time to evaluate.

PR #1254, opened 2024-08-08
Add 90s Explainer to README.md

harper-carroll:
Just added YouTube to ReadMe

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1254)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ NativeLink is an efficient, high-performance build cache and remote execution sy
 
 NativeLink is trusted in production environments to reduce costs and developer iteration times--handling over **one billion requests** per month for its customers, including large corporations such as **Samsung**.
 
+[![NativeLink Explained in 90 seconds](https://img.youtube.com/vi/WLpqFuyLMUQ/0.jpg)](https://www.youtube.com/watch?v=WLpqFuyLMUQ)
+
 ## 🔑 Key Features
 
 1. **Advanced Build Cache**:
@@ -61,7 +63,7 @@ curl -O \
 docker run \
     -v $(pwd)/basic_cas.json:/config \
     -p 50051 \
-    ghcr.io/tracemachina/nativelink:v0.5.1 \
+    ghcr.io/tracemachina/nativelink:v0.5.0 \
     config
 ```
 
@@ -78,7 +80,7 @@ Invoke-WebRequest `
 docker run `
     -v ${PWD}/basic_cas.json:/config `
     -p 50051 `
-    ghcr.io/tracemachina/nativelink:v0.5.1 `
+    ghcr.io/tracemachina/nativelink:v0.5.0 `
     config
 ```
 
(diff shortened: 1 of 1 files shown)


aaronmondal:
@blakehatch We'll need to check this against the Cloudflare deployment to ensure that it doesn't break this page (where this change will be visible as well):
 
- https://docs.nativelink.com/introduction/setup/

We'll need to double check that we don't accidentally reduce the performance and accessibility scores if for instance the image isn't lazily loaded. I think it should be fine, but let's make sure. There is a chance that we need to add some html tags around it or something.

blakehatch:
It's just an image of the thumbnail with a link to the video, shouldn't impact performance any more that the nativelink logo for instance @aaronmondal

MarcusSorealheis:
Question about breaking is different from performance but I agree it seems like a likely non-issue.

MarcusSorealheis:
I’m on my phone taxing. Is there an extra commit in here? That’s what it looks like but maybe there is an error in the existing README that this fixed. In that case we need to `git commit amend`. Should not take very long.

harper-carroll:
yeah there's 2 commits because the first made a weird change reverting 5.1 -> 5.0

aaronmondal (changes requested):
@blakehatch @MarcusSorealheis @krishmoran Please benchmark instead of guessing.

- This change adds 6 legacy-style external cookies from youtube which brings our perfect best-practice score from 100 to 75. Surely there is a better way to add this.
- The image is jpg when it should be webp. This makes the image almost 2x heavier than it should be which reduces performance and worsens user experience on slow internet connections.
- On desktop views the image shouldn't be lazily loaded to avoid large content shifts. On mobile it's ok to lazy load since it's not in the initial viewport.

If it's necessary and because it might be a nontrivial change we can gloss over the last point. The first two should be addressed before we merge though.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1254#-:-O3rxCwE49CEnmpB1jCi:b-d7r0bg)* status: 0 of 2 LGTMs obtained, and all files reviewed

MarcusSorealheis:
@harper-carroll and @aaronmondal I've pushed up a change that should fix this one. Let me know if this works for you. I think @blakehatch said he was working on a lighthouse test but I think now it should address one of the two issues.

MarcusSorealheis:
image

aaronmondal (approved):
Very nice! I can confirm that we're now back to green.

:lgtm: 

> Reviewed 1 of 1 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1254#-:-O46fT9XD7NAqb0Cq_a8:b-kahlpu)* status: 1 of 2 LGTMs obtained, and all files reviewed

aaronmondal (approved):
@blakehatch Needs your approval.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1254#-:-O46gLyr7m33QrxYbkGt:b-felr7t)* status: 1 of 2 LGTMs obtained, and all files reviewed

PR #1258, opened 2024-08-09
Declare version explicitly for `clap` library in `nativelink.rs`

### LINKED ISSUE #1257: NativeLink version is not displayed correctly when it's built with Bazel
Currently, when we build the application with `Bazel`, `--version` command doesn't work.
It works when the application is built with `Cargo` though.

error

The reason is because of this.

`clap` library gets the version information from the `CARGO_PKG_VERSION` environment variable.
https://docs.rs/clap_builder/4.5.14/src/clap_builder/macros.rs.html#16

Unfortunately, that variable is only set when the application is built with `Cargo`.
I've tried to pass that `CARGO_PKG_VERSION` environment variable explicitly via `.bazelrc` file but it didn't work.

The easiest way to solve this issue is to declare the version explicitly here in `nativelink.rs`.

https://github.com/TraceMachina/nativelink/blob/7777938294047377cb4ce9f4d8649c45055596ed/src/bin/nativelink.rs#L99

aleksdmladenovic:
# Description

Currently, the `--version` command doesn't work when the application is built with `Bazel`. The reason is because `Bazel` doesn't set the `CARGO_PKG_VERSION` environment variable natively. Solution is to declare `version` explicitly in `nativelink.rs`.
More details can be found in the issue below.

Fixes #1257 

## Type of change

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1258)

### DIFF
--- a/nativelink-scheduler/tests/utils/mock_scheduler.rs
+++ b/nativelink-scheduler/tests/utils/mock_scheduler.rs
@@ -19,11 +19,9 @@ use nativelink_error::{make_input_err, Error};
 use nativelink_metric::{MetricsComponent, RootMetricsComponent};
 use nativelink_scheduler::action_scheduler::ActionScheduler;
 use nativelink_scheduler::platform_property_manager::PlatformPropertyManager;
-use nativelink_util::{
-    action_messages::{ActionInfo, OperationId},
-    operation_state_manager::{
-        ActionStateResult, ActionStateResultStream, ClientStateManager, OperationFilter,
-    },
+use nativelink_util::action_messages::{ActionInfo, OperationId};
+use nativelink_util::operation_state_manager::{
+    ActionStateResult, ActionStateResultStream, ClientStateManager, OperationFilter,
 };
 use tokio::sync::{mpsc, Mutex};
 
--- a/src/bin/nativelink.rs
+++ b/src/bin/nativelink.rs
@@ -96,7 +96,7 @@ const METRICS_DISABLE_ENV: &str = "NATIVELINK_DISABLE_METRICS";
 #[derive(Parser, Debug)]
 #[clap(
     author = "Trace Machina, Inc. <nativelink@tracemachina.com>",
-    version,
+    version = "0.5.1",
     about,
     long_about = None
 )]


allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1258)* status: 0 of 2 LGTMs obtained, and 0 of 2 files reviewed, and 1 discussions need to be resolved

___
*[`src/bin/nativelink.rs` line 99 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1258#-O3s4odo7c6Kn16EHB-q:-O3s4odo7c6Kn16EHB-r:b8r4fro) ([raw file](https://github.com/TraceMachina/nativelink/blob/b21ff3f143d232af63624adbbec1ecab37326c31/src/bin/nativelink.rs#L99)):*
> ```XML
> #[clap(
>     author = "Trace Machina, Inc. <nativelink@tracemachina.com>",
>     version = "0.5.1",
> ```

@aaronmondal , doesn't this get pulled in when built with nix/cargo on the official releases?

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1258)* status: 0 of 2 LGTMs obtained, and 0 of 2 files reviewed, and 1 discussions need to be resolved

___
*[`src/bin/nativelink.rs` line 99 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1258#-O3s4odo7c6Kn16EHB-q:-O3t8C_C3-jE8pVAgHUb:b-feumzw) ([raw file](https://github.com/TraceMachina/nativelink/blob/b21ff3f143d232af63624adbbec1ecab37326c31/src/bin/nativelink.rs#L99)):*

@aaronmondal , doesn't this get pulled in when built with nix/cargo on the official releases?
</blockquote></details>

Yes. IMO it might actually be beneficial if the Bazel executable doesn't explicitly set a version. This way it's clearer that the Bazel build is not intended for production and clearly differentiable during testing. IIRC llvm does a similar thing  where the bazel build doesn't have a version attached.

I believe the reason it doesn't work in the .bazelrc is because we'd have to set this environment variable via an external mechanism first. This would probably add a lot of complexity. If we hardcode the value here it's one more value that we'd have to keep track of when we cut releases. I'm not sure whether that's worth it.

@allada Maybe we should close #1257 as "not planned"? From my end I'd say the current behavior is "intended behavior".

@aleksdmladenovic I'm aware that this might be a bit frustrating, but these version/hash/pinning things can lead to very tricky and subtle issues down the line, so I wanna be extra cautious with changes around this.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1258)* status: 0 of 2 LGTMs obtained, and 1 of 2 files reviewed, and 1 discussions need to be resolved

___
*[`src/bin/nativelink.rs` line 99 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1258#-O3s4odo7c6Kn16EHB-q:-O42t-6N-EjrCCISPfZd:b-ejnxni) ([raw file](https://github.com/TraceMachina/nativelink/blob/b21ff3f143d232af63624adbbec1ecab37326c31/src/bin/nativelink.rs#L99)):*

Yes. IMO it might actually be beneficial if the Bazel executable doesn't explicitly set a version. This way it's clearer that the Bazel build is not intended for production and clearly differentiable during testing. IIRC llvm does a similar thing  where the bazel build doesn't have a version attached.

I believe the reason it doesn't work in the .bazelrc is because we'd have to set this environment variable via an external mechanism first. This would probably add a lot of complexity. If we hardcode the value here it's one more value that we'd have to keep track of when we cut releases. I'm not sure whether that's worth it.

@allada Maybe we should close #1257 as "not planned"? From my end I'd say the current behavior is "intended behavior".

@aleksdmladenovic I'm aware that this might be a bit frustrating, but these version/hash/pinning things can lead to very tricky and subtle issues down the line, so I wanna be extra cautious with changes around this.
</blockquote></details>

Yeah, I'd say close as not planned.

PR #1267, opened 2024-08-18
Fix #1266

### LINKED ISSUE #1266: Show warning when redis store fails to connect
Currently it seems that nativelink will fail silently if it fails to connect to a given redis uri for a `RedisStore`. Instead, an error should be displayed when an attempt to upload or download data from the store does not execute.

vkobinski:
# Description

I believe this is part of the solution to the issue. The RedisStore StoreDriver implementation needs to check if the client_pool is connected. So I changed only the has_with_results, and if what I've is correct, I will update the rest of the code.

Fixes #1266

## Type of change

- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ X ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1267)

### DIFF
--- a/nativelink-store/src/default_store_factory.rs
+++ b/nativelink-store/src/default_store_factory.rs
@@ -53,7 +53,7 @@ pub fn store_factory<'a>(
             StoreConfig::experimental_s3_store(config) => {
                 S3Store::new(config, SystemTime::now).await?
             }
-            StoreConfig::redis_store(config) => RedisStore::new(config)?,
+            StoreConfig::redis_store(config) => RedisStore::new(config).await?,
             StoreConfig::verify(config) => VerifyStore::new(
                 config,
                 store_factory(&config.backend, store_manager, None).await?,
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -62,7 +62,7 @@ pub struct RedisStore {
 
 impl RedisStore {
     /// Create a new `RedisStore` from the given configuration.
-    pub fn new(config: &nativelink_config::stores::RedisStore) -> Result<Arc<Self>, Error> {
+    pub async fn new(config: &nativelink_config::stores::RedisStore) -> Result<Arc<Self>, Error> {
         if config.addresses.is_empty() {
             return Err(make_err!(
                 Code::InvalidArgument,
@@ -104,11 +104,12 @@ impl RedisStore {
             || Uuid::new_v4().to_string(),
             config.key_prefix.clone(),
         )
+        .await
         .map(Arc::new)
     }
 
     /// Used for testing when determinism is required.
-    pub fn new_from_builder_and_parts(
+    pub async fn new_from_builder_and_parts(
         builder: Builder,
         pub_sub_channel: Option<String>,
         temp_name_generator_fn: fn() -> String,
@@ -122,7 +123,7 @@ impl RedisStore {
             .build_subscriber_client()
             .err_tip(|| "while creating redis subscriber client")?;
         // Fires off a background task using `tokio::spawn`.
-        client_pool.connect();
+        client_pool.init().await?;
         subscriber_client.connect();
 
         Ok(Self {
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -179,6 +179,13 @@ impl StoreDriver for RedisStore {
         // TODO(caass): Optimize for the case where `keys.len() == 1`
         let pipeline = self.client_pool.next().pipeline();
 
+        if !pipeline.is_connected() {
+            return Err(Error::new(
+                Code::Unavailable,
+                "Could not connect to Redis Client".to_string(),
+            ));
+        }
+
         results.iter_mut().for_each(|result| *result = None);
 
         for (idx, key) in keys.iter().enumerate() {
--- a/nativelink-store/tests/redis_store_test.rs
+++ b/nativelink-store/tests/redis_store_test.rs
@@ -28,7 +28,7 @@ use nativelink_store::cas_utils::ZERO_BYTE_DIGESTS;
 use nativelink_store::redis_store::{RedisStore, READ_CHUNK_SIZE};
 use nativelink_util::buf_channel::make_buf_channel_pair;
 use nativelink_util::common::DigestInfo;
-use nativelink_util::store_trait::{StoreLike, UploadSizeInfo};
+use nativelink_util::store_trait::{StoreKey, StoreLike, UploadSizeInfo};
 use pretty_assertions::assert_eq;
 
 const VALID_HASH1: &str = "3031323334353637383961626364656630303030303030303030303030303030";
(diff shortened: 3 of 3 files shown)


vkobinski:
Looking for help to see if what i did is part of the solution before changing the rest of RedisStore.

adam-singer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1267)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved

___
*[`nativelink-store/src/redis_store.rs` line 182 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1267#-O4fQ-xk0iF-B6ceB55J:-O4fQ-xk0iF-B6ceB55K:b-pytle4) ([raw file](https://github.com/TraceMachina/nativelink/blob/2ce30b3a7946c9b6e56b980dd4b585cb1f11f541/nativelink-store/src/redis_store.rs#L182)):*
> ```XML
>         let pipeline = self.client_pool.next().pipeline();
> 
>         if !pipeline.is_connected() {
> ```

@caass at this stage if we are not connected who was responsible for setting up connection and reconnect/retry? Concerned throwing an error here might be more disruptive even if it is very clear we might not be able to proceed the pipeline by the time it tries to get executed. Maybe the running of pipelines should be handled outside of this code to directly control/reconnect logic outside of the store code.

vkobinski:
> > Reviewed 2 of 2 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1267)_ status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved
> 
> _[`nativelink-store/src/redis_store.rs` line 182 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1267#-O4fQ-xk0iF-B6ceB55J:-O4fQ-xk0iF-B6ceB55K:b-pytle4) ([raw file](https://github.com/TraceMachina/nativelink/blob/2ce30b3a7946c9b6e56b980dd4b585cb1f11f541/nativelink-store/src/redis_store.rs#L182)):_
> 
> > ```
> >         let pipeline = self.client_pool.next().pipeline();
> > 
> >         if !pipeline.is_connected() {
> > ```
> 
> @caass at this stage if we are not connected who was responsible for setting up connection and reconnect/retry? Concerned throwing an error here might be more disruptive even if it is very clear we might not be able to proceed the pipeline by the time it tries to get executed. Maybe the running of pipelines should be handled outside of this code to directly control/reconnect logic outside of the store code.

In this part of the fred.rs repository the code tries to connect each client of the RedisPool, but never test if the connection is actually successfull.
https://github.com/aembke/fred.rs/blob/8f011ddaf382c9570704155e93b396c3afb2e931/src/clients/pool.rs#L223-L233

caass (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1267)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved

___
*[`nativelink-store/src/redis_store.rs` line 182 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1267#-O4fQ-xk0iF-B6ceB55J:-O4rMuQv12IxgEPs1AYx:b9lcod0) ([raw file](https://github.com/TraceMachina/nativelink/blob/2ce30b3a7946c9b6e56b980dd4b585cb1f11f541/nativelink-store/src/redis_store.rs#L182)):*
<details><summary><i>Previously, vkobinski (Victor Kobinski) wrote…</i></summary><blockquote>

In this part of the fred.rs repository the code tries to connect each client of the RedisPool, but never test if the connection is actually successfull.
https://github.com/aembke/fred.rs/blob/8f011ddaf382c9570704155e93b396c3afb2e931/src/clients/pool.rs#L223-L233

</blockquote></details>

So, the way the connection works right now is that it just spawns some tasks off into the void that try to connect. When those (potentially) fail, those failures are unhandled. That logic happens above, in the initializer. See [`ClientLike::connect`](https://docs.rs/fred/latest/fred/interfaces/trait.ClientLike.html#method.connect).

Checking here with `pipeline.is_connected()` is definitely one way to verify the connection, and I wonder if a better way to do it would be by using [`ClientLike::init`](https://docs.rs/fred/latest/fred/interfaces/trait.ClientLike.html#method.init), which returns a task that completes when the connection either succeeds or fails.

To put it another way: we are currently using `ClientLike::connect`, which is semantically equivalent to

```rust
tokio::spawn(async move {
    let connection = client.try_connect().await?;
    connection.wait_for_termination().await
```

which means the task won't complete until the connection is dropped.

Using `init` instead would give us something more like

```rust
tokio::spawn(async move {
    client.try_connecect().await
```
Which would allow us to have more fine-grained control over connections.

In particular, there's a method [`RedisPool::connect_pool`](https://docs.rs/fred/latest/fred/clients/struct.RedisPool.html#method.connect_pool) that looks like it'd be useful for this purpose.

In short: I really like the direction you're going with this PR! I think the best way forward would actually be to add a `connection_status` field to the `RedisStore` that contains the result of calling `RedisPool::connect_pool`. And then when we go to grab a connection from the pool, we'd first check to make sure all our connections were healthy.

If that sort of fix seems to difficult given the tools we have in `fred`, then we can use this as a stopgap measure :)
___
*[`nativelink-store/tests/redis_store_test.rs` line 335 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1267#-O4rN7qfC1BIVBPcTk52:-O4rN7qfC1BIVBPcTk53:b5hq8n8) ([raw file](https://github.com/TraceMachina/nativelink/blob/2ce30b3a7946c9b6e56b980dd4b585cb1f11f541/nativelink-store/tests/redis_store_test.rs#L335)):*
> ```XML
> 
> #[nativelink_test]
> async fn unsucessfull_redis_connection() -> Result<(), Error> {
> ```

Thank you for adding a test!

vkobinski:
>Using init instead would give us something more like
>```
>tokio::spawn(async move {
>    client.try_connecect().await
>```

Hi, just for using your suggestion i have made some progress. In the main repo the connection is "tested" via: 

```
client_pool.connect()?;
```

By using this:

```
client_pool.init().await?;
```

Now the tests are failing because there is no Redis server running on my PC, but if i run the redis-server, the tests works.
The tests output this error if the redis-server is off:

```
running 7 tests
test upload_and_get_data ... ok
test upload_and_get_data_with_prefix ... ok
test unsucessfull_redis_connection ... FAILED
test upload_empty_data ... FAILED
test test_large_downloads_are_chunked ... ok
test yield_between_sending_packets_in_update ... ok
test upload_empty_data_with_prefix ... FAILED

failures:

---- unsucessfull_redis_connection stdout ----
Error: Error { code: Internal, messages: ["IO Error: Os { code: 111, kind: ConnectionRefused, message: \"Connection refused\" }"] }

---- upload_empty_data stdout ----
Error: Error { code: Internal, messages: ["IO Error: Os { code: 111, kind: ConnectionRefused, message: \"Connection refused\" }"] }

---- upload_empty_data_with_prefix stdout ----
Error: Error { code: Internal, messages: ["IO Error: Os { code: 111, kind: ConnectionRefused, message: \"Connection refused\" }"] }
```

vkobinski:
@caass what do you think the RedisStore should do if the connections do not succeed? Just output something or actually fail the operation?

aaronmondal:
Fixed by https://github.com/TraceMachina/nativelink/commit/a815ba0cb781a2ddc5d2afd4c97ef676326311c0

PR #1269, opened 2024-08-21
Fix potential infinite loop in `RedisStore::update`

caass:
# Description

There was a logic bug in our original implementation of the redis store that persisted across the rewrite to fred -- there could be an infinite loop in `RedisStore::update`. This PR fixes that.

See: https://reviewable.io/reviews/TraceMachina/nativelink/1188#-O2pu9LV5ux4ILuT6MND

## Type of change

  not work as expected)

## How Has This Been Tested?

The test suite has been validated to pass and behave as expected.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 bazel-*
 target/
 .vscode/
+.zed
 .cache
 .terraform*
 .config
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -298,39 +298,79 @@ impl StoreDriver for RedisStore {
         // As a result of this, there will be a span of time where a key in Redis has only partial data. We want other
         // observers to notice atomic updates to keys, rather than partial updates, so we first write to a temporary key
         // and then rename that key once we're done appending data.
-        //
-        // TODO(caass): Remove potential for infinite loop (https://reviewable.io/reviews/TraceMachina/nativelink/1188#-O2pu9LV5ux4ILuT6MND)
-        'outer: loop {
-            let mut expecting_first_chunk = true;
-            let pipe = client.pipeline();
+        let mut is_first_chunk = true;
+        let mut eof_reached = false;
+        let mut pipe = client.pipeline();
 
-            while expecting_first_chunk || !reader.is_empty() {
-                let chunk = reader
-                    .recv()
-                    .await
-                    .err_tip(|| "Failed to reach chunk in update in redis store")?;
+        while !eof_reached {
+            pipe = client.pipeline();
+            let mut pipe_size = 0;
+            const MAX_PIPE_SIZE: usize = 5 * 1024 * 1024; // 5 MB
 
-                if chunk.is_empty() {
-                    if is_zero_digest(key.borrow()) {
-                        return Ok(());
-                    }
+            let chunk = reader
+                .recv()
+                .await
+                .err_tip(|| "Failed to reach chunk in update in redis store")?;
 
-                    // Reader sent empty chunk, we're done here.
-                    break 'outer;
-                }
+            if chunk.is_empty() {
+                // There are three cases where we receive an empty chunk:
+                // 1. The first chunk of a zero-digest key. We're required to treat all zero-digest keys as if they exist
+                //    and are empty per the RBE spec, so we can just return early as if we've pushed it -- any attempts to
+                //    read the value later will similarly avoid the network trip.
+                // 2. This is an empty first chunk of a non-zero-digest key. In this case, we _do_ need to push up an
+                //    empty key, but can skip the rest of the process around renaming since there's only the one operation.
+                // 3. This is the last chunk (EOF) of a regular key. In that case we can skip pushing this chunk.
+                //
+                // In all three cases, we're done pushing data and can move it from the temporary key to the final key.
+                if is_first_chunk && is_zero_digest(key.borrow()) {
+                    // Case 1, a zero-digest key.
+                    return Ok(());
+                } else if is_first_chunk {
+                    // Case 2, an empty non-zero-digest key.
+                    pipe.append::<(), _, _>(&temp_key, "")
+                        .await
+                        .err_tip(|| "While appending to temp key in RedisStore::update")?;
+                };
 
-                // Queue the append, but don't execute until we've received all the chunks.
-                pipe.append::<(), _, _>(&temp_key, chunk)
-                    .await
-                    .err_tip(|| "Failed to append to temp key in RedisStore::update")?;
-                expecting_first_chunk = false;
-
-                // Give other tasks a chance to run to populate the reader's
-                // buffer if possible.
-                tokio::task::yield_now().await;
+                // Note: setting `eof_reached = true` and calling `continue` is semantically equivalent to `break`.
+                // Since we need to use the `eof_reached` flag in the inner loop, we do the same here
+                // for consistency.
+                eof_reached = true;
+                continue;
+            } else {
+                // Not EOF, but we've now received our first chunk.
+                is_first_chunk = false;
             }
 
-            // Here the reader is empty but more data is expected.
+            // Queue the append, but don't execute until we've received all the chunks.
+            pipe_size += chunk.len();
+            pipe.append::<(), _, _>(&temp_key, chunk)
+                .await
+                .err_tip(|| "Failed to append to temp key in RedisStore::update")?;
+
+            // Opportunistically grab any other chunks already in the reader.
+            while let Some(chunk) = reader
+                .try_recv()
+                .transpose()
+                .err_tip(|| "Failed to reach chunk in update in redis store")?
+            {
+                if chunk.is_empty() {
+                    eof_reached = true;
+                    break;
+                } else {
+                    pipe_size += chunk.len();
+                    pipe.append::<(), _, _>(&temp_key, chunk)
+                        .await
+                        .err_tip(|| "Failed to append to temp key in RedisStore::update")?;
+                }
+
+                // Stop appending if the pipeline is already holding 5MB of data.
+                if pipe_size >= MAX_PIPE_SIZE {
+                    break;
+                }
+            }
+
+            // We've exhausted the reader (or hit the 5MB cap), but more data is expected.
             // Executing the queued commands appends the data we just received to the temp key.
             pipe.all::<()>()
                 .await
@@ -338,8 +378,10 @@ impl StoreDriver for RedisStore {
         }
 
         // Rename the temp key so that the data appears under the real key. Any data already present in the real key is lost.
-        client
-            .rename::<(), _, _>(&temp_key, final_key.as_ref())
+        pipe.rename::<(), _, _>(&temp_key, final_key.as_ref())
+            .await
+            .err_tip(|| "While queueing key rename in RedisStore::update()")?;
+        pipe.all::<()>()
             .await
             .err_tip(|| "While renaming key in RedisStore::update()")?;
 
--- a/nativelink-store/tests/redis_store_test.rs
+++ b/nativelink-store/tests/redis_store_test.rs
@@ -13,10 +13,11 @@
 // limitations under the License.
 
 use std::collections::VecDeque;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread::panicking;
 
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use fred::bytes_utils::string::Str;
 use fred::error::RedisError;
 use fred::mocks::{MockCommand, Mocks};
@@ -30,6 +31,7 @@ use nativelink_util::buf_channel::make_buf_channel_pair;
 use nativelink_util::common::DigestInfo;
 use nativelink_util::store_trait::{StoreLike, UploadSizeInfo};
 use pretty_assertions::assert_eq;
+use tokio::sync::watch;
 
 const VALID_HASH1: &str = "3031323334353637383961626364656630303030303030303030303030303030";
 const TEMP_UUID: &str = "550e8400-e29b-41d4-a716-446655440000";
@@ -42,32 +44,72 @@ fn make_temp_key(final_name: &str) -> String {
     format!("temp-{TEMP_UUID}-{{{final_name}}}")
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 struct MockRedisBackend {
     /// Commands we expect to encounter, and results we to return to the client.
     // Commands are pushed from the back and popped from the front.
     expected: Mutex<VecDeque<(MockCommand, Result<RedisValue, RedisError>)>>,
+
+    tx: watch::Sender<MockCommand>,
+    rx: watch::Receiver<MockCommand>,
+
+    failing: AtomicBool,
+}
+
+impl Default for MockRedisBackend {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 
 impl MockRedisBackend {
     fn new() -> Self {
-        Self::default()
+        let (tx, rx) = watch::channel(MockCommand {
+            cmd: "".into(),
+            subcommand: None,
+            args: vec![],
+        });
+        Self {
+            expected: Mutex::default(),
+            tx,
+            rx,
+            failing: AtomicBool::new(false),
+        }
     }
 
     fn expect(&self, command: MockCommand, result: Result<RedisValue, RedisError>) -> &Self {
         self.expected.lock().unwrap().push_back((command, result));
         self
     }
+
+    async fn wait_for(&self, command: MockCommand) {
+        self.rx
+            .clone()
+            .wait_for(|cmd| *cmd == command)
+            .await
+            .expect("the channel isn't closed while the struct exists");
+    }
 }
 
 impl Mocks for MockRedisBackend {
     fn process_command(&self, actual: MockCommand) -> Result<RedisValue, RedisError> {
+        self.tx
+            .send(actual.clone())
+            .expect("the channel isn't closed while the struct exists");
+
         let Some((expected, result)) = self.expected.lock().unwrap().pop_front() else {
             // panic here -- this isn't a redis error, it's a test failure
+            self.failing.store(true, Ordering::Relaxed);
             panic!("Didn't expect any more commands, but received {actual:?}");
         };
 
-        assert_eq!(actual, expected);
+        if actual != expected {
+            self.failing.store(true, Ordering::Relaxed);
+            assert_eq!(
+                actual, expected,
+                "mismatched command, received (left) but expected (right)"
+            );
+        };
 
         result
     }
@@ -96,8 +138,8 @@ impl Mocks for MockRedisBackend {
 
 impl Drop for MockRedisBackend {
     fn drop(&mut self) {
-        if panicking() {
-            // We're already panicking, let's make debugging easier and let future devs solve problems one at a time.
+        if panicking() || self.failing.load(Ordering::Relaxed) {
+            // We're already failing, let's make debugging easier and let future devs solve problems one at a time.
             return;
         }
 
@@ -110,7 +152,7 @@ impl Drop for MockRedisBackend {
         assert_eq!(
             *expected,
             VecDeque::new(),
-            "Didn't receive all expected commands."
+            "Didn't receive all expected commands, expected (left)"
         );
 
         // Panicking isn't enough inside a tokio task, we need to `exit(1)`
@@ -186,9 +228,9 @@ async fn upload_and_get_data() -> Result<(), Error> {
         RedisStore::new_from_builder_and_parts(builder, None, mock_uuid_generator, String::new())?
     };
 
-    store.update_oneshot(digest, data.clone()).await?;
+    store.update_oneshot(digest, data.clone()).await.unwrap();
 
-    let result = store.has(digest).await?;
+    let result = store.has(digest).await.unwrap();
     assert!(
         result.is_some(),
         "Expected redis store to have hash: {VALID_HASH1}",
@@ -196,7 +238,8 @@ async fn upload_and_get_data() -> Result<(), Error> {
 
     let result = store
         .get_part_unchunked(digest, 0, Some(data.len()))
-        .await?;
+        .await
+        .unwrap();
 
     assert_eq!(result, data, "Expected redis store to have updated value",);
 
@@ -266,9 +309,9 @@ async fn upload_and_get_data_with_prefix() -> Result<(), Error> {
         )?
     };
 
-    store.update_oneshot(digest, data.clone()).await?;
+    store.update_oneshot(digest, data.clone()).await.unwrap();
 
-    let result = store.has(digest).await?;
+    let result = store.has(digest).await.unwrap();
     assert!(
         result.is_some(),
         "Expected redis store to have hash: {VALID_HASH1}",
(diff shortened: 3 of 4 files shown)


caass (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 2 discussions need to be resolved

___
*[`.gitignore` line 4 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nViKf4_LzErWCELdr:-O4nViKf4_LzErWCELds:b-vi50gq) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/.gitignore#L4)):*
> ```ignore
> target/
> .vscode/
> .zed
> ```

FYI you can't use LLDB in zed so I'm back to println debugging...
___
*[`nativelink-store/src/redis_store.rs` line 293 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nW1Zg7NO3smRk5Hfa:-O4nW1Zg7NO3smRk5Hfb:b3tm34x) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/src/redis_store.rs#L293)):*
> ```XML
> 
>                 eof_reached = true;
>                 continue;
> ```

This is semantically equivalent to `break`, but I felt that it was more clear to specifically indicate that an empty chunk indicates `EOF`.
___
*[`nativelink-store/tests/redis_store_test.rs` line 56 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWS5cEhTK7euxA0xL:-O4nWS5cEhTK7euxA0xM:b-czb9kb) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L56)):*
> ```XML
>     rx: watch::Receiver<MockCommand>,
> 
>     failing: AtomicBool,
> ```

FYI this wound up being more useful for getting helpful error messages than anything else, since we were struggling with getting tokio to only tell us about one error at a time.
___
*[`nativelink-store/tests/redis_store_test.rs` line 472 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWcfr8IqLPbPFKq0N:-O4nWcfsDv506GL1uxH0:b-k5jt3m) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L472)):*
> ```XML
> async fn yield_between_sending_packets_in_update() -> Result<(), Error> {
>     let data_p1 = Bytes::from(vec![b'A'; 6 * 1024]);
>     let data_p2 = Bytes::from(vec![b'B'; 4 * 1024]);
> ```

This made debugging easier, since at a glance I could tell what data had been sent and what hadn't.
___
*[`nativelink-store/tests/redis_store_test.rs` line 553 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWqLb3ZYbe16OkN6K:-O4nWqLb3ZYbe16OkN6L:bdu10x6) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L553)):*
> ```XML
>         },
>         async {
>             let mut store_rx = mocks.subscribe();
> ```

Nit this API could be more clear.
___
*[`nativelink-util/src/buf_channel.rs` line 209 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWuHI6mkOfD8eG1XW:-O4nWuHI6mkOfD8eG1XX:bisg3zp) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-util/src/buf_channel.rs#L209)):*
> ```XML
>     }
> 
>     fn recv_inner(&mut self, data: Result<Bytes, Error>) -> Result<Bytes, Error> {
> ```

Nit I don't really like this function name, something better might be like `process_data`.

caass (commented):
+@allada +@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O4nY_mk16FLwvLPL9Pm:bchhibo)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, asan / ubuntu-22.04, macos-13, pre-commit-checks, ubuntu-22.04, vale, and 1 discussions need to be resolved (waiting on @adam-singer and @allada)

adam-singer (approved):
:lgtm:

Do we have a way to trigger this in a test bounded up to a real redis instance vs mock tests? 

Also wasn't clear if the mock tests could trigger this condition, if so can we add a test proving it won't trigger or we don't regresss back to it?

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O4pvhDKEGwq5CokAjs6:bldsraw)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 2 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 293 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nW1Zg7NO3smRk5Hfa:-O4pv4k_ElOHd4CpH3uz:bguyp45) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/src/redis_store.rs#L293)):*

This is semantically equivalent to `break`, but I felt that it was more clear to specifically indicate that an empty chunk indicates `EOF`.
</blockquote></details>

sg, can we leave that as a comment so forward reading of the code makes sense why its laid out with a boolean/continue as the boolean is used later in the `'inner` reader loop.
___
*[`nativelink-store/src/redis_store.rs` line 294 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4puR3I5piA7EaWZYKW:-O4puR3I5piA7EaWZYKX:b-adq63) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L294)):*
> ```XML
>                 eof_reached = true;
>                 continue;
>             }
> ```

nit: can the assignment `is_first_chunk = false` be done in the a `else {}` of this condition, I think that reads it being specifically tied to the condition of  `chunk.is_empty()` not being empty and signals that a first chunk has been ready. I do understand its logically the same. The passthrough / fall down conditions/conclusions have are a bit much to follow. Not saying that it needs to be rewritten logically, hopefully less overhead to cognitively read it. 
___
*[`nativelink-store/tests/redis_store_test.rs` line 472 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWcfr8IqLPbPFKq0N:-O4ptTZYDBB_EbwmMx9o:b32v57) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L472)):*

This made debugging easier, since at a glance I could tell what data had been sent and what hadn't.
</blockquote></details>

+1

allada (dismissed):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 7 discussions need to be resolved

___
*[`nativelink-store/src/redis_store.rs` line 280 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4ptByAE6e-6hl8BH9i:-O4ptByAE6e-6hl8BH9j:bvr7dwy) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L280)):*
> ```XML
> 
>         while !eof_reached {
>             let pipe = client.pipeline();
> ```

nit: move this down below to where it is used.
___
*[`nativelink-store/src/redis_store.rs` line 303 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pvwLSDhAGRJlqjznE:-O4pvwLSDhAGRJlqjznF:bv20m0r) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L303)):*
> ```XML
> 
>             // Opportunistically grab any other chunks already in the reader.
>             'inner: while let Some(chunk) = reader
> ```

We should probably also flush if the `pipe`'s buffer gets larger than ~5MB (we can hard code it).
___
*[`nativelink-store/src/redis_store.rs` line 310 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pvYElETHM-ag-vIvx:-O4pvYElETHM-ag-vIvy:bk0uwwx) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L310)):*
> ```XML
>                 if chunk.is_empty() {
>                     eof_reached = true;
>                     break 'inner;
> ```

nit: No need for `'inner`.
___
*[`nativelink-store/src/redis_store.rs` line 326 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pv63H1CYjtuAR0X09:-O4pv63H1CYjtuAR0X0A:bodffap) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L326)):*
> ```XML
> 
>         // Rename the temp key so that the data appears under the real key. Any data already present in the real key is lost.
>         client
> ```

lets try and get this onto a pipeline too, this will save a roundtrip on nearly every upload, since it is rare for `try_recv` to return None, then next iteration return an empty slice.
___
*[`nativelink-store/tests/redis_store_test.rs` line 56 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWS5cEhTK7euxA0xL:-O4pxYRQ-3patZ_jEN7d:b-j83ibc) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L56)):*

FYI this wound up being more useful for getting helpful error messages than anything else, since we were struggling with getting tokio to only tell us about one error at a time.
</blockquote></details>

I'm a bit confused, do you mean you were struggling to get tests to only print one of the test failures at a time? We have asserts in there, so it should hard fail.
___
*[`nativelink-store/tests/redis_store_test.rs` line 557 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pwxZT0F7yNW1FzQm0:-O4pwxZT0F7yNW1FzQm1:bht7eku) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/tests/redis_store_test.rs#L557)):*
> ```XML
>         },
>         async {
>             tx.send(data_p1).await?;
> ```

fyi: I now prefer `.unwrap()` in tests, since `unwrap()`  gives line numbers an `?` does not.
___
*[`nativelink-util/src/buf_channel.rs` line 209 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pyECsD3E3U9o50YDz:-O4pyECsD3E3U9o50YE-:b-m2cvpv) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-util/src/buf_channel.rs#L209)):*
> ```XML
>     }
> 
>     fn recv_inner(&mut self, data: Result<Bytes, Error>) -> Result<Bytes, Error> {
> ```

Can we revert all of this and do it in a follow up PR? This change has a very high chance of breaking things and I don't want this entire PR to be reverted because of an unrelated change.

caass (commented):
@adam-singer we can test it the same way we tested the last redis change -- and I added a regression test.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O4rLX2fDooiYDhT4K_r:b27v7ob)* status: 1 of 2 LGTMs obtained, and all files reviewed, and pending CI: Analyze (javascript-typescript), Bazel Dev / ubuntu-22.04, Publish image, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), pre-commit-checks, ubuntu-20.04 / stable, vale, and 5 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 303 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pvwLSDhAGRJlqjznE:-O4rB2fM8jK4cPBLt4l-:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L303)):*

We should probably also flush if the `pipe`'s buffer gets larger than ~5MB (we can hard code it).
</blockquote></details>

Done.
___
*[`nativelink-store/src/redis_store.rs` line 310 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pvYElETHM-ag-vIvx:-O4rB6o2AOzi9N5nABOa:bb74ce2) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L310)):*

nit: No need for `'inner`.
</blockquote></details>

Sure
___
*[`nativelink-store/src/redis_store.rs` line 326 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pv63H1CYjtuAR0X09:-O4rCNRH-L54_fWuB9ZC:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-store/src/redis_store.rs#L326)):*

lets try and get this onto a pipeline too, this will save a roundtrip on nearly every upload, since it is rare for `try_recv` to return None, then next iteration return an empty slice.
</blockquote></details>

Done.
___
*[`nativelink-store/src/redis_store.rs` line 281 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4rLewwCBhnC_ZGfNlr:-O4rLewwCBhnC_ZGfNls:b-hlyyxh) ([raw file](https://github.com/TraceMachina/nativelink/blob/be34eca8f29c759b330a34a892107d597c23145e/nativelink-store/src/redis_store.rs#L281)):*
> ```XML
> 
>         while !eof_reached {
>             pipe = client.pipeline();
> ```

This weird double-declaration of `pipe` is because sending commands on a pipeline doesn't actually clear it -- so we need to drop the old pipeline every iteration.

The reason to put the pipeline here is so that the logic in the EOF handling uses a fresh pipe, and not one that contains stale data.
___
*[`nativelink-store/tests/redis_store_test.rs` line 56 at r4](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4nWS5cEhTK7euxA0xL:-O4rCOqY-blaho3U1ynM:bn4zrzv) ([raw file](https://github.com/TraceMachina/nativelink/blob/049220e77a6ffe452c8a331e45b2b33b449c64ec/nativelink-store/tests/redis_store_test.rs#L56)):*

I'm a bit confused, do you mean you were struggling to get tests to only print one of the test failures at a time? We have asserts in there, so it should hard fail.
</blockquote></details>

It had to do with `spawn`ing tokio stuff, where panics inside a task that wasn't awaited would silently die without killing the test.

[Context here](https://stackoverflow.com/questions/75520495/how-to-make-tokio-test-fail-if-thread-panicked)
___
*[`nativelink-util/src/buf_channel.rs` line 209 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4pyECsD3E3U9o50YDz:-O4rD80OBhRFeCZF-fnt:b-7f9kw8) ([raw file](https://github.com/TraceMachina/nativelink/blob/6d0ec8d767c813aaffe3a0c8caa6c51c4c03a58d/nativelink-util/src/buf_channel.rs#L209)):*

Can we revert all of this and do it in a follow up PR? This change has a very high chance of breaking things and I don't want this entire PR to be reverted because of an unrelated change.
</blockquote></details>

Unfortunately this change was a pre-req for everything else :( if you want, I can split this out into its own PR and we can test that first, but it would need to land before the rest of the redis stuff can land.

adam-singer (changes requested):
> Reviewed 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 6 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 309 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4vjKrc5YJqucZqfAXD:-O4vjKrc5YJqucZqfAXE:b4qp3wx) ([raw file](https://github.com/TraceMachina/nativelink/blob/be34eca8f29c759b330a34a892107d597c23145e/nativelink-store/src/redis_store.rs#L309)):*
> ```XML
>                         .err_tip(|| "While appending to temp key in RedisStore::update")?;
>                 } else {
>                     // Case 3, regular EOF.
> ```

Sorry, what I meant was `eof_reached = true;` goes into this block, not an empty block

caass (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 6 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 309 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4vjKrc5YJqucZqfAXD:-O4zioqyA6Qkxh6e_a_G:b-3o4ct6) ([raw file](https://github.com/TraceMachina/nativelink/blob/be34eca8f29c759b330a34a892107d597c23145e/nativelink-store/src/redis_store.rs#L309)):*

Sorry, what I meant was `eof_reached = true;` goes into this block, not an empty block
</blockquote></details>

Right, but don't we actually want to set `eof_reached = true` in both cases 2 and 3?

adam-singer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 6 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 309 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4vjKrc5YJqucZqfAXD:-O5-jEFB1vnn97bgleuy:b-veygyy) ([raw file](https://github.com/TraceMachina/nativelink/blob/be34eca8f29c759b330a34a892107d597c23145e/nativelink-store/src/redis_store.rs#L309)):*

Right, but don't we actually want to set `eof_reached = true` in both cases 2 and 3?
</blockquote></details>

ah ok, never mind then, the `else {}` here is pointless, it read differently to me in a different patch, remove the `else {}` condition here.

adam-singer (approved):
@caass did you test it the same way or is that still to be done?


:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O5ZENas4KC_EpHLBepd:b-1x3czc)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 6 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 309 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4vjKrc5YJqucZqfAXD:-O5ZEDdXCLZ7br2LnJC3:b-hpaqvc) ([raw file](https://github.com/TraceMachina/nativelink/blob/fd5a815e281a70c464da9aa53ca5012f7b1b6a50/nativelink-store/src/redis_store.rs#L309)):*

ah ok, never mind then, the `else {}` here is pointless, it read differently to me in a different patch, remove the `else {}` condition here.
</blockquote></details>

@caass ^^

caass (commented):
@adam-singer I tested it on cloud and it seemed to do the same thing as `main`

> Reviewed 3 of 3 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O7-uT8XCaXYV4Dd9Y-N:b-xmbzpr)* status: 1 of 2 LGTMs obtained, and all files reviewed, and pending CI: pre-commit-checks, and 5 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/redis_store.rs` line 309 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-O4vjKrc5YJqucZqfAXD:-O7-uEwvA81C5TEA292h:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/fd5a815e281a70c464da9aa53ca5012f7b1b6a50/nativelink-store/src/redis_store.rs#L309)):*

@caass ^^
</blockquote></details>

Done.

caass (commented):
> Dismissed @adam-singer and @allada from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269)* status: 1 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, NativeLink.com Cloud / Remote Cache (Legacy Dockerfile Test), Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable (waiting on @allada)

caass (commented):
-@allada 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1269#-:-O701RIx20KP3naj-MjP:b-r9zead)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

PR #1268, opened 2024-08-19
Migrate `docs` build system from pnpm to `bun`

SchahinRohani:
# Description

- Migrate the docs build system from pnpm to bun
- Set up GitHub workflow with Deno Deploy
- Configure Biome for code quality
- Remove unnecessary packages (Playform Compression causes issues)
- Refactore some of the Docs code for Code Quality.
- Update documentation to reflect new setup
- Add necessary dependencies in flake.nix

Fixes https://github.com/TraceMachina/nativelink/issues/1082

Preview [Deploy NativeLink Docs](https://github.com/TraceMachina/nativelink/actions/runs/10681909546/job/29606549890?pr=1268) Github Action Workflow 
Preview the deployed [NativeLink Docs](https://nativelink-docs-1n1ygpj7p441.deno.dev/) 

## Type of change

Please delete options that aren't relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist

- [x] Updated documentation if needed
- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1268)

Issue #1262, opened 2024-08-13
Broken nix flake on `aarch64-darwin`


SchahinRohani:
### **Description**

I'm encountering issues when running ```nix flake show --allow-import-from-derivation``` on MacOS (aarch64-darwin). The command results in an error that halts the process. It appears that there are issues in building the nixpkgs-patched with the wrong architecture.

### **Warnings & Errors**

```bash
error:
       … while evaluating the attribute 'optionalValue.value'

         at /nix/store/l04ifn08y80fcd3iblpzfr0fj0pqc58z-source/modules.nix:856:5:

          855|
          856|     optionalValue =
             |     ^
          857|       if isDefined then { value = mergedValue; }

       … while evaluating a branch condition

         at /nix/store/l04ifn08y80fcd3iblpzfr0fj0pqc58z-source/modules.nix:857:7:

          856|     optionalValue =
          857|       if isDefined then { value = mergedValue; }
             |       ^
          858|       else {};

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: a 'aarch64-linux' with features {} is required to build '/nix/store/jrhk2sdi1dij7hka434irlb1rnw4xv51-nixpkgs-patched.drv', but I am a 'aarch64-darwin' with features {apple-virt, benchmark, big-parallel, nixos-test}
```
### **Environment**

Operating System: MacOS
Architecture: aarch64-darwin
Nix Version: nix (Nix) 2.19.2
Nativelink Version: 0.5.1

### **Steps to Reproduce**
Run ```nix flake show --allow-import-from-derivation``` on MacOS with aarch64-darwin architecture.
Observe the warnings and errors during the evaluation.

### **Expected Behavior**
The ```nix flake show --allow-import-from-derivation``` command should complete successfully or provide actionable feedback without encountering fatal errors.

aaronmondal:
I can reproduce on `x86_64-linux`. This is caused by the fact that it's not possible to crosscompile for darwin, not even between `aarch64-darwin` and `x86_64-darwin`. I thought I cut these paths entirely in #1233 but apparently I overlooked something.

The solution to this is to remove the darwin flake outputs from the `packages` section on systems that don't support them. I.e.

- remove both if the host is linux
- remove `x86_64-darwin` if the host is `aarch64-darwin`
- remove `aarch64-darwin` if the host is `x86_64-darwin`

@SchahinRohani Do you want to try solving this issue?

aaronmondal:
[...]

PR #1279, opened 2024-08-23
Refactor `flake.nix` 

SchahinRohani:
# Description

Refactoring the flake into `tools` for better accessability.

Depends on this PR #1268.

## How Has This Been Tested?

The command `nix flake check`and `nix flake show` are throwing errors, so i need to refactor the flake first to be able to test it properly.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1279)


adam-singer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1279)* status: 0 of 2 LGTMs obtained, and 0 of 46 files reviewed, and pending CI: Docs Deployment, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), and 1 discussions need to be resolved

___
*[`tools/images/toolchain-drake.nix` line 10 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50I-esAAdGFOiaHPu2:-O50I-esAAdGFOiaHPu3:b7rsbax) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/images/toolchain-drake.nix#L10)):*
> ```Nix
>   fromImage = pullImage {
>     imageName = "localhost:5001/toolchain-drake";
>     imageDigest = ""; # DO NOT COMMIT DRAKE IMAGE_DIGEST VALUE
> ```

Keep in mind that https://github.com/TraceMachina/nativelink/blob/main/tools/toolchain-drake/toolchain-drake.sh#L12 does try to patch this file durning a toolchain build, it does a patch and then reverts the change when its done. So if this files path changes it will need to be reflected here.

aaronmondal (changes requested):
TBH I'm not a fan of this. The only thing this accomplishes is to reduce a few lines in the flake.nix at the cost of much higher complexity pushed into other directories, increased maintenance overhead for the factored-out parts and a higher requirement to know about the directory layout to do meaningful changes (i.e. you now have to change a bunch of files and know where things are instead of just editing a single file).

At first it might look like a nice cleanup to have a clearer separation for the different parts, but in many places this separation is "off" or wrong. Also in its current state this regresses fairly heavily on `import` vs `callPackage` usage.

I'm aware that the current state of the `flake.nix` file can be overwhelming, but the most important improvement to this is not part of this PR, which would be to move nativelink into it's own `default.nix` file. The reason this wasn't done yet is mainly because we have fairly specific requirements for crosscompilation that doesn't fit too well with the classic layouts in nixpkgs and other flake repos.

I'd be fine with moving the images like `siso-chromium` and `drake` somewhere else, but even for images in general it becomes unclear whether `nativelink-worker-init` should be in images or considered a part of a nativelink package.

I'd prefer if we dropped this PR and made smaller, more gradual changes (perhaps starting with only drake and siso which are "safe" to move around) instead of refactoring this many things at once.

> Reviewed 26 of 26 files at r1, 22 of 22 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-:-O50gCmw0PA1iK6MKd4W:b-jc20a)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Docs Deployment, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), and 19 discussions need to be resolved

___
*[`.gitignore` line 21 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50RxS7-it8PTSQyKoJ:-O50RxS7-it8PTSQyKoK:b-142obk) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/.gitignore#L21)):*
> ```ignore
> lre.bazelrc
> rust-project.json
> .flake.nix
> ```

what is this file?
___
*[`flake.nix` line 221 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50S1WC0G7c_xmjdqYH:-O50S1WC0G7c_xmjdqYI:bse90p7) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/flake.nix#L221)):*
> ```Nix
>         generate-toolchains = pkgs.callPackage ./tools/toolchains/generate.nix {inherit rbe-configs-gen;};
> 
>         buck2-toolchain = import ./tools/buildsystem/buck2.nix {
> ```

Prefer the `pkgs.callPackage` syntax.

When you have e.g. this:

```nix
x = pkgs.callPackage ./myfile.nixx {}
```

You can change the `myfile.nix` inputs arbitrarily without modifying the line above:

```nix
{
some,
inputs,
}

# use the inputs
```

```nix
{
some,
new,
inputs,
}

#use the inputs, doesn't require changing the top-level callPackage file.
```

Also this behaves better with crosscompilation since you can globablly override/patch `pkgs.callPackage` to e.g. `pkgs.pkgcCross.sometarget.callPackage` or `patched-pkgs.callPackage` without changing the implementation file.
___
*[`flake.nix` line 241 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50SpYwEHPQlLcGrsgO:-O50SpYwEHPQlLcGrsgP:b5h13dl) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/flake.nix#L241)):*
> ```Nix
>         nativelink-image = import ./tools/images/nativelink.nix {
>           inherit self;
>           inherit buildImage;
> ```

In this case the above `pkgs.callPackage` argument still holds.

In general, when you have multiple inputs that can be `inherit`ed, use this syntax instead:

```nix
inherit
  a
  b
  c
  ;
```

This is more idiomatic and allows 
___
*[`flake.nix` line 272 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50TSDS495311dgHy_u:-O50TSDS495311dgHy_v:b-nw8023) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/flake.nix#L272)):*
> ```Nix
>       in rec {
>         _module.args.pkgs = let
>           nixpkgs-patched = import ./tools/patches/apply.nix {
> ```

Let's keep the previous implementation here for not to make it as obvious as possible which patches are applied. I think it's a good idea to move this out somewhere, but I'll need some time to think about whether this is the optimal approach.
___
*[`flake.nix` line 367 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50Txre4utDhppzATNV:-O50Txre4utDhppzATNW:b-l32mkf) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/flake.nix#L367)):*
> ```Nix
>         };
> 
>         # Example for another devShell called "test"
> ```

No need for this example. Let's just remove it.
___
*[`tools/buildsystem/buck2.nix` line 2 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50UE6W0mp40dMvMs9G:-O50UE6W0mp40dMvMs9H:b-rz3d9) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/buildsystem/buck2.nix#L2)):*
> ```Nix
> {
>   pkgs,
> ```

Don't add `pkgs` to inputs. Instead, use the `callPackage` syntax and in this case do for instance this:

```nix
{
  buildImage,
  self,
  coreutils
  bash
  go
  diffutils
  gnutar
  gzip
  python3Full
  unzip
  zstd
  cargo-bloat
  mold-wrapped
  reindeer
  lld_16
  clang_16
}: let
  buck2-nightly-rust-version = "2024-04-28";
  buck2-nightly-rust = pkgs.rust-bin.nightly.${buck2-nightly-rust-version};
  buck2-rust = buck2-nightly-rust.default.override {extensions = ["rust-src"];};
in
  pkgs.callPackage ../nativelink/create-worker-experimental.nix {
    inherit buildImage self;
    imageName = "buck2-toolchain";
    packagesForImage = [
      coreutils
      bash
      go
      diffutils
      gnutar
      gzip
      python3Full
      unzip
      zstd
      cargo-bloat
      mold-wrapped
      reindeer
      lld_16
      clang_16
      buck2-rust
    ];
  }
```

I know that this looks like a lot of duplication, but it let's us do things like this:

```nix
mytoolchain = (pkgs.callPackage ./mytoolchain.nix {
  buck2-rust;
  lld = pkgs.lld_18; # Override a single input
})
```

Note how this now lets you override specific inputs as opposed to "all or nothing":

```nix
mytoolchain = pkgs.callPackage ./myshell.nix {
  # Only override clang.
  inherit (pkgsCross.musl64) clang;
};

# Or override everything
mytoolchain = pkgs.pkgsCross.musl64.callPackage ./mytoolchain.nix {};

# Here we have no way to selectively change tools
mytoolchain = import ./mytoolchain.nix {
  pkgs = pkgs.pkgsCross.musl64;
}
```
___
*[`tools/images/nativelink.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50Z8YT0ie1k-nHKHX1:-O50Z8YU6uuJHGUfZdgv:bcghput) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/images/nativelink.nix#L1)):*
> ```Nix
> {
> ```

Let's not move this one. The way that crosscompilation interacts with this is dependent on the exec platform of `buildImage`. This is not ideal at the moment, but factoring it out of the main flake.nix makes it harder to fix this issue.
___
*[`tools/images/siso-chromium.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50bYlM33j-VokIe6Br:-O50bYlM33j-VokIe6Bs:b-1km7dw) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/images/siso-chromium.nix#L1)):*
> ```Nix
> {
> ```

@adam-singer @SchahinRohani I wonder, should these live in `tools/images` or would a top-level `images` directory be more appropriate? I think I'd prefer a top-level directory which would also make it easier for other users to see where the images live, as this is a fairly common thing to look at when browsing a repo.
___
*[`tools/images/toolchain-drake.nix` line 10 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50I-esAAdGFOiaHPu2:-O50c6Nk08hD0xFo9DhK:bpowxyb) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/images/toolchain-drake.nix#L10)):*

Keep in mind that https://github.com/TraceMachina/nativelink/blob/main/tools/toolchain-drake/toolchain-drake.sh#L12 does try to patch this file durning a toolchain build, it does a patch and then reverts the change when its done. So if this files path changes it will need to be reflected here.
</blockquote></details>

Ah I totally forgot about this. Maybe we should leave some sort of reference to the patching file here?
___
*[`tools/patches/apply.nix` line 8 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50cSDs3HlT2f6izp2F:-O50cSDs3HlT2f6izp2G:b-rxxv2x) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/patches/apply.nix#L8)):*
> ```Nix
>   name = "nixpkgs-patched";
>   src = self.inputs.nixpkgs;
>   # To apply a patch on a pkg, go into the nixpkgs repository
> ```

nit: Use the full 80 character line length for comments and write out `package`.
___
*[`tools/shell/base.nix` line 2 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50dXg86t6diK3AKNKT:-O50dXg86t6diK3AKNKU:b-z0bjbe) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/shell/base.nix#L2)):*
> ```Nix
> {
>   config,
> ```

This seems like an unnecessary addition of complexity. The idiomatic approach here would be to put every package in the inputs. Since the devShell changes frequently that would mean that we'd have to change this file in multiple places all the time.

There is also not too much to gain from using different devshells as outlined in https://github.com/TraceMachina/nativelink/issues/1262#issuecomment-2301170142

Let's keep this in the top-level flake to make it easy for users to add/remove things to the dev environment.
___
*[`tools/shell/test.nix` line 2 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50e8YtBReMU1U-l5J9:-O50e8YtBReMU1U-l5JA:byp18fd) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/shell/test.nix#L2)):*
> ```Nix
> {pkgs, ...}:
> pkgs.mkShell {
> ```

Remove
___
*[`tools/nativelink/create-worker.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50eEbtBRu9M52JBXDl:-O50eEbu-HBzUvN0-M_i:bkwp9zx) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/nativelink/create-worker.nix#L1)):*
> ```Nix
> {
> ```

I'm not convinced that it's necessary to add a new level of indirection for this. Keep in mind that every additional directory increases friction in terminal-based editors.

Also, this is technically a utility to create an image, so I'd probably either leave it in `tools` or put it in images.
___
*[`tools/nativelink/create-worker-experimental.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50ewQZCp5Tl3WYZNW3:-O50ewQZCp5Tl3WYZNW4:b-9chp70) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/nativelink/create-worker-experimental.nix#L1)):*
> ```Nix
> {
> ```

as above
___
*[`tools/nativelink/worker-init.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50f4EuCujjwW9INeRx:-O50f4EuCujjwW9INeRy:b-g3enn7) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/nativelink/worker-init.nix#L1)):*
> ```Nix
> {
> ```

This is an image.
___
*[`tools/patches/nixpkgs_bun.diff` line 29 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50dK2U4rAXpCPXbtKL:-O50dK2VDB10l8xcRl_G:b4663n8) ([raw file](https://github.com/TraceMachina/nativelink/blob/62ca54baa4688e80a1c8c6be2df5aa55512fa3ee/tools/patches/nixpkgs_bun.diff#L29)):*
> ```Diff
>          url = "https://github.com/oven-sh/bun/releases/download/bun-v${version}/bun-darwin-x64-baseline.zip";
> -        hash = "sha256-5PLk8q3di5TW8HUfo7P3xrPWLhleAiSv9jp2XeL47Kk=";
> +        hash = "sha256-gaDYzANmTTsjIlTXqqCWcOT7NTscuEUINsrjEb0eTRs==";
> ```

I believe this needs to be fixed in #1268 
___
*[`tools/toolchains/customClang.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50fUo-96j7vn7YNgCf:-O50fUo-96j7vn7YNgCg:bsov468) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/toolchains/customClang.nix#L1)):*
> ```Nix
> {
> ```

A `toolchains` directory is IMO a very dangerous thing to add as we have many different kinds and forms of toolchains. The separation between them is not always clear and there is a risk of creating "too much separation" between the different things that make up a toolchain.
___
*[`tools/toolchains/generate.nix` line 2 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50fyan4eVkHJrRHZ0K:-O50fyao06m-NRuy_i_G:bk6caiz) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/toolchains/generate.nix#L2)):*
> ```Nix
> {
>   bazel_7,
> ```

This one should actually live in `local-remote-execution`.
___
*[`tools/toolchains/llvmStdenv.nix` line 1 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1279#-O50g2ghDOjrDkB__yjG:-O50g2gi7JKL-edeRlvL:b-ywnsnb) ([raw file](https://github.com/TraceMachina/nativelink/blob/5903a8e82ce4f441882a41e8a8d12ba6e47b1ca0/tools/toolchains/llvmStdenv.nix#L1)):*
> ```Nix
> {
> ```

same as with the customClang comment. Let's not move this for now because I need time to think about what the best approach is for this.

palfrey:
So much has changed that we'd need to restart this one from scratch, so closing.

PR #1290, opened 2024-08-29
Use ultracdc instead of fastcdc

### LINKED ISSUE #692: Vendored fastcdc implementation is redundant
The code at https://github.com/TraceMachina/nativelink/blob/main/nativelink-util/src/fastcdc.rs might be unnecessary.

Potential alternatives:

- https://github.com/nlfiedler/fastcdc-rs
- https://github.com/florian-g2/fastcdc-rs-alt

There is also an interesting algorithm called `UltraCDC` from some of the `FastCDC` authors that seems worth investigating:

- https://github.com/PlakarLabs/go-cdc-chunkers/blob/main/chunkers/ultracdc/ultracdc.go

leodziki:
UltraCDC outperforms FastCDC in speed and deduplication

# Description

UltraCDC offers superior deduplication efficiency by optimizing chunk boundaries, which reduces storage needs. It provides faster performance compared to FastCDC, thanks to improvements in processing speed. Additionally, its adaptability to various data types and efficient chunk size selection enhance its scalability and overall effectiveness.

Fixes https://github.com/TraceMachina/nativelink/issues/692

## Type of change

## How Has This Been Tested?

`cargo test --all` passes as expected; `bazel test //... `succeeds as well.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290)


allada (commented):
Questions:
1. why the rename?
2. Why change the rust library this is based on to a go library?
3. What performance is this improving?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-:-O5Tl_Mt6TBhTVKMS7fs:bt1oavs)* status: 0 of 2 LGTMs obtained, and 0 of 7 files reviewed, and pending CI: Remote / large-ubuntu-22.04

leodziki:
1. Why the Rename? 
The rename from `fastcdc` to `ultracdc` reflect enhancements in speed by implementing `ultracdc` core algorithm.

2. Why change the rust library this is based on to a go library?
There is no rust library related to `ultracdc`.
There is an interesting algorithm called `UltraCDC` from some of the `FastCDC` authors that seems worth investigating:
[https://github.com/PlakarLabs/go-cdc-chunkers/blob/main/chunkers/ultracdc/ultracdc.go](https://github.com/PlakarLabs/go-cdc-chunkers/blob/main/chunkers/ultracdc/ultracdc.go)
After careful analysis, this approach has been translated into a Rust-based library.

3. What performance is this improving?
**Efficiency and Speed:** The transition could lead to performance improvements in terms of processing speed, memory usage, or scalability
**Optimization:** `ultracdc` could include optimizations that weren't available or possible in `fastcdc`, such as better chunking algorithms, more efficient data handling, or improvements in how data is processed and transferred.

leodziki:
Hope to review the implementation of UltraCDC algorithm approach soon.

allada (commented):
The reason I'm concerned here is because it looks like very few changes have been made from the current `fastcdc.rs` algorithm and the `ultracdc` algorithm posted here.

Can you delete `fastcdc.rs` in this PR so we can see the diff?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-:-O5gSCNL7dpGgM4g5MYK:ba3rphm)* status: 0 of 2 LGTMs obtained, and 0 of 7 files reviewed, and pending CI: Remote / large-ubuntu-22.04

leodziki:
Removed fastcdc.rs to better compare differences with the newly implemented ultracdc algorithm in this PR.
The main difference in this UltraCDC implementation is the use of a rolling hash function to determine chunk boundaries. 
The rolling hash value is compared against a dynamically selected mask to decide when to split the chunk. 
This approach allows for more flexible and potentially more accurate chunking compared to the simpler hash-based method used in FastCDC.

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290)* status: 0 of 2 LGTMs obtained, and 0 of 9 files reviewed, and 7 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xsUPJ4gVWby3cpuAP:-O5xsUPJ4gVWby3cpuAQ:b-3hdudj) (no related file):*
I'm concerned because this is appears to be significantly slower now (but have not measured it). We are now computing a rolling hash on every byte making the algorithm now `O(n * m)` where `n` is number of bytes and `m` is `window_size`. This function is already quite slow. I'm not sure it is a good idea to make it significantly slower.

If we remove the copies and make it so the rolling hash can be updated instead of recomputed, it can probably be accepted.

___
___
*[`nativelink-util/src/ultracdc.rs` line 74 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xrh8XCuxio7ECk2EC:-O5xrh8XCuxio7ECk2ED:b-bmkcah) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L74)):*
> ```XML
>             // Turn our bits into a bitmask we can use later on for more
>             // efficient bitwise operations.
>             mask_hard: 2u32.pow(bits + 1) - 1,
> ```

nit: Make these `u64` and do the casting here so we don't pay for casting below.
___
*[`nativelink-util/src/ultracdc.rs` line 84 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xqGQw5Edh_qqKcc4_:-O5xqGQw5Edh_qqKcc4a:bu6y3ul) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L84)):*
> ```XML
>     }
> 
>     pub fn rolling_hash(&self, data: &BytesMut, window_size: usize) -> u64 {
> ```

nit: This should take `&[u8]` instead of `BytesMut`.
___
*[`nativelink-util/src/ultracdc.rs` line 88 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xo9AO7lQ3vmRXcN1F:-O5xo9AO7lQ3vmRXcN1G:b-8nextz) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L88)):*
> ```XML
>         let mut hash_val = 0u64;
>         for (i, &byte) in data.iter().enumerate().take(window_size) {
>             hash_val += (byte as u64) << (i & 0x1F);
> ```

it looks like this is vulnerable to overflows. Can we instead use one of the safe functions like `.saturating_add`, `.overflowing_add` or similar.
___
*[`nativelink-util/src/ultracdc.rs` line 113 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xrC-3A4mLSD8ZibBi:-O5xrC-3A4mLSD8ZibBj:b-7irfh) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L113)):*
> ```XML
>         while i < buf.len() {
>             let byte = buf[i] as usize;
>             chunk.put_u8(buf[i]);
> ```

nit: We don't want to copy out data, instead use the `.split_*` functions to create the dynamic slices and return them.
___
*[`nativelink-util/src/ultracdc.rs` line 115 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xsFmN6iLjaXxfKdsd:-O5xsFmN6iLjaXxfKdse:bebo757) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L115)):*
> ```XML
>             chunk.put_u8(buf[i]);
> 
>             self.state.hash = (self.state.hash >> 1) + TABLE[byte];
> ```

this appears to be unused now.
___
*[`nativelink-util/src/ultracdc.rs` line 129 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xqQ5e4xpFQu7ZeiGN:-O5xqQ5e4xpFQu7ZeiGO:b-crfjdz) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L129)):*
> ```XML
>                 let chunk_len = chunk.len();
> 
>                 let bytes_mut = BytesMut::from(&chunk[chunk_len - window_size.min(chunk_len)..]);
> ```

nit: use a slice instead, so we don't need to copy the data.

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 84 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xqGQw5Edh_qqKcc4_:-O6D4cH42n4db0OBoRcM:bb6un1j) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L84)):*

nit: This should take `&[u8]` instead of `BytesMut`.
</blockquote></details>

Done

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 88 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xo9AO7lQ3vmRXcN1F:-O6D5KbQ17O6cKJ0h5b3:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L88)):*

it looks like this is vulnerable to overflows. Can we instead use one of the safe functions like `.saturating_add`, `.overflowing_add` or similar.
</blockquote></details>

Done.

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 115 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xsFmN6iLjaXxfKdsd:-O6D6_ns0vQrJzFEEFNY:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L115)):*

this appears to be unused now.
</blockquote></details>

Done.

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 74 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xrh8XCuxio7ECk2EC:-O6FJpek2ILPiaYXuuih:bb6un1j) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L74)):*

nit: Make these `u64` and do the casting here so we don't pay for casting below.
</blockquote></details>

Done

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 129 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xqQ5e4xpFQu7ZeiGN:-O6FOdUK7b27E6drtU_D:bb6un1j) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L129)):*

nit: use a slice instead, so we don't need to copy the data.
</blockquote></details>

Done

leodziki:
*[`nativelink-util/src/ultracdc.rs` line 113 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O5xrC-3A4mLSD8ZibBi:-O6FTabPBPo2fzO9uNo1:bb6un1j) ([raw file](https://github.com/TraceMachina/nativelink/blob/4c30580f218509964fce2e0eccdce35b8e015c6f/nativelink-util/src/ultracdc.rs#L113)):*

nit: We don't want to copy out data, instead use the `.split_*` functions to create the dynamic slices and return them.
</blockquote></details>

Done

leodziki:

I'm concerned because this is appears to be significantly slower now (but have not measured it). We are now computing a rolling hash on every byte making the algorithm now `O(n * m)` where `n` is number of bytes and `m` is `window_size`. This function is already quite slow. I'm not sure it is a good idea to make it significantly slower.

If we remove the copies and make it so the rolling hash can be updated instead of recomputed, it can probably be accepted.
</blockquote></details>

I understand your concern about the potential performance impact of the current implementation. You're right that computing the rolling hash for every byte could lead to a time complexity of O(n m), where n is the number of bytes and m is the window size.

To optimize this, we can maintain the rolling hash in a way that only updates it when necessary, rather than recalculating it for every byte. This can be achieved by adjusting the hash incrementally as bytes are added and removed from the window, ensuring that we only perform constant-time updates.

leodziki:
I hope to review the revision based on the feedback. cc: @allada, @MarcusSorealheis

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290)* status: 0 of 2 LGTMs obtained, and 0 of 9 files reviewed, and 4 discussions need to be resolved

*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O6MR21lCMn7s2R6gnDX:-O6MR21lCMn7s2R6gnDY:bvkbzyp) (no related file):*
I'm still really failing to see why to change this. From what I can tell this is adding a significant amount of overhead to the existing algorithm to only gain the ability for a dynamically selected mask. The existing algorithm is a rolling checksum. Can you perform some measurements to show it performing better on any benchmark along with speed?

___

leodziki:
Thank you for sharing your thoughts. I understand your concerns about the potential overhead introduced by the changes, especially in the context of the existing rolling checksum algorithm. I will conduct some benchmarks to compare the performance of the current implementation against the proposed changes, focusing on both speed and efficiency. This data will help us make a more informed decision moving forward.

leodziki:
> > _[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1290)_ status: 0 of 2 LGTMs obtained, and 0 of 9 files reviewed, and 4 discussions need to be resolved
> 
> _[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/1290#-O6MR21lCMn7s2R6gnDX:-O6MR21lCMn7s2R6gnDY:bvkbzyp) (no related file):_ I'm still really failing to see why to change this. From what I can tell this is adding a significant amount of overhead to the existing algorithm to only gain the ability for a dynamically selected mask. The existing algorithm is a rolling checksum. Can you perform some measurements to show it performing better on any benchmark along with speed?

Thank you for sharing your thoughts. I understand your concerns about the potential overhead introduced by the changes, especially in the context of the existing rolling checksum algorithm. I will conduct some benchmarks to compare the performance of the current implementation against the proposed changes, focusing on both speed and efficiency. This data will help us make a more informed decision moving forward.

palfrey:
Closing. If we can get benchmark data, we can reopen this.

PR #1300, opened 2024-08-31
Implement DropGuard to ensure future Completion

### LINKED ISSUE #1156: EvictingMap manages state between await boundaries
If you look at the following code:
https://github.com/TraceMachina/nativelink/blob/d9922b370ab680602e7669a1480b6fa6694aaa1e/nativelink-util/src/evicting_map.rs#L437

You'll notice that we change the state of struct after this await call. This is quite dangerous because our future can die at any point `.await` is called. This is quite tricky to manage and easy to overlook. In reality, what we want to do is to prevent this future (and others similar) to not be able to be dropped in the middle of anywhere in the future (ie: always complete future). The way this is normally done is via `tokio::spawn`, but this is a very heavy handed way to handle this.

What I believe would be a good solution is to make a future wrapper that holds the state of if the inner future completed. If it was not completed, implement a drop function that will move the future to a background spawn and ignore the results. This in theory should give us the best of both worlds. We'll be able to keep the future on the stack, but in the rare event the future gets dropped in the middle of a critical state managed future, it'll still execute but as a background spawn.

leodziki:
# Description

To manage state transitions safely across `await` boundaries, we introduced a `DropGuard` wrapper. This ensures that even if a future is dropped before completion, it will continue executing in the background via `tokio::spawn`. This approach allows us to avoid the pitfalls of incomplete state transitions in critical futures without the overhead of spawning all futures immediately.

The `DropGuard` checks if the inner future has completed; if not, it safely spawns the future in the background during its drop. This solution strikes a balance between stack-based future management and the necessity to complete crucial futures.

Fixes https://github.com/TraceMachina/nativelink/issues/1156

## Type of change

  not work as expected)

## How Has This Been Tested?

`cargo test --all` and `bazel test //...` passes as expected.

## Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1300)

### DIFF
--- a/nativelink-scheduler/tests/utils/mock_scheduler.rs
+++ b/nativelink-scheduler/tests/utils/mock_scheduler.rs
@@ -17,12 +17,10 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use nativelink_error::{make_input_err, Error};
 use nativelink_metric::{MetricsComponent, RootMetricsComponent};
-use nativelink_util::{
-    action_messages::{ActionInfo, OperationId},
-    known_platform_property_provider::KnownPlatformPropertyProvider,
-    operation_state_manager::{
-        ActionStateResult, ActionStateResultStream, ClientStateManager, OperationFilter,
-    },
+use nativelink_util::action_messages::{ActionInfo, OperationId};
+use nativelink_util::known_platform_property_provider::KnownPlatformPropertyProvider;
+use nativelink_util::operation_state_manager::{
+    ActionStateResult, ActionStateResultStream, ClientStateManager, OperationFilter,
 };
 use tokio::sync::{mpsc, Mutex};
 
--- a/nativelink-util/BUILD.bazel
+++ b/nativelink-util/BUILD.bazel
@@ -16,6 +16,7 @@ rust_library(
         "src/common.rs",
         "src/connection_manager.rs",
         "src/digest_hasher.rs",
+        "src/drop_protected_future.rs",
         "src/evicting_map.rs",
         "src/fastcdc.rs",
         "src/fs.rs",
--- a/nativelink-util/src/drop_protected_future.rs
+++ b/nativelink-util/src/drop_protected_future.rs
@@ -1,0 +1,68 @@
+// Copyright 2024 The NativeLink Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+use tracing::{Instrument, Span};
+
+use crate::origin_context::{ContextAwareFuture, OriginContext};
+use crate::spawn;
+#[derive(Clone)]
+pub struct DropProtectedFuture<F, T>
+where
+    T: Send + 'static,
+    F: Future<Output = T> + Send + 'static + Unpin + Clone,
+{
+    future: ContextAwareFuture<F>,
+}
+
+impl<F, T> DropProtectedFuture<F, T>
+where
+    T: Send + 'static,
+    F: Future<Output = T> + Send + 'static + Unpin + Clone,
+{
+    pub fn new(f: F, span: Span, ctx: Option<Arc<OriginContext>>) -> Self {
+        Self {
+            future: ContextAwareFuture::new(ctx, f.instrument(span)),
+        }
+    }
+}
+
+impl<F, T> Future for DropProtectedFuture<F, T>
+where
+    T: Send + 'static,
+    F: Future<Output = T> + Send + 'static + Unpin + Clone,
+{
+    type Output = F::Output;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match Pin::new(&mut self.future).poll(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(output) => Poll::Ready(output),
+        }
+    }
+}
+
+impl<F, T> Drop for DropProtectedFuture<F, T>
+where
+    T: Send + 'static,
+    F: Future<Output = T> + Send + 'static + Unpin + Clone,
+{
+    fn drop(&mut self) {
+        spawn!("DropProtectedFuture::drop", self.future.clone());
+    }
+}
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -26,10 +26,12 @@ use lru::LruCache;
 use nativelink_config::stores::EvictionPolicy;
 use nativelink_metric::MetricsComponent;
 use serde::{Deserialize, Serialize};
-use tracing::{event, Level};
+use tracing::{event, info_span, Level};
 
+use crate::drop_protected_future::DropProtectedFuture;
 use crate::instant_wrapper::InstantWrapper;
 use crate::metrics_utils::{Counter, CounterWithTime};
+use crate::origin_context::ActiveOriginContext;
 
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 pub struct SerializedLRU<K> {
@@ -98,7 +100,7 @@ impl<T: LenEntry + Send + Sync> LenEntry for Arc<T> {
 }
 
 #[derive(MetricsComponent)]
-struct State<K: Ord + Hash + Eq + Clone + Debug, T: LenEntry + Debug> {
+struct State<K: Ord + Hash + Eq + Clone + Debug + Send, T: LenEntry + Debug> {
     lru: LruCache<K, EvictionItem<T>>,
     btree: Option<BTreeSet<K>>,
     #[metric(help = "Total size of all items in the store")]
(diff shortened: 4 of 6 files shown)


leodziki:
Hope to review the implementation of DropGuard soon

allada (changes requested):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1300)* status: 0 of 2 LGTMs obtained, and 1 of 5 files reviewed, and 8 discussions need to be resolved

___
*[`nativelink-util/src/drop_guard.rs` line 1 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gOqso3U22xfMeWDDA:-O5gOqso3U22xfMeWDDB:bok2db8) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L1)):*
> ```XML
> use std::future::Future;
> ```

nit: Needs copyright.
___
*[`nativelink-util/src/drop_guard.rs` line 5 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gKOYvA_r_UQ97hJhs:-O5gKOYvA_r_UQ97hJht:b-njh44t) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L5)):*
> ```XML
> use std::task::{Context, Poll};
> 
> pub struct DropGuard<F: Future> {
> ```

nit: `F: Future + Send + 'static` because our future must be able to be transferred into a spawn
___
*[`nativelink-util/src/drop_guard.rs` line 5 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gOeBb3A3HpJXzI9ZD:-O5gOeBb3A3HpJXzI9ZE:b-rzlv3j) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L5)):*
> ```XML
> use std::task::{Context, Poll};
> 
> pub struct DropGuard<F: Future> {
> ```

nit: Rename to `DropProtectedFuture` (and rename file).
___
*[`nativelink-util/src/drop_guard.rs` line 6 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gK2Ad7NQAy0Nz9rJF:-O5gK2Ad7NQAy0Nz9rJG:b-4yu4ft) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L6)):*
> ```XML
> 
> pub struct DropGuard<F: Future> {
>     future: Option<Pin<Box<F>>>,
> ```

nit: We should be able to do this with a `F`. We should not need to `Box` and cannot `Pin` here (because we may xfer our future to a spawn).
___
*[`nativelink-util/src/drop_guard.rs` line 10 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gNM2oEq0LrsrmTliF:-O5gNM2oEq0LrsrmTliG:bbi3kdh) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L10)):*
> ```XML
> 
> impl<F: Future> DropGuard<F> {
>     pub fn new(future: F) -> Self {
> ```

We need to be careful that we don't loose our global nativelink `Context` here.

1. We need to have the user pass in a span here.
2. The future that we store needs to be wrapped in`ContextAwareFuture::new` (plus the span the user passes in).
3. (see below in Drop call)
___
*[`nativelink-util/src/drop_guard.rs` line 30 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gMzcU5kR86tIIfg3x:-O5gMzcU5kR86tIIfg3y:bh7augp) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L30)):*
> ```XML
>             }
>         } else {
>             panic!("Future already completed");
> ```

nit: use `unreachable`
___
*[`nativelink-util/src/drop_guard.rs` line 39 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gN7EB-Zeo6j-Pv1BQ:-O5gN7EB-Zeo6j-Pv1BR:b1cnqvq) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L39)):*
> ```XML
>         if let Some(future) = self.future.take() {
>             // Block on the future to ensure it completes.
>             futures::executor::block_on(future);
> ```

This is unsafe, instead just forward it into a `spawn!()` and do not wait for the future.

Normally we'd require a `background_spawn!()`, but in this case we want to use `spawn!` directly and suppress the linter. If we use `background_spawn!` we'll create double work, because `ContextAwareFuture` handles all the context stuff we need to be careful with.
___
*[`nativelink-util/src/evicting_map.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gP6o_-HE8bjBsFdlH:-O5gOuhp8nvDwqhi35QT:bh544lf) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/evicting_map.rs#L438)):*
> ```XML
>             };
> 
>             let fut = state.put(key, eviction_item);
> ```

nit: We don't want to do this here, we want to make this entire future/function call protected. This can be done by rewriting the function signature to:
```rs
    fn inner_insert_many(
        &self,
        mut state: &mut State<K, T>,
        inserts: impl IntoIterator<Item = (K, T)>,
        seconds_since_anchor: i32,
    ) -> impl Future<Output=Vec<T>> + Send {
        DropSafeFuture::new(async move {
          // ...
        })
```

leodziki (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1300)* status: 0 of 2 LGTMs obtained, and 0 of 7 files reviewed, and pending CI: Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, NativeLink.com Cloud / Remote Cache (Legacy Dockerfile Test), Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 5 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-util/src/evicting_map.rs` line 447 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O6uHLSW4tzPNdLqTTmO:-O6uHLSW4tzPNdLqTTmP:b8ljxq) ([raw file](https://github.com/TraceMachina/nativelink/blob/fcc7fd522f4801c11405c2d5a33f50a1e8d511a3/nativelink-util/src/evicting_map.rs#L447)):*
> ```XML
>                     };
> 
>                     if let Some(old_item) = state.put(key, eviction_item).await {
> ```

I'm submitting this WIP to clarify unclear points.

As we see, there is `self` and `state` referenced here & mutated n this `async move {}` block.

We are gonna implement `Send + `static` trait on the Future itself.

How shall we handle this?

Should we change this functions parameter as `Arc<Mutex>` for state and add `Arc<Self>` as an additional parameter for `self`?
Or should we implement `Clone` trait on `self` and `state`?

Waiting for clarifying further walks from this.

Thanks.
___
*[`nativelink-util/src/evicting_map.rs` line 450 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O6uHi6mDPz9xeGx2ELz:-O6uHi6n-aneAwVw4nWA:b-875k1z) ([raw file](https://github.com/TraceMachina/nativelink/blob/fcc7fd522f4801c11405c2d5a33f50a1e8d511a3/nativelink-util/src/evicting_map.rs#L450)):*
> ```XML
>                         replaced_items.push(old_item);
>                     }
>                     state.sum_store_size += new_item_size;
> ```

Here.
___
*[`nativelink-util/src/evicting_map.rs` line 452 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O6uHdlo5DjZX27pAtpO:-O6uHdlo5DjZX27pAtpP:b-875k1z) ([raw file](https://github.com/TraceMachina/nativelink/blob/fcc7fd522f4801c11405c2d5a33f50a1e8d511a3/nativelink-util/src/evicting_map.rs#L452)):*
> ```XML
>                     state.sum_store_size += new_item_size;
>                     state.lifetime_inserted_bytes.add(new_item_size);
>                     self.evict_items(state.deref_mut()).await;
> ```

Here.
___
*[`nativelink-util/src/drop_guard.rs` line 10 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gNM2oEq0LrsrmTliF:-O6uH-DcCKa3TvMHCmXP:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L10)):*

We need to be careful that we don't loose our global nativelink `Context` here.

1. We need to have the user pass in a span here.
2. The future that we store needs to be wrapped in`ContextAwareFuture::new` (plus the span the user passes in).
3. (see below in Drop call)
</blockquote></details>

Done.
___
*[`nativelink-util/src/drop_guard.rs` line 39 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1300#-O5gN7EB-Zeo6j-Pv1BQ:-O6uGzr-96bsN9ZtEton:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/f8c62b8ae79fc8a353f243ce4a7c0d1d0b591f57/nativelink-util/src/drop_guard.rs#L39)):*

This is unsafe, instead just forward it into a `spawn!()` and do not wait for the future.

Normally we'd require a `background_spawn!()`, but in this case we want to use `spawn!` directly and suppress the linter. If we use `background_spawn!` we'll create double work, because `ContextAwareFuture` handles all the context stuff we need to be careful with.
</blockquote></details>

Done.

leodziki:
Waiting for opinion, @allada

PR #1304, opened 2024-09-01
Fix bug in redis store when zero data stored but data does not exist

### LINKED ISSUE #1286: Redis `.get()` does not return `NotFound` and instead returns empty payload
In a recent redis refactor it appears we no longer check for existence of items before returning an empty query. This is because Redis returns empty value if the item does not exist instead of a nil-like struct.

allada:
Note: This is unlikely to have been observed from the RBE API.

We are supposed to check if an item exists even if it has zero bytes stored in it, but since redis is quite tricky to do this, we were only checking existance if it was a zero byte digest or if the length was zero bytes. In this case we might have a zero-byte proto that is actually valid that does not follow the CAS rules (hash of contents equals key).

This is in prep for a Redis Scheduler.

fixes #1286

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1304)

### DIFF
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -109,12 +109,16 @@ impl RedisStore {
 
     /// Used for testing when determinism is required.
     pub fn new_from_builder_and_parts(
-        builder: Builder,
+        mut builder: Builder,
         pub_sub_channel: Option<String>,
         temp_name_generator_fn: fn() -> String,
         key_prefix: String,
     ) -> Result<Self, Error> {
         let client_pool = builder
+            .set_performance_config(PerformanceConfig {
+                broadcast_channel_capacity: 4096,
+                ..Default::default()
+            })
             .build_pool(CONNECTION_POOL_SIZE)
             .err_tip(|| "while creating redis connection pool")?;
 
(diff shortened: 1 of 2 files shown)


allada (commented):
+@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1304#-:-O5fVDKH7ECXsCUqGM7k:b-bn1n1c)* status: 0 of 1 LGTMs obtained, and 0 of 2 files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / ubuntu-22.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, NativeLink.com Cloud / Remote Cache (Legacy Dockerfile Test), Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale, windows-2022 / stable (waiting on @adam-singer)

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1304)* status: 0 of 1 LGTMs obtained, and 0 of 2 files reviewed, and pending CI: pre-commit-checks (waiting on @adam-singer)

___
*[`nativelink-store/src/redis_store.rs` line 118 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1304#-O5fWgG35oJqdofZmzpo:-O5fWgG35oJqdofZmzpp:b9tn4am) ([raw file](https://github.com/TraceMachina/nativelink/blob/8f26eb7633b8114afad3fbe1caf6797ded3855cc/nativelink-store/src/redis_store.rs#L118)):*
> ```XML
>     ) -> Result<Self, Error> {
>         let client_pool = builder
>             .set_performance_config(PerformanceConfig {
> ```

fyi: Gunna sneak this in here, it is not really related to the other PRs, but in theory could happen.

adam-singer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1304#-:-O5sNSCyD1cYFhRzDBF2:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

PR #1310, opened 2024-09-02
Add KeepAlive updating to ApiWorkerScheduler

allada:
This ensures that actions are periodically updated with with a keep
alive message, which will be heavily used with a distributed scheduling
model.

towards #359

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1310)

### DIFF
--- a/nativelink-scheduler/src/api_worker_scheduler.rs
+++ b/nativelink-scheduler/src/api_worker_scheduler.rs
@@ -23,9 +23,12 @@ use nativelink_metric::{
     group, MetricFieldData, MetricKind, MetricPublishKnownKindData, MetricsComponent,
     RootMetricsComponent,
 };
-use nativelink_util::action_messages::{ActionStage, OperationId, WorkerId};
-use nativelink_util::operation_state_manager::WorkerStateManager;
+use nativelink_util::action_messages::{OperationId, WorkerId};
+use nativelink_util::operation_state_manager::{UpdateOperationType, WorkerStateManager};
 use nativelink_util::platform_properties::PlatformProperties;
+use nativelink_util::spawn;
+use nativelink_util::task::JoinHandleDropGuard;
+use tokio::sync::mpsc::{self, UnboundedSender};
 use tokio::sync::Notify;
 use tonic::async_trait;
 use tracing::{event, Level};
@@ -81,6 +84,8 @@ struct ApiWorkerSchedulerImpl {
     allocation_strategy: WorkerAllocationStrategy,
     /// A channel to notify the matching engine that the worker pool has changed.
     worker_change_notify: Arc<Notify>,
+    /// A channel to notify that an operation is still alive.
+    operation_keep_alive_tx: UnboundedSender<(OperationId, WorkerId)>,
 }
 
 impl ApiWorkerSchedulerImpl {
@@ -103,6 +108,20 @@ impl ApiWorkerSchedulerImpl {
             timestamp
         );
         worker.last_update_timestamp = timestamp;
+        for operation_id in worker.running_action_infos.keys() {
+            if self
+                .operation_keep_alive_tx
+                .send((operation_id.clone(), *worker_id))
+                .is_err()
+            {
+                event!(
+                    Level::ERROR,
+                    ?operation_id,
+                    ?worker_id,
+                    "OperationKeepAliveTx stream closed"
+                );
+            }
+        }
         Ok(())
     }
 
@@ -177,7 +196,7 @@ impl ApiWorkerSchedulerImpl {
         &mut self,
         worker_id: &WorkerId,
         operation_id: &OperationId,
-        action_stage: Result<ActionStage, Error>,
+        update: UpdateOperationType,
     ) -> Result<(), Error> {
         let worker = self.workers.get_mut(worker_id).err_tip(|| {
             format!("Worker {worker_id} does not exist in SimpleScheduler::update_action")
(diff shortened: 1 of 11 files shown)


allada (commented):
+@adam-singer 
(also part of the chain)

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-:-O5nsDZ92nCa_hJOtgyx:b8jdf0u)* status: 0 of 1 LGTMs obtained, and 0 of 18 files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Publish image, Publish nativelink-worker-init, asan / ubuntu-22.04, macos-13, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, vale (waiting on @adam-singer)

adam-singer (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 6 of 6 files at r3, 4 of 4 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, 11 of 11 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-:-O5sZqgZ7ibqCQ48ZtMk:bnfp4nl)* status: 1 of 1 LGTMs obtained, and all files reviewed, and 3 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 271 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sSkxWDNHI95zHky5R:-O5sSkxWDNHI95zHky5S:b-em3pld) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L271)):*
> ```XML
>                 return Err(make_err!(
>                     Code::Internal,
>                     "Failed to update action after {} retries with no error set in MatchingEngineActionStateResult::changed",
> ```

Should include `awaited_action.operation_id()` and maybe `awaited_action.state().stage`
___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 317 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sTnr93mkZ0ozW5LD8:-O5sTnr93mkZ0ozW5LD9:b-if97az) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L317)):*
> ```XML
> 
>     /// Weak reference to self.
>     weak_self: Weak<Self>,
> ```

Can we comment a bit more on why we need a weak reference here over `Arc<Self>`. is the type/borrow checker barking at you for this?  Is this more of an optimization or a common semantic of  passing around `self.clone()`. Looking at the usage I kind of get it, tho I'm assuming its being used for prevention of leaking referencing as returning instances that hold a reference to `Simple SchhedulerStateManager`.  
___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 380 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sX1HNCobH-ANG988Q:-O5sX1HNCobH-ANG988R:bso0wxl) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L380)):*
> ```XML
>                 )
>             })?;
>         let worker_should_update_before = last_worker_updated
> ```

Is there a discrete number of times a timestamp can be updated? I'm guessing because of rescheduling/retry and number of states that exist there should be. I wonder if there is a situation where timestamps get updated but work or state changes don't get applied?

allada (commented):
> Reviewed 18 of 18 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1310)* status: 1 of 1 LGTMs obtained, and all files reviewed, and pending CI: pre-commit-checks, and 1 discussions need to be resolved

___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 271 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sSkxWDNHI95zHky5R:-O62I87f4rpAHgMtAgII:b-896fix) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L271)):*

Should include `awaited_action.operation_id()` and maybe `awaited_action.state().stage`
</blockquote></details>

Done.
___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 317 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sTnr93mkZ0ozW5LD8:-O62I_bD0QblN-49bQrc:b-v346f6) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L317)):*

Can we comment a bit more on why we need a weak reference here over `Arc<Self>`. is the type/borrow checker barking at you for this?  Is this more of an optimization or a common semantic of  passing around `self.clone()`. Looking at the usage I kind of get it, tho I'm assuming its being used for prevention of leaking referencing as returning instances that hold a reference to `Simple SchhedulerStateManager`.
</blockquote></details>

More to reduce risk. I don't like handing out shared pointers to things that are not the "owner" of the object to prevent problems.
Done.
___
*[`nativelink-scheduler/src/simple_scheduler_state_manager.rs` line 380 at r6](https://reviewable.io/reviews/TraceMachina/nativelink/1310#-O5sX1HNCobH-ANG988Q:-O62Iwh13Q63ZXkSWWPy:b-js8is) ([raw file](https://github.com/TraceMachina/nativelink/blob/1765e278f34fb599603f38e657a302abf9d90f81/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L380)):*

Is there a discrete number of times a timestamp can be updated? I'm guessing because of rescheduling/retry and number of states that exist there should be. I wonder if there is a situation where timestamps get updated but work or state changes don't get applied?
</blockquote></details>

The worker can update u32 times before problems happen.

In theory once `assign_operation` is called, the state would be changed to `Queued` preventing this path from being hit.

This code path is also safe to execute quite a lot, it won't deadlock because of the sleeps/timeouts above, but the operation could in theory end up never updating, but that would just trigger this call to try and change the state again, so we should be safe.

PR #1357, opened 2024-09-25
Fix the dates on case studies.

MarcusSorealheis:
# Description

The publication date on one of our case studies was wrong. 

## Type of change

Please delete options that aren't relevant.

- [x] This change requires a documentation update

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1357)

PR #1358, opened 2024-09-25
Remove date from case study page

blakehatch:
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1358)

### DIFF
--- a/web/platform/src/layouts/MarkdownLayout.astro
+++ b/web/platform/src/layouts/MarkdownLayout.astro
@@ -35,7 +35,7 @@ const formattedDate = pubDate.toLocaleDateString('en-US', {
     <LinearGradient text={frontmatter.title} class="text-2xl md:text-xl md:self-center" />
     <div class="text-sm self-start md:self-center md:text-center">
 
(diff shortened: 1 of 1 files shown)


SchahinRohani:
*[`web/platform/src/layouts/MarkdownLayout.astro` line 38 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1358#-O7aezYu0ZSmU3F8LBnC:-O7aezYu0ZSmU3F8LBnD:b-e5uomc) ([raw file](https://github.com/TraceMachina/nativelink/blob/f31015d96f47aef6daf63e405364c38679f29df6/web/platform/src/layouts/MarkdownLayout.astro#L38)):*
> ```Astro
>     <div class="text-sm self-start md:self-center md:text-center">
> 
>     Date: {formattedDate}<br /> Est. read time: {frontmatter.readTime}
> ```

I believe that each post should keep its date, as it provides important context for readers. 
Removing it might cause confusion. Could we instead find a better way to template the date?

MarcusSorealheis:
I think that we should close this one given the pre-existence of #1357, which is also the safer solution.

PR #1377, opened 2024-09-30
Fix clippy::missing-panics-doc

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1377)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -46,7 +46,7 @@ build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
 
 # TODO(aaronmondal): Extend these flags until we can run with clippy::pedantic.
 # TODO(aaronmondal): Fix dependency_on_unit_never_type_fallback.
-build --@rules_rust//:clippy_flags=-Dwarnings,-Adependency_on_unit_never_type_fallback,-Dclippy::uninlined_format_args,-Dclippy::manual_string_new
+build --@rules_rust//:clippy_flags=-Dwarnings,-Adependency_on_unit_never_type_fallback,-Dclippy::uninlined_format_args,-Dclippy::manual_string_new,-Dclippy::missing-panics-doc
 build --@rules_rust//:clippy.toml=//:clippy.toml
 
 test --@rules_rust//:rustfmt.toml=//:.rustfmt.toml
--- a/nativelink-metric/nativelink-metric-macro-derive/src/lib.rs
+++ b/nativelink-metric/nativelink-metric-macro-derive/src/lib.rs
@@ -190,6 +190,9 @@ impl<'a> ToTokens for MetricStruct<'a> {
     }
 }
 
+/// # Panics
+///
+/// This function will panic if the `MetricsComponent` macro is applied to anything other than a struct.
 #[proc_macro_derive(MetricsComponent, attributes(metric))]
 pub fn metrics_component_derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
--- a/nativelink-scheduler/src/worker.rs
+++ b/nativelink-scheduler/src/worker.rs
@@ -119,6 +119,13 @@ fn reduce_platform_properties(
 }
 
 impl Worker {
+    /// Creates a new `Worker` with the specified ID, platform properties, sender, and timestamp.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the system time is earlier than the Unix epoch.
+    /// The panic occurs due to the call to `SystemTime::now().duration_since(UNIX_EPOCH).unwrap()`,
+    /// which will fail if the system clock is misconfigured or set to a time before the Unix epoch.
     pub fn new(
         id: WorkerId,
         platform_properties: PlatformProperties,
--- a/nativelink-service/src/worker_api_server.rs
+++ b/nativelink-service/src/worker_api_server.rs
@@ -50,6 +50,18 @@ pub struct WorkerApiServer {
 }
 
 impl WorkerApiServer {
+    /// Creates a new `WorkerApiServer` with the given configuration and schedulers.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the system time is earlier than the Unix epoch.
+    /// The panic occurs in two situations:
+    /// - When retrieving the system time for scheduler timeouts (`SystemTime::now().duration_since(UNIX_EPOCH)`).
+    /// - During server creation, if the system time is invalid, it will trigger an internal error.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the server could not be created with the given configuration and schedulers.
     pub fn new(
         config: &WorkerApiConfig,
         schedulers: &HashMap<String, Arc<dyn WorkerScheduler>>,
--- a/nativelink-store/src/s3_store.rs
+++ b/nativelink-store/src/s3_store.rs
@@ -260,6 +260,16 @@ where
     I: InstantWrapper,
     NowFn: Fn() -> I + Send + Sync + Unpin + 'static,
 {
+    /// Creates a new S3 store with the specified configuration and time function.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the application name provided to the AWS SDK is invalid.
+    /// The panic occurs with the message `"valid app name"` in the call to `AppName::new()`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there is an issue with creating the S3 client or other components.
     pub async fn new(
         config: &nativelink_config::stores::S3Store,
         now_fn: NowFn,
--- a/nativelink-store/src/shard_store.rs
+++ b/nativelink-store/src/shard_store.rs
@@ -45,6 +45,16 @@ pub struct ShardStore {
 }
 
 impl ShardStore {
+    /// Creates a new `ShardStore` with the given configuration and stores.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the `weights` vector is empty when attempting to modify the last element
+    /// using `unwrap()`. This panic should not occur as long as the `config.stores` is not empty, as validated earlier.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the number of `config.stores` does not match the number of `stores`, or if `config.stores` is empty.
     pub fn new(
         config: &nativelink_config::stores::ShardStore,
         stores: Vec<Store>,
--- a/nativelink-util/src/buf_channel.rs
+++ b/nativelink-util/src/buf_channel.rs
@@ -275,6 +275,14 @@ impl DropCloserReadHalf {
     ///
     /// On error the state of the stream is undefined and the caller should not
     /// attempt to use the stream again.
+    /// # Panics
+    ///
+    /// This function will panic if the `recent_data` is not empty after draining it.
+    /// Specifically, it panics with the message `"Recent_data should be empty"`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `bytes_received` exceeds `max_recent_data_size`.
     pub fn try_reset_stream(&mut self) -> Result<(), Error> {
         if self.bytes_received > self.max_recent_data_size {
             return Err(make_err!(
@@ -313,6 +321,11 @@ impl DropCloserReadHalf {
     }
 
     /// Peek the next set of bytes in the stream without consuming them.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if `queued_data` is empty after attempting to receive data.
+    /// The panic occurs with the message `"Should have data in the queue"`.
     pub async fn peek(&mut self) -> &Result<Bytes, Error> {
         if self.queued_data.is_empty() {
             let chunk = self.recv().await;
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -210,6 +210,11 @@ where
     /// and return the number of items that were processed.
     /// The `handler` function should return `true` to continue processing the next item
     /// or `false` to stop processing.
+    ///
+    /// # Panics
+    ///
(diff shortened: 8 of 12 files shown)


aaronmondal (changes requested):
This seems to have pointed out a few areas where we could potentially improve the code. I'd be a bit hesitant to leave this as "just doc changes". WDYT about trying to remove most of the panics in separate PRs?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-:-O82Aykr54UaSQweNxPb:b-hqvqas)* status: 0 of 1 LGTMs obtained, and all files reviewed, and 16 discussions need to be resolved

___
*[`nativelink-scheduler/src/worker.rs` line 128 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O81wDtv-Hs9gjcNl5LW:-O81wDtv-Hs9gjcNl5LX:b-xl3ma2) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-scheduler/src/worker.rs#L128)):*
> ```XML
>     /// This function will panic if the system time is earlier than the Unix epoch.
>     /// The panic occurs due to the call to `SystemTime::now().duration_since(UNIX_EPOCH).unwrap()`,
>     /// which will fail if the system clock is misconfigured or set to a time before the Unix epoch.
> ```

Don't explain code-level internals.



_Suggestion:_
```XML
which might occur if the system clock is misconfigured or set to a time before the Unix epoch.
```
___
*[`nativelink-service/src/worker_api_server.rs` line 64 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O81yvLm2saGwxBM4q7o:-O81wxnP5Wz4WiWpQGq0:b-qynf6f) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-service/src/worker_api_server.rs#L64)):*
> ```XML
>     /// # Errors
>     ///
>     /// Returns an error if the server could not be created with the given configuration and schedulers.
> ```

This comment is incorrect. The panic only occurs during creation of schedulers, but the one at `self::new_with_now_fn` is an error.

- `expect` panics in the unexpected case
- `map_error` maps to an error :D

You probably want something like this:

```rust
/// Creates a new `WorkerApiServer` with the given configuration and schedulers.
///
/// # Panics
///
/// This function panics if the system time is earlier than the Unix epoch when retrieving the system
/// time for scheduler timeouts (`SystemTime::now().duration_since(UNIX_EPOCH)`).
///
/// # Errors
///
/// Returns an error in the following situations:
/// - If the server could not be created with the given configuration and schedulers.
/// - If the system time is invalid during server creation (i.e., it's before the Unix epoch).
```
___
*[`nativelink-store/src/s3_store.rs` line 268 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O82-WuF2vopRN6eP5Rd:-O82-WuF2vopRN6eP5Re:bhaua1k) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-store/src/s3_store.rs#L268)):*
> ```XML
>     ///
>     /// This function will panic if the application name provided to the AWS SDK is invalid.
>     /// The panic occurs with the message `"valid app name"` in the call to `AppName::new()`.
> ```

Omit explanations at the code-level. Also, since "nativelink" is hardcoded, it should be mentioned that hitting this panic should be impossible



_Suggestion:_
```XML
    /// This function will panic if the application name "nativelink" isn't accepted by the AWS SDK which should never happen.
```
___
*[`nativelink-store/src/shard_store.rs` line 57 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O8233Za4bywEx432Jjt:-O8233Zb07pqI8TnCT8O:b9pefgv) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-store/src/shard_store.rs#L57)):*
> ```XML
>     /// # Errors
>     ///
>     /// Returns an error if the number of `config.stores` does not match the number of `stores`, or if `config.stores` is empty.
> ```

For multiple error cases use a list to clarify how many cases could occur:

```rust
/// Returns an error in the following cases:
/// - If the number of stores in `config.stores` does not match the length of the `stores` vector.
/// - If `config.stores` is empty.
```

The panic in this function is so impossible to hit that we should probably try to find another way to implement this without unwrap.
___
*[`nativelink-store/src/shard_store.rs` line 84 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O824Zxp-HEDCjRniovO:-O824Zxp-HEDCjRniovP:bl9xbk9) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-store/src/shard_store.rs#L84)):*
> ```XML
>                 *state += weight;
>                 Some(*state)
>             })
> ```

I believe constructive approach should let us remove this panic:

```rust
let weights: Vec<u32> = {
    let mut cumulative_weight = 0u32;
    let len = config.stores.len();

    config
        .stores
        .iter()
        .enumerate()
        .map(|(i, shard_config)| {
            let weight = (u32::MAX as u64 * shard_config.weight.unwrap_or(1) as u64 / total_weight) as u32;
            cumulative_weight = if i == len - 1 {
                u32::MAX
            } else {
                cumulative_weight.saturating_add(weight)
            };
            cumulative_weight
        })
        .collect()
};

```
___
*[`nativelink-util/src/buf_channel.rs` line 280 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O825a8S9VKzLQICbMJp:-O825a8S9VKzLQICbMJq:b9t63zz) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/buf_channel.rs#L280)):*
> ```XML
>     /// # Panics
>     ///
>     /// This function will panic if the `recent_data` is not empty after draining it.
> ```

```rust
/// This function will panic if internal invariants are violated, such as:
///
/// - If `recent_data` is not empty after draining. This indicates a bug in the implementation.
/// - If the sum of the sizes of `recent_data` does not equal `bytes_received`. This also indicates a bug.
```
___
*[`nativelink-util/src/buf_channel.rs` line 281 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O825LM9Dp3fYvH0T2S0:-O825LM9Dp3fYvH0T2S1:btkx0g5) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/buf_channel.rs#L281)):*
> ```XML
>     ///
>     /// This function will panic if the `recent_data` is not empty after draining it.
>     /// Specifically, it panics with the message `"Recent_data should be empty"`.
> ```

No need reiterate the string.
___
*[`nativelink-util/src/buf_channel.rs` line 286 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O826elg8NnSLNTA23KD:-O826elg8NnSLNTA23KE:bbltmja) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/buf_channel.rs#L286)):*
> ```XML
>     ///
>     /// Returns an error if `bytes_received` exceeds `max_recent_data_size`.
>     pub fn try_reset_stream(&mut self) -> Result<(), Error> {
> ```

I'm actually not sure whether we should leave this function like this. Asserts in non-test code are not ideal and this function could potentially be in the critical path. This is just draft-level, but maybe something like this could work to handle this without panics:

```rust
pub fn try_reset_stream(&mut self) -> Result<(), Error> {
    if self.bytes_received > self.max_recent_data_size {
        return Err(make_err!(
            Code::Internal,
            "Cannot reset stream, max_recent_data_size exceeded"
        ));
    }

    let mut data_sum = 0u64;
    let mut chunks = Vec::with_capacity(self.recent_data.len());

    for chunk in self.recent_data.iter().rev() {
        data_sum += chunk.len() as u64;
        chunks.push(chunk.clone());
    }

    if data_sum != self.bytes_received {
        return Err(make_err!(
            Code::Internal,
            "Sum of recent_data bytes does not equal bytes_received"
        ));
    }

    self.recent_data.clear();
    for chunk in chunks {
        self.queued_data.push_front(Ok(chunk));
    }
    self.bytes_received = 0;
    Ok(())
}
```

cc @allada
___
*[`nativelink-util/src/buf_channel.rs` line 328 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O8285y_Aqh3gavhzgQE:-O8285y_Aqh3gavhzgQF:undefined) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/buf_channel.rs#L328)):*
> ```XML
>     ///
>     /// This function will panic if `queued_data` is empty after attempting to receive data.
>     /// The panic occurs with the message `"Should have data in the queue"`.
> ```


___
*[`nativelink-util/src/evicting_map.rs` line 217 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O828BAz8sqkhah9lmA5:-O828BAz8sqkhah9lmA6:b-ixdp49) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/evicting_map.rs#L217)):*
> ```XML
>     ///
>     /// This function will panic if the `btree` index is not properly initialized and is `None` after attempting to rebuild it.
>     /// The panic occurs due to calling `unwrap()` on a `None` value.
> ```

Hmm unwrapping here seems like a bad idea as well. Maybe we can find a better way? However, we probably need benchmarking infra if we want to change this.
___
*[`nativelink-util/src/fastcdc.rs` line 59 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O829PV3-EkCLhBqni1F:-O829PV3-EkCLhBqni1G:undefined) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/fastcdc.rs#L59)):*
> ```XML
>     ///
>     /// This function will panic if:
>     /// - `min_size` is not less than `avg_size`, in which case it panics with the message `"Expected {min_size} < {avg_size}"`.
> ```


___
*[`nativelink-util/src/fastcdc.rs` line 60 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O829Ruw4bKHEoMZSVfQ:-O829Ruw4bKHEoMZSVfR:undefined) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/fastcdc.rs#L60)):*
> ```XML
>     /// This function will panic if:
>     /// - `min_size` is not less than `avg_size`, in which case it panics with the message `"Expected {min_size} < {avg_size}"`.
>     /// - `avg_size` is not less than `max_size`, in which case it panics with the message `"Expected {avg_size} < {max_size}"`.
> ```


___
*[`nativelink-util/src/metrics_utils.rs` line 331 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O829eexDzKYVWToD8t6:-O829eexDzKYVWToD8t7:b-8a05gs) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/metrics_utils.rs#L331)):*
> ```XML
>     /// # Panics
>     ///
>     /// This function will panic if the system time is earlier than the Unix epoch.
> ```

As above
___
*[`nativelink-util/src/origin_context.rs` line 327 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O829zgzCWAchOjd0FBy:-O829zgzCWAchOjd0FBz:undefined) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-util/src/origin_context.rs#L327)):*
> ```XML
>     ///
>     /// This function will panic if the `ActiveOriginContext` is not set (i.e., it returns `None`).
>     /// The panic occurs with the message `"OriginContext must be set"`.
> ```





_Suggestion:_
```XML
    /// This function will panic if the `ActiveOriginContext` is not set.
```
___
*[`nativelink-worker/src/worker_utils.rs` line 34 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O82AC501576Mq3RnCAe:-O82AC501576Mq3RnCAf:undefined) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-worker/src/worker_utils.rs#L34)):*
> ```XML
> ///
> /// This function will panic if the process fails and the exit status code cannot be retrieved
> /// (i.e., `process_output.status.code().unwrap()` returns `None`). This might happen if the process was terminated by a signal and no exit code was available.
> ```


___
*[`nativelink-worker/src/worker_utils.rs` line 41 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1377#-O82AZXLCIwravL6bP-V:-O82AZXLCIwravL6bP-W:b2ni8l0) ([raw file](https://github.com/TraceMachina/nativelink/blob/198d940a7736984f45acc76d83279ac1e0b08dc0/nativelink-worker/src/worker_utils.rs#L41)):*
> ```XML
> /// - A worker property query command cannot be parsed.
> /// - A worker property query command fails to execute.
> /// - The process output cannot be converted to UTF-8.
> ```

nit



_Suggestion:_
```XML
/// Returns an error if:
/// - A worker property query command can't be parsed.
/// - A worker property query command fails to run.
/// - The process output can't be converted to UTF-8.
```

PR #1383, opened 2024-10-02
Fix S3Store retry might cause poisoned data

allada:
If using S3Store + VerifyStore is in use, S3 store could receive data that VerifyStore deemed invalid due to retry logic.

This only effects S3Store + VerifyStore due to the way AwsS3Sdk crate works, we need to hold recent data in the BufChannel, in the event VerifyStore got an invalid hash, the retry logic in S3 would trigger, but instead of being seen as "invalid" it would actually have stored the sent data and sent it, causing S3 to to still receive the invalid data. This PR causes BuffChannel logic to set a flag making the next read in S3 store always trigger the error condition.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1383)

### DIFF
--- a/nativelink-store/src/s3_store.rs
+++ b/nativelink-store/src/s3_store.rs
@@ -427,6 +427,9 @@ where
         // Note(allada) If the upload size is not known, we go down the multipart upload path.
         // This is not very efficient, but it greatly reduces the complexity of the code.
         if max_size < MIN_MULTIPART_SIZE && matches!(upload_size, UploadSizeInfo::ExactSize(_)) {
+            let UploadSizeInfo::ExactSize(sz) = upload_size else {
+                unreachable!("upload_size must be UploadSizeInfo::ExactSize here");
+            };
             reader.set_max_recent_data_size(
                 u64::try_from(self.max_retry_buffer_per_request)
                     .err_tip(|| "Could not convert max_retry_buffer_per_request to u64")?,
@@ -434,9 +437,6 @@ where
             return self
                 .retrier
                 .retry(unfold(reader, move |mut reader| async move {
-                    let UploadSizeInfo::ExactSize(sz) = upload_size else {
-                        unreachable!("upload_size must be UploadSizeInfo::ExactSize here");
-                    };
                     // We need to make a new pair here because the aws sdk does not give us
                     // back the body after we send it in order to retry.
                     let (mut tx, rx) = make_buf_channel_pair();
--- a/nativelink-util/src/buf_channel.rs
+++ b/nativelink-util/src/buf_channel.rs
@@ -49,6 +49,7 @@ pub fn make_buf_channel_pair() -> (DropCloserWriteHalf, DropCloserReadHalf) {
         DropCloserReadHalf {
             rx,
             queued_data: VecDeque::new(),
+            last_err: None,
             eof_sent,
             bytes_received: 0,
             recent_data: Vec::new(),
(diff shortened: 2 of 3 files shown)


allada (commented):
+@adam-singer 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1383#-:-O8ApJJ_68BeiplQ2-VC:b-bn1n1c)* status: 0 of 1 LGTMs obtained, and 0 of 3 files reviewed, and pending CI: pre-commit-checks (waiting on @adam-singer)

___
*[`nativelink-store/src/s3_store.rs` line 437 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1383#-O8AoaiBCfIPydc_kOcI:-O8AoaiBCfIPydc_kOcJ:b-4zsti0) ([raw file](https://github.com/TraceMachina/nativelink/blob/b5a4d928a817a7bdf7466cf01253fb1d92ee880f/nativelink-store/src/s3_store.rs#L437)):*
> ```XML
>                 .retrier
>                 .retry(unfold(reader, move |mut reader| async move {
>                     let UploadSizeInfo::ExactSize(sz) = upload_size else {
> ```

fyi: Just seemed weird to do this in retry. Not related to this PR though.

adam-singer (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1383#-:-O8AwiZJ6CSVXUzV7KkQ:bnfp4nl)* status: 1 of 1 LGTMs obtained, and all files reviewed, and pending CI: Installation / macos-13, Remote / large-ubuntu-22.04, and 1 discussions need to be resolved

___
*[`nativelink-util/src/buf_channel.rs` line 222 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1383#-O8AwDCK9YRnVo0pPzja:-O8AwDCK9YRnVo0pPzjb:b-8fbjro) ([raw file](https://github.com/TraceMachina/nativelink/blob/b71f04fc732af87560b859b825d3da54ccc7ec10/nativelink-util/src/buf_channel.rs#L222)):*
> ```XML
>                 self.queued_data.clear();
>                 self.recent_data.clear();
>                 self.bytes_received = 0;
> ```

Should this be part of some sort of "reset stream" or "reset channel" type function? Would there be other cases where we need to reset/clear? 



_Code quote:_
```XML
                self.queued_data.clear();
                self.recent_data.clear();
                self.bytes_received = 0;
```

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1383)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

___
*[`nativelink-util/src/buf_channel.rs` line 222 at r2](https://reviewable.io/reviews/TraceMachina/nativelink/1383#-O8AwDCK9YRnVo0pPzja:-O8Ct_InDwcOiCYbbfq5:br2gudm) ([raw file](https://github.com/TraceMachina/nativelink/blob/b71f04fc732af87560b859b825d3da54ccc7ec10/nativelink-util/src/buf_channel.rs#L222)):*

Should this be part of some sort of "reset stream" or "reset channel" type function? Would there be other cases where we need to reset/clear?
</blockquote></details>

I originally had it this way, but then realized the only way an error can happen is if it drops early. We don't expose a way for an error to be sent from the sender side.

PR #1396, opened 2024-10-07
Fix clippy::stable_sort_primitive

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1396)


adam-singer (commented):
Minor concern here based on reading https://rust-lang.github.io/rfcs/1884-unstable-sort.html#alternatives , if we remove clippy i and this pattern of sort is observed we might start using it in cases where its not primitive data. Is this lint really worth it?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1396#-:-O8biepH7rMTQkHavUG0:b-92mahe)* status: 0 of 1 LGTMs obtained, and 0 of 2 files reviewed

SchahinRohani (commented):
Thank you for your patience @adam-singer! This change is part of addressing various warnings as we work towards enabling `clippy::pedantic` to raise the overall code quality across the project.

Regarding the sorting algorithm, if we decide to use it in other cases or prefer a different approach for non-primitive data types, I think we can simply discuss and address it as needed, right?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1396#-:-O9RX-qA1w6EfL3Fhnty:b-8681bw)* status: 0 of 1 LGTMs obtained, and 0 of 2 files reviewed

aaronmondal (approved):
I'd argue that we should add a TODO to benchmark this case if performance between stable and unstable sort matters to us here, but I'd say that we should use unstable sort as our "default" as it's only a performance concern, not a correctness concern. This would be in line with clippy and is also the behavior in C++.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1396#-:-OAssbACCCa01oPhBzBR:b-6kfso1)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

SchahinRohani (commented):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1396)* status: 1 of 1 LGTMs obtained, and all files reviewed, and pending CI: Remote / large-ubuntu-22.04

PR #1417, opened 2024-10-17
Fix clippy::used_underscore_binding

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1417)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -45,7 +45,7 @@ build --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
 build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
 
 # TODO(aaronmondal): Extend these flags until we can run with clippy::pedantic.
-build --@rules_rust//:clippy_flags=-Dwarnings,-Dclippy::uninlined_format_args,-Dclippy::manual_string_new,-Dclippy::manual_let_else,-Dclippy::single_match_else,-Dclippy::redundant_closure_for_method_calls,-Dclippy::semicolon_if_nothing_returned,-Dclippy::unreadable_literal,-Dclippy::range_plus_one,-Dclippy::inconsistent_struct_constructor,-Dclippy::match_wildcard_for_single_variants,-Dclippy::implicit_clone
+build --@rules_rust//:clippy_flags=-Dwarnings,-Dclippy::uninlined_format_args,-Dclippy::manual_string_new,-Dclippy::manual_let_else,-Dclippy::single_match_else,-Dclippy::redundant_closure_for_method_calls,-Dclippy::semicolon_if_nothing_returned,-Dclippy::unreadable_literal,-Dclippy::range_plus_one,-Dclippy::inconsistent_struct_constructor,-Dclippy::match_wildcard_for_single_variants,-Dclippy::implicit_clone,-Dclippy::used_underscore_binding
 build --@rules_rust//:clippy.toml=//:clippy.toml
 
 test --@rules_rust//:rustfmt.toml=//:.rustfmt.toml
--- a/nativelink-store/src/s3_store.rs
+++ b/nativelink-store/src/s3_store.rs
@@ -49,7 +49,6 @@ use nativelink_metric::MetricsComponent;
 use nativelink_util::buf_channel::{
     make_buf_channel_pair, DropCloserReadHalf, DropCloserWriteHalf,
 };
-use nativelink_util::fs;
 use nativelink_util::health_utils::{HealthStatus, HealthStatusIndicator};
 use nativelink_util::instant_wrapper::InstantWrapper;
 use nativelink_util::retry::{Retrier, RetryResult};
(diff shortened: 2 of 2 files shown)


aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1417)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, and 1 discussions need to be resolved

___
*[`nativelink-store/src/s3_store.rs` line 175 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1417#-O9VIMxWCs3Dj_qa1I5U:-O9VIMxXC3p-S-wcS827:b-vhxp8x) ([raw file](https://github.com/TraceMachina/nativelink/blob/aba982280353c963ad9632a423775b51f3b052a0/nativelink-store/src/s3_store.rs#L175)):*
> ```XML
>     ) -> Result<ConnectionWithPermit<MaybeHttpsStream<TcpStream>>, Error> {
>         let retry_stream_fn = unfold(self.connector.clone(), move |mut connector| async move {
>             match connector.call(req.clone()).await {
> ```

IIUC the issue here is that the semaphore isn't shared? Is it really correct to remove this then? If it is it probably shouldn't be called `ConnectionWithPermit` anymore.

SchahinRohani (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1417)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, and 1 discussions need to be resolved

___
*[`nativelink-store/src/s3_store.rs` line 175 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1417#-O9VIMxWCs3Dj_qa1I5U:-O9WTw-UAMHcyo4hJUbA:b-tto513) ([raw file](https://github.com/TraceMachina/nativelink/blob/aba982280353c963ad9632a423775b51f3b052a0/nativelink-store/src/s3_store.rs#L175)):*

IIUC the issue here is that the semaphore isn't shared? Is it really correct to remove this then? If it is it probably shouldn't be called `ConnectionWithPermit` anymore.
</blockquote></details>

As i understand, the _permit wasnt used at all, thats why i removed it. 

@allada do you know how to handle this?

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1417)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, and 1 discussions need to be resolved

___
*[`nativelink-store/src/s3_store.rs` line 175 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1417#-O9VIMxWCs3Dj_qa1I5U:-O9uuKRA6EPcQALcrwBz:b-p41dyz) ([raw file](https://github.com/TraceMachina/nativelink/blob/aba982280353c963ad9632a423775b51f3b052a0/nativelink-store/src/s3_store.rs#L175)):*
<details><summary><i>Previously, SchahinRohani (Schahin) wrote…</i></summary><blockquote>

As i understand, the _permit wasnt used at all, thats why i removed it. 

@allada do you know how to handle this?
</blockquote></details>

This is actually used. `SemaphorePermit` is used for it's `drop` function. The semaphore is telling the semaphore counter that there is still a lease out on a file descriptor. When the semaphore is dropped it lets the counter know there's one available again waking up one of the waiters (if there are any).


So, in this case it'll need to have an exception to this linter.

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1417)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, and 1 discussions need to be resolved

___
*[`nativelink-store/src/s3_store.rs` line 175 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1417#-O9VIMxWCs3Dj_qa1I5U:-OAocPyL7TKA2i08ElPE:b-m9blg4) ([raw file](https://github.com/TraceMachina/nativelink/blob/aba982280353c963ad9632a423775b51f3b052a0/nativelink-store/src/s3_store.rs#L175)):*

This is actually used. `SemaphorePermit` is used for it's `drop` function. The semaphore is telling the semaphore counter that there is still a lease out on a file descriptor. When the semaphore is dropped it lets the counter know there's one available again waking up one of the waiters (if there are any).


So, in this case it'll need to have an exception to this linter.
</blockquote></details>

Looks like a bug. If we remove the underscore from the `_permit` we get dead-code warnings from rustc. So in it's current state this really doesn't do anything.

PR #1418, opened 2024-10-17
Fix clippy::from_iter_instead_of_collect

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended


aaronmondal (changes requested):
Looks like we should allow this case as upstream want's to only trigger this if the turbofish syntax isn't needed https://github.com/rust-lang/rust-clippy/issues/7213

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1418#-:-O9VHbsNCoHe4f8dIzU4:b-9uym2g)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04

PR #1423, opened 2024-10-21
Fix clippy::cast_possible_truncation

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

### DIFF
--- a/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
+++ b/nativelink-scheduler/src/awaited_action_db/awaited_action.rs
@@ -199,13 +199,20 @@ impl AwaitedActionSortKey {
     #[rustfmt::skip]
     const fn new(priority: i32, insert_timestamp: u32) -> Self {
         // Shift `new_priority` so [`i32::MIN`] is represented by zero.
-        // This makes it so any nagative values are positive, but
+        // This makes it so any negative values are positive, but
         // maintains ordering.
         const MIN_I32: i64 = (i32::MIN as i64).abs();
-        let priority = ((priority as i64 + MIN_I32) as u32).to_be_bytes();
+        let adjusted_priority = priority as i64 + MIN_I32;
 
-        // Invert our timestamp so the larger the timestamp the lower the number.
-        // This makes timestamp descending order instead of ascending.
+        // Manually check if `adjusted_priority` is within the bounds of `u32`
+        #[allow(clippy::cast_possible_truncation)]
+        let priority = if adjusted_priority >= u32::MIN as i64 && adjusted_priority <= u32::MAX as i64 {
+            (adjusted_priority as u32).to_be_bytes()
+        } else {
+            panic!("Failed to convert priority to u32");
+        };
+
+        // Invert the timestamp to make it descending order instead of ascending.
         let timestamp = (insert_timestamp ^ u32::MAX).to_be_bytes();
 
         AwaitedActionSortKey(u64::from_be_bytes([
@@ -215,10 +222,13 @@ impl AwaitedActionSortKey {
     }
 
     fn new_with_unique_key(priority: i32, insert_timestamp: &SystemTime) -> Self {
-        let timestamp = insert_timestamp
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs() as u32;
+        let timestamp = u32::try_from(
+            insert_timestamp
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs(),
+        )
+        .expect("Failed to convert");
         Self::new(priority, timestamp)
     }
 
--- a/nativelink-util/src/retry.rs
+++ b/nativelink-util/src/retry.rs
@@ -120,6 +120,7 @@ impl Retrier {
         }
     }
 
+    #[allow(clippy::cast_possible_truncation)]
     fn get_retry_config(&self) -> impl Iterator<Item = Duration> + '_ {
         ExponentialBackoff::new(Duration::from_millis(self.config.delay as u64))
             .map(|d| (self.jitter_fn)(d))
--- a/.bazelrc
+++ b/.bazelrc
@@ -45,7 +45,7 @@ build --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
 build --aspects=@rules_rust//rust:defs.bzl%rust_clippy_aspect
 
 # TODO(aaronmondal): Extend these flags until we can run with clippy::pedantic.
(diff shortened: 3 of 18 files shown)


nativelink-scheduler/src/awaited_action_db/awaited_action.rs
```diff
@@ -239,29 +239,37 @@ impl MetricsComponent for AwaitedActionSortKey {
...
 impl AwaitedActionSortKey {
-    #[rustfmt::skip]
     const fn new(priority: i32, insert_timestamp: u32) -> Self {
         // Shift `new_priority` so [`i32::MIN`] is represented by zero.
         // This makes it so any negative values are positive, but
         // maintains ordering.
-        const MIN_I32: i64 = (i32::MIN as i64).abs();
-        let priority = ((priority as i64 + MIN_I32) as u32).to_be_bytes();
+        let priority_u32 = 0x8000_0000_u32.wrapping_add_signed(priority);
+        let priority = priority_u32.to_be_bytes();
```
palfrey, line 246: This seems odd to me? What's the `0x8000_0000_u32` magic value there?

nativelink-store/src/gcs_client/client.rs
```diff
@@ -532,7 +535,11 @@ impl GcsOperations for GcsClient {
 
                     let mut rng = rand::rng();
                     let jitter_factor = rng.random::<f64>().mul_add(0.4, 0.8);
-                    retry_delay = (retry_delay as f64 * jitter_factor) as u64;
+                    retry_delay =
+                        Duration::from_secs_f64((retry_delay as f64 * jitter_factor) / 1000.0)
+                            .as_millis()
+                            .try_into()
+                            .unwrap_or(u64::MAX);
```
palfrey: Why the two conversions? Can't you just do `try_into` for u64?

nativelink-store/src/redis_store.rs
```diff
@@ -177,18 +177,22 @@ impl RedisStore {
                 spec.retry.jitter = DEFAULT_RETRY_JITTER;
             }
 
+            let to_ms = |secs: f32| -> u32 {
+                Duration::from_secs_f32(secs)
+                    .as_millis()
```
palfrey, line 182: Again the double conversion here?

nativelink-util/src/retry.rs
```diff
@@ -130,9 +130,9 @@ impl Retrier {
     }
 
     fn get_retry_config(&self) -> impl Iterator<Item = Duration> + '_ {
-        ExponentialBackoff::new(Duration::from_millis(self.config.delay as u64))
+        ExponentialBackoff::new(Duration::from_secs_f32(self.config.delay))
             .map(|d| (self.jitter_fn)(d))
-            .take(self.config.max_retries) // Remember this is number of retries, so will run max_retries + 1.
+            .take(self.config.max_retries)
```
palfrey: Can you readd the comment please!

nativelink-util/src/retry.rs
```diff
@@ -130,9 +130,9 @@ impl Retrier {
     }
 
     fn get_retry_config(&self) -> impl Iterator<Item = Duration> + '_ {
-        ExponentialBackoff::new(Duration::from_millis(self.config.delay as u64))
+        ExponentialBackoff::new(Duration::from_secs_f32(self.config.delay))
```
palfrey, line 133: Nice catch!

palfrey (changes requested):
A few items, but otherwise good, thanks!

nativelink-scheduler/src/awaited_action_db/awaited_action.rs
SchahinRohani: Good catch! `0x8000_0000` is 2^31 (or 2,147,483,648 in decimal). We're shifting the signed i32 range `[i32::MIN, i32::MAX]` to the unsigned u32 range `[0, u32::MAX]` to preserve ordering when we convert to bytes for sorting.

I can make this clearer by either:
1. Using a named constant: `const PRIORITY_OFFSET: u32 = 2_147_483_648;`
2. Making it self-documenting: `i32::MIN.unsigned_abs()`
3. Adding a comment explaining the shift

nativelink-store/src/gcs_client/client.rs
SchahinRohani: Good question! The issue is that `Duration::as_millis()` returns `u128`, not `u64`, so we need the `try_into()` to convert `u128 → u64`.

We can't do a direct cast from `f64 → u64` because that triggers the `clippy::cast_possible_truncation` lint. Using the `Duration` conversion is a workaround - it lets us go `f64 → Duration → u128 → u64` without any direct float-to-int casts that Clippy complains about.

If you have a cleaner way to do this without triggering the lint, I'm all ears! 😅

palfrey: Yeah, that makes sense. I've just played around a bit and not getting much better answers. For that first line, what do you think of `Duration::from_millis(retry_delay).mul_f64(jitter_factor)` instead? Removes the whole `/ 1000` which feels odd to me.

nativelink-scheduler/src/awaited_action_db/awaited_action.rs
palfrey: If you add a comment, I think that's enough. It makes sense with your explanation :)

nativelink-store/src/redis_store.rs
palfrey: This now makes more sense, and the clearer use of the `from_secs_f32` specific function LGTM

nativelink-store/src/gcs_client/client.rs
SchahinRohani: I like that more. Better than the `/ 1000`  which also feels odd to me 😄

nativelink-scheduler/src/awaited_action_db/awaited_action.rs
SchahinRohani: Changed it to a more idiomatic way.

palfrey (approved):
:lgtm:

> @palfrey reviewed 28 of 28 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tracemachina/nativelink/1423#-:-ObhrHlT4gw5LHvJXK2C:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

palfrey:
> Rebased it and resolved the conflicts @palfrey

Thanks!

PR #1436, opened 2024-10-24
Add nativelink version in BUILD.bazel

SchahinRohani:
# Description

This PR adds the nativelink version for bazel builds in `nativelink -V`

## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1436)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -13,6 +13,9 @@ rust_binary(
     srcs = [
         "src/bin/nativelink.rs",
     ],
+    rustc_env = {
+        "CARGO_PKG_VERSION": "0.5.3",
+    },
     deps = [
         "//nativelink-config",
         "//nativelink-error",


aaronmondal (changes requested):
See reasoning in #1258  

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1436#-:-O9zAg8t2GQP8PFehYFA:b-862r9c)* status: 0 of 1 LGTMs obtained, and 0 of 1 files reviewed

SchahinRohani:
How about setting it to something like "dev-version" instead of letting it blank which results in "0.0.0"?

allada (commented):
I'm not a huge fan of doing it this way, I'd rather make a genrule that reads the `Cargo.toml`, populates the env, then injects it.

example: https://github.com/bazelbuild/rules_rust/blob/c2014b6e3cb2530cc9a5eb2cf652b8cd30544486/examples/version_stamping/BUILD.bazel

I also don't see a lot of value from doing this right now.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1436#-:-O9zGBC2DikxG1fFeevs:buu7de1)* status: 0 of 1 LGTMs obtained, and 0 of 1 files reviewed

SchahinRohani:
The reason I opened this PR is that we should set the variable instead of leaving it blank. By fixing a generic "dev" Bazel version, we do not need to maintain this setting until we decide on a more suitable solution in the future.

The discussion regarding the implementation of an ideal solution, such as an environment generate rule through Bazel or finding another maintainable approach, should be out of scope for this PR. This PR focuses on setting an appropriate variable for the dev CLI version and serves as a minor improvement for maintenance and accessibility for developers who might look into it. And by having a lot of minor improvements, the overall code quality just gets higher.

Let me know what you think @aaronmondal @allada

aaronmondal:
I agree with @allada that we should use some sort of dedicated mechanism here, but since this is such a nuanced topic I'd prefer to keep the existing behavior for now as it seems undesirable to change this behavior temporarily only to change it again once we figure out a better scheme.

PR #1447, opened 2024-10-31
Create .valeignore

MarcusSorealheis:
we need to leave the blog's voice unemcumbered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1447)

### DIFF
--- a/.valeignore
+++ b/.valeignore
@@ -1,0 +1,1 @@
+web/platform/src/content/posts/


aaronmondal (changes requested):
You can disable vale already directly in the blogs via around the scopes that trigger vale, e.g via `<!-- vale Style.Rule = NO -->` and `<!-- vale Style.Rule = YES -->`.

See: https://vale.sh/docs/topics/config/#markdown-amp-html

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1447#-:-OAZrybM3vz9NzgBfsaA:b-1dkm41)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Coverage, Installation / macos-13, Installation / macos-14, Remote / large-ubuntu-22.04, macos-13

MarcusSorealheis:
I don't think Vale is for blogs.

aaronmondal:
> I don't think Vale is for blogs.

Some style issues are certainly not good for blogs, but we largely disabled those globally already. The errors that trigger are mostly to correct spelling mistakes. Almost all cases fall under one of these categories:

- Contractions like "that is" -> "that's"
- Spelling issues for programs where the warnings should be read as "sometool" -> "`sometool`"
- Words that are correct but vale isn't aware of (e.g CUDA, LLVM, etc) which we can add to https://github.com/TraceMachina/nativelink/blob/main/.github/styles/config/vocabularies/TraceMachina/accept.txt

If we add global vale ignores we run the risk of introducing typos in these documents, which I'd argue would be worse than the slight inconvenience sometimes caused by the warnings.

Also note that the language that we "enforce" here is harder to generate via LLMs. In other words, vale-linted documents have a higher chance to look "human" to search engines.

SchahinRohani:
Every writer has his own writing style. I am not sure if it's good to force those writers with vale to change their style.
A global .valeignore could be a good solution to this...

MarcusSorealheis:
Vale for blogs is not a slight inconvenience and contractions are a feature
of literature, not a bug. A blog is somewhere in the middle of reference
and literature.

aaronmondal:
After looking into this more I think I found a way that could make everyone happy here.

This file is the one that configures styles:

https://github.com/TraceMachina/nativelink/blob/8a3869cdddb9202de26bb0ab272519ace73c98f6/.vale.ini#L1-L27

Note the `[*.{md,mdx}]` section. I believe if we add a section like `[path/to/blogposts/*.{md,mdx}]` we can disable things like contractions "globally" for blogs but keep the typo checking intact. Would that work? I totally get that we shouldn't enforce strict style in freeform text, just trying to prevent things like actual typos and to ensure that if we use "unusual" words that we add them to the `accept.txt` file and/or use backticks correctly where appropriate.

PR #1461, opened 2024-11-05
Implement native OpenTelemetry infrastructure

aaronmondal:
This commit adds the OTLP exports to Nativelink and extends the `nativelink` deployments in the operator with OpenTelemetryCollector sidecars. The exposed traces, metrics and logs are published through Kafka to NATS Jetstream.


aaronmondal:
cc @allada @SchahinRohani You might want to play around with this while it's in preview.

@allada One thing that we'll need to figure out is where to put the `OtlpServer` in the nativelink-config. At the moment it requires a dummy listener which might not play well with workers.

@SchahinRohani You might want to look into OTLP, Kafka topics and NATS Jetstream. This initial implementation doesn't add structure, but at least it provides a central point to aggregate the logs, traces and metrics of nativelink deployments.

I'll polish this a bit, but for now we have the following (assuming the pod for the nativelink-cas is e.g. kubectl port-forward nativelink-cas-ff6544bb8-v4w86)

```
# Telemetry (CPU usage etc)
kubectl port-forward nativelink-cas-ff6544bb8-v4w86 8888
curl localhost:8888/metrics

# The info that was previously in the experimental_prometheus endpoint
kubectl port-forward nativelink-cas-ff6544bb8-v4w86 8888
curl localhost:8889/metrics

# The aggregated stream across cas, scheduler and worker deployments
kubectl exec -n nats-system nats-box-5d4d987f5b-thbfs -- nats stream info TELEMETRY
```

PR #1462, opened 2024-11-05
add basic visitor tracking.

MarcusSorealheis:
# Description

Adds basic visitor tracking. It does lead to a weird server side React warning about `useEffect` that I could use a tip on. I'm not sure if it was there already. First commit to the rewritten website.


MarcusSorealheis:
I've found a bug in a closure, one moment.

SchahinRohani (changes requested):
We are mainly using the [Qwik Framework](https://qwik.dev/). We have also enabled react, but we only have one react component on the website.
Qwik might look similar to react with the tsx syntax, but it has another approach than react [(resumability vs hydration)](https://qwik.dev/docs/concepts/resumable/).
So you need to be aware of....

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1462#-:-OAz6X-8BFsHdLnjJE0N:b7fc8xj)* status: 0 of 1 LGTMs obtained, and all files reviewed, and 2 discussions need to be resolved

___
*[`web/platform/src/components/qwik/components/visitors.tsx` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1462#-OAz76eV7FSYnsVMENas:-OAz76eW08mAooyqOaau:bfgzy5) ([raw file](https://github.com/TraceMachina/nativelink/blob/3d6d280b8f71facc7249ad13e4b4b26a1b2894af/web/platform/src/components/qwik/components/visitors.tsx#L42)):*
> ```XML
>         }
>       }
>       ce(`https://sc.lfeeder.com/lftracker_v1_${ss}.js`);
> ```

For thirdparty scripts we should use [Partytown](https://partytown.builder.io/), since it won't slow down the website performance.
It will use web worker under the hood, so the main thread doesn't get blocked.
For that we don't need a qwik component, since we also use astro as the meta framework for routing.

To integrate partytown in astro you'll need to add the dependency with `bun i -D @astrojs/partytown`.

Then you need to import it inthe `astro.config.ts`:

```typescript
...
import tailwindcss from "@tailwindcss/vite";                                 
import partytown from "@astrojs/partytown";     //  <--- import the dependency                

...
 integrations: [                                        
   ...
   partytown()  // <-- add the dependency as an integration
 ],
...   
```
___
*[`web/platform/src/layouts/Layout.astro` line 26 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1462#-OAz81o4AP5eSMK1fQ60:-OAz81o4AP5eSMK1fQ61:blnxa4z) ([raw file](https://github.com/TraceMachina/nativelink/blob/3d6d280b8f71facc7249ad13e4b4b26a1b2894af/web/platform/src/layouts/Layout.astro#L26)):*
> ```Astro
> 		<title>{title}</title>
>     	<ViewTransitions />
> 		<Visitors />
> ```

Let's create a directory  for third-party scripts something like `src/thirdParty`, and inside you can create a`lfeeder.astro` file.

An example for google tag manager would be:

```astro
---
type Props = { measurementId: string }
const props = Astro.props as Props
const measurementId = props.measurementId
---
<script
  type='text/partytown'
  src={`https://www.googletagmanager.com/gtag/js?id=${measurementId}`}></script>
<script

```

Then you could include it in the `layout.astro` - head like this:

```astro
---
import GoogleAnalytics from './GoogleAnalytics.astro'
---
<html lang='en'>
  <head>
    <meta charset='utf-8' />
    <GoogleAnalytics measurementId={'G-XXXXXXXXXX'} />
    <title>My page</title>
    ·
```

PR #1475, opened 2024-11-11
Update coverage for config and error

### LINKED ISSUE #1401: ☔ Help us improve code coverage
We recently added support for branch-based code coverage and now publish metrics at https://tracemachina.github.io/nativelink. The goal of this issue is to improve the coverage metrics.

We're targeting 80% function, region and branch coverage. For further information see the [OSSF silver criterion on coverage](https://github.com/coreinfrastructure/best-practices-badge/blob/71e94bf7c551c4a1d9678607702726cb8e8b4d50/criteria/criteria.yml#L1071) which we'd like to fulfill in good faith.

- When browsing through the code coverage reports you might find some simple cases that are easily testable. These are "good first issues".
- In some simpler cases it might be more desirable to remove "unnecessary" code via refactors instead of adding more tests.
- We're testing coverage against the production Nix executables. Due to certain restrictions in Nix's sandboxes, some tests are currently disabled via a `#[cfg_attr(feature = "nix", ignore)]` feature. Making these tests runnable in nix would improve coverage.
- Some cases can feel like they're impossible to write tests for. These might suggest that the code in question needs to be refactored towards better testability. However, be warned that such refactors can range from trivial to highly complex.
- Prefer high-quality tests with low coverage over low-quality tests with high coverage. We're trying to improve code quality by placing a higher emphasis on testability, not by score-chasing.

leodziki:
Update coverage rate of nativelink-config and nativelink-error to 80~100% Passed cargo test and bazel test

# Description
Currently, coverage rate has been upgraded to 80-100% in nativelink-error and nativelink-config packages.
Bazel test and cargo test both passed. And Pre-commit passed in direnv and everything is perfect.
Fixed everything about the review

Fixes #1401 

## Type of change
  not work as expected)
- [x ] This change requires a documentation update

## How Has This Been Tested?
Bazel test, cargo test and pre-commit

## Checklist
- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1475)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1772,6 +1772,7 @@ dependencies = [
  "pretty_assertions",
  "serde",
  "serde_json5",
+ "serde_test",
  "shellexpand",
 ]
 
@@ -2836,6 +2837,15 @@ checksum = "32a6b754515e1a7bd79fc2edeaecee526fc80cb3a918607e5ca149225a3a9586"
 dependencies = [
  "pest",
  "pest_derive",
+ "serde",
+]
+
+[[package]]
+name = "serde_test"
+version = "1.0.177"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f901ee573cab6b3060453d2d5f0bae4e6d628c23c0a962ff9b5f1d7c8d4f1ed"
+dependencies = [
  "serde",
 ]
 
--- a/nativelink-config/BUILD.bazel
+++ b/nativelink-config/BUILD.bazel
@@ -41,6 +41,7 @@ rust_test_suite(
         "@crates//:pretty_assertions",
         "@crates//:serde",
         "@crates//:serde_json5",
+        "@crates//:serde_test",
     ],
 )
 
--- a/nativelink-config/Cargo.toml
+++ b/nativelink-config/Cargo.toml
@@ -12,3 +12,4 @@ shellexpand = { version = "3.1.0", default-features = false, features = ["base-0
 
 [dev-dependencies]
 pretty_assertions = { version = "1.4.1", features = ["std"] }
+serde_test = "1.0"
--- a/nativelink-config/tests/deserialization_test.rs
+++ b/nativelink-config/tests/deserialization_test.rs
@@ -14,7 +14,9 @@
 
 use nativelink_config::serde_utils::{
     convert_data_size_with_shellexpand, convert_duration_with_shellexpand,
-    convert_optional_numeric_with_shellexpand, convert_optional_string_with_shellexpand,
+    convert_numeric_with_shellexpand, convert_optional_numeric_with_shellexpand,
+    convert_optional_string_with_shellexpand, convert_string_with_shellexpand,
+    convert_vec_string_with_shellexpand,
 };
 use serde::Deserialize;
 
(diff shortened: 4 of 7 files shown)


leodziki:
I would appreciate it if you could take a moment to review my pull request. Your feedback would be invaluable, and I’m particularly interested in any thoughts you have on the implementation and overall impact of the changes made.
@blakehatch, @adam-singer

adam-singer (commented):
> Reviewed 3 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1475)* status: 0 of 2 LGTMs obtained, and 3 of 5 files reviewed, and 2 discussions need to be resolved

___
*[`nativelink-config/tests/deserialization_test.rs` line 28 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1475#-OBSOOJY6bYEfb6DvxEW:-OBSOOJY6bYEfb6DvxEX:bkkyj2d) ([raw file](https://github.com/TraceMachina/nativelink/blob/9229e6b5dbe97659abbabce0a9892df2d0650457/nativelink-config/tests/deserialization_test.rs#L28)):*
> ```XML
> const MAP_TYPE_ERROR: &str = "invalid type: map, expected a string containing json data";
> 
> const DURATION_HUMAN_READABLE: &str = r#"{"duration": "1m 10s"}"#;
> ```

Using these consts makes looking up what the example content is a double navigation. I think they should continue to be inlined until duplication is unmanageable 
___
*[`nativelink-metric-collector/src/lib.rs` line 19 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1475#-OBSOuGV1WHmvKjSxBMX:-OBSOuGV1WHmvKjSxBMY:b-1z8mlp) ([raw file](https://github.com/TraceMachina/nativelink/blob/9229e6b5dbe97659abbabce0a9892df2d0650457/nativelink-metric-collector/src/lib.rs#L19)):*
> ```XML
> 
> mod metrics_collection;
> pub mod metrics_visitors;
> ```

why is this public?

leodziki:
Thanks for your review
1. The double iteration issue you mentioned is generally minor unless it significantly affects readability or productivity. Inlining test values improves readability by keeping the context readily accessible within the test code. If constants are declared separately, it requires extra scrolling or navigation, disrupting the flow. Duplication is only a concern when the same data is reused across multiple tests; for unique values, inlining is simpler and cleaner. Constant declarations become necessary when duplication complicates maintenance. Therefore, inlining is a practical approach for unique or infrequently reused values, while centralizing constants should only be considered when duplication becomes problematic.
2. This is not directly related to nativelink-config and error.
I changed it for other test code. I made private public to test nativelink-metric-collector.
(e.g. #[cfg(test)] mod tests in src/metrics_visitors.rs). This allows you to test private functionality directly without making it public, but it requires that your tests be in the src/ directory, not tests/.
So I'm going to do this for code cleanliness.
I think this way I can keep the scope of the module strictly in production, while still allowing access to the tests. If it's better to keep it private, you could consider making #[cfg(test)] and pub(crate) or #[cfg_attr(test, pub)] public only in the test configuration, but I don't think there's any other problem with making it public.
This doesn't directly affect nativelink-error and config.

aaronmondal (commented):
@leodziki Thanks for finding some test cases!

https://github.com/TraceMachina/nativelink/commit/b6cf6590211a01125ca662c395eb9dce0a8f7d3d changed the serde_utils and tests fairly significantly. Could you split off your changes to serde_utils.rs into a separate PR and integrate your test cases into the new structure?

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1475#-:-OBVOZ8N2pb2umgomi7r:b-51a2z4)* status: 0 of 2 LGTMs obtained, and 3 of 5 files reviewed, and pending CI: Bazel Dev / macos-13, and 2 discussions need to be resolved

leodziki:
Hello, @aaronmondal 
Thank you for your review
I will address this by splitting off my changes to serde_utils.rs into a separate PR and incorporating my test cases into the updated structure. Thank you for the feedback.

Issue #1108, opened 2024-07-08
FilesystemStore should support native `StoreKey` instead of changing it to digest internally


allada:
Inside `FilesystemStore` we always turn the `StoreKey` into a digest, but in reality it should be using `.as_str()` for it's final destination storage name. This will make it much more usable for other future things (like BEP).

`MemoryStore` already has this optimization, so it can be used as a baseline on how to do this.

MarcusSorealheis:
more guidance: https://gist.github.com/MarcusSorealheis/17307f41b6c632cb751c397280c70ed4

KGrewal1:
Is there any greater specification for this? (eg how backwards compat should be handled / pitfalls from last 2 attempts?)

MarcusSorealheis:
> Is there any greater specification for this? (eg how backwards compat should be handled / pitfalls from last 2 attempts?)

@KGrewal1 Welcome to NativeLink. Here is a high-level overview of contributing: https://github.com/tracemachina/nativelink/blob/main/CONTRIBUTING.md

It should be entirely backwards compatible. As in, an existing store configuration should work withNativeLink including this capability. It should be caught by the CI tests. 

There is also a design document required for the submission. Pitfalls from the first 2 attempts were no design document and too much reinvention of implementation. You should look at the structure of `MemoryStore` as a good reference for starting this work.

MarcusSorealheis:
If you have any more questions I'll be happy to answer them here. We also have a Slack channel with a link on the website.

KGrewal1:
Is there more guidance on how you prefer / what you want in design documentation (eg HackMD format or examples), and regarding lifetimes (were notes on both the previous PRs and the MemoryStore implementation), guess that is out of scope for this (as required by 'static bound on the map and so larger change as lots of 'static trait bounds as well) the Slack seems to be invitation only

aleksdmladenovic:
/attempt https://github.com/TraceMachina/nativelink/issues/1108

MarcusSorealheis:
The design document needs to be publicly viewable.

KGrewal1:
https://hackmd.io/@HPxqrI77Q02izyGstTHbdg/ByS50OhGJx should be public now I think - re the CI run, seems stuck making pipelines for the local run

KGrewal1:
[...]

PR #1489, opened 2024-11-21
Support native StoreKey in FilesystemStore

KGrewal1:
# Description

Fixes # 1108

/claim #1108

## Design Document

https://hackmd.io/@HPxqrI77Q02izyGstTHbdg/ByS50OhGJx

## Type of change

## How Has This Been Tested?

Passes all of current test suite

## Checklist

- [] Updated documentation if needed
- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -35,7 +35,9 @@ use nativelink_util::buf_channel::{
 use nativelink_util::common::{fs, DigestInfo};
 use nativelink_util::evicting_map::{EvictingMap, LenEntry};
 use nativelink_util::health_utils::{HealthRegistryBuilder, HealthStatus, HealthStatusIndicator};
-use nativelink_util::store_trait::{StoreDriver, StoreKey, StoreOptimizations, UploadSizeInfo};
+use nativelink_util::store_trait::{
+    StoreDriver, StoreKey, StoreKeyBorrow, StoreOptimizations, UploadSizeInfo,
+};
 use nativelink_util::{background_spawn, spawn_blocking};
 use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt, SeekFrom};
 use tokio::time::{sleep, timeout, Sleep};
@@ -48,6 +50,15 @@ use crate::cas_utils::is_zero_digest;
 const DEFAULT_BUFF_SIZE: usize = 32 * 1024;
 // Default block size of all major filesystems is 4KB
 const DEFAULT_BLOCK_SIZE: u64 = 4 * 1024;
+
+pub const STR_FOLDER: &str = "s";
+pub const DIGEST_FOLDER: &str = "d";
+
+#[derive(Clone, Copy, Debug)]
+pub enum FileType {
+    Digest,
+    String,
+}
 
 #[derive(Debug, MetricsComponent)]
 pub struct SharedContext {
@@ -70,22 +81,21 @@ enum PathType {
     Custom(OsString),
 }
 
-// Note: We don't store the full path of the file because it would cause
-// a lot of needless memeory bloat. There's a high chance we'll end up with a
-// lot of small files, so to prevent storing duplicate data, we store an Arc
-// to the path of the directory where the file is stored and the packed digest.
-// Resulting in usize + sizeof(DigestInfo).
-type FileNameDigest = DigestInfo;
+/// [`EncodedFilePath`] stores the path to the file
+/// including the context, path type and key to the file.
+/// The whole [`StoreKey`] is stored as opposed to solely
+/// the [`DigestInfo`] so that it is more usable for things
+/// such as BEP -see Issue #1108
 pub struct EncodedFilePath {
     shared_context: Arc<SharedContext>,
     path_type: PathType,
-    digest: FileNameDigest,
+    key: StoreKey<'static>,
 }
 
 impl EncodedFilePath {
     #[inline]
     fn get_file_path(&self) -> Cow<'_, OsStr> {
-        get_file_path_raw(&self.path_type, self.shared_context.as_ref(), &self.digest)
+        get_file_path_raw(&self.path_type, self.shared_context.as_ref(), &self.key)
     }
 }
 
@@ -93,14 +103,14 @@ impl EncodedFilePath {
 fn get_file_path_raw<'a>(
     path_type: &'a PathType,
     shared_context: &SharedContext,
-    digest: &DigestInfo,
+    key: &StoreKey<'a>,
 ) -> Cow<'a, OsStr> {
     let folder = match path_type {
         PathType::Content => &shared_context.content_path,
         PathType::Temp => &shared_context.temp_path,
         PathType::Custom(path) => return Cow::Borrowed(path),
     };
-    Cow::Owned(to_full_path_from_digest(folder, digest))
+    Cow::Owned(to_full_path_from_key(folder, key))
 }
 
 impl Drop for EncodedFilePath {
@@ -131,9 +141,23 @@ impl Drop for EncodedFilePath {
     }
 }
 
+/// This creates the file path from the [`StoreKey`]. If
+/// it is a string, the string, prefixed with [`STR_PREFIX`]
+/// for backwards compatibility, is stored.
+///
+/// If it is a [`DigestInfo`], it is prefixed by [`DIGEST_PREFIX`]
+/// followed by the string representation of a digest - the hash in hex,
+/// a hyphen then the size in bytes
+///
+/// Previously, only the string representation of the [`DigestInfo`] was
+/// used with no prefix
 #[inline]
-fn to_full_path_from_digest(folder: &str, digest: &DigestInfo) -> OsString {
-    format!("{folder}/{digest}").into()
+fn to_full_path_from_key(folder: &str, key: &StoreKey<'_>) -> OsString {
+    match key {
+        StoreKey::Str(str) => format!("{folder}/{STR_FOLDER}/{str}"),
+        StoreKey::Digest(digest_info) => format!("{folder}/{DIGEST_FOLDER}/{digest_info}"),
+    }
+    .into()
 }
 
 pub trait FileEntry: LenEntry + Send + Sync + Debug + 'static {
@@ -300,7 +324,7 @@ impl Debug for FileEntryImpl {
     }
 }
 
-fn make_temp_digest(digest: &mut DigestInfo) {
+fn make_temp_digest(mut digest: DigestInfo) -> DigestInfo {
     static DELETE_FILE_COUNTER: AtomicU64 = AtomicU64::new(0);
     let mut hash = *digest.packed_hash();
     hash[24..].clone_from_slice(
@@ -309,6 +333,11 @@ fn make_temp_digest(digest: &mut DigestInfo) {
             .to_le_bytes(),
     );
     digest.set_packed_hash(*hash);
+    digest
+}
+
+fn make_temp_key(key: &StoreKey) -> StoreKey<'static> {
+    StoreKey::Digest(make_temp_digest(key.borrow().into_digest()))
 }
 
 impl LenEntry for FileEntryImpl {
@@ -362,16 +391,15 @@ impl LenEntry for FileEntryImpl {
                 return;
             }
             let from_path = encoded_file_path.get_file_path();
-            let mut new_digest = encoded_file_path.digest;
-            make_temp_digest(&mut new_digest);
+            let new_key = make_temp_key(&encoded_file_path.key);
 
             let to_path =
-                to_full_path_from_digest(&encoded_file_path.shared_context.temp_path, &new_digest);
+                to_full_path_from_key(&encoded_file_path.shared_context.temp_path, &new_key);
 
             if let Err(err) = fs::rename(&from_path, &to_path).await {
                 event!(
                     Level::WARN,
-                    digest = ?encoded_file_path.digest,
+                    key = ?encoded_file_path.key,
                     ?from_path,
                     ?to_path,
                     ?err,
@@ -380,23 +408,30 @@ impl LenEntry for FileEntryImpl {
             } else {
                 event!(
                     Level::INFO,
-                    digest = ?encoded_file_path.digest,
+                    key = ?encoded_file_path.key,
                     ?from_path,
                     ?to_path,
                     "Renamed file",
                 );
                 encoded_file_path.path_type = PathType::Temp;
-                encoded_file_path.digest = new_digest;
+                encoded_file_path.key = new_key;
             }
         }
     }
 }
 
 #[inline]
-pub fn digest_from_filename(file_name: &str) -> Result<DigestInfo, Error> {
+fn digest_from_filename(file_name: &str) -> Result<DigestInfo, Error> {
     let (hash, size) = file_name.split_once('-').err_tip(|| "")?;
     let size = size.parse::<i64>()?;
     DigestInfo::try_new(hash, size)
+}
+
+pub fn key_from_file(file_name: &str, file_type: FileType) -> Result<StoreKey<'_>, Error> {
+    match file_type {
+        FileType::String => Ok(StoreKey::new_str(file_name)),
+        FileType::Digest => digest_from_filename(file_name).map(StoreKey::Digest),
+    }
 }
 
 /// The number of files to read the metadata for at the same time when running
@@ -404,21 +439,24 @@ pub fn digest_from_filename(file_name: &str) -> Result<DigestInfo, Error> {
 const SIMULTANEOUS_METADATA_READS: usize = 200;
 
 async fn add_files_to_cache<Fe: FileEntry>(
-    evicting_map: &EvictingMap<DigestInfo, Arc<Fe>, SystemTime>,
+    evicting_map: &EvictingMap<StoreKeyBorrow, Arc<Fe>, SystemTime>,
     anchor_time: &SystemTime,
     shared_context: &Arc<SharedContext>,
     block_size: u64,
+    rename_fn: fn(&OsStr, &OsStr) -> Result<(), std::io::Error>,
 ) -> Result<(), Error> {
+    #[expect(clippy::too_many_arguments)]
     async fn process_entry<Fe: FileEntry>(
-        evicting_map: &EvictingMap<DigestInfo, Arc<Fe>, SystemTime>,
+        evicting_map: &EvictingMap<StoreKeyBorrow, Arc<Fe>, SystemTime>,
         file_name: &str,
+        file_type: FileType,
         atime: SystemTime,
         data_size: u64,
         block_size: u64,
         anchor_time: &SystemTime,
         shared_context: &Arc<SharedContext>,
     ) -> Result<(), Error> {
-        let digest = digest_from_filename(file_name)?;
+        let key = key_from_file(file_name, file_type)?;
 
         let file_entry = Fe::create(
             data_size,
@@ -426,7 +464,7 @@ async fn add_files_to_cache<Fe: FileEntry>(
             RwLock::new(EncodedFilePath {
                 shared_context: shared_context.clone(),
                 path_type: PathType::Content,
-                digest,
+                key: key.borrow().into_owned(),
             }),
         );
         let time_since_anchor = anchor_time
@@ -434,7 +472,7 @@ async fn add_files_to_cache<Fe: FileEntry>(
             .map_err(|_| make_input_err!("File access time newer than now"))?;
         evicting_map
             .insert_with_time(
-                digest,
+                key.into_owned().into(),
                 Arc::new(file_entry),
                 time_since_anchor.as_secs() as i32,
             )
@@ -442,8 +480,21 @@ async fn add_files_to_cache<Fe: FileEntry>(
         Ok(())
     }
 
-    let mut file_infos: Vec<(String, SystemTime, u64)> = {
-        let (_permit, dir_handle) = fs::read_dir(format!("{}/", shared_context.content_path))
+    async fn read_files(
+        folder: Option<&str>,
+        shared_context: &SharedContext,
+    ) -> Result<Vec<(String, SystemTime, u64, bool)>, Error> {
+        // Note: In Dec 2024 this is for backwards compatibility with the old
+        // way files were stored on disk. Previously all files were in a single
+        // folder regardless of the StoreKey type. This allows old versions of
+        // nativelink file layout to be upgraded at startup time.
+        // This logic can be removed once more time has passed.
+        let read_dir = if let Some(folder) = folder {
+            format!("{}/{folder}/", shared_context.content_path)
+        } else {
+            format!("{}/", shared_context.content_path)
+        };
+        let (_permit, dir_handle) = fs::read_dir(read_dir)
             .await
             .err_tip(|| "Failed opening content directory for iterating in filesystem store")?
             .into_inner();
@@ -457,6 +508,9 @@ async fn add_files_to_cache<Fe: FileEntry>(
                     .metadata()
                     .await
                     .err_tip(|| "Failed to get metadata in filesystem store")?;
+                // We need to filter out folders - we do not want to try to cache the s and d folders.
+                let is_file =
+                    metadata.is_file() || !(file_name == STR_FOLDER || file_name == DIGEST_FOLDER);
                 let atime = match metadata.accessed() {
                     Ok(atime) => atime,
                     Err(err) => {
@@ -470,53 +524,136 @@ async fn add_files_to_cache<Fe: FileEntry>(
                         );
                     }
                 };
-                Result::<(String, SystemTime, u64), Error>::Ok((file_name, atime, metadata.len()))
+                Result::<(String, SystemTime, u64, bool), Error>::Ok((
+                    file_name,
+                    atime,
+                    metadata.len(),
+                    is_file,
+                ))
             })
             .buffer_unordered(SIMULTANEOUS_METADATA_READS)
             .try_collect()
-            .await?
-    };
-
-    file_infos.sort_by(|a, b| a.1.cmp(&b.1));
-    for (file_name, atime, data_size) in file_infos {
-        let result = process_entry(
-            evicting_map,
-            &file_name,
-            atime,
-            data_size,
-            block_size,
-            anchor_time,
-            shared_context,
-        )
-        .await;
-        if let Err(err) = result {
-            event!(
-                Level::WARN,
-                ?file_name,
-                ?err,
-                "Failed to add file to eviction cache",
-            );
-            // Ignore result.
-            let _ =
-                fs::remove_file(format!("{}/{}", &shared_context.content_path, &file_name)).await;
-        }
+            .await
     }
+
+    /// Note: In Dec 2024 this is for backwards compatibility with the old
+    /// way files were stored on disk. Previously all files were in a single
+    /// folder regardless of the [`StoreKey`] type. This moves files from the old cache
+    /// location to the new cache location, under [`DIGEST_FOLDER`].
+    async fn move_old_cache(
+        shared_context: &Arc<SharedContext>,
+        rename_fn: fn(&OsStr, &OsStr) -> Result<(), std::io::Error>,
+    ) -> Result<(), Error> {
+        let file_infos = read_files(None, shared_context).await?;
+
+        let from_path = shared_context.content_path.to_string();
+
+        let to_path = format!("{}/{DIGEST_FOLDER}", shared_context.content_path);
+
+        for (file_name, _, _, _) in file_infos.into_iter().filter(|x| x.3) {
+            let from_file: OsString = format!("{from_path}/{file_name}").into();
+            let to_file: OsString = format!("{to_path}/{file_name}").into();
+
+            if let Err(err) = rename_fn(&from_file, &to_file) {
+                event!(
+                    Level::WARN,
+                    ?from_file,
+                    ?to_file,
+                    ?err,
+                    "Failed to rename file",
+                );
+            } else {
+                event!(Level::INFO, ?from_file, ?to_file, "Renamed file",);
+            }
+        }
+        Ok(())
+    }
+
+    async fn add_files_to_cache<Fe: FileEntry>(
+        evicting_map: &EvictingMap<StoreKeyBorrow, Arc<Fe>, SystemTime>,
+        anchor_time: &SystemTime,
+        shared_context: &Arc<SharedContext>,
+        block_size: u64,
+        folder: &str,
+    ) -> Result<(), Error> {
+        let file_infos = read_files(Some(folder), shared_context).await?;
+        let file_type = match folder {
+            STR_FOLDER => FileType::String,
+            DIGEST_FOLDER => FileType::Digest,
+            _ => panic!("Invalid folder type"),
+        };
+
+        let path_root = format!("{}/{folder}", shared_context.content_path);
+
+        for (file_name, atime, data_size, _) in file_infos.into_iter().filter(|x| x.3) {
+            let result = process_entry(
+                evicting_map,
+                &file_name,
+                file_type,
+                atime,
+                data_size,
+                block_size,
+                anchor_time,
+                shared_context,
+            )
+            .await;
+            if let Err(err) = result {
+                event!(
+                    Level::WARN,
+                    ?file_name,
+                    ?err,
+                    "Failed to add file to eviction cache",
+                );
+                // Ignore result.
+                let _ = fs::remove_file(format!("{path_root}/{file_name}")).await;
+            }
+        }
+        Ok(())
+    }
+
+    move_old_cache(shared_context, rename_fn).await?;
+
+    add_files_to_cache(
+        evicting_map,
+        anchor_time,
+        shared_context,
+        block_size,
+        DIGEST_FOLDER,
+    )
+    .await?;
+
+    add_files_to_cache(
+        evicting_map,
+        anchor_time,
+        shared_context,
+        block_size,
+        STR_FOLDER,
+    )
+    .await?;
     Ok(())
 }
 
 async fn prune_temp_path(temp_path: &str) -> Result<(), Error> {
-    let (_permit, dir_handle) = fs::read_dir(temp_path)
-        .await
-        .err_tip(|| "Failed opening temp directory to prune partial downloads in filesystem store")?
-        .into_inner();
+    async fn prune_temp_inner(temp_path: &str, subpath: &str) -> Result<(), Error> {
+        let (_permit, dir_handle) = fs::read_dir(format!("{temp_path}/{subpath}"))
+            .await
+            .err_tip(|| {
+                "Failed opening temp directory to prune partial downloads in filesystem store"
+            })?
+            .into_inner();
 
-    let mut read_dir_stream = ReadDirStream::new(dir_handle);
-    while let Some(dir_entry) = read_dir_stream.next().await {
-        let path = dir_entry?.path();
-        if let Err(err) = fs::remove_file(&path).await {
-            event!(Level::WARN, ?path, ?err, "Failed to delete file",);
+        let mut read_dir_stream = ReadDirStream::new(dir_handle);
+        while let Some(dir_entry) = read_dir_stream.next().await {
+            let path = dir_entry?.path();
+            if let Err(err) = fs::remove_file(&path).await {
+                event!(Level::WARN, ?path, ?err, "Failed to delete file",);
+            }
         }
+        Ok(())
     }
+
+    prune_temp_inner(temp_path, STR_FOLDER).await?;
+    prune_temp_inner(temp_path, DIGEST_FOLDER).await?;
     Ok(())
 }
 
@@ -525,7 +662,7 @@ pub struct FilesystemStore<Fe: FileEntry = FileEntryImpl> {
     #[metric]
     shared_context: Arc<SharedContext>,
     #[metric(group = "evicting_map")]
-    evicting_map: Arc<EvictingMap<DigestInfo, Arc<Fe>, SystemTime>>,
+    evicting_map: Arc<EvictingMap<StoreKeyBorrow, Arc<Fe>, SystemTime>>,
     #[metric(help = "Block size of the configured filesystem")]
     block_size: u64,
     #[metric(help = "Size of the configured read buffer size")]
@@ -546,18 +683,25 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
         sleep_fn: fn(Duration) -> Sleep,
         rename_fn: fn(&OsStr, &OsStr) -> Result<(), std::io::Error>,
     ) -> Result<Arc<Self>, Error> {
+        async fn create_subdirs(path: &str) -> Result<(), Error> {
+            fs::create_dir_all(format!("{path}/{STR_FOLDER}"))
+                .await
+                .err_tip(|| format!("Failed to create directory {path}/{STR_FOLDER}"))?;
+            fs::create_dir_all(format!("{path}/{DIGEST_FOLDER}"))
+                .await
+                .err_tip(|| format!("Failed to create directory {path}/{DIGEST_FOLDER}"))
+        }
+
         let now = SystemTime::now();
 
         let empty_policy = nativelink_config::stores::EvictionPolicy::default();
         let eviction_policy = spec.eviction_policy.as_ref().unwrap_or(&empty_policy);
         let evicting_map = Arc::new(EvictingMap::new(eviction_policy, now));
 
-        fs::create_dir_all(&spec.temp_path)
-            .await
-            .err_tip(|| format!("Failed to temp directory {:?}", &spec.temp_path))?;
-        fs::create_dir_all(&spec.content_path)
-            .await
-            .err_tip(|| format!("Failed to content directory {:?}", &spec.content_path))?;
+        // Create temp and content directories and the s and d subdirectories.
+
+        create_subdirs(&spec.temp_path).await?;
+        create_subdirs(&spec.content_path).await?;
 
         let shared_context = Arc::new(SharedContext {
             active_drop_spawns: AtomicU64::new(0),
@@ -570,7 +714,14 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
         } else {
             spec.block_size
         };
-        add_files_to_cache(evicting_map.as_ref(), &now, &shared_context, block_size).await?;
+        add_files_to_cache(
+            evicting_map.as_ref(),
+            &now,
+            &shared_context,
+            block_size,
+            rename_fn,
+        )
+        .await?;
         prune_temp_path(&shared_context.temp_path).await?;
 
         let read_buffer_size = if spec.read_buffer_size == 0 {
@@ -595,7 +746,7 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
 
     pub async fn get_file_entry_for_digest(&self, digest: &DigestInfo) -> Result<Arc<Fe>, Error> {
         self.evicting_map
-            .get(digest)
+            .get::<StoreKey<'static>>(&digest.into())
             .await
             .ok_or_else(|| make_err!(Code::NotFound, "{digest} not found in filesystem store"))
     }
@@ -604,7 +755,7 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
         self: Pin<&'a Self>,
         mut entry: Fe,
         mut resumeable_temp_file: fs::ResumeableFileSlot,
-        final_digest: DigestInfo,
+        final_key: StoreKey<'static>,
         mut reader: DropCloserReadHalf,
     ) -> Result<(), Error> {
         let mut data_size = 0;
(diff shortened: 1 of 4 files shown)


KGrewal1:
Unsure what is happening with the local runner on CI,  seems to be a Tekton issue

nativelink-store/src/filesystem_store.rs
```diff
@@ -70,37 +73,31 @@ enum PathType {
     Custom(OsString),
 }
 
-// Note: We don't store the full path of the file because it would cause
-// a lot of needless memeory bloat. There's a high chance we'll end up with a
```
MarcusSorealheis, line 74: Because we are removing this comment, I would like to replace it with another comment block that explains the behavior now.

nativelink-store/src/filesystem_store.rs
```diff
@@ -132,8 +129,13 @@ impl Drop for EncodedFilePath {
 }
 
 #[inline]
```
MarcusSorealheis, line 154: Perhaps a comment block of a few lines here explaining how the prefixing works, and how we generate these values?

MarcusSorealheis: a before and after would be great but the before usage and explanation will eventually be phased out.

nativelink-store/src/filesystem_store.rs
```diff
@@ -848,9 +867,16 @@ impl<Fe: FileEntry> StoreDriver for FilesystemStore<Fe> {
             return Ok(());
         }
 
-        let entry =
-            self.evicting_map.get(&digest).await.ok_or_else(|| {
-                make_err!(Code::NotFound, "{digest} not found in filesystem store")
+        let entry = self
```
MarcusSorealheis: How have we evaluated whether or not we have increased heap allocation overhead? I'm not sure if you used a tool like Valgrind or something similar. Could you elaborate? I'm not saying that there is a problem here. I'm just saying that it's possible. 

I will also try to find some time this week to look into it, but this one is important to complete the task because we will be adding this code to a large scale system.

MarcusSorealheis (changes requested):
BTW, with the exception of the missing performance data, your design document is basically done. Thanks for the hardwork there. Now we just need to cover all the important areas of the code, in addition to including the perf data in the document.

nativelink-store/src/filesystem_store.rs
KGrewal1: I can actually just remove this  `into_owned()` using the method described here https://blinsay.com/blog/compound-keys/ however as the wrapper and inner structs are reversed (cannot easily make the currently existing StoreKey a wrapper struct around the actual enum, and cannot easily make everywhere storing StoreKey instead use StoreKeyWrapper<'static>, instead use unsafe to transmute the reference between the two, ensuring the wrapper is transparent so they are guaranteed to have the same layout in memory - I have not yet squashed commits so if wanted, this can be easily reverted

MarcusSorealheis: I want more co-founder @allada and @aaronmondal to take a look before we adopt this change. @allada is out until Monday but he might get a chance to review this change before we then. I appreciate that we have both looks so no need to squash just yet. 

Thanks again. Great work thus far!

KGrewal1: This does acc also work to remove the todo on the has_with_results for filesystem_store.rs and memory_store.rs but will push on top so easy reverts still. No worries if it takes a bit!

KGrewal1: Have a safe version now too, so should be easier to review!

aaronmondal (changes requested):
Very nice!

1. Let's add an explicit test for a string-key and a digest-key to check that `to_full_path_from_key` does what it's supposed to.
2. Regarding the deprecation thing, IIUC we'd want to remove the `Str` variant entirely from the `StoreKey`, right? That would allow us to keep moving just the digest info around instead of having to adjust for the `Cow` case as well. Regarding backwards-compatibility I'd say:
  - If this is just about staying backwards-compatible with existing artifacts in a filesystem store, don't worry about it. Caches are meant to be cleared once in a while anyways, so we don't need to account for the case where a user updates nativelink and tries to reuse an existing cache directory. If it's easy to remove, lets just get rid of it entirely.
  - If other parts of the codebase significantly depend on it and it's too hard to remove the `Str` case entirely, let's print an actual deprecation warning if for whatever reason a Str-style key ends up in the filesystem store so that we can better figure out which other parts we need to fix later.

I'm not entirely sure on point `2`. +@allada will likely want to chime in as well. @KGrewal1 Since you're already familiar with this part of the code, what are your thoughts on removing the `Str` variant?

> Reviewed 1 of 2 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-ODgfgiZEnF61R6jJRzg:b-nod4h6)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 5 discussions need to be resolved (waiting on @allada)

___
*[`-- commits` line 18 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODg_ZTaBS63ZBW1vua9:-ODg_ZTbC13Hp_qd-eBW:b-kxvdys):*
nit: Remember to squash at the end and remove this. Commit title lgtm :+1:
___
*[`nativelink-store/src/filesystem_store.rs` line 322 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODgarpv9FeTLlSaXpA7:-ODgarpv9FeTLlSaXpA8:b-xyosr1) ([raw file](https://github.com/TraceMachina/nativelink/blob/f85582f543da1920ee71b5337cad43659a8fd8a1/nativelink-store/src/filesystem_store.rs#L322)):*
> ```XML
> }
> 
> fn make_temp_digest(mut digest: DigestInfo) -> DigestInfo {
> ```

Interesting choice! Since this is just 40 bytes I'd assume that we get less cache misses by moving here.

KGrewal1:
If the aim is to remove the string variant from StoreKey, I'm not really sure what the aim of the bounty was for - as far as I could understand the aim was to keep using the string key as much as possible? The previous code already handled the digest variant fully?

KGrewal1:
Regarding moves with the make temp digest, it's a small function that is used once, so if this was limiting would assume it would get inlined - can add inline explicitly, but prefer to trust the compiler unless I am very sure it's a win

KGrewal1:
*[`-- commits` line 18 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODg_ZTaBS63ZBW1vua9:-ODgugC06n3pWiQfreJH:bbacuec):*

nit: Remember to squash at the end and remove this. Commit title lgtm :+1:
</blockquote></details>

Keeping it separate for now (in case any change needs to be reverted), but will squash before any merge

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 322 at r3](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODgarpv9FeTLlSaXpA7:-ODgvFE_1JmCRU4pgfPY:bu8tkym) ([raw file](https://github.com/TraceMachina/nativelink/blob/f85582f543da1920ee71b5337cad43659a8fd8a1/nativelink-store/src/filesystem_store.rs#L322)):*

Interesting choice! Since this is just 40 bytes I'd assume that we get less cache misses by moving here.
</blockquote></details>

As mentioned in GH, small function that is called in one place, so would expect it to be inlined - if not, passing a relatively small copy type value by value as opposed to by reference is still probably preferable

nativelink-store/src/filesystem_store.rs
KGrewal1: Done.

nativelink-store/src/filesystem_store.rs
KGrewal1: Done.

KGrewal1 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: 0 of 2 LGTMs obtained, and 3 of 4 files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Bazel Dev / ubuntu-24.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / macos-14, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 5 discussions need to be resolved (waiting on @allada)

aaronmondal (commented):
Hmm tbh I'm actually not sure why the StoreKey looks like it does in the first place. I must be overlooking something (and I don't have the context from reviews at previous attempts). Intuitively, it seems to me like the Cow variant only causes unnecessary potential for fragmentation and cache misses. I wonder why we'd want to dynamically determine the key type during runtime at all.

@allada Why does `StoreKey::Str` exist? Is it so that we can reuse stores for e.g. BEP data? If so, would it make sense to lift the supported format to nativelink-config so that it's set on a per-store basis at configuration time (i.e. in the `config.json5`)? While we'd lose the ability to reuse a store for different formats, it should lead to more predictable runtime behavior and afaik we use dedicated stores like a `"BEP_STORE"` for bep data anyways.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-ODguZZf-Dpa6bNHg4Mn:bc1bl9f)* status: 0 of 2 LGTMs obtained, and 3 of 4 files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Bazel Dev / ubuntu-24.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, Local / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / macos-14, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 3 discussions need to be resolved (waiting on @allada)

allada (changes requested):
Sorry for the delay, I've been busy.

Overall looks really good. Just the one blocker is to have `Str` and `Digest` live in different folders. Everything else looks great :-)

`StoreKey::Str` exists to support non-digest paths. It turns out all our stores support arbitrary strings as keys and being able to use the same store API for BEP and event, and scheduler backends was super useful (eg: redis scheduler uses `StoreKey::Str`).

> Reviewed 3 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> Dismissed @MarcusSorealheis from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-ODqVO2Q9sOxqpmTpPgy:brq9zus)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 5 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/filesystem_store.rs` line 79 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqWP67DjduWmEzanxk:-ODqWP67DjduWmEzanxl:b9qvn3m) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L79)):*
> ```XML
> 
> /// [`EncodedFilePath`] stores the path to the file
> /// inclusing the context, path type and key to the file.
> ```

nit: `inclusing` -> `including`.
___
*[`nativelink-store/src/filesystem_store.rs` line 150 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqWtBO1Px2wyhf8Xjj:-ODqWtBP93mV73wT00DN:bom6flj) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L150)):*
> ```XML
> #[inline]
> fn to_full_path_from_key(folder: &str, key: &StoreKey<'_>) -> OsString {
>     // appropriately prefix file name
> ```

nit: Remove this comment, it doesn't really say anything that the above comment or code doesn't describe.
___
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODqXw8I2-y_8H0m3oNP:b3uylwa) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*
> ```XML
>     // appropriately prefix file name
>     match key {
>         StoreKey::Str(str) => format!("{folder}/{STR_PREFIX}{str}"),
> ```

After thinking about this for a bit, I think we should put `Str` and `Digest` into different folders. I suggest:
`StoreKey::Str` -> `{folder}/s/{str}`
`StoreKey::Digest` -> `{folder}/d/{digest}`

We'll then need to make a reverse-compatibility layer that will read `StoreKey::Digest` from both `{folder}` and `{folder}/d` at load time, and move all the files to the new location (ie: migrate them).
___
*[`nativelink-store/tests/filesystem_store_test.rs` line 1267 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqbuDY23kk-0ynU1yr:-ODqbuDY23kk-0ynU1ys:bj3uc5k) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/tests/filesystem_store_test.rs#L1267)):*
> ```XML
>     assert!(digest_result.is_none());
> 
>     // repeat with a string typed key
> ```

nit: Please caps and period comments (a style we use in our code base).
___
*[`nativelink-util/src/store_trait.rs` line 149 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqayPC-1kmFko-xMaz:-ODqayPC-1kmFko-xMb-:b-43kr7a) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-util/src/store_trait.rs#L149)):*
> ```XML
> /// A wrapper struct for [`StoreKey`] to work around
> /// lifetime limitations in `HashMap::get()` as described in
> /// <https://github.com/rust-lang/rust/issues/80389>
> ```

Good catch! I was wondering why this gave problems.

aaronmondal (commented):
So I played around a bit witht the monomorphization approach and got something that's like 2-4x faster. I don't think it should be part of this PR and it's not *that* related to this issue, but this still should give us reason to reevaluate whether we should make the kind of supported digest type be a selection at configuration time (don't let this block you though, but probably something interesting to think about):

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use std::borrow::Cow;
use std::fmt::Debug;
use nativelink_util::common::DigestInfo;
use nativelink_store::filesystem_store::{digest_from_filename, to_full_path_from_key};
use nativelink_store::filesystem_store::{DIGEST_PREFIX, STR_PREFIX};
use nativelink_error::{Error, make_err, Code};
use nativelink_util::store_trait;
use std::ffi::OsString;

type StringKey = Cow<'static, str>;
type DigestKey = DigestInfo;

pub trait KeyType: Debug + Clone + Eq {
    fn file_name(&self) -> String;
    fn from_file_name(s: &str) -> Result<Self, Error> where Self: Sized;
}

impl KeyType for Cow<'_, str> {
    fn file_name(&self) -> String {
        let mut result = String::with_capacity(STR_PREFIX.len() + self.len());
        result.push_str(STR_PREFIX);
        result.push_str(self);
        result
    }

    fn from_file_name(s: &str) -> Result<Self, Error> {
        s.strip_prefix(STR_PREFIX)
            .map(|val| Cow::Owned(val.to_owned()))
            .ok_or_else(|| make_err!(Code::InvalidArgument, "Invalid string key format"))
    }
}

impl KeyType for DigestInfo {
    fn file_name(&self) -> String {
        let digest_str = self.to_string();
        let mut result = String::with_capacity(DIGEST_PREFIX.len() + digest_str.len());
        result.push_str(DIGEST_PREFIX);
        result.push_str(&digest_str);
        result
    }

    fn from_file_name(s: &str) -> Result<Self, Error> {
        let s = s.strip_prefix(DIGEST_PREFIX).unwrap_or(s);
        digest_from_filename(s)
    }
}

#[derive(Debug, Clone, Eq, PartialEq)]
pub struct StoreKey<T: KeyType>(pub T);

impl<T: KeyType> StoreKey<T> {
    pub fn new(inner: T) -> Self {
        Self(inner)
    }
}

impl StoreKey<StringKey> {
    pub fn new_str(s: &'static str) -> StoreKey<StringKey> {
        Self(Cow::Borrowed(s))
    }
}


pub fn to_full_path_from_key_template<T: KeyType>(folder: &str, key: &StoreKey<T>) -> OsString {
    let file_name = key.0.file_name();
    let mut path = OsString::with_capacity(folder.len() + 1 + file_name.len());
    path.push(folder);
    path.push("/");
    path.push(&file_name);
    path
}


fn create_test_data_enum() -> (store_trait::StoreKey<'static>, store_trait::StoreKey<'static>) {
    let digest = DigestInfo::try_new("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", 32).unwrap();
    let str_key = store_trait::StoreKey::new_str("test_file_very_long_name_to_prevent_sso");
    let digest_key = store_trait::StoreKey::Digest(digest);
    (str_key, digest_key)
}

fn create_test_data_template() -> (StoreKey<StringKey>, StoreKey<DigestKey>) {
    let digest = DigestInfo::try_new("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", 32).unwrap();
    let str_key = StoreKey::new(Cow::Owned("test_file_very_long_name_to_prevent_sso".to_owned()));
    let digest_key = StoreKey::new(digest);
    (str_key, digest_key)
}

fn bench_path_generation(c: &mut Criterion) {
    let (enum_str_key, enum_digest_key) = create_test_data_enum();
    let (template_str_key, template_digest_key) = create_test_data_template();
    let folder = "/some/base/path";

    let mut group = c.benchmark_group("path_generation");
    
    group.bench_function("enum_string_key", |b| b.iter(|| {
        black_box(to_full_path_from_key(
            black_box(folder),
            black_box(&enum_str_key)
        ))
    }));

    group.bench_function("enum_digest_key", |b| b.iter(|| {
        black_box(to_full_path_from_key(
            black_box(folder),
            black_box(&enum_digest_key)
        ))
    }));

    group.bench_function("template_string_key", |b| b.iter(|| {
        black_box(to_full_path_from_key_template(
            black_box(folder),
            black_box(&template_str_key)
        ))
    }));

    group.bench_function("template_digest_key", |b| b.iter(|| {
        black_box(to_full_path_from_key_template(
            black_box(folder),
            black_box(&template_digest_key)
        ))
    }));
    
    group.finish();
}

fn bench_comparison(c: &mut Criterion) {
    let (enum_str_key1, enum_digest_key1) = create_test_data_enum();
    let (enum_str_key2, enum_digest_key2) = create_test_data_enum();
    let (template_str_key1, template_digest_key1) = create_test_data_template();
    let (template_str_key2, template_digest_key2) = create_test_data_template();

    let mut group = c.benchmark_group("comparison");
    
    group.bench_function("enum_string_key_eq", |b| b.iter(|| {
        black_box(enum_str_key1 == enum_str_key2)
    }));

    group.bench_function("enum_digest_key_eq", |b| b.iter(|| {
        black_box(enum_digest_key1 == enum_digest_key2)
    }));

    group.bench_function("template_string_key_eq", |b| b.iter(|| {
        black_box(template_str_key1 == template_str_key2)
    }));

    group.bench_function("template_digest_key_eq", |b| b.iter(|| {
        black_box(template_digest_key1 == template_digest_key2)
    }));

    group.bench_function("create_enum_string", |b| b.iter(|| {
        black_box(store_trait::StoreKey::new_str("test"))
    }));

    group.bench_function("create_template_string", |b| b.iter(|| {
        black_box(StoreKey::new_str("test"))
    }));
    
    group.finish();
}

use rand::rngs::StdRng;
use rand::{SeedableRng, RngCore, Rng};

fn create_randomized_keys(size: usize) -> (Vec<store_trait::StoreKey<'static>>, 
                                         Vec<StoreKey<StringKey>>, 
                                         Vec<StoreKey<DigestKey>>) {
    let mut rng = StdRng::seed_from_u64(42);
    
    let random_strings: Vec<String> = (0..size).map(|_| {
        format!("test_file_very_long_name_{}", rng.gen::<u64>())
    }).collect();
    
    let random_digests: Vec<DigestInfo> = (0..size).map(|_| {
        let mut hash = [0u8; 32];
        rng.fill_bytes(&mut hash);
        DigestInfo::try_new(&hex::encode(hash), 32).unwrap()
    }).collect();

    let mixed_keys: Vec<store_trait::StoreKey<'static>> = (0..size)
        .map(|i| {
            if rng.gen_bool(0.5) {
                store_trait::StoreKey::new_str(&random_strings[i % (size/2)])
                    .into_owned()
            } else {
                store_trait::StoreKey::Digest(random_digests[i % (size/2)])
            }
        })
        .collect();

    let string_keys: Vec<StoreKey<StringKey>> = random_strings
        .iter()
        .map(|s| StoreKey::new(Cow::Owned(s.to_owned())))
        .collect();

    let digest_keys: Vec<StoreKey<DigestKey>> = random_digests
        .iter()
        .map(|d| StoreKey::new(*d))
        .collect();

    (mixed_keys, string_keys, digest_keys)
}

fn bench_sequential_access(c: &mut Criterion) {
    let size = 10_000;
    let folder = "/some/base/path";
    
    let (mixed_keys, string_keys, digest_keys) = create_randomized_keys(size);
    
    let mut group = c.benchmark_group("sequential_access");
    group.sample_size(50);
    
    group.bench_function("enum_mixed_access", |b| b.iter(|| {
        let mut sum = 0;
        for key in mixed_keys.iter() {
            let path = black_box(to_full_path_from_key(
                black_box(folder),
                black_box(key)
            ));
            sum += path.len();
        }
        black_box(sum)
    }));

    group.bench_function("template_sequential_access", |b| b.iter(|| {
        let mut sum = 0;
        for key in string_keys.iter() {
            let path = black_box(to_full_path_from_key_template(
                black_box(folder),
                black_box(key)
            ));
            sum += path.len();
        }
        for key in digest_keys.iter() {
            let path = black_box(to_full_path_from_key_template(
                black_box(folder),
                black_box(key)
            ));
            sum += path.len();
        }
        black_box(sum)
    }));
    
    group.finish();
}

criterion_group!(
    benches,
    bench_path_generation,
    bench_comparison,
    bench_sequential_access,
);
criterion_main!(benches);
```

```
path_generation/enum_string_key
                        time:   [129.91 ns 130.74 ns 131.94 ns]
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
path_generation/enum_digest_key
                        time:   [266.10 ns 267.19 ns 268.46 ns]
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe
path_generation/template_string_key
                        time:   [35.336 ns 35.537 ns 35.795 ns]
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
path_generation/template_digest_key
                        time:   [157.35 ns 158.06 ns 158.87 ns]

comparison/enum_string_key_eq
                        time:   [4.2017 ns 4.2484 ns 4.3138 ns]

Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
comparison/enum_digest_key_eq
                        time:   [1.7993 ns 1.8080 ns 1.8168 ns]

Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild
comparison/template_string_key_eq
                        time:   [3.3234 ns 3.3338 ns 3.3452 ns]

Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe
comparison/template_digest_key_eq
                        time:   [1.3477 ns 1.3504 ns 1.3533 ns]

Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
comparison/create_enum_string
                        time:   [2.9898 ns 3.0017 ns 3.0167 ns]

Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
comparison/create_template_string
                        time:   [1.6738 ns 1.6793 ns 1.6853 ns]

Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  6 (6.00%) high mild
  2 (2.00%) high severe

sequential_access/enum_mixed_access
                        time:   [2.0962 ms 2.1023 ms 2.1092 ms]

Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) high mild
  1 (2.00%) high severe
sequential_access/template_sequential_access
                        time:   [1.9032 ms 1.9170 ms 1.9372 ms]

Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) high mild
  1 (2.00%) high severe
```

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-ODqkJn280u-3LHAAGZu:b4f2yyx)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 5 discussions need to be resolved (waiting on @allada)

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODridTv7k-BAW5UYCd8:b-4jz9iq) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*

After thinking about this for a bit, I think we should put `Str` and `Digest` into different folders. I suggest:
`StoreKey::Str` -> `{folder}/s/{str}`
`StoreKey::Digest` -> `{folder}/d/{digest}`

We'll then need to make a reverse-compatibility layer that will read `StoreKey::Digest` from both `{folder}` and `{folder}/d` at load time, and move all the files to the new location (ie: migrate them).
</blockquote></details>

I think this would be a breaking change as `key_from_filename` is in the public API, and would have no way of distinguishing which folder the filename was in? (Or need to be modified to take a `Path` / `Into<Path>` so it is clear it takes the path relative to the folder above the string and digest folders?) Though also doesn't seem used outside the crate except for testing?

KGrewal1:
Given this is to interact with an on disk cache and bar the array access most of these are in the ns region I'd have assumed latency on actually doing disk operations would be far higher (maybe not for a ram disk)? Access over array difference is bigger but difference between is still around 200 us so comparable magnitude to disk access?

allada (commented):
Yeah, the disk access is likely going to be 1000x+ slower than the string/heap cost. In addition, the actual thing we should be comparing here is the cost to hash & compare native strings/digests & enum variants.

There are also 100x more lookups in the map than there are inserts and lookups only have to pay for a hash function + comparison. I will bet that comparisons are apx the same cost going from `StoreKey == StoreKey` when the types are the same (which the overwhelming common case).

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-ODtL4Ze3IiODwkIkX9F:bdqtt2t)* status: 0 of 2 LGTMs obtained, and 2 of 4 files reviewed, and pending CI: Cargo Dev / macos-13, and 1 discussions need to be resolved

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: 0 of 2 LGTMs obtained, and 2 of 4 files reviewed, and pending CI: Cargo Dev / macos-13, and 1 discussions need to be resolved

___
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODtOLUn1mnGURJ8mY5v:b5jt36a) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*
<details><summary><i>Previously, KGrewal1 (Kirpal Grewal) wrote…</i></summary><blockquote>

I think this would be a breaking change as `key_from_filename` is in the public API, and would have no way of distinguishing which folder the filename was in? (Or need to be modified to take a `Path` / `Into<Path>` so it is clear it takes the path relative to the folder above the string and digest folders?) Though also doesn't seem used outside the crate except for testing?
</blockquote></details>

`key_from_filename` is only exposed for tests, so it won't be breaking (plus nativelink follows breaking definition to be "if a user upgrades nativelink will they have to change anything or see behavioral changes between previous version").

Looking at the code, it should be fairly easy to address this, we can take in the store key type in the function as well. The `file_infos` will just need to also store and pass the key type to `process_entry` (or better yet `file_infos` might create the key first and pass the key directly into `process_entry`).

KGrewal1 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: 0 of 2 LGTMs obtained, and 2 of 4 files reviewed, and pending CI: Cargo Dev / macos-13, and 1 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODuOsJb2wHUXW4uYAJe:b8hwa5s) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*

`key_from_filename` is only exposed for tests, so it won't be breaking (plus nativelink follows breaking definition to be "if a user upgrades nativelink will they have to change anything or see behavioral changes between previous version").

Looking at the code, it should be fairly easy to address this, we can take in the store key type in the function as well. The `file_infos` will just need to also store and pass the key type to `process_entry` (or better yet `file_infos` might create the key first and pass the key directly into `process_entry`).
</blockquote></details>

I'm still unsure how to implement this: in `add_files_to_cache` the stream of filename / timestamp pairs are sorted by the `SystemTime` - both the digest keys and string keys I assume should be treated the same by this, however the `ReadDirStream`only produces the filename - I can add an extra field to distinguish which folder the filename came from, but there is then also the added complication of having to loop over both directories (or 3 if there is a need to find legacy cache files) - especially with the async stream over the files - would want to avoid allocating another `Vec` (plus issues regarding recursively making these subfolders)

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODuXxQg9csqZQknQ8kJ:b-ssvi0s) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*
<details><summary><i>Previously, KGrewal1 (Kirpal Grewal) wrote…</i></summary><blockquote>

I'm still unsure how to implement this: in `add_files_to_cache` the stream of filename / timestamp pairs are sorted by the `SystemTime` - both the digest keys and string keys I assume should be treated the same by this, however the `ReadDirStream`only produces the filename - I can add an extra field to distinguish which folder the filename came from, but there is then also the added complication of having to loop over both directories (or 3 if there is a need to find legacy cache files) - especially with the async stream over the files - would want to avoid allocating another `Vec` (plus issues regarding recursively making these subfolders)
</blockquote></details>

Pushed a version which currently ignores outdated cache files (ie ones stored in old location) but otherwise uses subfolders to distinguish the types

allada (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: 0 of 2 LGTMs obtained, and 2 of 4 files reviewed, and pending CI: Cargo Dev / macos-13, and 3 discussions need to be resolved

___
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODydJQY6lrhCLkvudqs:b-swi9p4) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*
<details><summary><i>Previously, KGrewal1 (Kirpal Grewal) wrote…</i></summary><blockquote>

Pushed a version which currently ignores outdated cache files (ie ones stored in old location) but otherwise uses subfolders to distinguish the types
</blockquote></details>

Yeah, this logic is one-time startup cost, so doing additional allocations and such are fine. Whatever it takes to make the code more readable and easier to use.
___
*[`nativelink-store/src/filesystem_store.rs` line 484 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODye4fV1Dp0wBYyY4Km:-ODye4fV1Dp0wBYyY4Kn:b-1axu3r) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L484)):*
> ```XML
> 
>     async fn read_files(
>         filetype: FileType,
> ```

nit: can we pass in the `folder` here and remove `filetype` here.

The caller knows the `FileType`, so just store `Digest` and `Str` as different variables and zip the `FileType` during the for-loop.

You should also be able to support the reverse compatibility mode trivially with this change.
___
*[`nativelink-store/src/filesystem_store.rs` line 538 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODyf18V7Z2gyj0zs6bM:-ODyf18V7Z2gyj0zs6bN:b-ur5rpv) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L538)):*
> ```XML
> 
>     file_infos.extend(digest_infos);
>     file_infos.sort_by(|a, b| a.2.cmp(&b.2));
> ```

nit: Side note: I have been wanting to remove all the `atime` stuff, as it really doesn't provide any value and costs a lot to maintain, so don't worry too much about perfection here, even feel free to remove the tests that verify this.

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 484 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODye4fV1Dp0wBYyY4Km:-ODzgZ6v5OumyB4h7oss:bq21ytz) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L484)):*

nit: can we pass in the `folder` here and remove `filetype` here.

The caller knows the `FileType`, so just store `Digest` and `Str` as different variables and zip the `FileType` during the for-loop.

You should also be able to support the reverse compatibility mode trivially with this change.
</blockquote></details>

Changed - looking through all three locations for files now

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 538 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODyf18V7Z2gyj0zs6bM:-ODzl6fI1B5oBVuqY2aQ:bfhcip3) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L538)):*

nit: Side note: I have been wanting to remove all the `atime` stuff, as it really doesn't provide any value and costs a lot to maintain, so don't worry too much about perfection here, even feel free to remove the tests that verify this.
</blockquote></details>

Kept the atime values but only per folder - it didn't make the code much more difficult so felt better to keep it in given that

KGrewal1:
*[`nativelink-store/src/filesystem_store.rs` line 152 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODqXw8I2-y_8H0m3oNO:-ODw8OkWDKYsbRAsWg8q:b-5qxd55) ([raw file](https://github.com/TraceMachina/nativelink/blob/f7fe0294de6586937ac5756fc6592f6865656c76/nativelink-store/src/filesystem_store.rs#L152)):*

Yeah, this logic is one-time startup cost, so doing additional allocations and such are fine. Whatever it takes to make the code more readable and easier to use.
</blockquote></details>

Fully working now with the subfolders and reading the old cache

KGrewal1:
> _[`nativelink-store/src/filesystem_store.rs` line 538 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODyf18V7Z2gyj0zs6bM:-ODzl6fI1B5oBVuqY2aQ:bfhcip3) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L538)):_
> Previously, allada (Nathan (Blaise) Bruer) wrote…
> 
> > nit: Side note: I have been wanting to remove all the `atime` stuff, as it really doesn't provide any value and costs a lot to maintain, so don't worry too much about perfection here, even feel free to remove the tests that verify this.
> 
> Kept the atime values but only per folder - it didn't make the code much more difficult so felt better to keep it in given that

I have removed this as the tests were working locally but failing in CI - corresponding test was `oldest_entry_evicted_with_access_times_loaded_from_disk()` that ensured that on startup only the newest file stayed in cache

allada (changes requested):
Sorry for the back and fourth. Everything looks good, if we just get that reverse-compatibility + upgrade path sorted we are all done.

Thanks :smile: 

> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-OE04F939ONExII9YWeU:bpwdu30)* status: 0 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Cargo Dev / macos-13, Installation / macos-13, Installation / macos-14, NativeLink.com Cloud / Remote Cache / macos-14, macos-13, and 2 discussions need to be resolved (waiting on @allada)

___
*[`nativelink-store/src/filesystem_store.rs` line 490 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-OE01OouDzmr3Ymf9RWZ:-OE01OouDzmr3Ymf9RW_:b-daflc5) ([raw file](https://github.com/TraceMachina/nativelink/blob/199e4635aa78c1a00773d230c40068c2740967f0/nativelink-store/src/filesystem_store.rs#L490)):*
> ```XML
>             format!("{}/{folder}/", shared_context.content_path) // read the current cache
>         } else {
>             format!("{}/", shared_context.content_path) // read old cache files
> ```

nit: Please remove these comments but add one more above this:
```rs
// Note: In Dec 2024 this is for backwards compatibility with the old
// way files were stored on disk. Previously all files were in a single
// folder regardless of the StoreKey type. This allows old versions of
// nativelink file layout to be upgraded at startup time.
// This logic an be removed once more time has passed.
```
___
*[`nativelink-store/src/filesystem_store.rs` line 582 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-OE03bZO7vdJhQLZRwfO:-OE03bZO7vdJhQLZRwfP:br2xmlk) ([raw file](https://github.com/TraceMachina/nativelink/blob/199e4635aa78c1a00773d230c40068c2740967f0/nativelink-store/src/filesystem_store.rs#L582)):*
> ```XML
>     }
> 
>     add_files_to_cache(evicting_map, anchor_time, shared_context, block_size, None).await?;
> ```

After reading this and thinking about it, I think we slightly over thought this. I think we can just move all files in the legacy folder and to the `DIGEST_FOLDER` then do the rest of the logic. This would likely simplify the code a bit too.

KGrewal1:
If we're moving the files, it may be best to do it in the `new_with_timeout_and_rename_fn` so the appropriate `rename_fn` is used?

MarcusSorealheis:
Can you identify?

Marcus Eagan

KGrewal1:
Have a working (locally at least) version using the specified rename function to move old cache files from their previous location to under `DigestFolder`

aaronmondal:
> assume this broke with dependency upgrade
I can reproduce with cargo. Working on finding a fix. In the meantime, the Bazel test still works.

KGrewal1:
> > assume this broke with dependency upgrade
> 
> I can reproduce with cargo. Working on finding a fix. In the meantime, the Bazel test still works.
> 
> ```shell
> bazel test ...
> ```

Thanks - managed to get a working version now too (the advantage of `cargo test` was so I could have had more fine grained control of which tests to run / no-capture to see debug output, compared to `bazel test //...`)

allada (dismissed):
Thanks! :lgtm:

> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-OE0nWlZ6Wb_33tB097-:b-ftzx47)* status: 2 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, and 1 discussions need to be resolved

___
*[a discussion](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-OE0nZbE2TCbUeHro6ab:-OE0nZbE2TCbUeHro6ac:bzgrm2p) (no related file):*
@aaronmondal, can you please double check this.

KGrewal1 (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: 2 of 2 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-store/src/filesystem_store.rs` line 538 at r7](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-ODyf18V7Z2gyj0zs6bM:-ODzwBrpAvmTvEWUQTYG:brflphu) ([raw file](https://github.com/TraceMachina/nativelink/blob/1003b6a1b6b28c7b397ee15626e009557441fcf4/nativelink-store/src/filesystem_store.rs#L538)):*
<details><summary><i>Previously, KGrewal1 (Kirpal Grewal) wrote…</i></summary><blockquote>

I have removed this as the tests were working locally but failing in CI - corresponding test was `oldest_entry_evicted_with_access_times_loaded_from_disk()` that ensured that on startup only the newest file stayed in cache 
</blockquote></details>

This actually appears to succeed locally and fail in CI - the corresponding test is `oldest_entry_evicted_with_access_times_loaded_from_disk() ` to ensure ordering was done on `atime` on insertion - as such will remove the test and corresponding sort

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-OE0uqhNCFgS2FpANpd-:bnfp4nl)* status: 2 of 2 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

aaronmondal (approved):
:lgtm:

> Reviewed 1 of 2 files at r10, 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-OE0xePs1t_ob3P2GzHz:bnfp4nl)* status: 2 of 2 LGTMs obtained, and all files reviewed, and pending CI: Analyze (javascript-typescript), Cargo Dev / ubuntu-22.04, Coverage, asan / ubuntu-22.04, pre-commit-checks, ubuntu-20.04 / stable, ubuntu-22.04, and 2 discussions need to be resolved

___
*[`-- commits` line 4 at r13](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-OE0wTuM-K5vOIFgCaEd:-OE0wTuM-K5vOIFgCaEe:b3fwv24):*
nit: Looks like a leftover comment

aaronmondal (approved):
@KGrewal1 Great work :heart:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489#-:-OE0yFoc6QpwZQb2Kvzx:bdq4t2x)* status: 2 of 2 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Bazel Dev / ubuntu-24.04, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / macos-14, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 2 discussions need to be resolved

aaronmondal (approved):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1489)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained, and all files reviewed

PR #1503, opened 2024-11-27
Fix clippy::implicit_hasher

SchahinRohani:
## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)


aaronmondal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Cargo Dev / macos-13, Installation / macos-13, NativeLink.com Cloud / Remote Cache / macos-14, Remote / large-ubuntu-22.04, macos-13, and 1 discussions need to be resolved

___
*[`nativelink-util/src/common.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-OCjqgo7Ee6RWpie2OVA:-OCjqgo7Ee6RWpie2OVB:ba9ux0u) ([raw file](https://github.com/TraceMachina/nativelink/blob/75629591042c69f933c5740ae9322b7999fe6550/nativelink-util/src/common.rs#L438)):*
> ```XML
> 
> impl<K: std::cmp::Eq + std::hash::Hash, T, S: BuildHasher> HashMapExt<K, T> for HashMap<K, T, S> {
>     fn try_map<F, U>(self, f: F) -> Result<HashMap<K, U>, Error>
> ```

Doesn't this need to be `Result<HashMap<K, U, S>, Error>`?

SchahinRohani (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)* status: 0 of 1 LGTMs obtained, and 1 of 3 files reviewed, and pending CI: pre-commit-checks, and 1 discussions need to be resolved

___
*[`nativelink-util/src/common.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-OCjqgo7Ee6RWpie2OVA:-OCndBbbEp-dF5oHy6YO:b-1sihpc) ([raw file](https://github.com/TraceMachina/nativelink/blob/75629591042c69f933c5740ae9322b7999fe6550/nativelink-util/src/common.rs#L438)):*

Doesn't this need to be `Result<HashMap<K, U, S>, Error>`?
</blockquote></details>

Only if we want to keep the hasher type, yes. But in general i think this implementation could be optimized, since this operation is pure and shouldn't throw an error at all.

For every given HashMap and transformation we convert it to a new HashMap and this is always returning the same Result<HashMap<K, U>>.  

But this is out of scope for this PR, just fixing the linter warning.

SchahinRohani (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)* status: 0 of 1 LGTMs obtained, and 1 of 3 files reviewed, and pending CI: pre-commit-checks, and 1 discussions need to be resolved

___
*[`nativelink-util/src/common.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-OCjqgo7Ee6RWpie2OVA:-OCngYzL53FHyRX8g5l7:b-lm7fw6) ([raw file](https://github.com/TraceMachina/nativelink/blob/75629591042c69f933c5740ae9322b7999fe6550/nativelink-util/src/common.rs#L438)):*
<details><summary><i>Previously, SchahinRohani (Schahin) wrote…</i></summary><blockquote>

Only if we want to keep the hasher type, yes. But in general i think this implementation could be optimized, since this operation is pure and shouldn't throw an error at all.

For every given HashMap and transformation we convert it to a new HashMap and this is always returning the same Result<HashMap<K, U>>.  

But this is out of scope for this PR, just fixing the linter warning.
</blockquote></details>

Forget to mention, that we pass a transformation `std::ops::Fn(T)` into the try_map, so we know, the hasher type, so why keep it?

aaronmondal (changes requested):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / macos-14, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved

___
*[`nativelink-util/src/common.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-OCjqgo7Ee6RWpie2OVA:-OCnidMs4k1yYbkYoznf:bmriqib) ([raw file](https://github.com/TraceMachina/nativelink/blob/75629591042c69f933c5740ae9322b7999fe6550/nativelink-util/src/common.rs#L438)):*
<details><summary><i>Previously, SchahinRohani (Schahin) wrote…</i></summary><blockquote>

Forget to mention, that we pass a transformation `std::ops::Fn(T)` into the try_map, so we know, the hasher type, so why keep it?
</blockquote></details>

The `try_map` function implements the interface for `HashMapExt`, so we can't make this infallible, except if we ensure that we only make use of infallible functions, call `.map` instead of `.try_map` at callsites and remove this impl entirely.

We know `T`, but that only gives us the value type for the resulting hashmap. What we're after here is to inherit the hashing algorithm  `S` which is orthogonal to that.

I'd say we have two options:

1. Remove this impl altogether and ensure that we only use infallible functions in the entire codebase if we use `.map` on hashmaps. I don't think that would be a good idea as it would make it impossible to project fallible functions to hashmaps.
2. Properly make the hashing algorithm configurable by passing `S` to the inner hashmap. Keep in mind that this is about making the hashing algorithm itself configurable, not the value type of the map (which is configurable via `T` anyways).

If we don't do `2` it would mean that we forget the hashing algorithm and always fall back to SipHash which is a bug that this lint exposed. The correct value for `S` at callsites will almost always be https://docs.rs/gxhash/latest/gxhash/struct.GxBuildHasher.html

SchahinRohani (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503)* status: 0 of 1 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Cargo Dev / macos-13, Cargo Dev / ubuntu-22.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / macos-14, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, ubuntu-20.04 / stable, ubuntu-22.04, ubuntu-22.04 / stable, windows-2022 / stable, and 1 discussions need to be resolved

___
*[`nativelink-util/src/common.rs` line 438 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-OCjqgo7Ee6RWpie2OVA:-OCnnSGj2cefph8wxTDF:b-c9vs11) ([raw file](https://github.com/TraceMachina/nativelink/blob/75629591042c69f933c5740ae9322b7999fe6550/nativelink-util/src/common.rs#L438)):*

The `try_map` function implements the interface for `HashMapExt`, so we can't make this infallible, except if we ensure that we only make use of infallible functions, call `.map` instead of `.try_map` at callsites and remove this impl entirely.

We know `T`, but that only gives us the value type for the resulting hashmap. What we're after here is to inherit the hashing algorithm  `S` which is orthogonal to that.

I'd say we have two options:

1. Remove this impl altogether and ensure that we only use infallible functions in the entire codebase if we use `.map` on hashmaps. I don't think that would be a good idea as it would make it impossible to project fallible functions to hashmaps.
2. Properly make the hashing algorithm configurable by passing `S` to the inner hashmap. Keep in mind that this is about making the hashing algorithm itself configurable, not the value type of the map (which is configurable via `T` anyways).

If we don't do `2` it would mean that we forget the hashing algorithm and always fall back to SipHash which is a bug that this lint exposed. The correct value for `S` at callsites will almost always be https://docs.rs/gxhash/latest/gxhash/struct.GxBuildHasher.html
</blockquote></details>

Let's take a step back. My question here is: Do we want to have more `flexibility` by passing impure/fallible functions or `reproducibility` and only pass "infallible"/pure functions ?  Or do we might want to have both?

Then we could just implement `.safe_map` & `.unsafe_map` fns.

aaronmondal (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1503#-:-OCoBCbJAgOq9qi9vngA:bnfp4nl)* status: 1 of 1 LGTMs obtained, and all files reviewed, and pending CI: Bazel Dev / macos-13, Cargo Dev / macos-13, Coverage, Installation / macos-13, Remote / large-ubuntu-22.04, macos-13

PR #1547, opened 2024-12-16
Add `GCSStore`

asr2003:
# Description

Fixes # (issue)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended


asr2003:
@SchahinRohani Here's the PR with approach we discussed to add GCS Store

MarcusSorealheis:
@asr2003 could convert to a Draft PR until ready?

nativelink-store/Cargo.toml
```diff
@@ -57,6 +57,10 @@ tokio-util = { version = "0.7.13" }
 tonic = { version = "0.12.3", features = ["transport", "tls"], default-features = false }
 tracing = { version = "0.1.41", default-features = false }
 uuid = { version = "1.11.0", default-features = false, features = ["v4", "serde"] }
+reqwest = { version = "0.12.9", features = ["json", "gzip", "stream"]}
+google-cloud-storage = "0.23.0"
+percent-encoding = "2.3.1"
```
MarcusSorealheis: why do we need to add this dependency?

nativelink-store/src/default_store_factory.rs
```diff
@@ -51,6 +52,7 @@ pub fn store_factory<'a>(
         let store: Arc<dyn StoreDriver> = match backend {
             StoreSpec::memory(spec) => MemoryStore::new(spec),
             StoreSpec::experimental_s3_store(spec) => S3Store::new(spec, SystemTime::now).await?,
+            StoreSpec::experimental_gcs_store(spec) => GCSStore::new(spec).await?,
```
MarcusSorealheis: Everywhere there is a divergence from the existing store, I'd like to understand why. I'm sure there is a good reason.

nativelink-store/src/gcs_store.rs
```diff
@@ -0,0 +1,554 @@
...
+/// streamed to Google Cloud Storage (GCS). Each chunk of data is broadcasted to all subscribers
+/// (e.g., for retry logic or parallel consumers). The `Stream` implementation makes it compatible
+/// with APIs like `Body::wrap_stream` which simplifies the upload process.
+///
+/// # Benefits
+/// - Converts the push-based `broadcast::Receiver` into a pull-based `Stream`, enabling compatibility
+///   with `Stream`-based APIs.
+/// - Handles error cases gracefully, converting `RecvError::Closed` to `None` to signal stream completion.
+/// - Allows multiple consumers to read from the same broadcasted data stream, useful for parallel processing.
+///
```
MarcusSorealheis: Please add a `# How is this Different from the S3 Store Implementation"` section for improved maintainability and review-ability. I'm sure there are good reasons, as I said, but I'd like to understand in English prior before juxtaposition in Reviewable or my IDE.

aaronmondal (commented):
Thanks a lot! Before we move on to a more in-depth review, please ensure the following:

- Don't import openssl or reqwest as they pull in too many dependencies and won't easily build into a statically linked executable. GCS supports gRPC, so let's go with a tonic implementation (I believe the protos are here: https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto and there is infra for it in the `nativelink-proto` directory).
- Please ensure that we have test coverage for core functionality. The scope of this should be at least what the S3Store has. 
- Regarding the failing CI jobs, something that might be helpful as well is to take a look at the https://www.nativelink.com/docs/contribute/guidelines#local-development-setup which comes with the exact versions and environment that our builds run in and the `bazel test //...` command which is what most of CI invokes. See also some details on toolchains etc here: https://www.nativelink.com/docs/contribute/bazel#test. Tbh I'm not sure whether the cargo build can generate protos at all :sweat_smile: My guess is that this part will only work with Bazel.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1547#-:-OEEwcPB1tviBRXMWG8s:bckxnxq)* status: 0 of 2 LGTMs obtained, and 0 of 11 files reviewed, and pending CI: Bazel Dev / ubuntu-24.04, Cargo Dev / ubuntu-22.04, Coverage, Installation / ubuntu-22.04, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), and 3 discussions need to be resolved

asr2003:
I have discussed with @SchahinRohani  evaluating all approaches `tonic/grpc` , ` http/rest`  and `google-cloud-storage` crate

As what we finalised is to use `google-cloud-storage` as it can simplify our work than reinventing the wheel It's been actively managing and have good community use of it

To work with bazel , I will agree with you @aaronmondal 
```
Don't import openssl or reqwest as they pull in too many dependencies and won't easily build into a statically linked executable. GCS supports gRPC, so let's go with a tonic implementation (I believe the protos are here: https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto and there is infra for it in the nativelink-proto directory).
and this
Tbh I'm not sure whether the cargo build can generate protos at all 😅 My guess is that this part will only work with Bazel.
```


Ahh, while the google-cloud-storage crate simplifies things, It seems the tonic/gRPC approach is a cleaner and more future-proof solution for interacting with GCS without pulling in unnecessary dependencies (reqwest/openssl).

If Bazel and CI integration for tonic/protobuf generation is feasible I will rewrite them to this. Can I ahead with this @aaronmondal

SchahinRohani:
First of all, thank you for your effort @asr2003.
We are looking for a grpc solution as @aaronmondal mentioned. Therefore the proposed solution is not suitable.

asr2003:
Dear reviewers,

Now `GCS Store` implementation closely mirrors the `S3` version but adapts it to the `GCS gRPC API` and `metadata` structure. It aligns with S3's structure and behavior while leveraging GCS-specific features. And it also eliminates manual generating of `protobufs `

- S3 manages concurrency using a bounded `mpsc::channel` and `FuturesUnordered`. GCS currently processes chunks sequentially in the provided implementation.. GCS lacks native support for concurrent part uploads in a single upload session which can be a performance bottleneck for very large files.
- GCS sequential resumable uploads are better suited for cases where data integrity and fault tolerance are more critical than performance and to support concurrency in uploads may require out of box solution
- In next iteration, I am going to add tests similar of s3 store 

This are my views and any reviews and suggestions are much appreciated.

aaronmondal:
@asr2003 Looks much nicer than the previous implementation!

Some things I'm noticing:

- I wasn't aware of the google-tonic-xxx crates. Since we're only talking to an external endpoint I think it's fine to not generated the protos via https://github.com/TraceMachina/nativelink/blob/main/nativelink-proto/BUILD.bazel, so I'd say we're good on that part.
- The authentication might need some work. I'm not sure how GCS handles this, but IIUC currently the only way to handle auth is via the config parameter file. The S3Store uses a credential provider: https://github.com/TraceMachina/nativelink/blob/f672af7d79ed8ab60e0b7f703aa625cba528e300/nativelink-store/src/s3_store.rs#L271-L294 It would be good to have some mechanism that works with an environment variable, similar to the S3 store as this makes it easier to handle the setup uniformly in cloud environments.
- Ensure that we have decent test coverage. Doesn't need to be a 100% perfect branch-based coverage setup, but the core functionality should be tested (IIRC the OSSF guidelines recommend 80%). The way we generate Coverage in CI is like this: https://www.nativelink.com/docs/contribute/guidelines#generating-code-coverage. This can be somewhat slow during development so you might want to take a look at this as well which is less precise but faster: https://bazel.build/configure/coverage

MarcusSorealheis:
In the first post on the issue, there is an issue template. You should complete it like the other issues.

MarcusSorealheis (changes requested):
> Please include a summary of the changes and the related issue. Please also

This is a placeholder for what you are supposed to add at the top of the issue.

asr2003:
@aaronmondal The GCS store  is tightly coupled with the `StorageClient<Channel>` , So that I am unable to mock, I think I need to rewrite Store to implement abstract Trait for StorageClient that enable us to mock GCS Store for testing but I am not sure whether that makes sense to go with but will try with it an update here

palfrey:
Obsoleted by https://github.com/TraceMachina/nativelink/pull/1645

PR #1549, opened 2024-12-17
Implement Google Cloud Storage Store

onyedikachi-david:
WIP

### DIFF
--- a/deployment-examples/gcs-store-example.yaml
+++ b/deployment-examples/gcs-store-example.yaml
@@ -1,0 +1,13 @@
+stores:
+  gcs_store:
+    type: "gcs_store"
+    bucket: "my-bucket"
+    key_prefix: "my-prefix/"
+    consider_expired_after_s: 3600
+    retry:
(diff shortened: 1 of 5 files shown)


onyedikachi-david:
@aaronmondal are you okay with this? so i can go ahead.

onyedikachi-david:
> @onyedikachi-david Generally looks like it's the right direction. Some things to be aware of:
> 
> - Please ensure that you're following our contribution guidelines and that things wiork with both Bazel and Cargo: https://www.nativelink.com/docs/contribute/guidelines, https://www.nativelink.com/docs/contribute/bazel. We have fairly strict equirements for perfect reproducibility. Without the Nix setup it'll be very hard to pass CI.
> - You can generate the proto libraries by extending this: https://github.com/TraceMachina/nativelink/blob/main/nativelink-proto/BUILD.bazel.
> - Tests should go into a separate file to optimize for incremental rebuilds.

Thanks for the feedback. I should be done on or after the weekend.

PR #1177, opened 2024-07-19
Fix:Prevent CAS and AC from using the same store and add integration …

RickDeb2004:
…test

# Description

This commit includes changes to prevent CAS and AC from being configured to use the same store, which could cause conflicts. The conflict check logic has been optimized using a hashmap for better performance. Additionally, an integration test script has been added to verify the conflict detection functionality.

Fixes # (issue)
issue: #768 
## Type of change

Please delete options that aren't relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

- [x] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?

Verified the service logs for correct error messages when CAS and AC are configured with the same store.
Created and executed an integration test script `simple_store_conflict_test.sh` to automate the verification process.

## Checklist

- [ ] Updated documentation if needed
- [ yes] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [yes ] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1177)

PR #1555, opened 2024-12-29
GH-768 Store config uniqueness enforced by StoreManager

### LINKED ISSUE #768: NativeLink Should Not Start if CAS and AC are the Same
If users try to start NativeLink configuring CAS and AC to use the same store, it should fail and provide a helpful error.

mkeen:
# Description

Ensuring that when new Stores are added to StoreManager, that `memory`, `experimental`, `filesystem`, `grpc`, and `redis_store` stores will not experience overlap within a `StoreManager` instance. The same data store may be used for multiple values for the listed types only if relevant `StoreSpec` configuration values differ (such as connection details, key namespace, etc). Dictated by `disallow_duplicates_digest` implementation being present on the relevant `StoreSpec` enum type.

Primary changes:

- Added a new helper: `make_and_add_store_to_manager` to make sure the `store_factory` function is not used for config that overlaps on enforced Store entities
- Fortified `add_store` so that duplicate-named items cannot be added
- Ensuring that unique-enforced `StoreSpec` entities are defined in `stores.rs` and enforced by `store_manager.rs`.
- Ensured that `StoreSpec` entities that don't need to be unique aren't encumbered by uniqueness enforcement

DRAFT STATUS: Need to finalize tests and do a bit more cleanup where possible

Fixes #768

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
  not work as expected)
- [ x ] This change requires a documentation update

## How Has This Been Tested?

## Checklist

- [ x ] Tests added/amended
- [ x ] `bazel test //...`  passes locally

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1555)


nativelink-store/src/default_store_factory.rs
```diff
@@ -42,12 +42,44 @@ use crate::verify_store::VerifyStore;
...
+}
+
+fn store_factory<'a>(
     backend: &'a StoreSpec,
     store_manager: &'a Arc<StoreManager>,
     maybe_health_registry_builder: Option<&'a mut HealthRegistryBuilder>,
 ) -> Pin<FutureMaybeStore<'a>> {
     Box::pin(async move {
+        if let Some(backend_config_digest) = backend.disallow_duplicates_digest() {
+            store_manager.digest_not_already_present(&backend_config_digest)?;
```
mkeen: this is redundant.. i will remove

MarcusSorealheis:
It's ironic the tests are failing here because even in our own tests we make use of a non-optimal configuration. Thank you for coming and working on this issue.

nativelink-service/tests/ac_server_test.rs
```diff
@@ -48,29 +49,46 @@ async fn insert_into_store<T: Message>(
...
         "main_ac",
-        store_factory(
-            &StoreSpec::memory(MemorySpec::default()),
-            &store_manager,
-            None,
-        )
-        .await?,
-    );
+        &StoreSpec::filesystem(FilesystemSpec {
+            content_path: current_dir
```
MarcusSorealheis: This is too complicated. Would be simpler to keep `MemorySpec` here. I know it isn't pure but we don't have consistent way to test `FilesystemStore`. We need one.

MarcusSorealheis (changes requested):
Beyond simplification requested, everything looks good.

nativelink-service/tests/ac_server_test.rs
```diff
@@ -48,29 +48,28 @@ async fn insert_into_store<T: Message>(
...
     let digest = DigestInfo::try_new(hash, action_size)?;
     store.update_oneshot(digest, store_data.freeze()).await?;
+
     Ok(data_len.try_into().unwrap())
 }
 
 async fn make_store_manager() -> Result<Arc<StoreManager>, Error> {
     let store_manager = Arc::new(StoreManager::new());
-    store_manager.add_store(
+    make_and_add_store_to_manager(
```
MarcusSorealheis, line 57: For my understanding, what is the reason why you moved away from `store_factory` rather than modifying `store_factory`?

nativelink-service/tests/bep_server_test.rs
```diff
@@ -35,7 +35,7 @@ use nativelink_proto::google::devtools::build::v1::{
     PublishBuildToolEventStreamRequest, PublishLifecycleEventRequest, StreamId,
 };
 use nativelink_service::bep_server::BepServer;
-use nativelink_store::default_store_factory::store_factory;
```
MarcusSorealheis, line 38: this seems heavy at first read but maybe it is an improvement to both the naming and the functionality. `store_factory` as a name says nothing about what its function is, but developers could go look at the definition. `make_and_add_store_to_manager` is definitely clearer.

nativelink-service/tests/store_overlap_rules_test.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2024 The NativeLink Authors. All rights reserved.
```
MarcusSorealheis, line 1: PERFECT!

src/bin/nativelink.rs
```diff
@@ -12,11 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
+
+
```
MarcusSorealheis, line 17: what are these spaces for?

src/bin/nativelink.rs
```diff
@@ -12,11 +12,21 @@
...
 use std::collections::{HashMap, HashSet};
+
+
+
```
MarcusSorealheis, line 21: and these? I mean empty lines.

src/bin/nativelink.rs
```diff
@@ -12,11 +12,21 @@
...
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+
+
+
+
```
MarcusSorealheis, line 29: and these?

src/bin/nativelink.rs
```diff
@@ -45,7 +55,7 @@ use nativelink_service::cas_server::CasServer;
 use nativelink_service::execution_server::ExecutionServer;
 use nativelink_service::health_server::HealthServer;
 use nativelink_service::worker_api_server::WorkerApiServer;
-use nativelink_store::default_store_factory::store_factory;
```
MarcusSorealheis, line 48: @allada and @aaronmondal you two are going to need to sign off on moving away from `store_factory`. I would say this new name is more self-describing for sure.

MarcusSorealheis (commented):
Looking good @mkeen There's a few things to tidy up and I also want one of two other maintainers to look at the rename and deletion of `store_factory`. After lots of thought, I'm in favor of that change, but we simply need to boost awareness.

I want to think just for a few more hours also about what the error text should say when there are conflicting store configurations. What you have is fine, but can it be improved. I'm not sure.

src/bin/nativelink.rs
mkeen: The big thing here guys is that `make_and_add_store_to_manager ` enforces that all factory-built stores are part of a StoreManager so it's less likely that an unmanaged Store would be hanging around. Still possible to build them separately and pass into `add_store` this way though (useful for testing and probably other reasons). The old `store_factory` is still there and used by the new `make_and_add_store_to_manager ` function, but `store_factory` is no longer public in this changeset.

src/bin/nativelink.rs
mkeen: I will remove these and make sure formatting gets run again. My mistake on these spaces here.

nativelink-service/tests/bep_server_test.rs
mkeen: `store_factory` is still used but I made private and think that `make_and_add_store_to_manager` is better as a front-facing API because it makes and associates the Store with a `StoreManager` -- rather than leaving them as separate steps. The door is still open to manually build any Store instance though and call `add_store` on `StoreManager`.

nativelink-service/tests/ac_server_test.rs
mkeen: Still using `store_factory` to build the Store, but wrapping up in `make_and_add_store_to_manager` because it makes building and adding a Store to a StoreManager a single step from a public-facing perspective.

nativelink-store/src/store_manager.rs
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 use std::collections::HashMap;
+use std::ptr;
```
mkeen, line 16: We don't need this.. will remove

nativelink-store/src/store_manager.rs
```diff
@@ -22,18 +24,56 @@ use parking_lot::RwLock;
...
+        if stores.contains_key(name) {
+            return Err(make_err!(
+                Code::AlreadyExists,
+                "a store with the name '{}' already exists",
+                name
+            ));
+        }
+
+        for existing_store in stores.values().into_iter() {
+            if ptr::eq(&store, existing_store) {
```
mkeen, line 50: Will remove this check. Thanks to rust's memory model and the fact that in this design, ownership of the struct instance is passed to the StoreManager on line 58. We don't have Store instance pointer being passed around with a risk of passing the same pointer N times. We have a single instance of which ownership is transferred exactly once in guaranteed fashion (thanks Rust).

So instance protection is no longer a goal or needed. But I do think the config overlap detection is valuable.

MarcusSorealheis: Thanks Rust is an underrated comment.

allada (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1555)* status: 0 of 2 LGTMs obtained, and 0 of 14 files reviewed, and pending CI: Bazel Dev / macos-13, Bazel Dev / macos-14, Bazel Dev / ubuntu-24.04, NativeLink.com Cloud / Remote Cache / macos-14, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, asan / ubuntu-22.04, docker-compose-compiles-nativelink (22.04), integration-tests (22.04), macos-13, pre-commit-checks, ubuntu-22.04, and 12 discussions need to be resolved

___
*[`nativelink-config/src/stores.rs` line 435 at r11](https://reviewable.io/reviews/TraceMachina/nativelink/1555#-OFT2Atu8r3RMKrvrdJY:-OFT2Atv39ElOmf-0ZcK:b-n1np06) ([raw file](https://github.com/TraceMachina/nativelink/blob/aff36b83090fa440aaa3ce90c6b5baa688f1162c/nativelink-config/src/stores.rs#L435)):*
> ```XML
> }
> 
> impl StoreSpec {
> ```

nit: We don't allow complex functions in the config. We use these configs as simple definitions with some macro magic (like number->string conversions and such in serde).

palfrey:
This was originally here to solve https://github.com/TraceMachina/nativelink/issues/768 which is now fixed by https://github.com/TraceMachina/nativelink/pull/1177. Given it's been a few years, the conflicts here and that issue is now solved, I'm closing this PR. Feel free to reopen if you wish to sort out the other features this PR provided.

Issue #659, opened 2024-02-08
💎 Implement Google Cloud Bucket Store


steedmicro:
We need to implement Google Cloud Bucket Store.
We can refer Amazon S3 Bucket Store code - `nativelink-store/src/s3_store.rs` as a reference which implements similar logic we are going to implement.
Unit tests and example config files need to be added.

aaronmondal:
IIRC when @allada and I talked about this a while ago we came to the conclusion that something like this might make sense:

1. Migrate the current aws sdk to a 1.x version as the one we currently use is outdated
2. Generalize the S3 store to something like a `blobstore`, i.e. a wrapper store where we can plug in different highly specialized/tuned backend implementations for AWS, GCP, Azure etc.
3. Implement the actual GCP store and potentially implement one for Azure as well

One major hurdle I can currently see is that these stores can be tricky to test. Ideally we'll want some sort of short lived buckets for integration tests at some point. Technically I'd have a setup for CubeFS in an ad-hoc K8s cluster available, but while that might be useful for some API tests, that setup most likely can't mirror the actual behavior of AWS, GCP etc with respect to things like HTTP/HTTP2 and other not-so-obvious pitfalls.

@steed924 Consider what i mentioned here as a sort of 'wishful thinking out loud' rather than requirements - the current S3 store is not exactly in optimal shape, so anything in this area will probably be better than the current implementation 😅

steed924:
Thanks. I will keep in mind that.  +@aaronmondal.

steed924:
FYI, while I was working on this issue, I encountered one problem while I was tryig to evaluate nativelink running with S3 bucket.
I was working on Ubuntu VPS and set aws credentials correctly and set environment variables for them also.
Even after I've tried run the command after `bazel clean --expunge` the result was the same.

Running the nativelink by `bazel run nativelink -- $(pwd)/nativelink-config/examples/s3_backend_with_local_fast_cas.json` seems to run server correctly.

But when I was going to evaluate the native link running by this command ` bazel test //...   --remote_instance_name=main   --remote_cache=grpc://127.0.0.1:50051   --remote_executor=grpc://127.0.0.1:50051   --remote_default_exec_properties=cpu_count=1`, it raises errors like this.

It raises an error while it tries to load the first chunk from ActionCache server, which is supported by S3 bucket.

I hope to hear your opinions.  @aaronmondal. Thanks.

[...]

PR #1639, opened 2025-03-24
Migrate S3Store to hyper 1.x

aaronmondal:
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1639)

PR #1645, opened 2025-03-26
Added implementation for Google Cloud Storage with REST

amankrx:
# Description

This PR implements the Google Cloud Store implementation and closely aligns with the AWS S3 implementation.
This is an update to my previous PR I raised using the gRPC implementation. There were multiple issues with that PR, and the primary issue was the stability. This PR fixes all the underlying issues we were facing with that PR. We are using the `google-cloud-storage` and `google-cloud-auth` crates. 

Note: To run the `gcs_backend.json5` example that I have provided, you should ensure that the path to the credentials file is specified in the `GOOGLE_APPLICATION_CREDENTIALS` environment variable. Check https://docs.rs/google-cloud-storage/0.24.0/google_cloud_storage/#automatically for more details.

Fixes https://github.com/TraceMachina/nativelink/issues/659
/claim https://github.com/TraceMachina/nativelink/issues/659

## Type of change

## How Has This Been Tested?

I did an E2E testing using the `gcs_backend.json5` example that I have provided. Along with that, I have set up a mocking framework for the GCS client and added a few tests for both the GCS store and the GCS client.

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)


MarcusSorealheis:
The only issue now is your private key.

amankrx:
> The only issue now is your private key.

Fixed it!

aaronmondal (changes requested):
I didn't get through all of it, but at the moment I'd say that this would introduce too much maintenance overhead considering that this is all a single store. If this is the amount of logic necessary to support a reliable GCS setup I'd honestly lean towards dropping this entirely.

cc @kubevalet AFAIU we already run on gcs via the filesystem store, right? Would it be viable to just recommend people to use that instead and improve documentation around that? I'm trying to understand the tradeoff that we'd be making here in terms of maintainability vs maintenance overhead. Are there usecases where one "has" to use GCS?

@amankrx Maybe my concerns are unreasonable. Have you explored ways to make this... well let's say "less code"?

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OMIy58TCyyhm6PoS6XG:bt2cbh)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 12 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdbhWCpQjGzNO4D7x:-OMIdbhWCpQjGzNO4D7y:b-m59sh1):*
"Add" instead of "Added"
___
*[`nativelink-store/Cargo.toml` line 31 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIfqB43rOHxxNJnvGU:-OMIfqB43rOHxxNJnvGV:b5guqwj) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/Cargo.toml#L31)):*
> ```TOML
> bytes = { version = "1.10.0", default-features = false }
> bytes-utils = { version = "0.1.4", default-features = false }
> chrono = "0.4.39"
> ```

Add this (and the other creates) with `x = { version = "y", default -features = false }` by default to prevent pulling in unnecessary deps.
___
*[`nativelink-store/src/gcs_client/client.rs` line 84 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMItvQJ-ew8S7OPgB8P:-OMItvQJ-ew8S7OPgB8Q:brraz4h) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L84)):*
> ```XML
>     pub async fn new_with_http_client(
>         spec: &GcsSpec,
>         client: hyper::Client<C>,
> ```

fyi this will need a hyper 1.x implementation after https://github.com/TraceMachina/nativelink/pull/1639. It might be easier to build on top of that so that we don't need migrate this one as well. The hyper 1.x migration is a bit complicated and we're still testing, but it should make it into main in the next few days.
___
*[`nativelink-store/src/gcs_client/client.rs` line 229 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIw1zm8LdpEZ4FyyXc:-OMIw1zm8LdpEZ4FyyXd:bs9rhf1) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L229)):*
> ```XML
>             return Err(make_err!(
>                 match status.as_u16() {
>                     401 | 403 => Code::PermissionDenied,
> ```

nit: Do these have typed names?
___
*[`nativelink-store/src/gcs_client/client.rs` line 315 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIwHnN83WwLhYv30Xe:-OMIwHnN83WwLhYv30Xf:b-2690sh) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L315)):*
> ```XML
>         }
> 
>         tokio::time::sleep(Duration::from_millis(200)).await; // Small delay to allow GCS to finalize
> ```

Let's make this a named const.
___
*[`nativelink-store/src/gcs_client/client.rs` line 642 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIx-aCEvKyh_Olscjg:-OMIx-aCEvKyh_Olscjh:b-3elj93) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L642)):*
> ```XML
> 
>         match status.as_u16() {
>             200 | 201 | 308 => Ok(()),
> ```

nit: as above
___
*[`nativelink-store/src/gcs_client/operations.rs` line 14 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIxaE_Bgv2rIDbZVzk:-OMIxaEaEDWjxL2hyCjK:bgzvyjv) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/operations.rs#L14)):*
> ```XML
> // See the License for the specific language governing permissions and
> // limitations under the License.
> ```

nit: Seems unnecessary to have this in a dedicated file.
___
*[`Cargo.lock` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdnGKDvXGoWUEvTl3:-OMIdnGKDvXGoWUEvTl4:btinl8z) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/Cargo.lock#L42)):*
> ```TOML
> 
> [[package]]
> name = "android_system_properties"
> ```

Whatever causes these `android_*` dependencies to be pulled in should probably be disabled. My guess is that it's some default feature in some cratej (`chrono` maybe?). Try adding the deps as `somedep = { version = "x.y.z", default-features = false, features = ["somefeature"] }` in Cargo.toml where you only enable the minimal set of features that we need.
___
*[`nativelink-store/src/gcs_store.rs` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIgGb-0GmCHOgvJfuv:-OMIgGb-0GmCHOgvJfuw:brg7rgl) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L42)):*
> ```XML
> };
> 
> struct ConnectionPool {
> ```

fyi We'll probably need to rework this as we're moving the entire `nativelink-store` setup to hyper 1.x (Current WIP branch at https://github.com/TraceMachina/nativelink/pull/1639). I general I like the explicitness of this approach though and I'll check whether this is something I should add to the aws store as well.
___
*[`nativelink-store/src/gcs_store.rs` line 204 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIipJr88UYoOwoZTt7:-OMIipJr88UYoOwoZTt8:bby2w06) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L204)):*
> ```XML
>     NowFn: Fn() -> I + Send + Sync + Unpin + 'static,
> {
>     async fn has_with_results(
> ```

This looks duplicated. We might get away with lifting this out of the aws implementation and reusing it in both places.
___
*[`nativelink-store/src/gcs_store.rs` line 272 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMImIt2Ah2dPo9HNCX8:-OMImIt2Ah2dPo9HNCX9:b-ieo040) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L272)):*
> ```XML
>         // For larger files, use resumable upload
>         // First, initiate the upload session
>         let upload_id = self
> ```

note to self: This looks wrong to me. I can't pinpoint it though. Will to evaluate in more detail.
___
*[`nativelink-store/src/gcs_store.rs` line 275 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIkE3N2DI7Ul7TG7AM:-OMIkE3N2DI7Ul7TG7AN:b39dbfd) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L275)):*
> ```XML
>             .retrier
>             .retry(unfold(object.clone(), move |object| async move {
>                 let conn = match self.connection_pool.acquire().await {
> ```

Hmm I'm not entirely sure on this one but I feel like the connection acquisition and retrial of that should be handled via the client side. Otherwise we'd have to "remember" to always acquire the permit in the calls. And it introduces fairly tight coupling between the store implementation and the underlying client.
___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OMIm_xc9xOVfxKZsaVy:b-bgzfff) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*
> ```XML
>             .retrier
>             .retry(unfold(
>                 (object.clone(), offset, end_offset),
> ```

I believe all of these `xx.clone()` calls in the unfolds (here and everywhere else) should be moves. It shouldn't be necessary to clone the objects or at least only do so in the minimal cases necessary.
___
*[`nativelink-store/src/gcs_client/connector.rs` line 83 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIxBIi-gXuI2gaoEYx:-OMIxBIi-gXuI2gaoEYy:b3wdgc6) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/connector.rs#L83)):*
> ```XML
>         Pin::new(&mut Pin::get_mut(self).connection).poll_shutdown(cx)
>     }
> }
> ```

fyi this will need to be changed after the hyper 1.x migration mentioned earlier.
___
*[`.gitignore` line 6 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdgDm6MPUeUb_dwqZ:-OMIdgDm6MPUeUb_dwq_:b3o83rq) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/.gitignore#L6)):*
> ```ignore
> .idea/
> .zed
> .idea/
> ```

This is duplicated from two lines above. Should be safe to remove.
___
*[`nativelink-store/src/gcs_client/auth.rs` line 35 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIrNlm9UFuD_SK5-gg:-OMIrNlm9UFuD_SK5-gh:bg1pnrm) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/auth.rs#L35)):*
> ```XML
> 
> #[derive(Debug, Serialize)]
> struct JwtClaims {
> ```

Is it actually necessary to roll our own implementation here? Are there reasons we can't use e.g. https://github.com/googleapis/google-cloud-rust/tree/main/src/auth or https://github.com/yoshidan/google-cloud-rust/tree/main/foundation/auth ?

For instance, if we used the official sdk, if we wanted to move the gcs store to non-experimental we could just wait until the official rust sdk reaches GA and then "bump a dependency".

I'm also worried that rolling our own here would require us to maintain perfect coverage and fuzzing as it would be "on us" to maintain the security guarantees of this implementation. This seems like a big lift. cc @kubevalet

amankrx (commented):
We can get this done with less code altogether. The primary issue is that we don't have a proper GCS SDK with a custom choice of client to use. So, the majority of the implementation is primarily the client itself. I'm not sure if we would be changing the client that much later on. So, it's up to us if we want a custom client like the one that the one I have provided, or to use third-party ones like https://github.com/abdolence/gcloud-sdk-rs
Also, it was initially less than 1000 lines of code PR, but the primary overhead was implementing traits and associated methods to add the unit tests. Apart from that, if it is absolutely necessary, I can work on reducing the unnecessary addition of extra modularity and concise it a bit.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OMJ-M72A6AZupOCPnCT:bjupvh1)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 12 discussions need to be resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdbhWCpQjGzNO4D7x:-OMJ1oA92r7gLb0YaGzC:b-kly7pp):*

"Add" instead of "Added"
</blockquote></details>

Sure. I will fix the commit message before the final merge.
___
*[`.gitignore` line 6 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdgDm6MPUeUb_dwqZ:-OMJ1w1Q2vt7aUcCNU9D:b-y1kkd6) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/.gitignore#L6)):*

This is duplicated from two lines above. Should be safe to remove.
</blockquote></details>

Ohh... This was added a while ago when I had raised the PR with the gRPC implementation. We didn't have `.idea` added at that time. I will remove it now.
___
*[`Cargo.lock` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdnGKDvXGoWUEvTl3:-OMJ2FcU7tCsPZIY167u:bypcudd) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/Cargo.lock#L42)):*

Whatever causes these `android_*` dependencies to be pulled in should probably be disabled. My guess is that it's some default feature in some cratej (`chrono` maybe?). Try adding the deps as `somedep = { version = "x.y.z", default-features = false, features = ["somefeature"] }` in Cargo.toml where you only enable the minimal set of features that we need.
</blockquote></details>

Yeah, looks like `chrono` added a number of crates unnecessarily. I will disable most of the features. 
___
*[`nativelink-store/src/gcs_store.rs` line 204 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIipJr88UYoOwoZTt7:-OMJ2ytmByM9_Yn6niOr:bfujdi0) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L204)):*

This looks duplicated. We might get away with lifting this out of the aws implementation and reusing it in both places.
</blockquote></details>

Yup, this is the same implementation we have for AWS. The primary implementation is what we have for the `has` method above. I would also argue that we can reuse the update and get_part methods, and the primary implementation can be shifted to a similar method we have as the `has` method. 
___
*[`nativelink-store/src/gcs_client/auth.rs` line 35 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIrNlm9UFuD_SK5-gg:-OMJ3wR_6nGGuFlCUcSU:bgrl75q) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/auth.rs#L35)):*

Is it actually necessary to roll our own implementation here? Are there reasons we can't use e.g. https://github.com/googleapis/google-cloud-rust/tree/main/src/auth or https://github.com/yoshidan/google-cloud-rust/tree/main/foundation/auth ?

For instance, if we used the official sdk, if we wanted to move the gcs store to non-experimental we could just wait until the official rust sdk reaches GA and then "bump a dependency".

I'm also worried that rolling our own here would require us to maintain perfect coverage and fuzzing as it would be "on us" to maintain the security guarantees of this implementation. This seems like a big lift. cc @kubevalet
</blockquote></details>

As I mentioned above, we can definitely use these crates, but they tend to pull in the `reqwest` library and we can't turn off that feature right away. If we decide to use that library, we wouldn't have to worry about the GCS client implementation though. 
___
*[`nativelink-store/src/gcs_client/client.rs` line 315 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIwHnN83WwLhYv30Xe:-OMJ4ePTAz15aBr79GXW:b-80tlqx) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L315)):*

Let's make this a named const.
</blockquote></details>

Sure!
___
*[`nativelink-store/src/gcs_client/operations.rs` line 14 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIxaE_Bgv2rIDbZVzk:-OMJ4qY2B9d9LEJXmKKS:bcxn8am) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/operations.rs#L14)):*

nit: Seems unnecessary to have this in a dedicated file.
</blockquote></details>

Can be moved to `client.rs`.

aaronmondal:
@amankrx Ah I think I get now where the aversion to reqwest is coming from :sweat_smile: You're probably referring to this: https://github.com/TraceMachina/nativelink/pull/1547#pullrequestreview-2506539052. What I meant by that is that we shouldn't *directly* depend on reqwest as we have a bunch of other mechanisms to make requests. If the google-cloud-sdk pulls that in it'd be fine as it would be similar to the aws sdk where we simply cannot control what official packages require. It's also possible to minimize the transitive dependencies by e.g. overriding the default features (although I wouldn't even consider it *that* problematic as long as we don't pull in any complex C/C++ dependencies).

So if we can get auth with the official sdk, even if that pulls in reqwest transitively it'd be totally fine as it's not us directly depending on reqwest.

This should significantly simplify this setup (and alleviate my security concerns as we can essentially treat it as "not our problem" if the official sdk has issues :laughing: This should help us "remove" the current `auth.rs` file.

If the new client implementation in #1639 is generic enough we should be able to remove the `connector.rs` from this PR as well (might need some adjustments and we don't actually know whether it works yet, but let's hope it does :crossed_fingers: - actually, did we ever try to point the aws implementation to GCS? Since we're doing REST now anyways, couldn't we potentially use almost 1:1 the aws store?).

Regarding the current `config.rs` the existing aws implementation will be changed/extended to support more methods anyways because of some requirements from #1630. This migtht allow us to consolidate this to some degree with something like a more generic "cloudSpec" or "authSpec" or something along these lines that we can reuse across the aws and gcp implementations.

There doesn't seem to be any easy way around the `mock.rs` implementation, but since that's test code I'd say it's fine.

With changes along these lines I'm optimistic that we can significantly reduce the maintenance overhead of the GCSStore.

MarcusSorealheis:
@amankrx the commit messages are identical @amankrx. Also, as a heads up you will need to squash the commits.

amankrx:
> @amankrx the commit messages are identical @amankrx. Also, as a heads up you will need to squash the commits.

Yup! I have added an extra commit now, such that it would be easier if we do revert the changes. The current implementation is based on the `google_cloud_store` and `google_cloud_auth` crates. The previous commit is the manual implementation without using these crates. 
Before merging the PR, I will squash the commits into a single commit. 

Also, there might be some CI failures. I will address some of the @aaronmondal 's comments as well by today. I just wanted to push these changes to get an idea if the current implementation seems to be in the right direction. I tried using AWS's SDK for GCS auth and stuff, but there were always some challenge with the trait bounds, or some other issue. If we really need that implementation, then I can dive a bit more deeper into it, but until then, please let me know if this implementation seems good enough? 
Also, regarding the large *lines of code*, I would say most of these are tests and mock implementations, the actual client and store implementation doesn't seem to be that large.

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 1 of 24 files reviewed, and pending CI: Bazel Dev / macos-15, Cargo Dev / macos-15, Coverage, Installation / macos-13, Installation / macos-14, Installation / macos-15, Local / lre-rs / macos-15, NativeLink.com Cloud / Remote Cache / macos-15, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, Web Platform Deployment / macos-15, macos-15, windows-2022 / stable, and 12 discussions need to be resolved

___
*[`Cargo.lock` line 42 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdnGKDvXGoWUEvTl3:-OOdFZ9z0Pfnldawx7cz:bfu1v4c) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/Cargo.lock#L42)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Yeah, looks like `chrono` added a number of crates unnecessarily. I will disable most of the features.
</blockquote></details>

Fixed it by removing most of the non-useful packages.
___
*[`nativelink-store/Cargo.toml` line 31 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIfqB43rOHxxNJnvGU:-OOdJPgC7vh_XJ5rnpba:b-896fja) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/Cargo.toml#L31)):*

Add this (and the other creates) with `x = { version = "y", default -features = false }` by default to prevent pulling in unnecessary deps.
</blockquote></details>

Done!
___
*[`nativelink-store/src/gcs_store.rs` line 204 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIipJr88UYoOwoZTt7:-OOdFrfY2FwrpeNzqcwZ:bxp69hy) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L204)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Yup, this is the same implementation we have for AWS. The primary implementation is what we have for the `has` method above. I would also argue that we can reuse the update and get_part methods, and the primary implementation can be shifted to a similar method we have as the `has` method.
</blockquote></details>

Reusing it at multiple places causes some issues like `has` method and `now_fn` to not get recognized. We can get away with it with a trait implementation, but that's just adding some extra complexity for not too big of a duplication. What do you suggest here? 
___
*[`nativelink-store/src/gcs_store.rs` line 272 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMImIt2Ah2dPo9HNCX8:-OOdJRmGDvKZw-JgIzgo:b-5jq577) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L272)):*

note to self: This looks wrong to me. I can't pinpoint it though. Will to evaluate in more detail.
</blockquote></details>

This works locally, but it would be great to get a follow up here!
___
*[`nativelink-store/src/gcs_store.rs` line 275 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIkE3N2DI7Ul7TG7AM:-OOdGiqH4BDVf9kB6Hlv:b-7ok9vq) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L275)):*

Hmm I'm not entirely sure on this one but I feel like the connection acquisition and retrial of that should be handled via the client side. Otherwise we'd have to "remember" to always acquire the permit in the calls. And it introduces fairly tight coupling between the store implementation and the underlying client.
</blockquote></details>

Removed the `ConnectionPool` from the GCS Store, and it is being handled in the client side now.
___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OOdGqm7-EAmUtnGYtiy:bi5qb21) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*

I believe all of these `xx.clone()` calls in the unfolds (here and everywhere else) should be moves. It shouldn't be necessary to clone the objects or at least only do so in the minimal cases necessary.
</blockquote></details>

Most of them are necessary otherwise, it is causing move errors. 
___
*[`nativelink-store/src/gcs_client/client.rs` line 84 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMItvQJ-ew8S7OPgB8P:-OOdH4nI9ljK3UTC-3XB:b-896fja) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L84)):*

fyi this will need a hyper 1.x implementation after https://github.com/TraceMachina/nativelink/pull/1639. It might be easier to build on top of that so that we don't need migrate this one as well. The hyper 1.x migration is a bit complicated and we're still testing, but it should make it into main in the next few days.
</blockquote></details>

Done!
___
*[`nativelink-store/src/gcs_client/client.rs` line 229 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIw1zm8LdpEZ4FyyXc:-OOdH9Jw-5UGCJktUUfK:bimfijr) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L229)):*

nit: Do these have typed names?
</blockquote></details>

Not sure about it! GCS returns the code as a `u64` and thus I am converting it to appropriate StatusCode.
___
*[`nativelink-store/src/gcs_client/client.rs` line 315 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIwHnN83WwLhYv30Xe:-OOdISsI3J-YqRoEfIlR:bqi5yyu) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L315)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Sure!
</blockquote></details>

Update! Moved most of these to named constants
___
*[`nativelink-store/src/gcs_client/client.rs` line 642 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIx-aCEvKyh_Olscjg:-OOdJZZCBJzMsx2Bc712:bfu9ujz) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L642)):*

nit: as above
</blockquote></details>

What exactly should we change here?
___
*[`.gitignore` line 6 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIdgDm6MPUeUb_dwqZ:-OOdIqIN9ofhiNo56vGx:b32n8sx) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/.gitignore#L6)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Ohh... This was added a while ago when I had raised the PR with the gRPC implementation. We didn't have `.idea` added at that time. I will remove it now.
</blockquote></details>

Removed it!
___
*[`nativelink-store/src/gcs_client/auth.rs` line 35 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIrNlm9UFuD_SK5-gg:-OOdIt-GDleqkLPdlrbd:b-dofyf0) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/auth.rs#L35)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

As I mentioned above, we can definitely use these crates, but they tend to pull in the `reqwest` library and we can't turn off that feature right away. If we decide to use that library, we wouldn't have to worry about the GCS client implementation though.
</blockquote></details>

Removed the manual auth implementation
___
*[`nativelink-store/src/gcs_client/connector.rs` line 83 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIxBIi-gXuI2gaoEYx:-OOdIwyP0OvDXIt6LxbB:b-896fja) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/connector.rs#L83)):*

fyi this will need to be changed after the hyper 1.x migration mentioned earlier.
</blockquote></details>

Done!
___
*[`nativelink-store/src/gcs_client/operations.rs` line 14 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIxaE_Bgv2rIDbZVzk:-OOdIzKp-lhyt87jr_Lj:bb5kmoo) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/operations.rs#L14)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Can be moved to `client.rs`.
</blockquote></details>

Update: Moved to `client.rs`

aaronmondal (commented):
I think this is looking quite nice! Maintainability-wise this seems in handleable territory. My main concern was the custom auth implementation, so not having that anymore alleviates most of those concerns :+1:

I think now it mostly comes down to getting the user-facing config part right. Ideally, users shouldn't need to "learn" multiple store configs and it seems that with your latest changes it's fairly easy to create a unified `BlobStore` at least on the config side. That'll give us the flexibility to easily change things around under the hood later.

+@jhpratt +@jaroeichler

> Reviewed 1 of 7 files at r2, 13 of 22 files at r5.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OOeQzxx1LGbPuS31osH:b-y3w2ev)* status: 0 of 2 LGTMs obtained, and 15 of 24 files reviewed, and pending CI: Remote / lre-cc / large-ubuntu-22.04, and 4 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-store/src/gcs_store.rs` line 204 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIipJr88UYoOwoZTt7:-OOeRERaBJLwjYukftC1:b8tvjkj) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L204)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Reusing it at multiple places causes some issues like `has` method and `now_fn` to not get recognized. We can get away with it with a trait implementation, but that's just adding some extra complexity for not too big of a duplication. What do you suggest here?
</blockquote></details>

cc @jhpratt Who might have some magical trick up their sleeve :magic_wand:
___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OOeQhiG2Cqfu4ogFnhK:b-opbfok) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Most of them are necessary otherwise, it is causing move errors.
</blockquote></details>

Ah I think the confusing part here is that the variable is named `object` instead of `object_path`. The latter makes it clearer that clones are harmless. (The former could give the impression that we're creating copies of the data).
___
*[`nativelink-store/src/gcs_client/client.rs` line 642 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIx-aCEvKyh_Olscjg:-OOeMWVl9htHHiOiZHuw:b8rzl4j) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_client/client.rs#L642)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

What exactly should we change here?
</blockquote></details>

Was probably referring to the const comment or the hyper 1.x comment.
___
*[`nativelink-config/src/stores.rs` line 748 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOeFAu73-WeRK7xhZIh:-OOeFAu73-WeRK7xhZIi:b-ihgd24) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-config/src/stores.rs#L748)):*
> ```XML
> #[derive(Serialize, Deserialize, Debug, Default, Clone)]
> #[serde(deny_unknown_fields)]
> pub struct CloudSpec {
> ```

Very nice! I think this is exactly the kind of change I think we need, which I originally mentioned here as well: https://github.com/TraceMachina/nativelink/issues/659#issuecomment-1938752804

I think we can actually consider going further and create a generic "ExperimentalBlobstore" or similar that has some sort of tuning parameter like a "provider" field implemented by an enum that switches the under-the-hood logic. On the code side this could potentially make it more obvious to see the places where the GCS implementation differs from the AWS implementation.

The concern here is the fact that we'll be breaking the config. It seems like we need this breakage either way, but let's try to keep the number of breaking commits minimal. I.e. I think we should first figure out a nice approach that also scales to other backends and do that change once for the config side. E.g. adding a field to an enum won't affect existing deployments, but renaming a field does.

Essentially, I'm thinking something like

```rust
ExperimentalBlobStore(ExperimentalBlobSpec)
```

Where

```rust
pub struct ExperimentalBlobSpec {
   // other fields
  provider: ExperimentalBlobStoreProvider
}

pub enum ExperimentalBlobStoreProvider {
  Aws,
  Gcs,
  Azure, // etc.
}
```

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 12 of 25 files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / macos-15, Bazel Dev / ubuntu-24.04, Cargo Dev / macos-15, Cargo Dev / ubuntu-24.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / macos-15, Installation / ubuntu-22.04, Installation / ubuntu-24.04, Local / lre-cc / ubuntu-24.04, Local / lre-rs / macos-15, Local / lre-rs / ubuntu-24.04, NativeLink.com Cloud / Remote Cache / macos-15, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, Web Platform Deployment / macos-15, Web Platform Deployment / ubuntu-24.04, asan / ubuntu-24.04, buildstream, integration-tests (24.04), macos-15, pre-commit-checks, ubuntu-24.04, ubuntu-24.04 / stable, windows-2022 / stable, and 4 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-config/src/stores.rs` line 748 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOeFAu73-WeRK7xhZIh:-OOhjh3PDat6XOxAPyJ1:b-4c1d8m) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-config/src/stores.rs#L748)):*

Very nice! I think this is exactly the kind of change I think we need, which I originally mentioned here as well: https://github.com/TraceMachina/nativelink/issues/659#issuecomment-1938752804

I think we can actually consider going further and create a generic "ExperimentalBlobstore" or similar that has some sort of tuning parameter like a "provider" field implemented by an enum that switches the under-the-hood logic. On the code side this could potentially make it more obvious to see the places where the GCS implementation differs from the AWS implementation.

The concern here is the fact that we'll be breaking the config. It seems like we need this breakage either way, but let's try to keep the number of breaking commits minimal. I.e. I think we should first figure out a nice approach that also scales to other backends and do that change once for the config side. E.g. adding a field to an enum won't affect existing deployments, but renaming a field does.

Essentially, I'm thinking something like

```rust
ExperimentalBlobStore(ExperimentalBlobSpec)
```

Where

```rust
pub struct ExperimentalBlobSpec {
   // other fields
  provider: ExperimentalBlobStoreProvider
}

pub enum ExperimentalBlobStoreProvider {
  Aws,
  Gcs,
  Azure, // etc.
}
```
</blockquote></details>

I have added an extra commit that implements this requirement. Please let me know if this works!

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 12 of 25 files reviewed, and pending CI: Bazel Dev / macos-15, Cargo Dev / macos-15, Cargo Dev / ubuntu-24.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / macos-15, NativeLink.com Cloud / Remote Cache / macos-15, Publish image, Publish nativelink-worker-init, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, Web Platform Deployment / macos-15, buildstream, windows-2022 / stable, and 4 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-config/src/stores.rs` line 748 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOeFAu73-WeRK7xhZIh:-OOho4XZ2aabA8hCBtMR:bx2n17y) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-config/src/stores.rs#L748)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

I have added an extra commit that implements this requirement. Please let me know if this works!
</blockquote></details>

That looks good. Let's pull this change out into a dedicated PR called `[Breaking] Change S3Store to generic ObjectStore`. I think regarding the `GcsConfig` and `S3Config` in the enum it would be nice to have just one config that implements all functionality, but I'd be fine to change this afterwards as well as it wouldn't be breaking anymore.

One thing to note is that I made a mistake calling this `BlobStore`. I believe `ObjectStore` would be more appropriate.

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 12 of 25 files reviewed, and pending CI: Bazel Dev / macos-15, Cargo Dev / macos-15, Coverage, Installation / macos-13, Installation / macos-14, Installation / macos-15, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, and 4 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-config/src/stores.rs` line 748 at r5](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOeFAu73-WeRK7xhZIh:-OOhpSaX8SPPwvEEBuYA:b907jus) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-config/src/stores.rs#L748)):*

That looks good. Let's pull this change out into a dedicated PR called `[Breaking] Change S3Store to generic ObjectStore`. I think regarding the `GcsConfig` and `S3Config` in the enum it would be nice to have just one config that implements all functionality, but I'd be fine to change this afterwards as well as it wouldn't be breaking anymore.

One thing to note is that I made a mistake calling this `BlobStore`. I believe `ObjectStore` would be more appropriate.
</blockquote></details>

Sure! I'll pull this out into a separate PR and point it to GCS PR for now (until it's merged). Also, I have two separate config because it would be easier to handle unique fields, for the respective cloud provider. This would also be relevant when we add AzureSpec, as that would have entire unique fields as per the naming convention (container, account_name, etc.) .
Yeah, I had this thought as well that having a BlobStore as a higher object might be confusing with the Azure Blob Store. I'll rename these as well.

jhpratt (changes requested):
> Reviewed 5 of 22 files at r5, 4 of 9 files at r9.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 23 of 25 files reviewed, and pending CI: pre-commit-checks, and 8 discussions need to be resolved (waiting on @jaroeichler)

___
*[`nativelink-store/src/gcs_store.rs` line 204 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIipJr88UYoOwoZTt7:-OOj2idu2GuR-v8hNZFb:bf5dejm) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L204)):*

cc @jhpratt Who might have some magical trick up their sleeve :magic_wand:
</blockquote></details>

Nothing immediately comes to mind aside from a macro, but that would likely be more confusing than it's worth. It's not a ton of duplication, after all.
___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OOj4HgK7viNRQoUO1AY:bm6q7n5) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*

Ah I think the confusing part here is that the variable is named `object` instead of `object_path`. The latter makes it clearer that clones are harmless. (The former could give the impression that we're creating copies of the data).
</blockquote></details>

For what it's worth I was about to leave a separate comment about the clones before seeing this. Something like `client` was cloned immediately prior to being moved into the closure & async block, so I don't see why it would need to be cloned again?
___
*[`nativelink-store/src/gcs_client/mocks.rs` line 33 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj6wUT4MfG71vOo6iM:-OOj6wUT4MfG71vOo6iN:b-jfihwb) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_client/mocks.rs#L33)):*
> ```XML
>     objects: RwLock<HashMap<String, MockObject>>,
>     // Flag to simulate failures
>     should_fail: RwLock<bool>,
> ```

For integers, atomics are preferable over a lock.
___
*[`nativelink-store/src/gcs_client/mocks.rs` line 107 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj7D0d25_rYmjHSy2t:-OOj7D0eDSiBMljWf0Wj:b3x3heg) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_client/mocks.rs#L107)):*
> ```XML
>         Self::None
>     }
> }
> ```

nit: `#[derive(Default)]` works on enums like this. You'll need to specify `#[default]` on the relevant variant.



_Code quote:_
```XML
impl Default for FailureMode {
    fn default() -> Self {
        Self::None
    }
}
```
___
*[`nativelink-store/src/gcs_store.rs` line 92 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj1JO48Rl7M0J5cq9A:-OOj1JO48Rl7M0J5cq9B:bcoas2c) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_store.rs#L92)):*
> ```XML
>             let mut rng = rand::rng();
>             let factor = min + (max - min) * rng.random::<f32>();
>             delay.mul_f32(factor)
> ```

Thoughts…`rng.random::<f32>()` returns [0, 1). If we subtract 0.5, this gives us [-0.5, 0.5). `(max - min)` is equivalent to `jitter_amt`, so `factor` could be written as `1. + jitter_amt * (rand::random::<f32>() - 0.5)`, no? That seems like it would be simpler than doing multiple calculations separately.



_Code quote:_
```XML
            let min = 1.0 - (jitter_amt / 2.0);
            let max = 1.0 + (jitter_amt / 2.0);
            let mut rng = rand::rng();
            let factor = min + (max - min) * rng.random::<f32>();
            delay.mul_f32(factor)
```
___
*[`nativelink-store/src/gcs_store.rs` line 112 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj0FvJ3odglrCorilT:-OOj0FvJ3odglrCorilU:bep0amv) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_store.rs#L112)):*
> ```XML
>                 .resumable_chunk_size
>                 .unwrap_or(CHUNK_SIZE)
>                 .min(CHUNK_SIZE),
> ```

Let's use `core::cmp::min` here — it's generally quite confusing to see it called this way.

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 0 of 2 LGTMs obtained, and 13 of 25 files reviewed, and pending CI: Analyze (javascript-typescript), Web Platform Deployment / ubuntu-24.04, asan / ubuntu-24.04, pre-commit-checks, ubuntu-24.04, vale, and 8 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OOnFLFDBRutIiZoQB19:b-hxljyv) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*
<details><summary><i>Previously, jhpratt (Jacob Pratt) wrote…</i></summary><blockquote>

For what it's worth I was about to leave a separate comment about the clones before seeing this. Something like `client` was cloned immediately prior to being moved into the closure & async block, so I don't see why it would need to be cloned again?
</blockquote></details>

Thanks! `client` clones were definitely unnecessary, and I removed those clones. Also, I renamed the `object` to `object_path`.  
___
*[`nativelink-store/src/gcs_store.rs` line 92 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj1JO48Rl7M0J5cq9A:-OOmZ0lL67mJ9gKhtszb:bcjxzcu) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_store.rs#L92)):*
<details><summary><i>Previously, jhpratt (Jacob Pratt) wrote…</i></summary><blockquote>

Thoughts…`rng.random::<f32>()` returns [0, 1). If we subtract 0.5, this gives us [-0.5, 0.5). `(max - min)` is equivalent to `jitter_amt`, so `factor` could be written as `1. + jitter_amt * (rand::random::<f32>() - 0.5)`, no? That seems like it would be simpler than doing multiple calculations separately.
</blockquote></details>

Indeed, Should've simplified this approach. I am duplicating the same approach to other stores as well.
___
*[`nativelink-store/src/gcs_store.rs` line 112 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj0FvJ3odglrCorilT:-OOmiEYu3lRAZjnfIOg1:b-896fja) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_store.rs#L112)):*
<details><summary><i>Previously, jhpratt (Jacob Pratt) wrote…</i></summary><blockquote>

Let's use `core::cmp::min` here — it's generally quite confusing to see it called this way.
</blockquote></details>

Done!
___
*[`nativelink-store/src/gcs_client/mocks.rs` line 33 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj6wUT4MfG71vOo6iM:-OOmWRoBDKPynwyOyzEw:b-fv1uji) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_client/mocks.rs#L33)):*
<details><summary><i>Previously, jhpratt (Jacob Pratt) wrote…</i></summary><blockquote>

For integers, atomics are preferable over a lock.
</blockquote></details>

Updated the CallCounts and `should_fail` here to use Atomics
___
*[`nativelink-store/src/gcs_client/mocks.rs` line 107 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OOj7D0d25_rYmjHSy2t:-OOlt4-Q6_wwLb1CxSU6:bt2vr0y) ([raw file](https://github.com/TraceMachina/nativelink/blob/232e2f3ede3b1e2a39c7beede1ba324906b00b94/nativelink-store/src/gcs_client/mocks.rs#L107)):*
<details><summary><i>Previously, jhpratt (Jacob Pratt) wrote…</i></summary><blockquote>

nit: `#[derive(Default)]` works on enums like this. You'll need to specify `#[default]` on the relevant variant.
</blockquote></details>

Fixed!

jhpratt (approved):
:lgtm:

> Reviewed 9 of 12 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OOyV9N17ImGQSgpSikt:bnfp4nl)* status: 1 of 2 LGTMs obtained, and 22 of 25 files reviewed, and 4 discussions need to be resolved (waiting on @jaroeichler)

___
*[`nativelink-store/src/gcs_store.rs` line 436 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OMIm_xc9xOVfxKZsaVx:-OOyRYu11wX1QLjGBAoU:b-qqw8sq) ([raw file](https://github.com/TraceMachina/nativelink/blob/c3b84c0ba0bbab934cd7455fc7fc925866b353aa/nativelink-store/src/gcs_store.rs#L436)):*
<details><summary><i>Previously, amankrx (Aman Kumar) wrote…</i></summary><blockquote>

Thanks! `client` clones were definitely unnecessary, and I removed those clones. Also, I renamed the `object` to `object_path`.
</blockquote></details>

Much better :smile:

MarcusSorealheis:
Wow @amankrx

jaroeichler (changes requested):
> Reviewed 1 of 19 files at r1, 4 of 22 files at r5, 6 of 12 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 6 discussions need to be resolved

___
*[`nativelink-store/Cargo.toml` line 52 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP134OmEAJAt-UL3gdT:-OP134OmEAJAt-UL3gdU:b-rvfazz) ([raw file](https://github.com/TraceMachina/nativelink/blob/1d72f7ffb35b70a977ce409eeeabc79c866dbdda/nativelink-store/Cargo.toml#L52)):*
> ```TOML
> google-cloud-auth = { version = "0.17.2", default-features = false }
> hex = { version = "0.4.3", default-features = false }
> http = { version = "1.2.0", default-features = false }
> ```

This is on `1.3.1` on `main`. Any reason to use the older version?
___
*[`Cargo.lock` line 751 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP10Yg4DJfgs8ECqH6p:-OP10Yg4DJfgs8ECqH6q:b-6m4tz4) ([raw file](https://github.com/TraceMachina/nativelink/blob/1d72f7ffb35b70a977ce409eeeabc79c866dbdda/Cargo.lock#L751)):*
> ```TOML
> [[package]]
> name = "bumpalo"
> version = "3.16.0"
> ```

Some of these look like rebase errors. Please checkout `Cargo.toml` from `main` and run `cargo generate-lockfile` again.

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 1 of 2 LGTMs obtained, and 23 of 25 files reviewed, and pending CI: Analyze (javascript-typescript), Analyze (python), Bazel Dev / macos-15, Bazel Dev / ubuntu-24.04, Cargo Dev / macos-15, Cargo Dev / ubuntu-24.04, Coverage, Installation / macos-13, Installation / macos-14, Installation / macos-15, Installation / ubuntu-22.04, Installation / ubuntu-24.04, Local / bazel / ubuntu-24.04, Local / lre-cc / ubuntu-24.04, Local / lre-rs / macos-15, Local / lre-rs / ubuntu-24.04, NativeLink.com Cloud / Remote Cache / macos-15, NativeLink.com Cloud / Remote Cache / ubuntu-24.04, Publish image, Publish nativelink-worker-init, Publish nativelink-worker-lre-cc, Remote / lre-cc / large-ubuntu-22.04, Remote / lre-rs / large-ubuntu-22.04, Web Platform Deployment / macos-15, Web Platform Deployment / ubuntu-24.04, asan / ubuntu-24.04, buildstream, integration-tests (24.04), macos-15, pre-commit-checks, ubuntu-24.04, ubuntu-24.04 / stable, windows-2022 / stable, and 6 discussions need to be resolved (waiting on @jaroeichler)

___
*[`Cargo.lock` line 751 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP10Yg4DJfgs8ECqH6p:-OP1A4oi9YlBZ85AMf4y:bqgpe4b) ([raw file](https://github.com/TraceMachina/nativelink/blob/1d72f7ffb35b70a977ce409eeeabc79c866dbdda/Cargo.lock#L751)):*

Some of these look like rebase errors. Please checkout `Cargo.toml` from `main` and run `cargo generate-lockfile` again.
</blockquote></details>

Updated to the latest main and generated the updated lockfiles!
___
*[`nativelink-store/Cargo.toml` line 52 at r10](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP134OmEAJAt-UL3gdT:-OP1AADo5dZFzQhk3LDu:b-84i4p2) ([raw file](https://github.com/TraceMachina/nativelink/blob/1d72f7ffb35b70a977ce409eeeabc79c866dbdda/nativelink-store/Cargo.toml#L52)):*

This is on `1.3.1` on `main`. Any reason to use the older version?
</blockquote></details>

This was most likely a rebase issue. I have updated it to `1.3.1`. Aside from `google_cloud_auth` crate, the rest of the crates can be latest version as it is. `google_cloud_auth` needs be this specific (or older version), because they stopped supporting the credential method that I am using right now. Also, the latest version are quite unstable, and is a WIP, so we should keep it to `0.17.2` for a while.

jaroeichler (approved):
Nice work!


:lgtm:

> Reviewed 1 of 22 files at r5, 6 of 12 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OP5QWDk5kJmwrPkzQ0n:b90cg2a)* status: 2 of 2 LGTMs obtained, and all files reviewed, and 5 discussions need to be resolved

___
*[`nativelink-store/src/gcs_client/types.rs` line 44 at r11](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP5JA0T5DxFpx1eb4mo:-OP5JA0T5DxFpx1eb4mp:b-b97h3n) ([raw file](https://github.com/TraceMachina/nativelink/blob/ffb619f564e49dfaf5679a64ac1406a6caa4f8f5/nativelink-store/src/gcs_client/types.rs#L44)):*
> ```XML
>     pub name: String,
>     pub bucket: String,
>     pub size: i64,
> ```

nit: Doesn't it make more sense to use `u64` or `usize` for all the size related variables in the Rust files? I think that would reduce a lot of the conversions in `client.rs`.

amankrx (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645)* status: 2 of 2 LGTMs obtained, and all files reviewed, and 5 discussions need to be resolved

___
*[`nativelink-store/src/gcs_client/types.rs` line 44 at r11](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-OP5JA0T5DxFpx1eb4mo:-OP5bEfdDkllUqXp-3r3:b-k2jjcb) ([raw file](https://github.com/TraceMachina/nativelink/blob/ffb619f564e49dfaf5679a64ac1406a6caa4f8f5/nativelink-store/src/gcs_client/types.rs#L44)):*

nit: Doesn't it make more sense to use `u64` or `usize` for all the size related variables in the Rust files? I think that would reduce a lot of the conversions in `client.rs`.
</blockquote></details>

It is basically mapped to the GCS Object. The size is specified as `i64` there: https://docs.rs/google-cloud-storage/0.24.0/google_cloud_storage/http/objects/struct.Object.html#structfield.size

amankrx:
Do we need any further updates here @aaronmondal? Maybe we can get this merged now. I'll rebase the Azure Blob Store on top of this PR.

aaronmondal (approved):
Great job @amankrx and thank you for being so patient with all the reviews! This was certainly a pretty difficult one with a bunch of back and forth and I feel like we now have a strong baseline for future stores as well :heart:


:lgtm:

> Reviewed 2 of 22 files at r5, 9 of 12 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OPC8scn4Dim4-Q9YMCL:b-u1f886)* status: 2 of 2 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

aaronmondal:
@amankrx (just needs a rebase after https://github.com/TraceMachina/nativelink/commit/f6269d19f392a90a7a63e9b9d3835d84f04868cd

amankrx:
Done! Thanks everyone for such a detailed and awesome review process!

aaronmondal (approved):
:lgtm:

> Reviewed 3 of 4 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1645#-:-OPCQMhy5BOJC82W8HpL:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained, and all files reviewed

PR #1669, opened 2025-04-03
Reduce use of monomorphization

jhpratt:
For reviewing, it's likely best to view without whitespace changes; indentation changed a fair amount.

# Description

While far from perfect, this pattern allows the compiler to build the primary function only once per implementation of the relevant type. We then manually call the method to monomorphize the variable. Note that this does not necessarily result in only a single copy of the function in the final binary, as other types may still be monomorphized.

This pattern isn't usable everywhere due to pre-existing limitations, primarily 'static lifetimes and interactions with async `Drop` handling.

## Type of change

strictly internal

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally — not done; `cargo` is sufficient

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1669)

### DIFF
--- a/nativelink-store/src/ac_utils.rs
+++ b/nativelink-store/src/ac_utils.rs
@@ -51,35 +51,40 @@ pub async fn get_size_and_decode_digest<T: Message + Default + 'static>(
     store: &impl StoreLike,
     key: impl Into<StoreKey<'_>>,
 ) -> Result<(T, u64), Error> {
-    let key = key.into();
-    // Note: For unknown reasons we appear to be hitting:
-    // https://github.com/rust-lang/rust/issues/92096
-    // or a smiliar issue if we try to use the non-store driver function, so we
-    // are using the store driver function here.
-    let mut store_data_resp = store
-        .as_store_driver_pin()
-        .get_part_unchunked(key.borrow(), 0, Some(MAX_ACTION_MSG_SIZE as u64))
-        .await;
-    if let Err(err) = &mut store_data_resp {
-        if err.code == Code::NotFound {
-            // Trim the error code. Not Found is quite common and we don't want to send a large
-            // error (debug) message for something that is common. We resize to just the last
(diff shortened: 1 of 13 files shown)


aaronmondal (approved):
Interestingly it appears that surprisingly this *increases* binary size:

```
# Previous commit
-r-xr-xr-x 1 root root 32613992 Jan  1  1970 result/bin/nativelink

# This PR
-r-xr-xr-x 1 root root 32673720 Jan  1  1970 result/bin/nativelink
```

No strong opinions on this though as it's like a 0.01% difference and appears to be functionally an NFC, so feel free to merge if you prefer this pattern.

:lgtm:

> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1669#-:-ON-PP-M7DIFo8zJGqgD:b-q544yq)* status: 1 of 2 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`nativelink-util/src/digest_hasher.rs` line 271 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1669#-ON-QD2XD3azfE386FB5:-ON-QD2XD3azfE386FB6:b-kbyxp) ([raw file](https://github.com/TraceMachina/nativelink/blob/11463c9be194c2a711d4d1232a5e59f7a40184a4/nativelink-util/src/digest_hasher.rs#L271)):*
> ```XML
>         size_hint: Option<u64>,
>     ) -> Result<(DigestInfo, fs::FileSlot), Error> {
>         async fn digest_for_file_inner(
> ```

nit: Since readers might not be used to this approach, consider adding a comment like `// This pattern reduces impl Trait monomorphization` to the functions using this pattern.

jhpratt:
Closing given the non-obviousness and non-existent improvement.

PR #1686, opened 2025-04-14
[Breaking] Remove `experimental_prometheus` and `disable_metrics`

aaronmondal:
These options were never useful as the metrics exposed didn't follow usable Prometheus structure. This change removes the public-facing configuration options related to the prometheus service. We keep the internals of the metric gathering infrastructure around as that's being migrated to a new OpenTelemetry setup.

Fixes GHSA-2gh3-rmm4-6rq5

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1686)

PR #1687, opened 2025-04-14
[DO NOT MERGE] Otel example

aaronmondal:
This is still under construction, but should give a general gist of things (and because we're removing a bunch of things in the main repo which might look scary :laughing:)

Note that the `disable_telemetry` option I added in this PR won't be in the final variant as the new monitoring setup will be entirely outside of the nativelink config and handled via environment variables as per https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1687)


aaronmondal:
cc @MarcusSorealheis @jaroeichler @jhpratt This is only intended as "explainer" for #1686

aaronmondal:
Superseded by https://github.com/TraceMachina/nativelink/pull/1693

PR #1699, opened 2025-04-19
Refactor flake modules

jaroeichler:
This change is breaking for flake module users. Removes `settings` in the configuration, e.g. use `nixos.path` instead of `nixos.settings.path` now.

Factors the logic behind `installation-script` out into `installation-script.nix` and doesn't use submodules anymore to reduce complexity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1699)


jaroeichler:
+@aaronmondal

MarcusSorealheis:
Good to call out but I don't think there are any flake module users other than the contributors.

MarcusSorealheis:
I could be wrong as it is open source and hard to say.

aaronmondal (approved):
Some of our internal stuff and some demos depend on this, but the breakage here is easy to work with, so that should be fine :+1:

Really good work. I know it was a surprisingly complex and subtle issue removing those deleted files, but this is much more structured while keeping that fine balance between maintainability and separation of concerns!

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1699#-:-OODq803DFc2AecJpMsg:bugx98e)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

PR #1732, opened 2025-04-30
[Breaking] Remove support for MacOS 13 on x86_64

aaronmondal:
This is slowing down CI too much. This change also increases some of the Ubuntu runner corecounts to improve CI runtime.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732)

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   coverage:
     name: Coverage
-    runs-on: ubuntu-24.04
+    runs-on: xlarge-ubuntu-24.04
     timeout-minutes: 45
     steps:
       - name: Checkout
--- a/.github/workflows/lre.yaml
+++ b/.github/workflows/lre.yaml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [large-ubuntu-22.04]
+        os: [xlarge-ubuntu-24.04]
         toolchain: [lre-cc, lre-rs]
     name: Remote / ${{ matrix.toolchain }} / ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -74,7 +74,6 @@ jobs:
         os:
           - ubuntu-22.04
           - ubuntu-24.04
-          - macos-13
           - macos-14
           - macos-15
     name: Installation / ${{ matrix.os }}


aaronmondal (commented):
@MarcusSorealheis @kubevalet The coverage change got us from 30 min to 13 min, so that seems desirable. The previous macos workflow which took 30-35 min is gone now. This makes the LRE remote workflow the slowest one in terms of wall-clock time.

With theses changes LRE-remote now runs in ~18 min rather than the previous ~26 min. I think that's a decent initial improvement, although it's not as big as the coverage change and IIRC it requires 2x or 4x the corecount of the previous job. The slowness here is mostly due to flux timings where some reconciliation loops are set to run every 2 min. We could potentially get another 1-2 min out of that workflow by tightening the timings, but I'm not sure whether we'd want to risk the additional potential instability (even if we did, we should probably separate that out from this PR). What this change will improve is the time that the lre-remote workflow takes in cases where rust code changes, as the runner will be able to much more quickly rebuild the images. I'd say it's worth a try, but it's not as obvious of an improvement as the coverage change.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732#-:-OP8u68a-J1-A5niNmE6:b-ebvqcx)* status: 0 of 1 LGTMs obtained, and 0 of 3 files reviewed

MarcusSorealheis (approved):
I think that 18 minutes is still too slow, but perfect is at odds with incremental improvement. Let's move forward here.

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732#-:-OPBLuDbAT9nnsl6CfQk:be7itac)* status: 0 of 1 LGTMs obtained, and all files reviewed

MarcusSorealheis (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732#-:-OPBMl9pDu_vmB32k8Ed:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

jaroeichler (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732)* status: 1 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`.github/workflows/nix.yaml` line 77 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1732#-OPBTSX8BhRa0yeuVQIe:-OPBTSX8BhRa0yeuVQIf:b-nj818s) ([raw file](https://github.com/TraceMachina/nativelink/blob/9cdbfef27c6bd82d838155833bfd2861eee3717f/.github/workflows/nix.yaml#L77)):*
> ```YAML
>           - ubuntu-22.04
>           - ubuntu-24.04
>           - macos-14
> ```

Is there any reason to not further reduce it to just the latest version?

aaronmondal (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1732)* status: 1 of 1 LGTMs obtained, and all files reviewed, and 1 discussions need to be resolved

___
*[`.github/workflows/nix.yaml` line 77 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1732#-OPBTSX8BhRa0yeuVQIe:-OPBgyoN1Bc7s3eAAfXh:bnx18pk) ([raw file](https://github.com/TraceMachina/nativelink/blob/9cdbfef27c6bd82d838155833bfd2861eee3717f/.github/workflows/nix.yaml#L77)):*

Is there any reason to not further reduce it to just the latest version?
</blockquote></details>

This job is the "final installation test". That is, this tests all the versions where we guarantee that nativelink will be installable as raw executable. TBH it's potentially a bit pointless given that we're fully self-contained, so we could certainly remove it as it doesn't really make a difference anyways. This was probably an artifact from when we were still dependent on glibc. But removing it would remove the "official support" for that, so I'm a bit hesitant here.

PR #1781, opened 2025-05-21
Fix integer overflow in compression_store.rs data retrieval logic

### LINKED ISSUE #1566: Thread 'tokio-runtime-worker' panicked
When using the cache via reclient, we encounter hanging builds.
The logs contain the following:
`thread 'tokio-runtime-worker' panicked at /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/bytes-1.7.1/src/bytes.rs:281:9:
range start must not be greater than end: 65536 <= 65535
thread 'tokio-runtime-worker' panicked at /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/bytes-1.7.1/src/bytes.rs:281:9:
range start must not be greater than end: 65536 <= 65535
thread 'tokio-runtime-worker' panicked at /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/bytes-1.7.1/src/bytes.rs:281:9:
range start must not be greater than end: 65536 <= 65535`

Version v0.5.3
Nativelink started in the docker of the VM

JannisFengler:
## Issue
`nativelink-store/src/compression_store.rs` had an integer overflow issue during data retrieval. On line 526, the code was calculating an end position by adding `start_pos + remaining_bytes_to_send as usize`, which could overflow when `remaining_bytes_to_send` was very large. When this overflow occurred, it would cause `end_pos` to wrap around to a smaller value, potentially making it smaller than `start_pos` in the slice operation, triggering an "attempt to add with overflow" panic.

## Fix
Changed the standard addition operation to use Rust's saturating addition:
```rust
// Before
let end_pos = start_pos + remaining_bytes_to_send as usize;

// After
let end_pos = start_pos.saturating_add(remaining_bytes_to_send as usize);
```

This prevents integer overflow by capping the result at the maximum possible value for the type instead of wrapping around.

## Testing
All tests now pass, including the previously failing regression test that demonstrated this issue.

## Related Issues
Fixes #1566

### DIFF
--- a/nativelink-store/src/compression_store.rs
+++ b/nativelink-store/src/compression_store.rs
@@ -521,8 +521,9 @@ impl StoreDriver for CompressionStore {
                         uncompressed_data_sz + uncompressed_chunk_sz as u64;
                     if new_uncompressed_data_sz >= offset && remaining_bytes_to_send > 0 {
                         let start_pos = offset.saturating_sub(uncompressed_data_sz) as usize;
+                        // Use saturating_add to prevent overflow when remaining_bytes_to_send is very large
                         let end_pos = cmp::min(
-                            start_pos + remaining_bytes_to_send as usize,
+                            start_pos.saturating_add(remaining_bytes_to_send as usize),
                             uncompressed_chunk_sz,
                         );
                         if end_pos != start_pos {
--- a/nativelink-store/tests/compression_store_test.rs
+++ b/nativelink-store/tests/compression_store_test.rs
@@ -510,3 +510,134 @@ async fn get_part_is_zero_digest() -> Result<(), Error> {
 
     Ok(())
 }
+
+// Regression test for the bug where start_pos > end_pos in the slice operation
+#[nativelink_test]
+async fn regression_test_range_start_not_greater_than_end() -> Result<(), Error> {
+    // Create a store with a small block size to trigger multiple blocks
+    const BLOCK_SIZE: u32 = 64 * 1024; // 64KB, same as DEFAULT_BLOCK_SIZE
+
+    let inner_store = MemoryStore::new(&MemorySpec::default());
+    let store_owned = CompressionStore::new(
+        &CompressionSpec {
+            backend: StoreSpec::Memory(MemorySpec::default()),
+            compression_algorithm: nativelink_config::stores::CompressionAlgorithm::Lz4(
+                nativelink_config::stores::Lz4Config {
+                    block_size: BLOCK_SIZE,
+                    ..Default::default()
+                },
+            ),
+        },
+        Store::new(inner_store.clone()),
+    )
+    .err_tip(|| "Failed to create compression store")?;
+    let store = Pin::new(&store_owned);
+
+    // Create a large buffer that spans multiple blocks
+    let data_size = BLOCK_SIZE as usize * 3; // 3 blocks
+    let mut data = vec![0u8; data_size];
+    let mut rng = SmallRng::seed_from_u64(42);
+    rng.fill(&mut data[..]);
+
+    let digest = DigestInfo::try_new(VALID_HASH, data_size).unwrap();
+    store.update_oneshot(digest, data.clone().into()).await?;
+
+    // Try to read exactly at block boundaries with various offsets
+    let boundary = u64::from(BLOCK_SIZE);
+
+    // These specific offsets test the case in the bug report where
+    // start_pos was 65536 and end_pos was 65535
+    for (offset, length) in &[
+        (boundary - 1, Some(2u64)),  // Read across block boundary
+        (boundary, Some(1u64)),      // Read exactly at block boundary
+        (boundary + 1, Some(10u64)), // Read just after block boundary
+        // Specifically test the case where offset >= block size
+        (u64::from(BLOCK_SIZE), Some(20u64)),
+        // Specifically test the case that caused the bug (65536 and 65535)
+        (u64::from(BLOCK_SIZE), Some(u64::from(BLOCK_SIZE) - 1)),
+        // More edge cases around the block boundary to thoroughly test the issue
+        (u64::from(BLOCK_SIZE) - 1, Some(1u64)), // Just before boundary
+        (u64::from(BLOCK_SIZE), Some(0u64)),     // Zero length at boundary
+        (u64::from(BLOCK_SIZE), Some(u64::MAX)), // Unlimited length at boundary
+        (u64::from(BLOCK_SIZE) * 2, Some(u64::from(BLOCK_SIZE) - 1)), // Same issue at next block
+    ] {
+        // First test with get_part_unchunked
+        let result = store.get_part_unchunked(digest, *offset, *length).await;
+
+        // The bug was causing a panic, so just checking that it doesn't panic
+        // means the fix is working
+        assert!(
+            result.is_ok(),
+            "Reading with get_part_unchunked at offset {offset} with length {length:?} should not fail"
+        );
+
+        let store_data = result.unwrap();
+
+        // Verify the data matches what we expect
+        let expected_len = cmp::min(
+            length.unwrap_or(u64::MAX) as usize,
+            data.len().saturating_sub(*offset as usize),
+        );
+        assert_eq!(
+            store_data.len(),
+            expected_len,
+            "Expected data length to match when reading at offset {} with length {:?}",
+            offset,
+            length
+        );
+
+        if expected_len > 0 {
+            let start = *offset as usize;
+            let end = start + expected_len;
+            assert_eq!(
+                &store_data[..],
+                &data[start..end],
+                "Expected data content to match when reading at offset {} with length {:?}",
+                offset,
+                length
+            );
+        }
+
+        // Now also test with the lower-level get_part method to ensure it doesn't panic
+        // This is closer to what the bytestream server would call
+        let (mut tx, mut rx) = make_buf_channel_pair();
+
+        // The error was happening in this method call
+        let get_part_result = store.get_part(digest, &mut tx, *offset, *length).await;
+        assert!(
+            get_part_result.is_ok(),
+            "Reading with get_part at offset {offset} with length {length:?} should not fail"
+        );
+
+        // Just to consume the stream and ensure it behaves as expected
+        let mut received_data = Vec::new();
+        while let Ok(chunk) = rx.consume(Some(1024)).await {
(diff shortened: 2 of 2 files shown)


aaronmondal (changes requested):
Thanks a lot for figuring this out!

+@jaroeichler +@jhpratt 

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-:-OQkfNTm9IqHDvshtMbU:byypbui)* status: 0 of 2 LGTMs obtained, and 0 of 2 files reviewed, and pending CI: Cargo Dev / macos-15, Installation / macos-14, Installation / macos-15, NativeLink.com Cloud / Remote Cache / macos-15, Publish image, Publish nativelink-worker-init, Remote / lre-cc / xlarge-ubuntu-24.04, Remote / lre-rs / xlarge-ubuntu-24.04, windows-2022 / stable, and 9 discussions need to be resolved (waiting on @jaroeichler and @jhpratt)

___
*[`nativelink-store/src/compression_store.rs` line 526 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkb6xNCUsDsv84NeUx:-OQkb6xOCkxQYPrrDejb:bsxklpu) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/src/compression_store.rs#L526)):*
> ```XML
>                         // Use saturating_add to prevent overflow when remaining_bytes_to_send is very large
>                         let end_pos = cmp::min(
>                             start_pos.saturating_add(remaining_bytes_to_send as usize),
> ```

Interesting. So we were overflowing `usize`? In that case, would a `try_from` have prevented this? We haven't fixed `clippy::cast_possible_truncation` yet: https://github.com/TraceMachina/nativelink/blob/d53363dca585e5a467fe38fef2c914928537b5c3/.bazelrc#L138-L140

In that case I'm slightly worried that while this fixes it, it masks the actual issue of the overflow wheras a `try_from` with a branch to properly handle the overflow case could be more explicit.
___
*[`nativelink-store/tests/compression_store_test.rs` line 518 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkcpdiBzYZRuTp3R3Z:-OQkcpdiBzYZRuTp3R3_:bx4xy9s) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L518)):*
> ```XML
> async fn regression_test_range_start_not_greater_than_end() -> Result<(), Error> {
>     // Create a store with a small block size to trigger multiple blocks
>     const BLOCK_SIZE: u32 = 64 * 1024; // 64KB, same as DEFAULT_BLOCK_SIZE
> ```

If it's the same as `DEFAULT_BLOCK_SIZE`, couldn't we use that instead?
___
*[`nativelink-store/tests/compression_store_test.rs` line 534 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkd64aBDttZoaIXWLM:-OQkd64aBDttZoaIXWLN:b-kp6j74) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L534)):*
> ```XML
>     )
>     .err_tip(|| "Failed to create compression store")?;
>     let store = Pin::new(&store_owned);
> ```

Let's inline the stores as they're only needed once.
___
*[`nativelink-store/tests/compression_store_test.rs` line 537 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkeWqECZREcBLzLhQG:-OQkeWqECZREcBLzLhQH:bdep0cv) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L537)):*
> ```XML
> 
>     // Create a large buffer that spans multiple blocks
>     let data_size = BLOCK_SIZE as usize * 3; // 3 blocks
> ```

Since we're trying to move to `clippy::cast_possible_truncation`, it would be nice to avoid too many uses of `as` in new code so that fixing it will be easier later on.
___
*[`nativelink-store/tests/compression_store_test.rs` line 550 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkdKSW6tpYBi92dmDj:-OQkdKSW6tpYBi92dmDk:b2jxwmm) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L550)):*
> ```XML
>     // These specific offsets test the case in the bug report where
>     // start_pos was 65536 and end_pos was 65535
>     for (offset, length) in &[
> ```

If this test was used in the old implementation I believe the first failure would prevent the other cases from running. I.e. if the example fails it hides the output of the cases after that.

Since we're reusing the stores it also seems like the different cases could influence each other.
___
*[`nativelink-store/tests/compression_store_test.rs` line 556 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQke8C86dKqVvQI8b43:-OQke8C86dKqVvQI8b44:b-1xfw1g) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L556)):*
> ```XML
>         // Specifically test the case where offset >= block size
>         (u64::from(BLOCK_SIZE), Some(20u64)),
>         // Specifically test the case that caused the bug (65536 and 65535)
> ```

Let's add the specific github issue here for easier lookup later.
___
*[`nativelink-store/tests/compression_store_test.rs` line 564 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkeE7v5OrHKFWSWfyI:-OQkeE7v5OrHKFWSWfyJ:bnttlu2) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L564)):*
> ```XML
>         (u64::from(BLOCK_SIZE) * 2, Some(u64::from(BLOCK_SIZE) - 1)), // Same issue at next block
>     ] {
>         // First test with get_part_unchunked
> ```

nit: Remove redundant comment
___
*[`nativelink-store/tests/compression_store_test.rs` line 576 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkeKck3IdlbUjHcy-d:-OQkeKck3IdlbUjHcy-e:bnttlu2) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L576)):*
> ```XML
>         let store_data = result.unwrap();
> 
>         // Verify the data matches what we expect
> ```

nit: Remove redundant comment
___
*[`nativelink-store/tests/compression_store_test.rs` line 606 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkexHG1YxziFAk5Vwu:-OQkexHG1YxziFAk5Vwv:be62zvs) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L606)):*
> ```XML
> 
>         // The error was happening in this method call
>         let get_part_result = store.get_part(digest, &mut tx, *offset, *length).await;
> ```

If the call above panicked, isn't this part (and the ones below it) essentially dead code that would be better as a separate test? I.e. a test for the higher level `get_part_unchunked` and another one for the lower-level `get_part` might be good here.

jaroeichler (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1781)* status: 0 of 2 LGTMs obtained, and all files reviewed, and 9 discussions need to be resolved (waiting on @jhpratt)

___
*[`nativelink-store/src/compression_store.rs` line 526 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkb6xNCUsDsv84NeUx:-OQlzu9S5Ehv49sIit41:b-dhojne) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/src/compression_store.rs#L526)):*

Interesting. So we were overflowing `usize`? In that case, would a `try_from` have prevented this? We haven't fixed `clippy::cast_possible_truncation` yet: https://github.com/TraceMachina/nativelink/blob/d53363dca585e5a467fe38fef2c914928537b5c3/.bazelrc#L138-L140

In that case I'm slightly worried that while this fixes it, it masks the actual issue of the overflow wheras a `try_from` with a branch to properly handle the overflow case could be more explicit.
</blockquote></details>

`try_from` addresses a different issue though. We should instead try to avoid these conversions. Can we define `remaining_bytes_to_send` as `usize` in the first place? In general, we have these kind of conversion issues in a lot of places, where functions related to memory sizes expect `usize` while we'd like to do things in `u64`.

For the addition, `saturating_add` is reasonable since this should mostly occur if `length` is `None` and then `remaining_bytes_to_send` is `u64::MAX`.
___
*[`nativelink-store/tests/compression_store_test.rs` line 534 at r1](https://reviewable.io/reviews/TraceMachina/nativelink/1781#-OQkd64aBDttZoaIXWLM:-OQm5G7B4XHRLSrX5wCH:b-nkei97) ([raw file](https://github.com/TraceMachina/nativelink/blob/9fa2192973e69b3a2975065aea9558a0796d53ee/nativelink-store/tests/compression_store_test.rs#L534)):*

Let's inline the stores as they're only needed once.
</blockquote></details>

nit: If you change anything in the above part, please also adjust it in the other tests in this file.

JannisFengler:
I will update this PR soon :)

palfrey:
> I will update this PR soon :)

Just wanted to check in and see how this one was going? Do you need any help from us on this?

PR #1712, opened 2025-04-23
Largely change from map-based to array-based config

jhpratt:
# Description

This includes a backwards-compatibility layer to avoid breaking anything. When the old format is fully removed, this can be trivially accomplished by removing `backcompat.rs` and removing the `#[serde(deserialize_with)]` annotations from the fields.

This does not change the config for bytestream, as it is far more involved than the other fields.

Part of #1649

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1712)

PR #1806, opened 2025-05-27
[Breaking] Change bytestream to array-based config

### LINKED ISSUE #1649: Rework the instance_name config api
Somewhat of a follow-up to https://github.com/TraceMachina/nativelink/issues/834.

Our current configuration looks like this:

```json5
    "services": {
      "cas": {
        "main": {
          "cas_store": "WORKER_FAST_SLOW_STORE"
        }
      },
      "ac": {
        "main": {
          "ac_store": "AC_MAIN_STORE"
        }
      },
      "execution": {
        "main": {
          "cas_store": "WORKER_FAST_SLOW_STORE",
          "scheduler": "MAIN_SCHEDULER"
        }
      },
      "capabilities": {
        "main": {
          "remote_execution": {
            "scheduler": "MAIN_SCHEDULER"
          }
        }
      },
      "bytestream": {
        "cas_stores": {
          "main": "WORKER_FAST_SLOW_STORE"
        }
      }
```

The `"main"` here refers to the `"instance_name"` as per https://github.com/bazelbuild/remote-apis/blob/2721568dea746d81a98a116ee222c8da50bbcf5c/build/bazel/remote/execution/v2/remote_execution.proto#L1405.

There are multiple issues with this approach:

- This has repeatedly been a source of confusion for users as it's not immediately clear that this is the instance name without consulting additional documentation.
- It's a dynamic key, which is incompatible with how K8s wants CRDs to look like. In other words, you can't create correct yaml with this.
- The default should be no instance name. This looks ugly as it requires an empty string key:
```json5
"cas": {
  "": {
    "cas_store": "WORKER_FAST_SLOW_STORE"
  }
},
```

The initially mentioned discussion was about a similar issue. In the same spirit, it seems better to use something along these lines:

```json5
    "services": {
      "cas": {
        "instance_name": "main",
        "cas_store": "WORKER_FAST_SLOW_STORE"
      },
      "ac": {
        "instance_name": "main",
        "ac_store": "AC_MAIN_STORE"
      },
      "execution": {
        "instance_name": "main",
        "cas_store": "WORKER_FAST_SLOW_STORE",
        "scheduler": "MAIN_SCHEDULER"
      },
      "capabilities": {
        "instance_name": "main",
        "remote_execution": {
          "scheduler": "MAIN_SCHEDULER"
        }
      },
      "bytestream": {
        "cas_stores": [
        {
          "instance_name": "main",
          "ref": "WORKER_FAST_SLOW_STORE"
        }]
      }
```

This solves all above mentioned issues because

- It's self-documenting
- We're no longer using a dynamic key for the instance name.
- If we make the `"instance_name"` field optional an default to an empty string if not set users can omit it entirely and we get this:

```json5
"cas": {
  "cas_store": "WORKER_FAST_SLOW_STORE"
},
```

jaroeichler:
In contrast to #1712, there's no backwards-compatibility layer.

Closes #1649

### DIFF
--- a/deployment-examples/docker-compose/local-storage-cas.json5
+++ b/deployment-examples/docker-compose/local-storage-cas.json5
@@ -48,11 +48,9 @@
         "ac_store": "AC_MAIN_STORE"
       }],
       "capabilities": [],
-      "bytestream": {
-        "cas_stores": {
-          "": "CAS_MAIN_STORE"
-        }
-      },
+      "bytestream": [{
+        "cas_store": "CAS_MAIN_STORE"
+      }],
       "fetch": {},
       "push": {},
     }
@@ -79,11 +77,9 @@
         "ac_store": "AC_MAIN_STORE"
       }],
       "capabilities": [],
-      "bytestream": {
-        "cas_stores": {
-          "": "CAS_MAIN_STORE"
-        }
-      },
+      "bytestream": [{
+        "cas_store": "CAS_MAIN_STORE"
+      }],
       "health": {}
     }
   }]
--- a/integration_tests/buildstream/buildstream_cas.json5
+++ b/integration_tests/buildstream/buildstream_cas.json5
@@ -126,11 +126,9 @@
           "scheduler": "MAIN_SCHEDULER"
         }
       }],
-      "bytestream": {
-        "cas_stores": {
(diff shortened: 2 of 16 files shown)


MarcusSorealheis:
cannot merge until conflicts resolved

PR #1646, opened 2025-03-26
Initial Remote asset support

palfrey:
# Description

This technically fixes #1533, in that it's now possible to run Buildstream with Nativelink. This has been done by doing the least possible implementation of the Remote Asset protocol that still passes the basic requirements of Buildstream i.e. Fetch always returns "not found" and Push throws away whatever it's given. There's also a few small config changes, because Buildstream needed an extra platform property.

I'm able to run a "hello world" Buildstream project based on the tutorial, but anything more substantial has not been tested, and so may well fail.

## Type of change

Please delete options that aren't relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Booting the docker-compose setup in deployment-examples/docker-compose and running `bst -c buildstream.conf build hello.bst` in another terminal with the contents of
[buildstream.tar.gz](https://github.com/user-attachments/files/19475641/buildstream.tar.gz) which resulted in [buildstream-results.log](https://github.com/user-attachments/files/19475651/buildstream-results.log)

## Checklist

- [ ] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1646)

/claim #1533

PR #1816, opened 2025-06-06
Implements the internals of the remote asset protocol

palfrey:
# Description

[Previously](https://github.com/TraceMachina/nativelink/pull/1646) I'd added support for the Remote Asset protocol, but purely in a stub "doesn't actually do anything form". This PR adds support for actually adding resources to the store and retrieving them, making Buildstream support usable.

## Type of change

## How Has This Been Tested?

Mainly with `buildstream-with-nativelink-test`. The "hello world" test as that now runs fully cached on the second run of that locally. More advanced Buildstream configs might still break in some way, but hopefully not.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1816)


nativelink-service/src/fetch_server.rs
```diff
@@ -59,9 +128,8 @@ impl FetchServer {
 impl Fetch for FetchServer {
     #[allow(clippy::blocks_in_conditions)]
     #[instrument(
-        err,
+        err(level = Level::WARN),
         ret(level = Level::INFO),
-        level = Level::ERROR,
```
palfrey, line 64: I'm a little uncertain about this one (and similar changes elsewhere in Remote Asset, but should probably be applied more widely). So, currently the setup is that there's either Ok or Errors, and the latter logs at error level. To my mind, there's actually 3 cases: Ok, Client Error, and Server Error (2xx, 4xx and 5xx from an HTTP perspective), and those should be logged as Info/Warning/Error.

Why do I want this? Well, buildstream on bootup submits invalid fetch and push requests to answer the question "do I get 404 or some other error". If it gets a 404, it assumes the functionality doesn't exist in a server, where as another error indicates it's there and all good. Without this logging change, we end up outputting a pair of errors every time, and really we should only be seeing errors in our integration tests when there's an actual bug.

WDYT?

MarcusSorealheis: I think this looks fine.

MarcusSorealheis: Could you please resolve the merge conflicts and then we can merge. We can iterate further if users need it.

palfrey: Conflicts resolved, build green

PR #1817, opened 2025-06-06
Add JSON5 formatting to pre-commit

palfrey:
# Description

This adds formatting for JSON5 files to the pre-commit, and applies it to all the current files. Their formatting was previously mostly good, but somewhat inconsistent. Note this removes the quote formatting around properties which makes the diff notably larger, but the tool I've been using doesn't support that as an option. I like it personally, but that's an open question if it's a preferred style or not.

## Type of change

  not work as expected)

## How Has This Been Tested?

Existing test suites.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1817)

### DIFF
--- a/deployment-examples/docker-compose/local-storage-cas.json5
+++ b/deployment-examples/docker-compose/local-storage-cas.json5
@@ -4,87 +4,100 @@
 // so objects are compressed, deduplicated and uses some in-memory
 // optimizations for certain hot paths.
 {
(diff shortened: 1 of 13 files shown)


MarcusSorealheis (approved):
Will be very helpful

PR #1779, opened 2025-05-17
Treat warnings in Bazel builds as errors

aaronmondal:
Bazel's incremental caching doesn't rerun actions that only emit warnings. That's inconvenient as these warnings will only be shown once during the initial invocation but not in following invocations which treat the action as successfull run. This leads to friction where a passing local build won't necessarily pass CI.

We now treat all warnings as hard errors. The more natural approach to temporarily skip over warnings in Bazel is via `-k` (or `--keep_going`).

PR #1840, opened 2025-06-26
Mark all warnings as errors so bazel fails

palfrey:
# Description

Obsoletes https://github.com/TraceMachina/nativelink/pull/1779

The main changes are:
* Replace every warning in `Cargo.toml` with `deny`
* Make the `generate-bazel-rc` tool no longer support `warn` so we'll get failures in the future if we try to add them back
* Fix the ordering of the flags in `.bazelrc` - the original generator code used purely alphabetical ordering, but I didn't realise the priorities were also encoded in the ordering, This is now fixed in the generator, and so some `allow` overrides we have are now actually working.
* Fix some of the previous warnings (including probably a few I could have not done if I realised the ordering problem earlier)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended


MarcusSorealheis (approved):
LGTM

PR #1860, opened 2025-07-08
Modified approach to Fix Zero Byte Uploads

MarcusSorealheis:
# Description

@chrisstaite reported issues with the zero bytes and the GCS store. I'm trying to understand why this wasn't resolved bya. previous test to address it. 

Fixes # (issue)

## Type of change

## How Has This Been Tested?

I wrote a test. It needs to be tested on a live GCP instance. I'm currently on a plane with limited bandwidth.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1860)

### DIFF
--- a/nativelink-store/src/gcs_store.rs
+++ b/nativelink-store/src/gcs_store.rs
@@ -31,6 +31,7 @@ use nativelink_util::retry::{Retrier, RetryResult};
 use nativelink_util::store_trait::{StoreDriver, StoreKey, UploadSizeInfo};
 use rand::Rng;
 use tokio::time::sleep;
+use tracing::warn;
 
 use crate::cas_utils::is_zero_digest;
 use crate::gcs_client::client::{GcsClient, GcsOperations};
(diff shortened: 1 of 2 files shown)


amankrx:
Maybe we can do a simple upload at the beginning for a zero byte uploads. The problem appears to be that the loop breaks without initializing the upload session. So, a simple zero-byte upload in the beginning should solve this issue.

nativelink-store/tests/gcs_store_test.rs
```diff
@@ -346,43 +346,44 @@ async fn get_part_with_range() -> Result<(), Error> {
...
-    let call_counts = mock_ops.get_call_counts();
-    assert_eq!(
-        call_counts.read_calls.load(Ordering::Relaxed),
-        0,
-        "No read operations should have been called for zero digest"
-    );
-
-    Ok(())
-}
+// I'll delete this but it did not actually test and prevent the issue we were addressing.
```
amankrx, line 349: I'm assuming this is working as expected right? 
The problem appears to be during zero byte uploads. The get operation should work fine in this case.
Let me test this out locally as well.

nativelink-store/src/gcs_store.rs
```diff
@@ -216,162 +217,233 @@ where
...
+    upload_size: UploadSizeInfo,
+) -> Result<(), Error> {
+    let object_path = self.make_object_path(&digest);
+
+    reader.set_max_recent_data_size(
+        u64::try_from(self.max_retry_buffer_size)
+            .err_tip(|| "Could not convert max_retry_buffer_size to u64")?,
+    );
+
+    // Special case for known zero-byte uploads: Start resumable and finalize immediately.
```
amankrx, line 233: Why do a resumable upload if the exact size is zero. Wouldn't it be better to do a simple upload for a zero byte uploads.

PR #1861, opened 2025-07-09
Log fallback calls to help with adding new gRPC bits

palfrey:
# Description

When you're adding new gRPC calls, it's sometimes hard to check you called the right thing v.s. maybe calling some other endpoint. This PR adds logging to the `fallback` function so we know what wasn't implemented.

## Type of change

  not work as expected)

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1861)

### DIFF
--- a/deployment-examples/docker-compose/Dockerfile
+++ b/deployment-examples/docker-compose/Dockerfile
@@ -29,7 +29,7 @@ RUN apt-get update \
         DEBIAN_FRONTEND=noninteractive \
         apt-get install --no-install-recommends -y \
             npm=9.2.0~ds1-2 \
-            git=1:2.43.0-1ubuntu7.2 \
+            git=1:2.43.0-1ubuntu7.3 \
             gcc=4:13.2.0-7ubuntu1 \
             g++=4:13.2.0-7ubuntu1 \
             python3=3.12.3-0ubuntu2 \
@@ -38,7 +38,7 @@ RUN apt-get update \
         DEBIAN_FRONTEND=noninteractive \
         apt-get install --no-install-recommends -y \
             npm=8.5.1~ds-1 \
-            git=1:2.34.1-1ubuntu1.12 \
+            git=1:2.34.1-1ubuntu1.14 \
             gcc=4:11.2.0-1ubuntu1 \
             g++=4:11.2.0-1ubuntu1 \
             python3=3.10.6-1~22.04.1 \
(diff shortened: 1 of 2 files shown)


deployment-examples/docker-compose/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN apt-get update \
         DEBIAN_FRONTEND=noninteractive \
         apt-get install --no-install-recommends -y \
             npm=9.2.0~ds1-2 \
-            git=1:2.43.0-1ubuntu7.2 \
+            git=1:2.43.0-1ubuntu7.3 \
```
palfrey, line 32: Because upstream released new versions

src/bin/nativelink.rs
```diff
@@ -72,9 +73,7 @@ use tokio_rustls::rustls::server::WebPkiClientVerifier;
 use tokio_rustls::rustls::{RootCertStore, ServerConfig as TlsServerConfig};
 use tonic::codec::CompressionEncoding;
 use tonic::service::Routes;
-#[cfg(target_family = "unix")]
```
MarcusSorealheis, line 75: I'm not sure why this was ever here. Does `tracing::warn` not work on Windows or something?  That's such a particular invocation of `target_family` that I never noticed.

src/bin/nativelink.rs
```diff
@@ -541,9 +540,11 @@ async fn inner_main(
             );
         }
 
-        svc = svc
-            // This is the default service that executes if no other endpoint matches.
-            .fallback((StatusCode::NOT_FOUND, "Not Found"));
+        // This is the default service that executes if no other endpoint matches.
+        svc = svc.fallback(|uri: Uri| async move {
+            warn!("No route for {uri}");
+            (StatusCode::NOT_FOUND, format!("No route for {uri}"))
```
MarcusSorealheis, line 546: An obvious improvement.

src/bin/nativelink.rs
palfrey: It works fine on Windows, but the only place it was previously called from is https://github.com/TraceMachina/nativelink/blob/main/src/bin/nativelink.rs#L889 which I presume didn't work on Windows for some reason

Issue #1859, opened 2025-07-08
Failing action retried on same worker causes file exists


chrisstaite-menlo:
When an action fails it is added back to the queue to re-try.  However, if the action is immediately re-tried on a worker that it previously failed on it attempts to create the work directory of the same name (named for the Operation ID) and fails due to it already existing.

Example:
```
File exists (os error 17) : Error creating action directory /tmp/nativelink/work/56127536-1155-4075-b181-726727f9a549
```

There's a few possible solutions here:

- Always clean up the work directory before returning the result (causes successful actions to be delayed)
- Always clean up the work directory before returning a failed result (causes retries of the action to be delayed)
- Wait for an existing attempt to be cleaned up before starting an operation (best, but required some lookup of previous actions)

MarcusSorealheis:
Ok, I'm trying to fix this because it looks like there's a race condition in [running_actions_manager.rs](https://github.com/TraceMachina/nativelink/blob/main/nativelink-worker/src/running_actions_manager.rs).

Could you share reproduction steps, in case I cannot reproduce this error before you get back to me?

MarcusSorealheis:
I'm also checking to see if this was introduced [here](https://github.com/TraceMachina/nativelink/pull/752).

dkostyrev:
I've encountered this exact issue when I noticed that one of our tests was failing reliably with Nativelink. We use Bazel and the test was running with an `eternal` timeout (3600s), it was failing due to [this check](https://github.com/TraceMachina/nativelink/blob/main/nativelink-worker/src/running_actions_manager.rs#L1821-L1828) and then restarting on the same worker, which in turn resulted in this error.

MarcusSorealheis:
@dkostyrev thank you for the added context. I haven't removed that check. And now that you mention the eternal timeout use case, I'm not sure that we solve for that. I didn't think about it but should have. Could you verify that my branch with the bugfix solves your issue?

I think the one significant difference is that in my fix's case, we won't retry until the dir is physically cleaned. In theory, it should work. I don't want to add a test like that to our CI for obvious reasons. 😆

MarcusSorealheis:
also, you are making my weekend warrior work feel all the more meaningful, so thank you. Love fixing stability/usability issues for users.

PR #1868, opened 2025-07-20
Fix directory collision on action retries by waiting for cleanup and removing stales

MarcusSorealheis:
# Description

Per a user's report, if an action fails on worker X, and the failed action is returned the queue, if worker x picks up the action again, there can be a directory collision because worker x may be immediately available due its most recent action failing and the queue being empty. 

The physical cleanup should happen before an action is eligible for retry. 

One tradeoff to consider here: there is some risk of impact to performance but I'm fine with is because it improves stability and predictability.

Fixes # ([1859](https://github.com/TraceMachina/nativelink/issues/1859))

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1868)

### DIFF
--- a/nativelink-worker/src/running_actions_manager.rs
+++ b/nativelink-worker/src/running_actions_manager.rs
@@ -19,14 +19,14 @@ use core::pin::Pin;
 use core::sync::atomic::{AtomicBool, Ordering};
 use core::time::Duration;
 use std::borrow::Cow;
-use std::collections::HashMap;
 use std::collections::vec_deque::VecDeque;
+use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 #[cfg(target_family = "unix")]
 use std::fs::Permissions;
 #[cfg(target_family = "unix")]
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::{Arc, Weak};
 use std::time::SystemTime;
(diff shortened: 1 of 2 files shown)


MarcusSorealheis:
I think the elephant in this Pull Request is the ability to reproduce the issue you were encountering @chrisstaite-menlo. In a couple respects, this is a risky change. That said, it's not worth it to even consider merging if your issue still occurs. 

I tried to add test a concurrent retry, but it was on the path to certain flakiness so I removed that for now. Open to feedback.

MarcusSorealheis:
One thing I did not consider is that this change potentially needs to wait for a check from @chrisstaite-menlo because Bazel and REAPI could be slightly different in their action behavior. I have not observed this issue in production with any customer, or at least identified it, but it's definitely a bug.

MarcusSorealheis:
I guess we don't need to wait on @chrisstaite-menlo to review because he is out on holiday but we do need at least two eyes on this one.

palfrey (approved):
I'm happy with this. There is the potential slowdowns with the locks, but I'd rather get something like this in first that optimises for safety and reliability first. Once the benchmarking is in, if we see any slowdowns due to this there's a multitude of alternative mutex options out there (including IIRC probably some specialised "check for value" ones) we can swap over to if needed.

MarcusSorealheis:
Agreed! Thanks for taking a look.  We are past the days of only std and parking_lot

PR #1871, opened 2025-07-23
Add nix command to create local nativelink images

palfrey:
# Description

It's occasionally useful to make a local version of the Nativelink image that can be used interchangeably with the released ones.

## Type of change

  not work as expected)

## How Has This Been Tested?

Local running of the command

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1871)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -352,7 +352,8 @@
               nativelink-x86_64-linux
               ;
 
-            inherit (pkgs.nativelink-tools) local-image-test publish-ghcr native-cli;
+            # Used by the CI
+            inherit (pkgs.nativelink-tools) local-image-test;
 
             default = nativelink;
 
@@ -498,6 +499,7 @@
               pkgs.lre.lre-cc.lre-cc-configs-gen
               pkgs.nativelink-tools.local-image-test
               pkgs.nativelink-tools.native-cli
+              pkgs.nativelink-tools.create-local-image
             ]
             ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [
               pkgs.darwin.apple_sdk.frameworks.CoreFoundation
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -352,8 +352,6 @@
               nativelink-x86_64-linux
               ;
 
-            inherit (pkgs.nativelink-tools) local-image-test publish-ghcr native-cli;
-
```
palfrey, line 356: This wasn't actually needed anymore as everything refers to those packages by their full names.

MarcusSorealheis (changes requested):
This needs documentation. The code is not sufficient. If it helps people make docker images easily, it should probably be featured as prominently as in the top-level `README`.

palfrey:
> I was trying to build the docker image locally yesterday, and it was failing due to the git version error.
> 
> ```
> 8.041 However the following packages replace it:
> 8.041   git-svn
> 8.041 
> 8.042 E: Version '1:2.34.1-1ubuntu1.14' for 'git' was not found
> ------
> Dockerfile:27
> ```

That's with `deployment-examples/docker-compose/Dockerfile` right? Whole other problem, but I'll have a look at that in a bit.

palfrey:
> This needs documentation. The code is not sufficient. If it helps people make docker images easily, it should probably be featured as prominently as in the top-level `README`.

The [top-level docs](https://github.com/TraceMachina/nativelink?tab=readme-ov-file#-prebuilt-images) refer to https://nativelink.com/docs/contribute/nix so I've made a fix there to use the easier command. There was existing things there, but it made a dynamically tagged image which was more of a pain to use, and this option gives you the same tag every time which is more useful for local running.

PR #1879, opened 2025-07-28
Make Redis connection errors actually fail as such

palfrey:
# Description

Previously, if Redis was having connection errors, you'd just get something like
```Error { code: DeadlineExceeded, messages: ["Timeout Error: Request timed out.", "In RedisStore::has_with_results::query"] }```
because the connection was failing silently behind the scenes, and then the thread that actually wanted to send something would timeout waiting for the connection to exist.

This PR explicitly waits for the connection to be up before trying to do anything with it.

## Type of change

  not work as expected)

## How Has This Been Tested?

`bazel test //...` mostly.

## Checklist

- [x] Tests added/amended

### CODE REVIEW

File: nativelink-store/Cargo.toml [hunk 1]
```diff
@@ -42,7 +42,7 @@ fred = { version = "10.1.0", default-features = false, features = [
   "i-redisearch",
   "i-scripts",
   "i-std",
-  "metrics",
+  "mocks",
```
palfrey, line 45: `metrics` wasn't actually used anymore.

We need `mocks` in here for consistency of the available attributes, especially as Cargo will compile the store without the `dev-dependency` flags, and Bazel will compile it with all the flags and so they get different test results if we try to toggle on test/non-test for the `mocks` attribute.

nativelink-store/tests/redis_store_test.rs
```diff
@@ -181,6 +182,31 @@ fn make_clients(mut builder: Builder) -> (RedisPool, SubscriberClient) {
     (client_pool, subscriber_client)
 }
 
+fn make_mock_store(mocks: &Arc<MockRedisBackend>) -> RedisStore {
```
palfrey, line 185: This all got refactored because I needed to do the store creation for the empty data tests and realised there was a lot of copy/paste here.

nativelink-store/src/redis_store.rs
```diff
@@ -293,6 +293,22 @@ impl RedisStore {
         })
     }
 
+    async fn get_client(&'_ self) -> Result<&'_ Client, Error> {
+        let config = self.client_pool.last().client_config();
+        if config.mocks.is_none() {
+            self.client_pool.wait_for_connect().await.err_tip(|| {
```
MarcusSorealheis: I understand why this is done and it definitely makes sense, given it was silently failing before. I'm curious about the overhead of `wait_for_connect()`. It only happens once, correct?

MarcusSorealheis: This is only a nit btw, and you can resolve. I'm curious in the answer only for my own understanding. Synchronous, blocking threads of execution are always interesting in high throughput scenarios like our own. I've added a fe of these recently as well, and they also piqued my interest.

nativelink-store/tests/redis_store_test.rs
MarcusSorealheis: I like to re-use the test set up as much as possible I'm also not against. Whatever makes the most sense for you. I think it's a bit early for us to be overly pedantic here on DRY vs not DRY. You've cleaned it up.

nativelink-store/tests/redis_store_test.rs
```diff
@@ -980,3 +847,23 @@ async fn dont_loop_forever_on_empty() -> Result<(), Error> {
...
+        command_timeout_ms: 10,
+        ..Default::default()
+    };
+    let store = RedisStore::new(spec).expect("Working spec");
+    let err = store
+        .has("1234")
+        .await
+        .expect_err("Wanted connection error");
+    assert_eq!(err.messages.len(), 2);
+    // err.messages[0] varies a bit, always something about lookup failures
```
MarcusSorealheis, line 863: varies a bit by OS, probably? I know we have people running on Mac runners and Windows runners (somehow).

MarcusSorealheis (approved):
LGTM!

nativelink-store/src/redis_store.rs
palfrey: The overhead is a set of `parking_lot::RwLock` being locked for read (one per client connection) plus a little bit of clone's and things internally in `fred` associated with that. There's a trivial improvement I can do to wait for just the individual connection, not the set, which will drop it to one lock v.s. one per connection, which I'll do.

I was going to note this is only kinda needed the first time, except that's not quite true. If there's been a disconnect later on and it's going through a reconnect, we'll see the same failure case without this, so I think we need to keep at least the first level "is this connected?" there.

palfrey: There's also some increased overhead in the disconnected scenarios, but I think that's acceptable given the other option is just failing to work :)

nativelink-store/tests/redis_store_test.rs
palfrey: Not just OS. My local linux box gets `Name does not resolve`, the CI Ubuntu gets `Temporary failure in name resolution`

nativelink-store/tests/redis_store_test.rs
palfrey: I'm not especially pedantic on these things, it's usually more at the point where I go "it'll be annoying to have to repeatedly copy the changes for this across many things", or in this case where it was hard to tell if they were all deliberately subtly different or if it was just a "these differences don't matter" case like here. This made it much easier for me to read the tests as well!

nativelink-store/tests/redis_store_test.rs
MarcusSorealheis: ahh, so by a network and other factors.

nativelink-store/src/redis_store.rs
MarcusSorealheis: I agree

PR #1882, opened 2025-07-28
Update git version used in Dockerfile due to drift

ayewo:
Currently, trying to standup Nativelink on AWS infra using [past code](https://github.com/TraceMachina/graveyard), as part of https://github.com/TraceMachina/nativelink/pull/1864.

When running this Docker command:
```sh
sudo docker build --build-arg OS_VERSION=$DISTRIB_RELEASE -t nativelink-runner -f ./deployment-examples/docker-compose/Dockerfile . &&
```
it fails with the following error: 
```sh
Reading state information...
Package git is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
   git-svn
E: Version '1:2.34.1-1ubuntu1.14' for 'git' was not found
```

A quick check shows `2.34.1-1ubuntu1.15` as the latest:
```sh
apt-cache policy git
git:
  Installed: 1:2.34.1-1ubuntu1.15
  Candidate: 1:2.34.1-1ubuntu1.15
  Version table:
 *** 1:2.34.1-1ubuntu1.15 500
        500 http://eu-west-2.ec2.archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages
        500 http://security.ubuntu.com/ubuntu jammy-security/main amd64 Packages
        100 /var/lib/dpkg/status
     1:2.34.1-1ubuntu1 500
        500 http://eu-west-2.ec2.archive.ubuntu.com/ubuntu jammy/main amd64 Packages
```

# Description

This change fixes the error.

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Please also list any relevant details for your test configuration

## Checklist

- [ ] Updated documentation if needed
- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [ ] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1882)

PR #1883, opened 2025-07-29
Drop 22.04 support entirely

palfrey:
# Description

As noted in https://github.com/TraceMachina/nativelink/pull/1882, we don't test 22.04 and we have specific package versions locked down. This will break when a new patch version of Ubuntu or the packages comes out and we don't test it. This PR drops 22.04 support entirely (although we probably still build just fine on it, we're just not checking it) and locks down 24.04 to a very specific revision

## Type of change

## How Has This Been Tested?

I've built all of the Dockerfile's I've changed by hand and made sure they still work and have corrected package versions.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1883)

### DIFF
--- a/tools/toolchain-drake/Dockerfile
+++ b/tools/toolchain-drake/Dockerfile
@@ -12,17 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:22.04@sha256:3c61d3759c2639d4b836d32a2d3c83fa0214e36f195a3421018dbaaf79cbe37f AS dependencies
-# hadolint ignore=DL3009
+FROM ubuntu:noble-20250925 AS dependencies
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
     apt-get install --no-install-recommends -y \
-        git=1:2.34.1-1ubuntu1.11 \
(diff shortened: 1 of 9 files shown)


palfrey:
> We already have so many tests. I want to deprecate and eventually remove one of the Mac versions, for instance. I think we should first conduct an inventory of what we need versus what we don't need.
> 
> Right now, the whole pipeline s too slow.

I've removed the 22.04 support entirely, as 24.04 is enough I'd say.

tools/toolchain-drake/Dockerfile
```diff
@@ -12,17 +12,18 @@
...
+        ca-certificates=20240203 \
+        && apt-get clean \
+        && rm -rf /var/lib/apt/lists/* \
+        && update-ca-certificates
 
 FROM dependencies AS final
 
 # hadolint ignore=DL3003
-RUN git clone https://github.com/blakehatch/drake.git && \
+RUN git clone https://github.com/RobotLocomotion/drake.git && \
```
palfrey, line 27: `blakehatch` version is very out of date and doesn't build with 24.04

PR #1892, opened 2025-08-08
Retry matching on failure

chrisstaite-menlo:
# Description

If the do_try_match algorithm fails then it does not re-try until there is another task or worker event.  However if the algorithm failed due to a worker disconnect then the action will permanently be stuck.

Ensure that the do_try_match gets to run again in a reasonable time by executing it after a short delay if the previous execution failed.

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1892)

### DIFF
--- a/nativelink-scheduler/src/simple_scheduler.rs
+++ b/nativelink-scheduler/src/simple_scheduler.rs
@@ -411,6 +411,7 @@ impl SimpleScheduler {
             let weak_inner = weak_self.clone();
             let task_worker_matching_spawn =
                 spawn!("simple_scheduler_task_worker_matching", async move {
+                    let mut last_match_successful = true;
                     // Break out of the loop only when the inner is dropped.
                     loop {
                         let task_change_fut = task_change_notify.notified();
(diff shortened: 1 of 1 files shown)


MarcusSorealheis (approved):

MarcusSorealheis:
> Looks like there's bugs with Rust 1.89 support in the code base...


Could you file an issue?

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -411,20 +411,34 @@ impl SimpleScheduler {
...
                     // Break out of the loop only when the inner is dropped.
                     loop {
                         let task_change_fut = task_change_notify.notified();
                         let worker_change_fut = worker_change_notify.notified();
                         tokio::pin!(task_change_fut);
                         tokio::pin!(worker_change_fut);
                         // Wait for either of these futures to be ready.
-                        let _ = futures::future::select(task_change_fut, worker_change_fut).await;
+                        let state_changed =
+                            futures::future::select(task_change_fut, worker_change_fut);
```
MarcusSorealheis, line 423: I wonder if we should be using `tokio::select!` here?

I think the pinning ceremony could lead to more problems. In recent work on this code, I observed significant overhead with `tokio::pin!`.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -411,20 +411,34 @@ impl SimpleScheduler {
...
+                        if last_match_successful {
+                            let _ = state_changed.await;
+                        } else {
+                            // If the last match failed, then run again after a short sleep.
+                            // This resolves issues where we tried to re-schedule a job to
+                            // a disconnected worker.  The sleep ensures we don't enter a
+                            // hard loop if there's something wrong inside do_try_match.
+                            let sleep_fut = tokio::time::sleep(Duration::from_millis(100));
+                            tokio::pin!(sleep_fut);
```
MarcusSorealheis, line 432: sleep here is good to avoid a hot loop for sure, but I favor exponential backoff with a _tiny_ bit of randomized jitter in cases like these. I'm working on a blogpost to this effect at the moment, as time permits.

MarcusSorealheis:
> Looks like there's bugs with Rust 1.89 support in the code base...

If no issue, point me to the offending lines and I will enter the operating room tonight.

MarcusSorealheis:
We'd need to comment out the `sqlite_update_metadata` benchmark.

MarcusSorealheis (approved):
LGTM you can disregard the comments now with the other fixes in. I think this is great.

MarcusSorealheis:
I n the future we could need exponential backoff but a sleep should be sufficient here.

PR #1394, opened 2024-10-04
Add graceful shutdown to worker instances

SchahinRohani:
# Description

This PR implements graceful shutdown for worker instances to ensure that the scheduler is properly notified when workers are shutting down.

## How Has This Been Tested?

`bazel test ...`

## Checklist

- [ ] Updated documentation if needed
- [ ] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1394)

PR #1899, opened 2025-08-12
Graceful worker shutdown

chrisstaite-menlo:
# Description

In [PR#1394](https://github.com/TraceMachina/nativelink/pull/1394) the worker shutdown was made to happen gracefully and wait for scheduled actions to complete.  However it missed three separate race conditions which can cause this to misbehave and timeouts to occur at the scheduler.

Firstly, there is the condition where a shutdown request is handled while an action is in-flight (being downloaded, but not yet registered with the running actions manager).  If there are no other actions then this can result in the complete_actions returning immediately even though there are still actions in-flight which then get dropped and have to time out at the scheduler.

Secondly, there is the race of sending the GoingAwayRequest to the scheduler and it assigning new work.  This means that if the scheduler assigns work in this period it will go to the in-flight list while the complete_actions returns and once again the action is dropped with no notification to the scheduler.

Finally, the complete_actions triggers as soon as the start_action_fut returns which means that the result is not uploaded to the Action Cache and the scheduler isn't notified of the result.

To resolve this we remove the complete_actions solution and instead introduce a new counter similar to the actions_in_transit which only decrements when the entire action is completed including uploading to the AC and notifying the scheduler.  This is asynchronously notified through a new Notify to avoid a busy loop.

## Type of change

## How Has This Been Tested?

Tests are passing, but not fully soak tested yet, will do that soon.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1899)

### DIFF
--- a/nativelink-worker/src/local_worker.rs
+++ b/nativelink-worker/src/local_worker.rs
@@ -190,6 +190,13 @@ impl<'a, T: WorkerApiClientTrait, U: RunningActionsManager> LocalWorkerImpl<'a, 
         let mut add_future_rx = UnboundedReceiverStream::new(add_future_rx).fuse();
 
         let mut update_for_worker_stream = update_for_worker_stream.fuse();
+        // A notify which is triggered every time actions_in_flight is subtracted.
+        let actions_notify = Arc::new(tokio::sync::Notify::new());
+        // A counter of actions that are in-flight, this is similar to actions_in_transit but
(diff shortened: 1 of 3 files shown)


MarcusSorealheis (approved):
LGTM also an improvement!

PR #1906, opened 2025-08-14
Retry on disconnect

chrisstaite-menlo:
# Description

This suite of changes hardens the Redis scheduler for scaling and ensures that operations are correctly queued.  The existing scheduler has issues during shutdown and re-queuing of failed operations.  Here we ensure that worker disconnects during action assignment do not count against attempts (worker scale down), scheduler shutdown correctly sets operations back to queued (scheduler scale down) and operation re-joining when an action has completed correctly schedules a new operation rather than leaving it permanently in the executing stage.

## Type of change

## How Has This Been Tested?

Tested on my build cluster which is now functioning without issues even when scaling.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1906)

### DIFF
--- a/nativelink-scheduler/src/simple_scheduler_state_manager.rs
+++ b/nativelink-scheduler/src/simple_scheduler_state_manager.rs
@@ -566,6 +566,7 @@ where
                         ActionStage::Queued
                     }
                 }
+                UpdateOperationType::UpdateWithDisconnect => ActionStage::Queued,
             };
             let now = (self.now_fn)().now();
             if matches!(stage, ActionStage::Queued) {
@@ -619,7 +620,11 @@ where
         action_info: Arc<ActionInfo>,
     ) -> Result<T::Subscriber, Error> {
         self.action_db
-            .add_action(new_client_operation_id, action_info)
+            .add_action(
+                new_client_operation_id,
+                action_info,
+                self.no_event_action_timeout,
+            )
             .await
             .err_tip(|| "In SimpleSchedulerStateManager::add_operation")
     }
--- a/nativelink-scheduler/src/simple_scheduler_state_manager.rs
+++ b/nativelink-scheduler/src/simple_scheduler_state_manager.rs
@@ -16,7 +16,6 @@ use core::ops::Bound;
 use core::time::Duration;
 use std::string::ToString;
 use std::sync::{Arc, Weak};
-use std::time::SystemTime;
 
 use async_lock::Mutex;
 use async_trait::async_trait;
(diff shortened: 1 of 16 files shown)


nativelink-scheduler/src/store_awaited_action_db.rs
```diff
@@ -492,74 +493,69 @@ where
...
+                            .last_worker_updated_timestamp()
+                            .duration_since(SystemTime::UNIX_EPOCH)
+                            .ok()
+                    })
+                    .and_then(|last_worker_updated| {
+                        last_worker_updated.checked_add(no_event_action_timeout)
+                    });
+                let awaited_action = if awaited_action.state().stage.is_finished()
+                    || worker_should_update_before
+                        .is_some_and(|timestamp| timestamp >= (self.now_fn)().elapsed())
```
chrisstaite-menlo: I'm staring at this line and I can't convince myself the comparison is the correct way around.  If it's not then I'm certain that this one https://github.com/TraceMachina/nativelink/blob/main/nativelink-scheduler/src/simple_scheduler_state_manager.rs#L454 is also the wrong way around.

chrisstaite-menlo: I was right... it was the wrong way round.  The only reason it wasn't causing issues was because the test is for `elapsed` rather than `now` which always returns `0` because there's never been any time elapsed between now and now...

chrisstaite-menlo: Fixed it in the latest commit.

MarcusSorealheis: `(self.now_fn)().elapsed()` yikes.. happens

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -439,19 +438,16 @@ where
...
+        let worker_should_update_before = awaited_action
             .last_worker_updated_timestamp()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .map_err(|e| {
+            .checked_add(self.no_event_action_timeout)
+            .ok_or_else(|| {
                 make_err!(
                     Code::Internal,
-                    "Failed to convert last_worker_updated to duration since epoch {e:?}"
+                    "Timestamp too big in SimpleSchedulerStateManager::timeout_operation_id"
```
MarcusSorealheis:
```suggestion
                    "Timestamp overflow for operation {} in SimpleSchedulerStateManager::timeout_operation_id",
                    operation_id
```

MarcusSorealheis: shouldn't happen often, but if it does, it might be helpful.

MarcusSorealheis (commented):
It's mostly fine. At some point, someone on the team will need to go through and instrument all of these features with telemetry but this is great and generally should not be delayed besides the operation id in the error log.

I added a suggestion but you could do it another way if you'd like.

Issue #1943, opened 2025-10-01
Eager Eviction Leads to Not Found Error


MarcusSorealheis:
```
INFO: Analyzed 262 targets (1791 packages loaded, 72437 targets configured).
ERROR: D:/workdir/servicemesh/src/third_party/zot/BUILD.bazel:62:14: OCI Image //src/third_party/zot:zot-windows-process failed: Failed to fetch blobs because they do not exist remotely.: 3 errors during bulk transfer:
com.google.devtools.build.lib.remote.common.CacheNotFoundException: Missing digest: 8529e83aa16541b76e5351d0f82fb57fa670fa9989a52085ba33a8787a751f7f/351 for bazel-out/x64_windows-opt/bin/src/third_party/zot/zot-windows-process.0.descriptor.json
com.google.devtools.build.lib.remote.common.CacheNotFoundException: Missing digest: 7c183cd3a038fc1a5f5a388cea63507cb7e6570e7528d0151b470c6f59fdf542/357 for bazel-out/x64_windows-opt/bin/src/third_party/zot/zot-windows-process.1.descriptor.json
com.google.devtools.build.lib.remote.common.CacheNotFoundException: Missing digest: c92df9d4b7c63e91d9e9ff65f947dbd641d286a9ed33dd48bad663897f9c2e9f/359 for bazel-out/x64_windows-opt/bin/src/third_party/zot/zot-windows-process.2.descriptor.json
INFO: Elapsed time: 265.487s, Critical Path: 64.65s
INFO: 2122 processes: 96 remote cache hit, 2016 internal, 10 local.
```

```
ERROR: Build did NOT complete successfully
2025-09-25T05:46:00.401435Z ERROR nativelink_service::bytestream_server: response: Error { code: NotFound, messages: ["Object c92df9d4b7c63e91d9e9ff65f947dbd641d286a9ed33dd48bad663897f9c2e9f-359 not found in either fast or slow store"] }
    at nativelink-service/src/bytestream_server.rs:369
    in nativelink_service::bytestream_server::read_stream
    in nativelink_service::bytestream_server::bytestream_read
    in nativelink_service::bytestream_server::read with request: ReadRequest { resource_name: "dev/blobs/c92df9d4b7c63e91d9e9ff65f947dbd641d286a9ed33dd48bad663897f9c2e9f/359", read_offset: 0, read_limit: 0 }
    in nativelink_util::task::http_executor
[...]

PR #1944, opened 2025-10-01
Add eviction grace period 

MarcusSorealheis:
# Description

Customers have reported problems with to prevent eager evictions and cache thrash. The grace period optionally prevents recently added cache entries from eviction.

Fixes #[1943](https://github.com/TraceMachina/nativelink/issues/1943)

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1944)


nativelink-config/src/stores.rs
```diff
@@ -850,6 +850,21 @@ pub struct EvictionPolicy {
...
+    /// in-flight actions in multi-worker deployments.
+    /// Default: 1800 (30 minutes). Zero means no grace period.
+    #[serde(
+        default = "default_eviction_grace_period",
+        deserialize_with = "convert_duration_with_shellexpand"
+    )]
+    pub eviction_grace_period_seconds: u32,
+}
+
+const fn default_eviction_grace_period() -> u32 {
```
palfrey: Can we add `secs/seconds` to this just so we're clear on units?

deployment-examples/docker-compose/multi-worker-with-grace-period.json5
```diff
@@ -0,0 +1,134 @@
...
+// them from being evicted while still needed by in-flight actions.
+{
+  stores: [
+    {
+      name: "CAS_FS",
+      filesystem: {
+        content_path: "/tmp/nativelink-multi-test/cas/content",
+        temp_path: "/tmp/nativelink-multi-test/cas/temp",
+        eviction_policy: {
+          max_bytes: 10000000000, // 10GB
```
palfrey: Nit: We should be able to use `10GB` directly (and similar for other units here) not as a comment

deployment-examples/multi-worker-with-grace-period.json5
palfrey, line 1: Does this belong here or in `nativelink-config/examples`? If we do the latter, it'll get automatically tested by `nativelink-config/tests/json5_test.rs`

nativelink-util/src/evicting_map.rs
```diff
@@ -168,6 +170,8 @@ pub struct EvictingMap<
     max_seconds: i32,
     #[metric(help = "Maximum number of items to keep in the store")]
     max_count: u64,
+    #[metric(help = "Grace period in seconds to prevent eviction of recently accessed items")]
+    eviction_grace_period_seconds: i32,
```
palfrey: Shouldn't this be `u32`?

nativelink-util/tests/evicting_map_test.rs
```diff
@@ -523,6 +531,7 @@ async fn get_evicts_on_time() -> Result<(), Error> {
             max_seconds: 5,
             max_bytes: 0,
             evict_bytes: 0,
+            eviction_grace_period_seconds: 0,
```
palfrey, line 535: Nit: These could use `..EvictionPolicy::default()` and stop needing to add new default fields

palfrey (changes requested):
Some minor items, otherwise good

nativelink-util/src/evicting_map.rs
MarcusSorealheis: yes

nativelink-util/tests/evicting_map_test.rs
MarcusSorealheis: I think i should do in a future test. I'm exhausted.

deployment-examples/docker-compose/multi-worker-with-grace-period.json5
MarcusSorealheis: pre-commit doesn't allow for the 10GB here but the API does.

nativelink-config/src/stores.rs
MarcusSorealheis: done

chrisstaite-menlo:
My main concern with this is that nativelink filesystem store does *really* badly when it reaches an out of disk space.  This isn't an issue with the existing implementation, but with this there's a possibility that it can happen.  I think if we're going to introduce this we also need to introduce a hard upper limit which starts evicting stuff that's in the grace period too.

chrisstaite-menlo (commented):
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1944)* status: 0 of 1 LGTMs obtained, and 0 of 5 files reviewed, and 7 discussions need to be resolved

___
*[`nativelink-util/src/evicting_map.rs` line 321 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1944#-Ob2xeMv2UDwtc28qRT7:-Ob2xeMv2UDwtc28qRT8:b-5r3r2b) ([raw file](https://github.com/TraceMachina/nativelink/blob/a36bcda0e1074b7a15090850e0f98c9040ea86aa/nativelink-util/src/evicting_map.rs#L321)):*
> ```XML
>         };
> 
>         let (should_evict_initial, _within_grace) = self.should_evict_with_grace_check(
> ```

Previously the work of calling `should_evict` was gated behind `self.max_bytes` and `self.evict_bytes`.  Now this work is done upfront unnecessarily.
___
*[`nativelink-util/src/evicting_map.rs` line 348 at r9](https://reviewable.io/reviews/TraceMachina/nativelink/1944#-Ob2yEqI8GUEPzlb3sFT:-Ob2yEqI8GUEPzlb3sFU:b1haqfl) ([raw file](https://github.com/TraceMachina/nativelink/blob/a36bcda0e1074b7a15090850e0f98c9040ea86aa/nativelink-util/src/evicting_map.rs#L348)):*
> ```XML
>                     // Track that grace period blocked an eviction
>                     state.grace_period_blocks.inc();
>                     info!(
> ```

This could be a very noisy log, it should probably be `debug` level.

MarcusSorealheis:
stale and unsafe now

PR #1945, opened 2025-10-01
Prevent UUID collisions

### LINKED ISSUE #1931: Error "Cannot upload same UUID simultaneously"
I've been trying out nativelink with 3 workers and a simple CAS, using reclient as the client, building a fairly large C++ project.

Recently I ran into an error (which was also sent to `reclient` in the response) that I don't understand:

```
ERROR nativelink_service::bytestream_server: error: status: InvalidArgument, message: "Cannot upload same UUID simultaneously : In ByteStreamServer::write", details: [], metadata: MetadataMap { headers: {} }
```

This was followed by more errors:
```
  2025-09-04T20:08:42.606341Z ERROR nativelink_service::bytestream_server: error: status: Internal, message: "Error receiving first message in stream : Could not unwrap first stream message", details: [], metadata: MetadataMap { headers: {} }
  2025-09-04T20:08:42.606356Z ERROR nativelink_service::bytestream_server: error: status: Internal, message: "status: Internal, message: \"Unexpected EOF decoding stream.\", details: [], metadata: MetadataMap { headers: {} } : Stream error at byte 1048576 : In ByteStreamServer::write", details: [], metadata: MetadataMap { headers: {} }
```

and a lot more of the "Unexpected EOF decoding stream".

Unfortunately I don't have the full logs anymore and I haven't been able to reproduce it.

MarcusSorealheis:
# Description

Append a timestamp to UUID to prevent collisions. 

Fixes #1931 

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1945)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -104,12 +104,14 @@ build --@rules_rust//:clippy_flag=-Dclippy::alloc_instead_of_core
 build --@rules_rust//:clippy_flag=-Dclippy::as_underscore
 build --@rules_rust//:clippy_flag=-Wclippy::dbg_macro
 build --@rules_rust//:clippy_flag=-Wclippy::decimal_literal_representation
+build --@rules_rust//:clippy_flag=-Dclippy::elidable_lifetime_names
 build --@rules_rust//:clippy_flag=-Aclippy::get_unwrap
 build --@rules_rust//:clippy_flag=-Dclippy::missing_const_for_fn
 build --@rules_rust//:clippy_flag=-Aclippy::missing_docs_in_private_items
 build --@rules_rust//:clippy_flag=-Wclippy::print_stdout
+build --@rules_rust//:clippy_flag=-Dclippy::semicolon_if_nothing_returned
 build --@rules_rust//:clippy_flag=-Dclippy::std_instead_of_core
-build --@rules_rust//:clippy_flag=-Wclippy::todo
+build --@rules_rust//:clippy_flag=-Dclippy::todo
 build --@rules_rust//:clippy_flag=-Aclippy::too_long_first_doc_paragraph
 build --@rules_rust//:clippy_flag=-Wclippy::unimplemented
 build --@rules_rust//:clippy_flag=-Aclippy::unwrap_in_result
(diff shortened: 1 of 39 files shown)


nativelink-service/src/bytestream_server.rs
```diff
@@ -177,6 +178,15 @@ impl Debug for ByteStreamServer {
 }
 
 impl ByteStreamServer {
+    /// Generate a unique UUID by appending a nanosecond timestamp to avoid collisions.
+    fn generate_unique_uuid(base_uuid: &str) -> String {
```
amankrx, line 182: I don't think this is necessary as UUIDs should always be unique. It is really unlikely that this error occurs due to UUID collision.

MarcusSorealheis: I agree that this is a symptom not a source of the issue. However, I'm trying to discover why.

MarcusSorealheis: Do you still feel this way given all tests are working?

MarcusSorealheis: By test, I mean customer tests

amankrx: We still can't be sure if this fixes the issue. The above UUID error was flaky. We should run the tests a couple of times before we merge this PR.

MarcusSorealheis: Agree

nativelink-service/src/bytestream_server.rs
```diff
@@ -227,23 +237,51 @@ impl ByteStreamServer {
...
+                if let Some(idle_stream) = maybe_idle_stream.1.take() {
+                    // Case 2: Stream exists but is idle, we can resume it
+                    let bytes_received = maybe_idle_stream.0.clone();
+                    info!(msg = "Joining existing stream", entry = ?entry.key());
+                    return idle_stream.into_active_stream(bytes_received, self);
+                }
+                // Case 3: Stream is active - generate a unique UUID to avoid collision
+                // Using nanosecond timestamp makes collision probability essentially zero
+                let original_uuid = entry.key().clone();
+                let unique_uuid = Self::generate_unique_uuid(&original_uuid);
```
palfrey, line 269: Wondering about this one. Can't we just make an entirely new UUID rather than doing the timestamp thing? I can see some benefits of the timestamp stuff (mostly to go "these two streams should have been the same thing" from eyeballing the logs), but a new fresh UUID is definitely unique and we don't have to even worry about the unlikely timestamp case.

palfrey: We're doing this so we can more easily track the real problem later on v.s. hiding it entirely with new uuids.

PR #1947, opened 2025-10-02
Existence cache remove callbacks

palfrey:
# Description

There's an issue with the existence cache if one or more of it's underlying stores evicts items. Namely, when they do so, they don't tell anyone about it, and so the existence cache still says "yup, got that" and then clients will fail when they try to retrieve the item. This is believed (but not certain) as the source of the `not found in either fast or slow store` errors that have been cropping up. This PR adds a callback mechanism for evictions so compound stores that need to know about their inner stores evicting things can get notified.

Consequences of this change:
* Various configs that were previously regarded as valid are no longer, primarily existence caches with `GrpcStore` as an internal store, as we don't get "I evicted this" feedback from those. These will now fail at runtime (FIXME: actually make this fail)
* A bunch of `StoreKey<'_>` have become `StoreKey<'static>` and the `EvictingMap` type has become more complicated to support using the borrow key there as the callback key. This has resulted in some extra `clone`'s and probably some slowdowns. This is believed to be the best option we have that gets us the improved stability from this PR.

## Type of change

  not work as expected)

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

### DIFF
--- a/nativelink-store/src/existence_cache_store.rs
+++ b/nativelink-store/src/existence_cache_store.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use core::ops::DerefMut;
 use core::pin::Pin;
 use std::borrow::Cow;
-use std::ops::DerefMut;
 use std::sync::Arc;
 use std::time::SystemTime;
 
@@ -58,7 +58,7 @@ pub struct ExistenceCacheStore<I: InstantWrapper> {
     // as if it immediately expires them, we should only apply the remove callbacks
     // afterwards. If this is None, we're not pausing; if it's Some it's the location to
     // store them in temporarily
-    pause_remove_callbacks: Arc<Mutex<Option<Vec<StoreKey<'static>>>>>
+    pause_remove_callbacks: Arc<Mutex<Option<Vec<StoreKey<'static>>>>>,
 }
 
 impl ExistenceCacheStore<SystemTime> {
@@ -106,7 +106,7 @@ impl<I: InstantWrapper> ExistenceCacheStore<I> {
         let existence_cache_store = Arc::new(Self {
             inner_store,
             existence_cache: EvictingMap::new(eviction_policy, anchor_time),
-            pause_remove_callbacks: Arc::new(Mutex::new(None))
+            pause_remove_callbacks: Arc::new(Mutex::new(None)),
         });
         let other_ref = existence_cache_store.clone();
         existence_cache_store
--- a/nativelink-store/src/existence_cache_store.rs
+++ b/nativelink-store/src/existence_cache_store.rs
@@ -14,6 +14,7 @@
 
 use core::pin::Pin;
 use std::borrow::Cow;
+use std::ops::DerefMut;
 use std::sync::Arc;
 use std::time::SystemTime;
 
@@ -29,7 +30,8 @@ use nativelink_util::instant_wrapper::InstantWrapper;
 use nativelink_util::store_trait::{
     RemoveItemCallback, Store, StoreDriver, StoreKey, StoreLike, UploadSizeInfo,
 };
-use tracing::debug;
+use parking_lot::Mutex;
+use tracing::{debug, info, trace};
 
 #[derive(Clone, Debug)]
 struct ExistenceItem(u64);
@@ -51,6 +53,12 @@ pub struct ExistenceCacheStore<I: InstantWrapper> {
     #[metric(group = "inner_store")]
     inner_store: Store,
     existence_cache: EvictingMap<DigestInfo, DigestInfo, ExistenceItem, I>,
+
+    // We need to pause them temporarily when inserting into the inner store
+    // as if it immediately expires them, we should only apply the remove callbacks
+    // afterwards. If this is None, we're not pausing; if it's Some it's the location to
+    // store them in temporarily
+    pause_remove_callbacks: Arc<Mutex<Option<Vec<StoreKey<'static>>>>>
 }
 
 impl ExistenceCacheStore<SystemTime> {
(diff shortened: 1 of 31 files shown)


nativelink-store/src/callback_utils.rs
```diff
@@ -0,0 +1,39 @@
...
+        Self {
+            callback_fn: callback.clone(),
+        }
+    }
+}
+
+#[async_trait]
+impl RemoveStateCallback<StoreKey<'static>> for RemoveItemCallbackHolder {
+    async fn callback(&self, key: &StoreKey<'static>) {
+        self.callback_fn.callback(key).await;
```
MarcusSorealheis, line 39: The `RemoveItemCallback` trait could use a module-level doc comment explaining its purpose (e.g., "For notifying caches of deletions from backing stores").

It's a pretty big piece of functionality.

nativelink-store/src/existence_cache_store.rs
```diff
@@ -47,7 +51,13 @@ impl LenEntry for ExistenceItem {
...
     #[metric(group = "inner_store")]
     inner_store: Store,
-    existence_cache: EvictingMap<DigestInfo, ExistenceItem, I>,
+    existence_cache: EvictingMap<DigestInfo, DigestInfo, ExistenceItem, I>,
+
+    // We need to pause them temporarily when inserting into the inner store
+    // as if it immediately expires them, we should only apply the remove callbacks
+    // afterwards. If this is None, we're not pausing; if it's Some it's the location to
+    // store them in temporarily
+    pause_remove_callbacks: Arc<Mutex<Option<Vec<StoreKey<'static>>>>>,
```
MarcusSorealheis, line 60: Are paused callbacks still flushed if an update fails mid way?

nativelink-store/src/shard_store.rs
```diff
@@ -167,7 +167,7 @@ impl StoreDriver for ShardStore {
             .collect();
         // Bucket each key into the store that it belongs to.
         keys.iter()
-        .cloned()
+            .cloned()
```
MarcusSorealheis: I'm not blaming you because it was here for almost two years. It think we could be more efficient than `.cloned().enumerate()`. The only reason I would even bring it up is the fact that `shard_store` really only matters for large workloads.

nativelink-store/src/existence_cache_store.rs
```diff
@@ -184,21 +231,37 @@ impl<I: InstantWrapper> StoreDriver for ExistenceCacheStore<I> {
...
+                }
+            }
+        }
         result
     }
 
     async fn get_part(
         self: Pin<&Self>,
-        key: StoreKey<'_>,
+        key: StoreKey<'static>,
```
MarcusSorealheis: Can you explain this change to me? No fault of your own, btw.

MarcusSorealheis: I wasn’t suggesting you change. I wanted to know what was the rationale, mostly.

palfrey: It was always one of those changes that I wasn't entirely happy with, but I thought it was needed in order to support some part of the evicting map changes. It turns out that I think it was needed earlier on, but not anymore, so I've been able to reset this.

nativelink-store/src/shard_store.rs
palfrey: Removed now the `'static` fixes are done.

nativelink-store/src/existence_cache_store.rs
palfrey: Yes. The update is in two phases - inner store and then the existence cache - and both don't do any returns of errors until after the `pause_remove_callbacks` has been reset.

nativelink-store/src/callback_utils.rs
palfrey: Yup, I'll add a few doc comments to the various new traits tomorrow.

.bazelrc
```diff
@@ -152,12 +154,15 @@ test --experimental_ui_max_stdouterr_bytes=104857600
 build:nightly --@rules_rust//rust/toolchain/channel=nightly
 
 # AddressSanitizer to detect memory leaks.
-build:asan --config=nightly
+build:asan --config=nightly --config=debug
 build:asan --@rules_rust//:extra_rustc_flags=-Zsanitizer=address
+# asan explicitly needs to _not_ use the Nix toolchains and so this
+# is needed to load the rust toolchains in Bazel
+build:asan --extra_toolchains=@rust_toolchains//:all
```
palfrey, line 161: Needed this to make it run locally

.bazelrc
```diff
@@ -152,12 +154,15 @@ test --experimental_ui_max_stdouterr_bytes=104857600
 build:nightly --@rules_rust//rust/toolchain/channel=nightly
 
 # AddressSanitizer to detect memory leaks.
-build:asan --config=nightly
+build:asan --config=nightly --config=debug
```
palfrey: I'd hoped this would add debug symbols to asan results as
`#0 0x56355b780794  (/home/palfrey/.cache/bazel/_bazel_palfrey/6b0cb87e287251de607ffc2b7ceb90ad/execroot/_main/bazel-out/k8-dbg/bin/nativelink-store/test-2651107275/integration_tests/existence_store_test_test+0x2fac794)` isn't very useful, but not so

.bazelrc
```diff
@@ -154,10 +156,13 @@ build:nightly --@rules_rust//rust/toolchain/channel=nightly
...
 build:asan --config=nightly
 build:asan --@rules_rust//:extra_rustc_flags=-Zsanitizer=address
+# asan explicitly needs to _not_ use the Nix toolchains and so this
+# is needed to load the rust toolchains in Bazel
+build:asan --extra_toolchains=@rust_toolchains//:all
 
 # ThreadSanitizer to detect data races. Tests under tsan shouldn't be cached
 # since they tend to produce irreproducible false negatives.
-build:tsan --config=nightly -c dbg
+build:tsan --config=nightly --config=debug
```
palfrey, line 165: Same effect, but clearer

PR #1948, opened 2025-10-03
Fixes all the examples in the stores config

palfrey:
# Description

Some of the examples in `stores.rs` were out of date. This PR brings them all up-to-date, lines up the JSON formatting, and uses byte syntax not `"number" // bytes comments` syntax for the example values.

## Type of change

## How Has This Been Tested?

`bazel test  //...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1948)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,8 @@
 [workspace]
-exclude = ["tools/generate-bazel-rc"]
+exclude = [
+  "nativelink-config/generate-stores-config",
+  "tools/generate-bazel-rc",
+]
 resolver = "2"
 
 [package]


MarcusSorealheis (approved):

nativelink-config/src/stores.rs
```diff
@@ -313,20 +311,18 @@ pub enum StoreSpec {
...
+    ///   "backend": {
+    ///     "memory": {
+    ///       "eviction_policy": {
+    ///         "max_bytes": "500mb",
     ///       }
     ///     }
+    ///   },
+    ///   // Note this is the existence store policy, not the backend policy
+    ///   "eviction_policy": {
+    ///     "max_seconds": "100",
```
ilsubyeega:
```suggestion
    ///     "max_seconds": 100,
```

as per `web/platform/src/content/docs/docs/config/production-config.mdx`

nativelink-config/src/stores.rs
```diff
@@ -441,19 +435,19 @@ pub enum StoreSpec {
...
-    ///     "lower_store": {
-    ///       "memory": {
-    ///         "eviction_policy": {
-    ///           "max_bytes": "${NATIVELINK_CAS_MEMORY_CONTENT_LIMIT:-100000000}"
-    ///         }
+    ///   "size": "128mib",
+    ///   "lower_store": {
+    ///     "memory": {
+    ///       "eviction_policy": {
+    ///         "max_bytes": "${NATIVELINK_CAS_MEMORY_CONTENT_LIMIT:-100000000}"
```
ilsubyeega: I'm not sure whether its available to do this or not, can we just use like `${NATIVELINK_CAS_MEMORY_CONTENT_LIMIT:-100000000}` to `${NATIVELINK_CAS_MEMORY_CONTENT_LIMIT:-100mb}`?

ilsubyeega (commented):
LGTM for me but some suggestions

nativelink-config/src/stores.rs
MarcusSorealheis: Oh wow, great catch!

nativelink-config/src/stores.rs
MarcusSorealheis: another good comment! @palfrey wdyt about supporting this format? There are risks but I think we need to weigh flexibility, consistency, and usability in the config. mixing math and letters is tough sometimes.

palfrey: We already support it :) It's done via [shellexpand](https://docs.rs/shellexpand/latest/shellexpand/) which happily supports this as an option. I've changed the example config to show this.

nativelink-config/src/stores.rs
palfrey: Nice! There were actually a few other items like this, so I've added in a generator tool to make https://github.com/palfrey/nativelink/blob/correct-store-docs/nativelink-config/examples/stores-config.json5 so we can catch things like this via the existing tests for valid config.

PR #1951, opened 2025-10-06
Move Bytestream to array config

palfrey:
# Description

Like https://github.com/TraceMachina/nativelink/pull/1806, but with backwards compatibility. Note, that when upgrading the config the `max_decoding_message_size` has moved into the `http` part of the config as it's set per-server.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended


nativelink-config/src/cas_server.rs
```diff
@@ -191,11 +191,6 @@ pub struct ByteStreamConfig {
     #[serde(default, deserialize_with = "convert_data_size_with_shellexpand")]
     pub max_bytes_per_stream: usize,
 
-    /// Maximum number of bytes to decode on each grpc stream chunk.
```
MarcusSorealheis, line 194: This is a breaking change in security boundaries. If someone had different message size limits for different services, they’ll lose that capability. 

It’s a fringe use case so should be fine but we should check with the Chromium users.

nativelink-config/examples/legacy_service_config.json5
```diff
@@ -0,0 +1,42 @@
+{
+  stores: [],
```
MarcusSorealheis, line 2: Should this be called legacy_bytestream_config.json5 or something?

MarcusSorealheis:
Tests should pass twice before merging

nativelink-config/examples/legacy_service_config.json5
palfrey: It's actually the legacy config for all the services, but I'll rename it as the legacy config

nativelink-config/src/cas_server.rs
palfrey: 👍🏻 In theory we could do something where we support setting all the http_config stuff at a per-service level, but that would require redoing the config _again_. It does feel like this being supported for bytestream but not any of the other services is a bit odd TBH.

PR #1958, opened 2025-10-10
Detect anonymous GCS auth and optionally quit

### LINKED ISSUE #1953: Detect GCS authentication failure and quit
Sometimes the GKE infrastructure doesn't set up authentication correctly and ends up with errors like this from the GCS store:

```
[6998/47774] 18m39.60s 2m54.15s[remote]: CXX obj/third_party/perfetto/protos/perfetto/trace/non_minimal_cpp/trace_packet.gen.oE1007 15:42:44.126075 1186210 exec.go:246] c14992e0-7d08-4d59-bd3d-3c84e5e0897a execute response: status code:7 message:"GCS operation failed: Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist). : Error while trying to read - bucket: <bucket> path: f287b769cbe8452b3f794ab5a4c63e2dbd7a4a1978180d0d0729a754d4fb4d9e-355 : Failed to run has() on slow store : Failed to get_part in get_part_unchunked : --- : Sender dropped before sending EOF : During first read of buf_channel::take() : Failed to read stream to completion in get_part_unchunked : During start_action : --- : Job cancelled because it attempted to execute too many times 4 > 3 times for operation_id: f2462435-5d62-4350-8d16-fe0a4dc7e91d, maybe_worker_id: Some(1f0a38bb-007d-6469-ae7f-7cee8905de2e)"
```

If this occurs then the worker should either try to get a new token or kill itself to be replaced with a worker that is actually configured properly.

chrisstaite-menlo:
# Description

The GCS store currently attempts to get authentication, but happily continues without if it's unavailable.  Sometimes a GKE worker fails to be configured correctly and that breaks all jobs scheduled to it.

Add a new configuration item which defaults to the existing behaviour but allows users to quit if authentication fails which will allow the system to recover automatically.

Update the GCS dependencies to their latest versions which provide most of the set-up functionality meaning we can drastically simplify our implementation.

Switch from using an async_trait to a modern impl Future trait and update the GcsStore to take a generic rather than requiring costly and unnecessary dynamic dispatch.

Fixes #1953

## Type of change

## How Has This Been Tested?

Tested on my cluster to ensure it functions.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1958)


nativelink-store/src/gcs_store.rs
```diff
@@ -40,8 +40,8 @@ use crate::gcs_client::types::{
 };
 
 #[derive(MetricsComponent, Debug)]
-pub struct GcsStore<NowFn> {
-    client: Arc<dyn GcsOperations>,
+pub struct GcsStore<Client: GcsOperations, NowFn> {
```
MarcusSorealheis, line 43: big improvement

PR #1961, opened 2025-10-10
Add timeout to health check

palfrey:
# Description

Especially when using Nativelink in Kubernetes, we've seen a number of instances of errors like `Readiness probe failed: Get "http://10.112.5.3:9090/status"` and the related `{"message":"Failed running service","err":"hyper::Error(IncompleteMessage)"}` in the logs. This PR adds a timeout setting to the health check to enable determining which of the health checks is causing the timeout issues.

After these changes, the log now looks like `WARN nativelink_util::health_utils: Timeout during health check, struct_name: "nativelink_store::redis_store::RedisStore"`

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1961)


nativelink-store/tests/redis_store_test.rs
```diff
@@ -866,3 +867,36 @@ fn test_connection_errors() {
         "Connection issue connecting to redis server with hosts: [\"non-existent-server:6379\"], username: None, database: 0"
     );
 }
+
+#[nativelink_test]
+fn test_health() {
```
palfrey, line 872: This turned out to be less broken than I originally thought, but leaving in as an additional useful test.

PR #1975, opened 2025-10-14
Notify execution complete

### LINKED ISSUE #1903: Report operation complete while uploading results
Currently the scheduler doesn't release workers to perform another action until the previous action has finished uploading its results.  This is very CPU inefficient.

In order to optimise this we can introduce a new message from the worker to the scheduler to inform it that the execution is complete and then allow it to schedule new work to start downloading and executing while the previous action continues to upload its results.  Obviously we can't simply report the action status early as this will cause the user to attempt to download the results before they are available hence the need for a new message.

chrisstaite-menlo:
# Description

When execution is complete, there's a large amount of IO still to be done.  In the mean time a new action could be starting.

Previously an attempt to implement this was quite complex and caused panics.  In this implementation a very simple mechanism is used which only executes on success and keeps track of which operations have been notified on the scheduler.  This massively simplifies things.

Fixes #1903

## Type of change

## How Has This Been Tested?

Ran up on my test cluster and it was running at 100% constantly.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1975)

### DIFF
--- a/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
+++ b/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
@@ -57,6 +57,9 @@ service WorkerApi {
 
     /// Informs the scheduler about the result of an execution request.
     rpc ExecutionResponse(ExecuteResult) returns (google.protobuf.Empty);
+
+    /// Notify that the execution has completed, but result is uploading.
+    rpc ExecutionComplete(ExecuteComplete) returns (google.protobuf.Empty);
 }
 
 /// Request object for keep alive requests.
@@ -121,6 +124,15 @@ message ExecuteResult {
     }
 
     reserved 9; // NextId.
+}
+
+/// The result of an ExecutionComplete.
+message ExecuteComplete {
+    /// ID of the worker making the request.
+    string worker_id = 1;
+
+    /// The operation ID that was executed.
+    string operation_id = 2;
 }
 
 /// Result sent back from the server when a node connects.
--- a/nativelink-scheduler/src/simple_scheduler_state_manager.rs
+++ b/nativelink-scheduler/src/simple_scheduler_state_manager.rs
@@ -629,6 +629,11 @@ where
                     }
                 }
                 UpdateOperationType::UpdateWithDisconnect => ActionStage::Queued,
+                // We shouldn't get here, but we just ignore it if we do.
+                UpdateOperationType::ExecutionComplete => {
+                    warn!("inner_update_operation got an ExecutionComplete, that's unexpected.");
+                    return Ok(());
+                }
             };
             let now = (self.now_fn)().now();
             if matches!(stage, ActionStage::Queued) {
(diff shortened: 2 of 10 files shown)


chrisstaite-menlo:
That sounds like a standard overloaded system with a worker keep alive set too low.  Notably #1977 should actually help with this as it doesn't require setting up a separate channel.

You may be seeing this with this change as it's specifically designed to keep worker utilisation high.

nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
```diff
@@ -123,6 +126,17 @@ message ExecuteResult {
     reserved 9; // NextId.
 }
 
+/// The result of an ExecutionComplete.
+
+/// The result of an ExecutionRequest.
```
palfrey: Why the two comments?

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -629,6 +629,8 @@ where
                     }
                 }
                 UpdateOperationType::UpdateWithDisconnect => ActionStage::Queued,
+                // We shouldn't get here, but we just ignore it if we do.
+                UpdateOperationType::ExecutionComplete => return Ok(()),
```
palfrey: Possibly worth logging, given it shouldn't occur, so we can debug and figure out why later

nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
chrisstaite-menlo: I assume this was the result of a branch change :)

nativelink-scheduler/src/simple_scheduler_state_manager.rs
chrisstaite-menlo: Sure.

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -629,6 +629,13 @@ where
                     }
                 }
                 UpdateOperationType::UpdateWithDisconnect => ActionStage::Queued,
+                // We shouldn't get here, but we just ignore it if we do.
+                UpdateOperationType::ExecutionComplete => {
+                    tractin::warn!(
```
palfrey: Assuming you meant `tracing`?

chrisstaite-menlo: This is what happens when you edit on the command line rather than in an IDE!

palfrey (approved):
:lgtm:

> @palfrey reviewed 7 of 10 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1975#-:-OblojNQBgiKOCWzZepm:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

PR #1976, opened 2025-10-14
Property replace

chrisstaite-menlo:
# Description

Currently the property modifier scheduler can add or remove properties.  It would be useful to replace a property too, for example Siso sets the label:action_large=1 property and we would like to switch that to cpu_count=2000 instead.

Implement an additional option to the property modifier to replace properties as well as add and remove.  This has the option to match values and the option to replace the value or pass it through.

## Type of change

## How Has This Been Tested?

Added new tests and tested on my local cluster.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1976)

### DIFF
--- a/nativelink-config/src/schedulers.rs
+++ b/nativelink-config/src/schedulers.rs
@@ -196,12 +196,29 @@ pub struct PlatformPropertyAddition {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


palfrey (approved):
Nice, this looks really useful!

PR #1977, opened 2025-10-14
Single worker stream

chrisstaite-menlo:
# Description

When there are multiple schedulers in high availability mode then the workers can end up communicating with the wrong scheduler and state can get confused.

Modify the communications such that there is a single bi-directional gRPC stream between the client and the worker.  This way they will always be one-to-one mapped even with a load balancer in front.

## Type of change

  not work as expected)

## How Has This Been Tested?

Running on my cluster this now functions perfectly with a one-to-one worker to scheduler and removes the races.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1977)


palfrey (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1977#-:-ObhYJ7dER2oJN60QEKQ:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 LGTMs obtained, and all files reviewed

palfrey:
https://github.com/TraceMachina/nativelink/pull/1975 is now merged, and I'm guessing that's the source of conflicts here

PR #1978, opened 2025-10-14
Fast-slow deadlock

chrisstaite:
# Description

There are a number of different semaphores in the system, for example the file open semaphore and the GCS connections semaphore.  When the fast-slow store interacts between these then it can cause deadlocks.

Synchronising the collection of semaphores throughout the system is incredibly hard with the interfaces that are in place.  Therefore we don't even try to.

Instead, add a check for the reader and writer being started for both sides of the fast-slow store and time out if they aren't.  This should catch deadlocks and kick the system back to life as a watchdog timer.  It's not the best solution, but it's something for now.

## Type of change

## How Has This Been Tested?

Tests are still passing, not tried the code out in production yet.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1978)


palfrey:
I'm trying to understand what the improvements here are. I'm seeing a lot of replacement of `with_connection` in GCS client with something that does a very similar Semaphore call, but replaces `.acquire()` with `.clone().acquire_owned()` which doesn't feel like an improvement?

There's then what feels like a separate chunk of changes in `fast_slow_store.rs` which feels like it's the core of the changes here, but lacking in any tests around it. It looks ok at a first glance, but I'd love to see some tests demoing the sorts of problems it's meant to solve? 

There's also the GCS store changes, which look like a fairly sensible set of "lets do the once a loop stuff once before the loop" changes.

This feels like at least two PRs, maybe three? Thoughts?

chrisstaite:
The issue with the GCS store is (as well as nested semaphore use) it doesn't hold a semaphore for the whole operation. Therefore detecting deadlocks externally is pretty much impossible compared to slow IO.

Once there's a single semaphore used we can add the watchdog timer to the fast-slow initial start to detect any deadlocks.

chrisstaite:
Quite right that a test was needed.  It was a little tricky to get working due to the triple buffering, but the test breaks on main and passes on this branch.

chrisstaite-menlo:
Hmm... without it I'm seeing a lot of deadlocked workers which start up, get issued a few actions and then sit there idling with no logs at debug level.  I think there might be an issue with the reqwest client retry logic used by gcloud-storage interacting with the retry logic in our layer on top.

nativelink-store/src/gcs_client/client.rs
```diff
@@ -41,6 +41,12 @@ use crate::gcs_client::types::{
     SIMPLE_UPLOAD_THRESHOLD, Timestamp,
 };
 
+#[derive(Debug)]
+pub struct UploadRef {
+    pub upload_ref: String,
+    pub(crate) _permit: OwnedSemaphorePermit,
```
MarcusSorealheis, line 47: This is key because if means that connections are only held during active uploads, I think.

nativelink-store/src/fast_slow_store.rs
```diff
@@ -185,8 +199,62 @@ impl FastSlowStore {
         let send_range = offset..length.map_or(u64::MAX, |length| length + offset);
         let mut bytes_received: u64 = 0;
 
-        let (mut fast_tx, fast_rx) = make_buf_channel_pair();
-        let (slow_tx, mut slow_rx) = make_buf_channel_pair();
+        // There's a strong possibility of a deadlock here as we're working with multiple
```
MarcusSorealheis, line 206: Important fix.

MarcusSorealheis (dismissed):
This was a complex one for me to review. I had only a few minor thoughts on it. I think we could potentially make this configurable in the future. It's the never ending challenge of trying to have generalizable piece of infrastructure but the use cases and scales are so different.

MarcusSorealheis:
stale and unsafe now

PR #1981, opened 2025-10-15
Fix removal state

chrisstaite:
# Description

The changes to the EvictionMap to add removal callbacks introduced a lot of memory allocations, async locks and dynamic dispatch.  This trashes the performance of the EvictionMap.

Fix the implementation to avoid all of the indirection through generics and move callbacks to outside of the locks to avoid deadlocks and issues with contention.

## Type of change

## How Has This Been Tested?

Tests still passing.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1981)

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,56 +20,56 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  nativelink-dot-com-cloud-rbe-main-legacy-dockerfile-test:
-    runs-on: ubuntu-24.04
-    environment: production
-    name: NativeLink.com Cloud / RBE on Main (Legacy Dockerfile Test)
-    if: github.ref == 'refs/heads/main'
-    steps:
-    - name: Checkout
-      uses: >- # v4.2.2
-        actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+  # nativelink-dot-com-cloud-rbe-main-legacy-dockerfile-test:
+  #   runs-on: ubuntu-24.04
+  #   environment: production
+  #   name: NativeLink.com Cloud / RBE on Main (Legacy Dockerfile Test)
+  #   if: github.ref == 'refs/heads/main'
+  #   steps:
+  #   - name: Checkout
+  #     uses: >- # v4.2.2
+  #       actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
-    - name: Set up AWS CLI
-      uses: >- # v4.1.0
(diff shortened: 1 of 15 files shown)


MarcusSorealheis:
@chrisstaite Could you show the performance data you collected or the tests you ran so we can repro?

nativelink-store/src/completeness_checking_store.rs
```diff
@@ -392,9 +392,9 @@ impl StoreDriver for CompletenessCheckingStore {
 
     fn register_remove_callback(
         self: Arc<Self>,
-        callback: &Arc<Box<dyn RemoveItemCallback>>,
+        callback: Arc<dyn RemoveItemCallback>,
     ) -> Result<(), Error> {
-        self.ac_store.register_remove_callback(callback)?;
+        self.ac_store.register_remove_callback(callback.clone())?;
```
MarcusSorealheis, line 397: I'm not convinced that this is a performance improvement. It may be, but we need evidence. cc @palfrey

MarcusSorealheis: Box types are not ideal

MarcusSorealheis:
> Using `FuturesUnordered` for all these Future loops is a much more efficient way to perform multiple tasks.

Yes, very much so.

MarcusSorealheis (approved):
LGTM

nativelink-store/src/redis_store.rs
```diff
@@ -660,7 +660,7 @@ impl StoreDriver for RedisStore {
 
     fn register_remove_callback(
         self: Arc<Self>,
-        _callback: &Arc<Box<dyn RemoveItemCallback>>,
+        _callback: Arc<dyn RemoveItemCallback>,
```
MarcusSorealheis, line 663: Something is wrong here

chrisstaite: What do you mean?  It's simply updating to the trait change.

chrisstaite: It's worth noting that this should probably return an error in the same way that the gRPC store does because the Redis LRU does not notify on object removal.

nativelink-store/src/redis_store.rs
```diff
@@ -660,7 +660,7 @@ impl StoreDriver for RedisStore {
...
     ) -> Result<(), Error> {
         // As redis doesn't drop stuff, we can just ignore this
```
chrisstaite, line 665: @palfrey this isn't true, Redis runs with a global LRU which automatically clears keys, probably should return an error in the same way as the gRPC store.  That's a breaking change though, so should be separate.

PR #1996, opened 2025-10-20
Unify all the service setups with a macro

palfrey:
# Description

The individual services all get setup with the same code, and it looks like the sort of thing you could unify with a function. Except, the functions that are being called have common signatures, but not common traits, because they're all generated by `protoc`. This PR instead unifies them with a macro, which also means everyone now gets the common `max_decoding_message_size` HTTP settings that were previously only applied to bytestream

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1996)


MarcusSorealheis (approved):
DRY is overrated but here it is ok.

PR #2002, opened 2025-10-22
Run coverage on standard runner

palfrey:
# Description

Fixes # (issue)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2002)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -362,37 +362,34 @@ impl LenEntry for FileEntryImpl {
     // target file location to the new temp file. `unref()` should only ever be called once.
     #[inline]
     async fn unref(&self) {
-        {
-            let mut encoded_file_path = self.encoded_file_path.write().await;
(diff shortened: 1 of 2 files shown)


nativelink-store/src/filesystem_store.rs
```diff
@@ -362,37 +362,34 @@ impl LenEntry for FileEntryImpl {
...
-            } else {
-                debug!(
-                    key = ?encoded_file_path.key,
-                    ?from_path,
-                    ?to_path,
-                    "Renamed file",
-                );
-                encoded_file_path.path_type = PathType::Temp;
-                encoded_file_path.key = new_key;
-            }
```
palfrey, line 395: This change removes an additional extra `{`/`}` pair from around the function. It should have zero actual effects to the code (other than making it look nicer) and is primarily here so the Coverage tool doesn't just load from cache as we need to check it'll work when rebuilding.

PR #2012, opened 2025-10-27
fix: guard shutting down in scheduler while SIGTERM

### LINKED ISSUE #2010: shutdown guard did not guard components shutting down gracefully completely
### Background
We configure a worker and a scheduler in same process, scheduler uses Redis as backend while worker connects to the scheduler in the same process, and we make it a cloneSet with HPA，but found Bazel client hanging on waiting actions while the pod (scheduler, worker) is already shutdown with 143 exit code.

### Problem

1. The guard implementation work like a reference counter, **if** worker / scheduler receive guard before t2, main will wait until them shutdown gracefully, but the condition itself is not guarded.

```Rust
    // nativelink.rs
    #[cfg(target_family = "unix")]
    #[expect(clippy::disallowed_methods, reason = "signal handler on main runtime")]
    runtime.spawn(async move {
        signal(SignalKind::terminate())
            .expect("Failed to listen to SIGTERM")
            .recv()
            .await;
        warn!("Process terminated via SIGTERM",);
        drop(shutdown_tx_clone.send(shutdown_guard.clone()));  //  <- t1: here send guard to shutdown_rx listened in worker/scheduler
        let () = shutdown_guard.wait_for(Priority::P0).await;  //  <- t2: here worker/scheduler may not receive guard yet
        warn!("Successfully shut down nativelink.",);
        std::process::exit(143);
    });
```
2. Worker will reconnect to scheduler with a new worker_id after shutdown.

zpzjzj:
# Description

1. Fix guard wait_for finished before receiver receive cloned guard
2. Add worker_scheduler reject further registration after shutting down

Fixes #2010 

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2012)

### DIFF
--- a/deployment-examples/docker-compose/worker.json5
+++ b/deployment-examples/docker-compose/worker.json5
@@ -41,6 +41,7 @@
             },
           },
         },
+        fast_direction: "get",
         slow: {
           ref_store: {
             name: "GRPC_LOCAL_STORE",
--- a/kubernetes/components/worker/worker.json5
+++ b/kubernetes/components/worker/worker.json5
@@ -40,6 +40,7 @@
             },
           },
         },
+        fast_direction: "get",
         slow: {
           ref_store: {
             name: "GRPC_LOCAL_STORE",
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -639,6 +639,26 @@ pub struct OntapS3ExistenceCacheSpec {
     pub backend: Box<ExperimentalOntapS3Spec>,
 }
 
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Copy, PartialEq)]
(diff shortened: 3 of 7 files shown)


MarcusSorealheis (approved):
I think the only thing we are missing is a test. We could spawn co-located setup, simulate SIGTERM mid-action, verify no hangs/reconnects. But that's so rare, I don't think it's necessary.

One nit would be to log rejections, maybe like: `warn!("Rejected worker add during shutdown: {}", worker_id);`

I think this is simple enough. It will only work on Unix. We don't advertise Windows support yet anyway, and I think it is ok because this is such an esoteric setup. Thank you for the contribution!

PR #2019, opened 2025-11-03
Fix fast store direction

chrisstaite:
# Description

The merge of the fast store directions with the populate stream didn't quite go right and meant that multiple requests to a read only fast store would end up iterating on get_part.  This will cause many requests to fast_store.has and be incredibly inefficient.  Although this doesn't fail, a large number of requests for the same item could cause a stack overflow and it will certainly slow performance significantly.

Create a test for this case, however it already passed because the logic wasn't actually flawed, just very inefficient.  Fix up the implementation for this case.

## Type of change

## How Has This Been Tested?

Tests still passing.  It's mostly an optimisation though.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2019)

### DIFF
--- a/.github/workflows/native-cargo.yaml
+++ b/.github/workflows/native-cargo.yaml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Rust cache
         # https://github.com/Swatinem/rust-cache/releases/tag/v2.8.1
-        uses: Swatinem/rust-cache@94162284cf9c7d6640f58b132f82f114d78d8ab0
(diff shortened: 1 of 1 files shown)


MarcusSorealheis:
> What actually happened here is the original PR lingered for months and then making it work with the latest code was non-trivial. I think a PR cull would be a good idea. Many of these are ancient and that causes issues as people forget what they were doing after all that time.

Totally fair! Do you still think this feature is useful for you? I think it's useful.

PR #2021, opened 2025-11-04
Directory Cache

### LINKED ISSUE #2020: can we reuse CAS directory in filesystem_store across working directory
Thanks for [the opt on evicting_map](https://github.com/TraceMachina/nativelink/pull/1875/files) .
We used v0.7.5 for remote-execution, found that fetch time is improved but still cost a lot comparing to execution time even when filesystem store is mostly hit, especially in cases which needs many inputs like linking and testing (We do Python remote test on pods with GPU).

Image

We used Buildfarm years ago, it will create a directory with same prefix of its digest, and will be hardlinked into working directory if needed. To make sure the directory is not changed, it will set directory as readonly, which can make some tests trying to write on local directory fail but may reduce the cost of creating directory in each job.

I'll try to collect metrics in fetch stage later, produce data of (time of reconstructing directory) / ( time of fetch across threads)

MarcusSorealheis:
# Description

Adds a high-performance directory cache using hardlinks to avoid repeated directory reconstruction from CAS. This provides up to a ~100x performance improvement for repeated builds with the same input directories.

Fixes #2020 

## Type of change

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2021)

### DIFF
--- a/.github/workflows/native-cargo.yaml
+++ b/.github/workflows/native-cargo.yaml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Rust cache
         # https://github.com/Swatinem/rust-cache/releases/tag/v2.8.1
-        uses: Swatinem/rust-cache@94162284cf9c7d6640f58b132f82f114d78d8ab0
+        uses: Swatinem/rust-cache@a84bfdc502f07db5a85dd9d7a30f91a931516cc5
 
       - name: Build on ${{ runner.os }}
         run: cargo build --all --profile=smol
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,5 @@ nixos.bazelrc
 rust-project.json
 darwin.bazelrc
 nativelink.bazelrc
-integration_tests/**/*.log
-toolchain-examples/*.log
+*.log
 buck-out/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2548,6 +2548,7 @@ dependencies = [
  "serde_json5",
  "tokio",
  "tonic 0.13.1",
+ "uuid",
  "walkdir",
 ]
 
--- a/nativelink-error/BUILD.bazel
+++ b/nativelink-error/BUILD.bazel
@@ -22,6 +22,7 @@ rust_library(
         "@crates//:serde_json5",
         "@crates//:tokio",
         "@crates//:tonic",
+        "@crates//:uuid",
         "@crates//:walkdir",
     ],
 )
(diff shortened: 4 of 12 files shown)


deployment-examples/docker-compose/Dockerfile
```diff
@@ -59,7 +59,7 @@ COPY --from=builder /root/nativelink-bin /usr/local/bin/nativelink
 ARG ADDITIONAL_SETUP_WORKER_CMD
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends curl=8.5.0-2ubuntu10.6 \
```
palfrey, line 62: https://github.com/TraceMachina/nativelink/actions/runs/19063163878/job/54447370861 looks like a momentary network failure, not a failure to install a particular curl version, so these should be reverted

nativelink-config/src/cas_server.rs
```diff
@@ -797,6 +797,40 @@ pub struct LocalWorkerConfig {
...
+#[serde(deny_unknown_fields)]
+pub struct DirectoryCacheConfig {
+    /// Maximum number of cached directories.
+    /// Default: 1000
+    #[serde(default = "default_directory_cache_max_entries")]
+    pub max_entries: usize,
+
+    /// Maximum total size in bytes for all cached directories (0 = unlimited).
+    /// Default: 10737418240 (10 GB)
+    #[serde(default = "default_directory_cache_max_size_bytes")]
```
palfrey:
```suggestion
    #[serde(default = "default_directory_cache_max_size_bytes", deserialize_with = "convert_data_size_with_shellexpand")]
```
in line with other data size fields

nativelink-util/BUILD.bazel
```diff
@@ -52,6 +53,7 @@ rust_library(
         "@crates//:bitflags",
         "@crates//:blake3",
         "@crates//:bytes",
+        "@crates//:filetime",
```
palfrey: What needs this?

nativelink-worker/src/directory_cache.rs
```diff
@@ -0,0 +1,535 @@
...
+        // Fast path: check if already in cache
+        {
+            let mut cache = self.cache.write().await;
+            if let Some(metadata) = cache.get_mut(&digest) {
+                // Update access time and ref count
+                metadata.last_access = SystemTime::now();
+                metadata.ref_count += 1;
+
+                event!(
+                    Level::DEBUG,
```
palfrey:
```suggestion
                debug!(
```
nit: Use the named helpers, rather than `Level`

PR #2024, opened 2025-11-04
Eviction breaking side case fix for filesystem store

palfrey:
# Description

There's a case where if you've got a filesystem store that's small enough, you can have a case where your entry gets evicted _during_ insertion, and right now this causes a deadlock. To some extent, this is mostly a bad config (i.e. you should have a bigger store), but I suspect the same error can occur with much larger stores, depending on the exact eviction priorities on items. More specifically, we shouldn't be able to cause deadlock no matter how bad the config is, so this needs fixing either way.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2024)

### DIFF
--- a/.github/workflows/custom-image.yaml
+++ b/.github/workflows/custom-image.yaml
@@ -37,7 +37,7 @@ jobs:
     steps:
       - name: Check trigger
         id: check
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             if (context.eventName === 'workflow_dispatch') {
@@ -99,7 +99,7 @@ jobs:
     timeout-minutes: 45
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           ref: ${{ needs.check-trigger.outputs.pr_sha }}
 
@@ -130,7 +130,7 @@ jobs:
 
       - name: Comment on PR
         if: github.event_name == 'issue_comment'
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             await github.rest.issues.createComment({


nativelink-store/tests/filesystem_store_test.rs
```diff
@@ -586,47 +596,6 @@ async fn file_gets_cleans_up_on_cache_eviction() -> Result<(), Error> {
     check_temp_empty(&temp_path).await
 }
 
-// Test to ensure that if we are holding a reference to `FileEntry` and the contents are
-// replaced, the `FileEntry` continues to use the old data.
-// `FileEntry` file contents should be immutable for the lifetime of the object.
-#[nativelink_test]
-async fn digest_contents_replaced_continues_using_old_data() -> Result<(), Error> {
```
palfrey, line 593: I don't think this test should actually pass and hence I've deleted it. The primary reason for this is that it creates a file with `HASH1` digest, and inserts `VALUE1`, which is fine. It then updates said digest with `VALUE2`, and somehow expects the file to not be overwritten with `VALUE2`.

AFAIK there are two problems with this (certainly to my mental model of the stores):
1) In general, if you write some value to a digest, this is what the digest should store. There are cases where you do a partial write to a digest value and then append when you have more of the data, but I don't think there are any real-world cases where a digest should ever get a brand new value.
2) The filesystem store in particular stores in files named via their digest and length. If you store a new value with the same digest and length, it'll be stored in the same place, and anything other than that should not be relied on. I _suspect_ this test might have worked before because you had cases where the content path contained the new value and the temp path contained the old value or something like that, but that would be a very specific implementation property and not something we should be relying on or testing.

I think my changes here change the underlying order of actions in some way, but not IMHO the external interface of those i.e. the filesystem store should still behave in the ways our users expect.

nativelink-store/tests/filesystem_store_test.rs
```diff
@@ -439,7 +449,7 @@ async fn file_continues_to_stream_on_content_replace_test() -> Result<(), Error>
     }
 
     // Replace content.
-    store.update_oneshot(digest1, VALUE2.into()).await?;
+    store.update_oneshot(digest1, VALUE1.into()).await?;
```
palfrey: Same sort of issue as with `digest_contents_replaced_continues_using_old_data`. I'm not totally sure this test should continue to exist.

nativelink-store/src/filesystem_store.rs
```diff
@@ -362,37 +365,49 @@ impl LenEntry for FileEntryImpl {
     // target file location to the new temp file. `unref()` should only ever be called once.
     #[inline]
     async fn unref(&self) {
```
palfrey, line 367: A lot of this is whitespace changes as it had a double-nested `{`/`}` for some reason.

MarcusSorealheis:
in CI it was a little bit flakey so we need to test live to understand.

PR #2030, opened 2025-11-05
fix root issue and permissions issue

### LINKED ISSUE #2020: can we reuse CAS directory in filesystem_store across working directory

MarcusSorealheis:
# Description

This is a follow to #2021 as it introduced a new functionality but did not handle some edge cases.

Fixes #2020 

## Type of change

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2030)

### DIFF
--- a/nativelink-worker/src/local_worker.rs
+++ b/nativelink-worker/src/local_worker.rs
@@ -272,7 +272,7 @@ impl<'a, T: WorkerApiClientTrait + 'static, U: RunningActionsManager> LocalWorke
                                     })
                                     .and_then(|action| {
                                         debug!(
-                                            operation_id = ?action.get_operation_id(),
+                                            operation_id = %action.get_operation_id(),
                                             "Received request to run action"
                                         );
                                         action
--- a/nativelink-worker/src/running_actions_manager.rs
+++ b/nativelink-worker/src/running_actions_manager.rs
@@ -1050,7 +1050,7 @@ impl RunningActionImpl {
                     }
                 },
                 maybe_exit_status = child_process_guard.wait() => {
-                    // Defuse our guard so it does not try to cleanup and make nessless logs.
+                    // Defuse our guard so it does not try to cleanup and make senseless logs.
                     drop(ScopeGuard::<_, _>::into_inner(child_process_guard));
                     let exit_status = maybe_exit_status.err_tip(|| "Failed to collect exit code of process")?;
                     // TODO(palfrey) We should implement stderr/stdout streaming to client here.
@@ -1077,6 +1077,8 @@ impl RunningActionImpl {
                         }
                         exit_code
                     });
+
+                    info!(?args, "Command complete");
 
                     let maybe_error_override = if let Some(side_channel_file) = maybe_side_channel_file {
                         process_side_channel_file(side_channel_file.clone(), &args, requested_timeout).await
--- a/nativelink-worker/tests/running_actions_manager_test.rs
+++ b/nativelink-worker/tests/running_actions_manager_test.rs
@@ -2893,6 +2893,14 @@ exit 1
         });
         assert_eq!(results?.error.unwrap().code, Code::DeadlineExceeded);
 
+        #[cfg(target_family = "unix")]
+        let command = "[\"sh\", \"-c\", \"sleep infinity\"]";
+        #[cfg(target_family = "windows")]
+        let command = "[\"cmd\", \"/C\", \"ping -n 99999 127.0.0.1\"]";
+
+        assert!(logs_contain(&format!("Executing command args={command}")));
+        assert!(logs_contain(&format!("Command complete args={command}")));
+
         assert!(!logs_contain(
             "Child process was not cleaned up before dropping the call to execute(), killing in background spawn"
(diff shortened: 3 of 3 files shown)


palfrey:
I've been testing with this and with `directory_cache` enabled on current `main` and I'm not able to see the problems being fixed here. I'm not currently able to reproduce any problems at all, with various combination of warm and cold caches, deleting the cache beforehand, etc.

(There's one bug I'm seeing `ERROR: /home/palfrey/src/trace/nativelink/BUILD.bazel:95:8: Testing //:dummy_test failed: lost inputs with digests: 654c4faeee4ceaa9ade121d4e00219e323e924192456537b07e91507e959de69/118` but that's only when I delete the cache, and restart the CAS, and that's only with the first build. It also appears with and without `directory_cache`)

nativelink-worker/src/directory_cache.rs
```diff
@@ -74,32 +76,61 @@ struct CachedDirectoryMetadata {
...
+    /// * `cas_store` - `FastSlowStore` for fetching data (must have `FilesystemStore` as fast store)
+    /// * `filesystem_store` - `FilesystemStore` for hardlinking (must be the fast store of `cas_store`)
+    pub async fn new(
+        config: DirectoryCacheConfig,
+        cas_store: Arc<FastSlowStore>,
+        filesystem_store: Arc<FilesystemStore>,
+    ) -> Result<Self, Error> {
         // Ensure cache root exists
-        fs::create_dir_all(&config.cache_root)
+        tokio_fs::create_dir_all(&config.cache_root)
```
palfrey: Why not standard `fs`?

nativelink-util/src/fs_util.rs
```diff
@@ -146,8 +142,12 @@ fn hardlink_directory_tree_recursive<'a>(
...
 ///
 /// # Platform Notes
-/// - Unix: Sets permissions to 0o555 (r-xr-xr-x)
+/// - Unix: Sets permissions to 0o555 (r-xr-xr-x) for directories, 0o444 for files
 /// - Windows: Sets `FILE_ATTRIBUTE_READONLY`
+///
+/// # Important
+/// Note that when using hardlinks, all hardlinked copies share the same inode
+/// and thus share permissions. Setting permissions on one affects all hardlinked copies.
 pub async fn set_readonly_recursive(dir: &Path) -> Result<(), Error> {
```
palfrey: This only appears to be used in tests now

nativelink-util/tests/fs_util_test.rs
```diff
@@ -0,0 +1,177 @@
+// Copyright 2024 The NativeLink Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
palfrey: Should be the FSL license (and I've just realised `fs_util.rs` has the same problem)

nativelink-worker/tests/directory_cache_test.rs
```diff
@@ -0,0 +1,154 @@
+// Copyright 2024 The NativeLink Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
palfrey: Again FSL license

nativelink-worker/src/directory_cache.rs
MarcusSorealheis: We're in an `async fn new()`, but actually creating the directory doesn't need to be async here because it's only done once during initialization, it's not I/O intensive (just one directory creation), and using `sync std::fs` would definitely be simpler to avoid the `.await`.

nativelink-util/tests/fs_util_test.rs
MarcusSorealheis: I'll fix this

nativelink-worker/tests/directory_cache_test.rs
MarcusSorealheis: and this

MarcusSorealheis:
> This seems strange. I think this breaks the RBE standard.

@chrisstaite-menlo Here is what I heard from another user that may want to stay anonymous but I hope not:

> Populating the entire work tree in one shot fails because the root already exists."

What I did was an attempt to get by that but I need to think of a different solution. I'm going to re-evaluate if this is truly an issue with the advent of some new flags.

PR #2031, opened 2025-11-06
Add testing for running action manager failure logging

palfrey:
# Description

We've been seeing various intermittent instances of `Child process was not cleaned up before dropping the call to execute(), killing in background spawn` from `running_actions_manager` with little in the way of logging. I haven't yet been able to reproduce this in tests, but this PR adds some new test checks around timeout logging in the same area as well as various other bits of extra logging around the failure cases to try and aid future investigations

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2031)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -462,7 +462,7 @@ NativeLink Code of Conduct is available in the
 You can generate branch-based coverage reports via:
 
 ```
-nix run .#nativelinkCoverageForHost
+nix build .#nativelinkCoverageForHost
 ```
 
 The `result` symlink contains a webpage with the visualized report.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -462,7 +462,7 @@ NativeLink Code of Conduct is available in the
 You can generate branch-based coverage reports via:
 
 ```
-nix run .#nativelinkCoverageForHost
+nix build .#nativelinkCoverageForHost
```
palfrey, line 465: We don't run it, we build coverage files. This is now the command we run in `.github/workflows/coverage.yaml`

amankrx (approved):

PR #2052, opened 2025-11-16
fix(worker): Resolve deadlock due to file permit exhaustion (#2051)

### LINKED ISSUE #2051: Deadlock in NativeLink Worker due to File Permit Exhaustion
## 1. Summary

A deadlock was identified in the NativeLink worker when processing concurrent remote execution jobs that involve a large number of input files. The application would become unresponsive, and `tokio-console` revealed that a significant number of tasks were permanently blocked waiting for file I/O permits. This happens when I cross build Chromium with `siso`.

The root cause was a classic deadlock scenario involving a global `Semaphore` used to limit concurrent file operations and other locks within the application. Specifically, the `hard_link` and `symlink` file operations were holding locks on other resources while waiting to acquire a file permit, leading to a circular dependency and a complete system halt.

## 2. Background

The `nativelink-util/src/fs.rs` module implements a global semaphore, `OPEN_FILE_SEMAPHORE`, to limit the number of concurrently open file descriptors. This is a standard pattern to prevent the application from exceeding the operating system's `ulimit` for open files.

All file system operations in `nativelink-util/src/fs.rs` are designed to first acquire a permit from this semaphore before proceeding.

## 3. Root Cause Analysis

The deadlock was diagnosed using `tokio-console`, which showed a large number of tasks stuck in `nativelink_util::fs::call_with_permit`. This function is the entry point for all file operations that are limited by the semaphore.

The specific deadlock scenario was as follows:

1.  **Semaphore Exhaustion:** A high volume of concurrent actions caused all available permits from `OPEN_FILE_SEMAPHORE` to be acquired by "Worker Tasks" performing file I/O (initially `hard_link`, later identified in `symlink` as well).

2.  **Circular Lock Dependency:**
    *   The **Worker Tasks**, while holding the file permits, needed to acquire a read lock on a `FileEntry` within the `FilesystemStore` to get the source path for a `hard_link`.
    *   Simultaneously, a "Manager Task" (likely the cache eviction thread) was holding a write lock on that same `FileEntry` because it was trying to evict the file.
    *   The **Manager Task** then needed to perform its own file operation (e.g., deleting the file), so it tried to acquire a permit from the already exhausted `OPEN_FILE_SEMAPHORE`.

3.  **Deadlock:**
    *   The Manager Task was blocked indefinitely, waiting for a file permit.
    *   The Worker Tasks were blocked indefinitely, waiting for the `FileEntry` lock held by the Manager Task.
    *   No progress could be made, and the application hung.

A similar issue was identified in the `symlink` function, where a misuse of `spawn_blocking` and `block_on` created a deadlock between Tokio's main runtime and its blocking thread pool when the file permit semaphore was exhausted.

rogerwang:
# Description

A deadlock occurs during concurrent actions with many input files. This is caused by tasks holding locks on `FileEntry` while waiting for file permits from the global `OPEN_FILE_SEMAPHORE`, leading to a circular dependency and application hang.

This was observed in two places:
1. `hard_link` operations holding a `FileEntry` read lock.
2. `symlink` operations misusing `spawn_blocking` and `block_on`.

This commit resolves the deadlock by:
- Modifying `download_to_directory` to release the `FileEntry` lock before calling `hard_link`.
- Refactoring `fs::symlink` to be fully asynchronous, removing the `spawn_blocking`/`block_on` anti-pattern.

Fixes #2051 

## Type of change

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2052)

### DIFF
--- a/nativelink-util/src/fs.rs
+++ b/nativelink-util/src/fs.rs
@@ -273,6 +273,7 @@ pub async fn create_dir_all(path: impl AsRef<Path>) -> Result<(), Error> {
 
 #[cfg(target_family = "unix")]
 pub async fn symlink(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> Result<(), Error> {
+    // TODO: add a test for #2051: deadlock with large number of files
     let _permit = get_permit().await?;
     tokio::fs::symlink(src, dst).await.map_err(Into::into)
 }
--- a/nativelink-worker/src/running_actions_manager.rs
+++ b/nativelink-worker/src/running_actions_manager.rs
@@ -154,6 +154,7 @@ pub fn download_to_directory<'a>(
                             .get_file_entry_for_digest(&digest)
                             .await
                             .err_tip(|| "During hard link")?;
+                        // TODO: add a test for #2051: deadlock with large number of files
                         let src_path = file_entry.get_file_path_locked(|src| async move { Ok(PathBuf::from(src)) }).await?;
                         fs::hard_link(&src_path, &dest)
                             .await


nativelink-util/src/fs.rs
```diff
@@ -273,14 +273,8 @@ pub async fn create_dir_all(path: impl AsRef<Path>) -> Result<(), Error> {
...
-    let src = src.as_ref().to_owned();
-    let dst = dst.as_ref().to_owned();
-    call_with_permit(move |_| {
-        tokio::runtime::Handle::current()
-            .block_on(tokio::fs::symlink(src, dst))
-            .map_err(Into::<Error>::into)
-    })
-    .await
+    let _permit = get_permit().await?;
+    tokio::fs::symlink(src, dst).await.map_err(Into::into)
```
MarcusSorealheis, line 278: wow. so simple. you are very much appreciated.

MarcusSorealheis: if many symlinks fire concurrently, you might still hit permit waits.

It would be good to know what scale of project you ran into this bug (I'm guessing Chromium), but I assume sufficiently large.

MarcusSorealheis: I also just checked out the tokio docs for [`block_on`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on). This was a clear screw up. sync-in-async. 🤦

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -154,8 +154,8 @@ pub fn download_to_directory<'a>(
                             .get_file_entry_for_digest(&digest)
                             .await
                             .err_tip(|| "During hard link")?;
-                        file_entry
-                            .get_file_path_locked(|src| fs::hard_link(src, &dest))
+                        let src_path = file_entry.get_file_path_locked(|src| async move { Ok(PathBuf::from(src)) }).await?;
+                        fs::hard_link(&src_path, &dest)
```
MarcusSorealheis, line 159: This one is clever. I initially made the synchronous because I think it was effectively a no-op, but that's not true.

MarcusSorealheis:
the test would only be necessary for `download_to_directory`. The other issue is that this bug only occurs at sufficient scale. So we might need to provision some infrastructure for an integration if we cannot get a consistent repro on one or two machines. We have some capacity for that.

nativelink-util/src/fs.rs
rogerwang: > if many symlinks fire concurrently, you might still hit permit waits.
> 
> It would be good to know what scale of project you ran into this bug (I'm guessing Chromium), but I assume sufficiently large.

yes. I mentioned in the bug report that I was cross building Chromium with siso, when it will upload and reconstruct the whole macOS SDK and standard libc++.

rogerwang: I added a file count metrics and it shows that the number of files are above 10k.

MarcusSorealheis: > yes. I mentioned in the bug report that I was cross building Chromium with siso, when it will upload and reconstruct the whole macOS SDK and standard libc++.

Ahh you're right. Unconscious knowledge. Less a guess. Thank you for the report.

MarcusSorealheis (approved):
Thank you for the TODO. We will get to it to detect regressions.

LGTM (if the tests pass and they should)

PR #2067, opened 2025-11-21
Recoverable connection pool

### LINKED ISSUE #2065: Fix Query Syntax Error in `redis_store`
The `nativelink-scheduler` is failing due to a RediSearch keyword conflict with the field name "state". This causes index creation errors and stalls the system.

MarcusSorealheis:
# Description
I have replaced the default fred Redis client pool with a custom `RecoverablePool` to address the system stalling issues caused by poor client recovery. `RecoverablePool` is really only a pool that manages a list of `fred::clients::Client instances.`

This is only meant to be a PoC

Fixes #2065 I believe that this issue is multi-faceted and related to connection storms, redis connection pools, and race conditions.

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2067)

### DIFF
--- a/nativelink-store/tests/redis_store_test.rs
+++ b/nativelink-store/tests/redis_store_test.rs
@@ -170,7 +170,7 @@ impl Drop for MockRedisBackend {
     }
 }
 
-fn make_clients(builder: Builder) -> (RecoverablePool, SubscriberClient) {
+fn make_clients(builder: &Builder) -> (RecoverablePool, SubscriberClient) {
     const CONNECTION_POOL_SIZE: usize = 1;
     let client_pool = RecoverablePool::new(builder.clone(), CONNECTION_POOL_SIZE).unwrap();
 
@@ -189,7 +189,7 @@ fn make_mock_store_with_prefix(mocks: &Arc<MockRedisBackend>, key_prefix: String
         mocks: Some(mocks),
         ..Default::default()
     });
-    let (client_pool, subscriber_client) = make_clients(builder);
+    let (client_pool, subscriber_client) = make_clients(&builder);
     RedisStore::new_from_builder_and_parts(
         client_pool,
         subscriber_client,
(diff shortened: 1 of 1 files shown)


nativelink-store/src/redis_store.rs
```diff
@@ -98,11 +98,87 @@ const DEFAULT_SCAN_COUNT: u32 = 10_000;
...
+
+    fn next(&self) -> Client {
+        let clients = self.clients.read();
+        let index = self
+            .counter
+            .fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+        clients[index % clients.len()].clone()
+    }
+
+    async fn replace_client(&self, old_client: &Client) -> Result<Client, Error> {
```
MarcusSorealheis, line 139: I need did some testing and I need to do some more thinking here. Here is the problem: Every call to `replace_client` eagerly builds and connects a brand-new Client before it acquires the write lock to see if the old handle is still present. When the target slot has already been replaced by another task ( weethe None branch at lines 157‑165), we now return that freshly-built client even though it never enters the pool. 

Any concurrent reconnect storm therefore spawns one extra TCP connection per racing task, only to drop it again after a single request 😭 . On a busy instance a single node flake can suddenly create dozens of unnecessary Redis connections, exhaust FD limits, and hammer the server with repeated AUTH/HELLO. We should only build a new client once we know we actually need to swap the slot (or fall back to handing out an existing pooled client, as the code previously did).

amankrx:
Tested this out, None of the jobs reached reached the workers, and all of them were stuck in scheduler state and eventually got killed due to an OOM error.

MarcusSorealheis:
@amankrx is that still accurate? I am pretty sure everything is passing on all large workloads. Could you confirm how your tests went?

amankrx (approved):

MarcusSorealheis:
Cached also @amankrx

Issue #1999, opened 2025-10-21
Add store optimisation for fast has


chrisstaite-menlo:
Currently the fast-slow store calls has and then calls get.  On an S3 or GCS store that's really inefficient as calling has is pretty much the same expense as calling get and having a 404 returned.

Add a store optimisation flag to avoid this and if it's set then don't call has in the fast-store population path.  Simply handle the Code::NotFound response correctly.

PR #2072, opened 2025-11-26
Add LazyNotFound Store Optimization, Support for fast_slow_store (S3, GCS slow_store targets)

mkeen:
# Description

As requested, this implements an optimization that skips `.has()` checks on streamed syncs from fast to supported slow stores (S3Store or GCSStore for now) in `FastSlowStore`.

Fixes [#1999](https://github.com/TraceMachina/nativelink/issues/1999)

## Type of change

  not work as expected)

## How Has This Been Tested?

Some light testing locally. But feedback is requested. I was thinking of writing a new mock/set of tests in `nativelink-store/tests/fast_slow_store_test.rs`. Let me know if that sounds ok or if there is a better place.

I don't believe a documentation change is needed for this optimization, but correct me if you disagree.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2072)

### DIFF
--- a/nativelink-store/src/fast_slow_store.rs
+++ b/nativelink-store/src/fast_slow_store.rs
@@ -269,14 +269,10 @@ impl FastSlowStore {
                     },
                 )
             }
-            Err(err) => {
-                if let Err(slow_err) = &slow_res {
-                    if slow_err.code == Code::NotFound {
-                        return Err(slow_err.clone());
-                    }
-                }
-                fast_res.merge(slow_res).merge(Err(err))
-            }
+            Err(err) => match slow_res {
+                Err(slow_err) if slow_err.code == Code::NotFound => Err(slow_err),
+                _ => fast_res.merge(slow_res).merge(Err(err)),
+            },
         }
     }
 


nativelink-store/src/fast_slow_store.rs
```diff
@@ -168,26 +168,31 @@ impl FastSlowStore {
...
-                )
-            })?;
-
-        self.metrics
-            .slow_store_hit_count
-            .fetch_add(1, Ordering::Acquire);
+            .inner_store(Some(key.borrow()))
+            .optimized_for(StoreOptimizations::LazyNotFound)
+        {
+            None
```
MarcusSorealheis: do we want debug logging here to understand how frequently we are terminating here in the case of issues?

MarcusSorealheis (changes requested):
would like to see tests

nativelink-store/src/fast_slow_store.rs
```diff
@@ -168,26 +169,36 @@ impl FastSlowStore {
         offset: u64,
         length: Option<u64>,
     ) -> Result<(), Error> {
-        let sz = self
+        let deterministic_sz = if self
```
MarcusSorealheis: I think having `sz` was a smell to begin with. I think it should be `size` so that it is searchable. `deterministic_size` is fine.

nativelink-util/src/store_trait.rs
```diff
@@ -127,6 +127,9 @@ pub enum StoreOptimizations {
 
     /// If the store will never serve downloads.
     NoopDownloads,
+
+    /// If the store will rely on lazily determining file existence for mutations.
+    LazyNotFound,
```
MarcusSorealheis: nit: I wonder about `LazyNotFound` but curious on your thoughts.

MarcusSorealheis (changes requested):
Sorry for the delay here. I'd been squeezing in work the past few days. 

Overall, this looks good. I do have concerns that I have mentioned. I know the original issue may have used `LazyNotFound` but can we be even more explicit in referring to it.

I'm thinking `LazyExistenceOnWrite` or something that explains what exactly we are lazing about. Let me know your thoughts.

nativelink-store/src/fast_slow_store.rs
```diff
@@ -230,9 +249,14 @@ impl FastSlowStore {
...
 
         let slow_store_fut = self.slow_store.get(key.borrow(), slow_tx);
-        let fast_store_fut =
-            self.fast_store
-                .update(key.borrow(), fast_rx, UploadSizeInfo::ExactSize(sz));
+        let fast_store_fut = self.fast_store.update(
+            key.borrow(),
+            fast_rx,
+            match deterministic_sz {
+                Some(sz) => UploadSizeInfo::ExactSize(sz),
```
MarcusSorealheis: if we are using sz everywhere in the codebase, that would be annoying and a massive smell but if we can improve it her, we can improve it everywhere.

nativelink-util/src/store_trait.rs
mkeen: I think that your point is true. Building on your thought, I think the intent is only half-captured via the current name since what we are doing foregoes finding out anything about the file's existence until we attempt to start reading from it in a fast to slow store streamed sync.

I like where you're going with `LazyExistenceOnWrite`, but the actual optimization happens on `reads`, and more pedantically, outgoing streams triggered by a caller. Should we go with `LazyExistenceOnSync`? Or maybe `SpeculativeRead`. I'm good with any of these including yours, or something else if there are other ideas.

MarcusSorealheis: `LazyExistenceOnSync` is great.

mkeen:
@chrisstaite-menlo Great feedback, thanks. Your thought re: `UploadSizeInfo` was particularly great.

I tried the chained if-let and things complained.

```
error[E0658]: `let` expressions in this position are unstable
   --> nativelink-store/src/fast_slow_store.rs:273:20
    |
273 |                 if let Err(slow_err) = &slow_res
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information
```

I also didn't see it being used in the codebase anywhere with a cursory glance. I think it was reverted back to unstable somewhere before 1.87. I tried it on a recent nightly also FWIW -- but if your CI will handle it differently than local builds I will attempt to add it and see if the build succeeds here. I was trying this:

```
Err(err) => {
    if let Err(slow_err) = &slow_res
        && slow_err.code == Code::NotFound
    {
        return Err(slow_err.clone());
    }
    fast_res.merge(slow_res).merge(Err(err))
}
```

Otherwise, if you give me a LGTM I will combine the commits and we should be all set.

MarcusSorealheis (approved):

PR #2076, opened 2025-11-27
Replace Fred with redis-rs

palfrey:
# Description

Replaces [fred](https://github.com/aembke/fred.rs) with [redis-rs](https://github.com/redis-rs/redis-rs), because there's a _lot_ of issues with the former and the latter is more maintained and more widely used (~12x download count on crates.io - ~4M vs. 47M)

## Type of change

  not work as expected)

in theory, this is a drop-in change. In practice, there's likely a number of underlying changes - which is kinda the point, but there's probably some undertested areas that we don't know about issues in yet.

## How Has This Been Tested?

`bazel test //...` and a bunch of time with the store tester

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2076)

### DIFF
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -55,6 +55,7 @@ use tokio::select;
 use tokio::sync::mpsc::{UnboundedReceiver, unbounded_channel};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tokio::time::{sleep, timeout};
+use tokio_stream::wrappers::UnboundedReceiverStream;
 use tracing::{debug, error, info, trace, warn};
 use url::Url;
 use uuid::Uuid;
@@ -1147,13 +1148,21 @@ impl RedisPatternSubscriber for MockPubSub {
 }
 
 impl RedisSubscriptionManager {
-    pub fn new<P>(mut pub_sub: P, pub_sub_channel: String) -> Self
+    pub fn new<P>(
+        mut pub_sub: P,
+        subscriber_channel: Option<UnboundedReceiver<PushInfo>>,
+        pub_sub_channel: String,
+    ) -> Self
     where
         P: RedisPatternSubscriber,
     {
         let subscribed_keys = Arc::new(RwLock::new(StringPatriciaMap::new()));
         let subscribed_keys_weak = Arc::downgrade(&subscribed_keys);
         let (tx_for_test, mut rx_for_test) = unbounded_channel();
+        let mut local_subscriber_channel: Pin<Box<dyn Stream<Item = PushInfo> + Send>> =
+            subscriber_channel
+                .and_then(|channel| Some(UnboundedReceiverStream::new(channel).boxed()))
+                .unwrap_or_else(|| stream::empty::<PushInfo>().boxed());
         Self {
             subscribed_keys,
             tx_for_test,
@@ -1200,6 +1209,29 @@ impl RedisSubscriptionManager {
                                         break;
                                     }
                                 },
+                                maybe_push_info = local_subscriber_channel.next() => {
+                                    if let Some(push_info) = maybe_push_info {
+                                        if push_info.data.len() != 1 {
+                                            error!(?push_info, "Expected exactly one message on subscriber_channel");
+                                            continue;
+                                        }
+                                        match push_info.data.first().unwrap() {
+                                            Value::SimpleString(s) => {
+                                                s.clone()
+                                            }
+                                            Value::BulkString(v) => {
+                                                String::from_utf8(v.to_vec()).expect("String message")
+                                            }
+                                            other => {
+                                                error!(?other, "Received non-string message in RedisSubscriptionManager");
+                                                continue;
+                                            }
+                                        }
+                                    } else {
+                                        error!("Error receiving message in RedisSubscriptionManager from subscriber_channel");
+                                        break;
+                                    }
+                                }
                             };
                             let Some(subscribed_keys) = subscribed_keys_weak.upgrade() else {
                                 warn!(
@@ -1302,6 +1334,7 @@ impl<C: Clone + ConnectionLike + Sync + Send + 'static, P: RedisPatternSubscribe
             };
             let sub = Arc::new(RedisSubscriptionManager::new(
                 pub_sub,
+                self.subscriber_channel.lock().take(),
                 pub_sub_channel.clone(),
             ));
             *subscription_manager = Some(sub.clone());


nativelink-util/Cargo.toml
```diff
@@ -17,7 +17,9 @@ base64 = { version = "0.22.1", default-features = false, features = ["std"] }
 bitflags = { version = "2.9.0", default-features = false }
 blake3 = { version = "1.8.0", features = ["mmap"], default-features = false }
 bytes = { version = "1.10.1", default-features = false }
-futures = { version = "0.3.31", default-features = false }
+futures = { version = "0.3.31", features = [
+  "async-await",
+], default-features = false }
```
palfrey, line 22: I'm guessing fred pulled this in before, but need to spell it out explicitly here now

MarcusSorealheis: hmmm. That's surprising.

MarcusSorealheis:
@amankrx I'm good with this after you have tested with some of the larger workloads. Once you've done so, please let us know.

nativelink-config/src/stores.rs
```diff
@@ -1150,6 +1150,7 @@ pub struct RedisSpec {
     ///
     /// Default: 4096
     #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
+    // DEPRECATED
```
palfrey: I need to clarify this (or probably just do some sort of runtime "this is deprecated, stop using it" thing)

nativelink-store/src/redis_store.rs
```diff
@@ -1318,157 +1305,103 @@ impl SchedulerStore for RedisStore {
...
-                    .as_i64()
-                    .err_tip(|| {
-                        formatcp!("'{VERSION_FIELD_NAME}' is not u64 in RedisStore::search_by_index_prefix::as_u64")
-                    })?
-            } else {
-                0
-            };
-            K::decode(version, bytes_data)
-                .err_tip(|| "In RedisStore::search_by_index_prefix::decode")
+            panic!("redis_map: {redis_map:?}");
```
palfrey: This needs fixing before merge. It's currently entirely untested, as noted by the CI passing despite this!

.github/workflows/native-bazel.yaml
```diff
@@ -63,42 +63,45 @@ jobs:
...
-  #         disk-cache: ${{ github.workflow }}-ubuntu-24.04
+      - name: Run Store tester with sentinel
+        run: |
+          bazel run //:redis_store_tester \
+            --extra_toolchains=@rust_toolchains//:all \
+            --verbose_failures -- --redis-mode sentinel --mode sequential
+        env:
+          RUST_LOG: trace
+          REDIS_HOST: localhost
+          MAX_LOOPS: 10 # running sequentially just to test all the actions work
```
palfrey, line 95: This works again primarily because MAX_LOOPS is set low enough to test all the functions, but not so high it hits concurrency issues

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -741,7 +741,15 @@ where
...
-                UpdateOperationType::UpdateWithActionStage(stage) => stage.clone(),
+                UpdateOperationType::UpdateWithActionStage(stage) => {
+                    if stage == &ActionStage::Executing
+                        && awaited_action.state().stage == ActionStage::Executing
+                    {
+                        warn!(state = ?awaited_action.state(), "Action already assigned");
+                        return Err(make_err!(Code::Aborted, "Action already assigned"));
+                    }
+                    stage.clone()
+                }
```
palfrey, line 752: I think this was a problem in general, but I've seen it when running `test_multiple_clients_subscribe_to_same_action` now which this fixes. There was a case with two clients subscribing to the same action fast enough to both try and execute on a worker, and this change makes which ever one goes second bug out because it's already assigned.

nativelink-store/src/mongo_store.rs
```diff
@@ -800,13 +800,6 @@ impl ExperimentalMongoSubscriptionManager {
 impl SchedulerSubscriptionManager for ExperimentalMongoSubscriptionManager {
     type Subscription = ExperimentalMongoSubscription;
 
-    fn notify_for_test(&self, value: String) {
-        let subscribed_keys_mux = self.subscribed_keys.read();
-        subscribed_keys_mux
-            .common_prefix_values(&value)
-            .for_each(ExperimentalMongoSubscriptionPublisher::notify);
-    }
-
```
palfrey, line 809: This was only actually used by the Redis test BTW

.github/workflows/native-bazel.yaml
MarcusSorealheis: makes sense

nativelink-scheduler/src/simple_scheduler_state_manager.rs
MarcusSorealheis: Fantastic

nativelink-store/src/mongo_store.rs
MarcusSorealheis: correct. The goal was going to be to expand.

nativelink-store/src/redis_store.rs
```diff
@@ -486,10 +275,231 @@ impl RedisStore {
...
+                Code::InvalidArgument,
+                "new_cluster only works for Cluster mode".to_string(),
+            ));
+        }
+        Self::set_spec_defaults(&mut spec)?;
+
+        let full_urls: Vec<_> = spec
+            .addresses
+            .iter_mut()
+            .map(|addr| format!("{addr}?protocol=resp3"))
```
MarcusSorealheis: Are you appending `?protocol=resp3` directly to each address? Is there any reason that we should use  proper URL parsing and merging instead? Working in this fashion directly risks breaking addresses that already contain queries, potentially losing auth or other params. If it's a non-issue we can ignore.

nativelink-store/src/redis_utils/ft_cursor_read.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+    let Value::Int(new_cursor_id) = value.next().unwrap() else {
+        return Err(RedisError::from((
+            ErrorKind::Parse,
+            "Expected cursor id as second element",
+        )));
+    };
+
+    Ok(RedisCursorData {
+        total: -1, // FIXME(palfrey): fill in value
```
MarcusSorealheis: If any upstream logic relies on total for metrics or pagination expectations, this becomes wrong or misleading. Even if unused today, leaving a sentinel value may mask future regressions.  Maybe we can fix this later.

nativelink-store/src/redis_store.rs
```diff
@@ -486,10 +275,231 @@ impl RedisStore {
...
+        }
+        if spec.retry.max_retries == 0 {
+            spec.retry.max_retries = 1;
+        }
+        trace!(?spec, "redis spec is after setting defaults");
+        Ok(())
+    }
+}
+
+impl RedisStore<ClusterConnection, PubSub> {
```
MarcusSorealheis, line 341: Looks like a cluster pubsub channel created but not consumed. `new_cluster` builds `ClusterClient` with `push_sender(tx)` and stores `subscriber_channel` in the struct, but  no code ever reads from subscriber_channel. That measn push messages will accumulate in an unbounded channel, so memory can grow without bound if Redis sends push events (and in practice it also means cluster-mode pubsub notifications are effectively unimplemented, despite comments implying it’s used).

Either:

1. wire subscriber_channel into `RedisSubscriptionManager` for cluster mode (consume `PushInfo` and map to pubsub events), or
2. only set `push_sender` if pubsub is actually enabled and a consumer is installed.

nativelink-store/src/redis_store.rs
palfrey: Fixed

nativelink-store/src/redis_utils/ft_cursor_read.rs
palfrey: Fixed

nativelink-store/src/redis_store.rs
palfrey: Cluster mode wasn't properly tested. I've fixed `redis_store_tester` to support cluster mode properly and added some bits to the docker-compose for that so it's got a working cluster. Which now breaks because of RESP3 messages, so fixing that...

palfrey: RESP3 support fixed, `redis_store_tester` with cluster mode testing added to CI. I've also added `subscriber_channel` support

PR #1648, opened 2025-03-31
Fix calculation of multipart S3 upload chunk size

byteOfChaos:
# Description

Fix an incorrect calculation of the chunk size for a multipart file upload to S3. Original code seemed to be targeting calculation of smallest chunk size for successful S3 file upload.

Added a few range checks - which may be out of place; I'll happily receive correction - at the top of the `update()` method and after calculation of chunk size and chunk count.

Fixes #1425

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Included a range check after calculation, requiring visual verification. For S3 store, there seems to only be integration tests that do not cover multipart upload.

I'd be happy to explore more thorough testing after receiving some guidance.

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1648)

PR #2091, opened 2025-12-11
Fix calculation of multipart S3 upload chunk size (continues #1648)

### LINKED ISSUE #1425: S3 Store's bytes per upload part calculation logic doesn't make sense
This is the code of S3 Store's bytes per upload part calculation.

https://github.com/TraceMachina/nativelink/blob/712608ccd91a088545b9e93b7faf1f48355c7c18/nativelink-store/src/s3_store.rs#L530-L533

We've declared constant variables as this.

https://github.com/TraceMachina/nativelink/blob/712608ccd91a088545b9e93b7faf1f48355c7c18/nativelink-store/src/s3_store.rs#L67-L77

It makes sense to use `clamp` to make the value to keep in range.

But it doesn't make sense to divide the `max_size` with `MIN_MULTIPART_SIZE`.

If that part was to keep the parts count to be less than `MAX_UPLOAD_PARTS`, shouldn't the code be like this?

```
 let bytes_per_upload_part = 
     (max_size / (MAX_UPLOAD_PARTS as u64 - 1)).clamp(MIN_MULTIPART_SIZE, MAX_MULTIPART_SIZE); 
```

Hope you to check this issue's validity. I will raise a PR if my solution is correct. Thank you.

MarcusSorealheis:
# Description

- Add check for maximum file size
- Calculate minimum chunk size (maximum chunk count) for multipart S3 file upload, checking for proper limit of chunk count before upload
- Make MAX_UPLOAD_PARTS u64 for uniformity and simplicity

Fixes: https://github.com/TraceMachina/nativelink/issues/1425
Continues: https://github.com/TraceMachina/nativelink/pull/1648

## Type of change

## How Has This Been Tested?

Testing it now.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2091)

### DIFF
--- a/nativelink-store/src/s3_store.rs
+++ b/nativelink-store/src/s3_store.rs
@@ -57,6 +57,10 @@ use tracing::{error, info};
 use crate::cas_utils::is_zero_digest;
 use crate::common_s3_utils::{BodyWrapper, TlsClient};
 
+// S3 object cannot be larger than this number. See:
+// https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+const MAX_UPLOAD_SIZE: u64 = 5 * 1024 * 1024 * 1024 * 1024; // 5TB.
+
 // S3 parts cannot be smaller than this number. See:
 // https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
 const MIN_MULTIPART_SIZE: u64 = 5 * 1024 * 1024; // 5MB.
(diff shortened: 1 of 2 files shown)


nativelink-store/src/s3_store.rs
```diff
@@ -57,6 +57,10 @@ use tracing::{error, info};
 use crate::cas_utils::is_zero_digest;
 use crate::common_s3_utils::{BodyWrapper, TlsClient};
 
+// S3 object cannot be larger than this number. See:
+// https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+const MAX_UPLOAD_SIZE: u64 = 5 * 1024 * 1024 * 1024 * 1024; // 5TB.
```
palfrey: That page details the max part size as 48.8 TiB

palfrey (changes requested):
One minor issue. Although, TBH if anyone's trying to upload 5TB objects in one go, problems will probably get hit anyways, let alone the 48.8TB ones!

palfrey (approved):

PR #2094, opened 2025-12-12
Replace rustls-pemfile to fix RUSTSEC-2025-0134

palfrey:
# Description

See https://rustsec.org/advisories/RUSTSEC-2025-0134.html

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2094)

### DIFF
--- a/nativelink-store/src/ontap_s3_store.rs
+++ b/nativelink-store/src/ontap_s3_store.rs
@@ -50,7 +50,8 @@ use nativelink_util::retry::{Retrier, RetryResult};
 use nativelink_util::store_trait::{RemoveItemCallback, StoreDriver, StoreKey, UploadSizeInfo};
 use parking_lot::Mutex;
 use rustls::{ClientConfig, RootCertStore};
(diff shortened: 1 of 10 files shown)


nativelink-store/src/ontap_s3_store.rs
```diff
@@ -104,7 +105,7 @@ pub fn load_custom_certs(cert_path: &str) -> Result<Arc<ClientConfig>, Error> {
     );
 
     // Parse certificates
-    let certs = extract_certs(&mut cert_reader)
+    let certs = CertificateDer::pem_reader_iter(&mut cert_reader)
```
MarcusSorealheis: is the error mapping still needed because of
`impl From<rustls_pki_types::pem::Error> for Error`

palfrey: Fixed

PR #2103, opened 2026-01-04
Add docs for configuring Worker Match Logging Interval

amankrx:
# Description
Added the docs to configure the `worker_match_logging_interval_s`. This should help with reducing the larger redis queries for production environments. 

Fixes # (issue)

## Type of change

- [X] This change is a documentation update

## How Has This Been Tested?

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2103)

### DIFF
--- a/web/platform/src/content/docs/docs/config/production-config.mdx
+++ b/web/platform/src/content/docs/docs/config/production-config.mdx
@@ -506,4 +506,4 @@ Setting it to `-1` turns off the logging entirely.
 ### What You Lose
 
 You won't see those worker matching logs anymore. If you need to debug why actions aren't getting assigned to workers, you'll have to turn this back on temporarily or look at other metrics.
-For most production setups at scale, that's a fair trade for better Redis performance.
+For most production setups at scale, that's a fair trade-off for better Redis performance.


web/platform/src/content/docs/docs/config/production-config.mdx
```diff
@@ -453,3 +453,57 @@ Here is the final CAS Config JSON without the 99 extra shards for writing to S3.
...
+
+**Maybe not, if you're:**
+- Running a smaller setup
+- Actively debugging worker assignment issues
+- Still tuning your deployment
+
+### What You Lose
+
+You won't see those worker matching logs anymore. If you need to debug why actions aren't getting assigned to workers, you'll have to turn this back on temporarily or look at other metrics.
+For most production setups at scale, that's a fair trade for better Redis performance.
```
MarcusSorealheis: I would you use the word trade off. 

Also, I wish to disable the query entirely because it's expensive.

web/platform/src/content/docs/docs/config/production-config.mdx
```diff
@@ -453,3 +453,57 @@ Here is the final CAS Config JSON without the 99 extra shards for writing to S3.
   ]
 }
 ```
+
+
+## Speed Up NativeLink by Turning Off a Hidden Redis Query
```
MarcusSorealheis, line 458: Are we turning of the query or turning off logging of the query?  Is the log the source of the query?

MarcusSorealheis (changes requested):
a few questions/suggestions

web/platform/src/content/docs/docs/config/production-config.mdx
amankrx: This logging is the only source of the query, so we can either remove this completely, or just turn off the logging, either way it is fine. I think, disabling the query would be ideal, or we can disable the use of the query by default, and the user can turn this on if they want for development builds.

web/platform/src/content/docs/docs/config/production-config.mdx
MarcusSorealheis: @amankrx `trade off` instead of `trade`

PR #2121, opened 2026-01-26
Be clearer about what property values workers are missing

palfrey:
# Description

https://github.com/TraceMachina/nativelink/pull/2118 added logging around missing worker properties, but you still have cases where you don't know what property value should actually be set on the workers! This PR improves that.

## Type of change

## How Has This Been Tested?

`bazel test //...` mostly.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2121)

### DIFF
--- a/nativelink-scheduler/src/worker_capability_index.rs
+++ b/nativelink-scheduler/src/worker_capability_index.rs
@@ -134,16 +134,16 @@ impl WorkerCapabilityIndex {
         action_properties: &PlatformProperties,
         full_worker_logging: bool,
     ) -> HashSet<WorkerId> {
-        if action_properties.properties.is_empty() {
-            // No properties required, all workers match
-            return self.all_workers.clone();
-        }
-
         if self.all_workers.is_empty() {
             if full_worker_logging {
                 info!("No workers available to match!");
             }
             return HashSet::new();
+        }
+
+        if action_properties.properties.is_empty() {
+            // No properties required, all workers match
+            return self.all_workers.clone();
         }
 
         let mut candidates: Option<HashSet<WorkerId>> = None;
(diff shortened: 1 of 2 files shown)


nativelink-scheduler/src/worker_capability_index.rs
```diff
@@ -132,18 +132,18 @@ impl WorkerCapabilityIndex {
...
                 info!("No workers available to match!");
             }
             return HashSet::new();
         }
 
+        if action_properties.properties.is_empty() {
+            // No properties required, all workers match
+            return self.all_workers.clone();
+        }
+
```
palfrey, line 148: These two got swapped over. In the "you have no workers case" the return value won't change, but it'll be a bit clearer as to why nothing is getting scheduled.

nativelink-scheduler/src/worker_capability_index.rs
```diff
@@ -166,7 +166,7 @@ impl WorkerCapabilityIndex {
                     if internal_candidates.is_empty() {
                         if full_worker_logging {
                             info!(
-                                "No candidate workers due to a lack of matching {name} = {value:?}"
+                                "No candidate workers due to a lack of matching '{name}' = {value:?}"
```
palfrey: Otherwise you got things like `due to a lack of key os` and this more clearly delineates between the key and the message.

nativelink-scheduler/src/worker_capability_index.rs
```diff
@@ -194,7 +194,9 @@ impl WorkerCapabilityIndex {
 
                     if internal_candidates.is_empty() {
                         if full_worker_logging {
-                            info!("No candidate workers due to a lack of key {name}");
+                            info!(
+                                "No candidate workers due to a lack of key '{name}'. Job asked for {value:?}"
+                            );
```
palfrey, line 207: This is the main addition, as errors like `No candidate workers due to a lack of key InputRootAbsolutePath` are hard to fix without knowing what value the jobs are looking for.

amankrx (approved):

PR #2152, opened 2026-02-13
Add GRPC timeouts and other improvements to detect dead connections

amankrx:
# Description

Add RPC timeouts and other additions to detect dead connections in workers. 

Fixes # (issue)

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2152)

### DIFF
--- a/nativelink-store/src/grpc_store.rs
+++ b/nativelink-store/src/grpc_store.rs
@@ -53,7 +53,7 @@ use parking_lot::Mutex;
 use prost::Message;
 use tokio::time::sleep;
 use tonic::{Code, IntoRequest, Request, Response, Status, Streaming};
-use tracing::error;
+use tracing::{error, trace, warn};
 use uuid::Uuid;
 
 // This store is usually a pass-through store, but can also be used as a CAS store. Using it as an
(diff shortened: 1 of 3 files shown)


nativelink-store/src/grpc_store.rs
```diff
@@ -294,51 +303,120 @@ impl GrpcStore {
             stream,
         )));
 
+        let write_start = std::time::Instant::now();
+        let instance_name = self.instance_name.clone();
+        let rpc_timeout = self.rpc_timeout;
+        debug!(
```
palfrey: Can we make this `trace` otherwise we're going to get a log on every write, and that's going to be a lot for some workers I think?

nativelink-store/src/grpc_store.rs
```diff
@@ -294,51 +303,120 @@ impl GrpcStore {
...
+                                }
+                            }
+                        });
+
+                    drop(local_state_locked);
+                    Some((result, local_state))
+                }
             }))
             .await?;
+        info!(
```
palfrey: Similarly on the `trace` level. Unless you've done some testing with this? Worried we're going to end up with a lot of these logs!

nativelink-store/src/grpc_store.rs
```diff
@@ -596,6 +674,12 @@ impl StoreDriver for GrpcStore {
             digest.packed_hash(),
             digest.size_bytes(),
         );
+        debug!(
```
palfrey: Another level query

palfrey (changes requested):

PR #2267, opened 2026-04-07
Set arg0 for process.

chrisstaite-menlo:
# Description

In #2237 the process path was canoncialised to work around the Rust stdlib instability when using current_dir.  However, this breaks RBE where the program path changes the compiler behaviour.

Update the builder to ensure that the arg0 remains relative even when the process path is resolved.

## Type of change

## How Has This Been Tested?

In the process of testing.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2267)


MarcusSorealheis:
Once this is worked out, we should cut v1.0.1.

palfrey (approved):

PR #2285, opened 2026-04-27
Add use_legacy_resource_names option to GrpcSpec

jsharpe:
# Description

Some older backends (e.g. Buildbarn) do not understand the modern ByteStream resource name format that includes the digest function:

  {instance}/blobs/{digest_function}/{hash}/{size}

They expect the original format without the digest function component:

  {instance}/blobs/{hash}/{size}

Add a `use_legacy_resource_names` boolean to GrpcSpec (default false) that, when enabled, omits the digest function from ByteStream resource names for both reads and writes. This fixes InvalidArgument errors when proxying to such backends.

## Type of change

  not work as expected)

## How Has This Been Tested?

Tested with a local instance of nativelink configured as a pull through proxy for a remote buildbarn instance.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2285)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -1218,6 +1218,22 @@ pub struct GrpcSpec {
     /// Default: 0 (disabled)
     #[serde(default, deserialize_with = "convert_duration_with_shellexpand")]
     pub rpc_timeout_s: u64,
+
+    /// Use legacy ByteStream resource name format, omitting the digest
+    /// function component from the path.
+    ///
+    /// Modern NativeLink generates resource names like:
+    ///   `{instance}/blobs/{digest_function}/{hash}/{size}`
+    ///
+    /// Older backends (e.g. Buildbarn pre-v0.3) expect the original format:
+    ///   `{instance}/blobs/{hash}/{size}`
+    ///
+    /// Set this to `true` when connecting to such backends to avoid
+    /// `InvalidArgument: Unsupported digest function` errors.
+    ///
+    /// Default: false
+    #[serde(default)]
+    pub use_legacy_resource_names: bool,
 }
 
 /// The possible error codes that might occur on an upstream request.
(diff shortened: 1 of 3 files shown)


nativelink-config/src/stores.rs
```diff
@@ -1218,6 +1218,22 @@ pub struct GrpcSpec {
...
+    ///   `{instance}/blobs/{digest_function}/{hash}/{size}`
+    ///
+    /// Older backends (e.g. Buildbarn pre-v0.3) expect the original format:
+    ///   `{instance}/blobs/{hash}/{size}`
+    ///
+    /// Set this to `true` when connecting to such backends to avoid
+    /// `InvalidArgument: Unsupported digest function` errors.
+    ///
+    /// Default: false
+    #[serde(default)]
```
palfrey:
```suggestion
    #[serde(default, deserialize_with = "convert_boolean_with_shellexpand")]
```
So then environment variables work when setting this

nativelink-store/src/grpc_store.rs
```diff
@@ -676,22 +678,31 @@ impl StoreDriver for GrpcStore {
...
+                .to_ascii_lowercase();
+            format!(
+                "{}/uploads/{}/blobs/{}/{}/{}",
+                &self.instance_name,
+                Uuid::new_v4().hyphenated().encode_lower(&mut buf),
+                digest_function,
+                digest.packed_hash(),
+                digest.size_bytes(),
+            )
+        };
```
palfrey, line 705: Can you add a test for this please? This might be a bit tricky, as our only test so far in `grpc_store_test` doesn't have a fake backing store, so I can do some work there first if needed to provide a pattern for that if you need it.

palfrey (changes requested):
Thanks for the contribution! See my comments, and if you could also sign the CLA please so we can use this code?

jsharpe:
I'm waiting on internal review of the CLA.

nativelink-store/src/grpc_store.rs
palfrey: https://github.com/TraceMachina/nativelink/pull/2287 adds a gRPC store test with an actual gRPC server, which should provide a good pattern for doing tests here.

palfrey (approved):

PR #2287, opened 2026-04-28
Add grpc test with an actual gRPC server

palfrey:
# Description

Adds a first test to the gRPC store with an actual gRPC server. Mostly here to support testing work in https://github.com/TraceMachina/nativelink/pull/2285 but also more generally useful.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2287)

### DIFF
--- a/nativelink-store/tests/grpc_store_test.rs
+++ b/nativelink-store/tests/grpc_store_test.rs
@@ -1,21 +1,39 @@
+use core::pin::Pin;
 use core::time::Duration;
+use std::sync::Arc;
 
+use async_lock::Mutex;
+use futures::{Stream, StreamExt};
 use nativelink_config::stores::{GrpcEndpoint, GrpcSpec, Retry, StoreType};
-use nativelink_error::Error;
+use nativelink_error::{Error, ResultExt};
 use nativelink_macro::nativelink_test;
 use nativelink_proto::build::bazel::remote::execution::v2::{
     FindMissingBlobsRequest, digest_function,
 };
+use nativelink_proto::google::bytestream::byte_stream_server::{ByteStream, ByteStreamServer};
+use nativelink_proto::google::bytestream::{
+    QueryWriteStatusRequest, QueryWriteStatusResponse, ReadRequest, ReadResponse, WriteRequest,
+    WriteResponse,
+};
 use nativelink_store::grpc_store::GrpcStore;
+use nativelink_util::background_spawn;
+use nativelink_util::buf_channel::make_buf_channel_pair;
+use nativelink_util::common::DigestInfo;
+use nativelink_util::store_trait::{StoreLike, UploadSizeInfo};
+use regex::Regex;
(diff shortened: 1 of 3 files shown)


nativelink-store/tests/grpc_store_test.rs
```diff
@@ -43,3 +66,109 @@ async fn fast_find_missing_blobs() -> Result<(), Error> {
...
+        unimplemented!();
+    }
+}
+
+async fn make_fake_bytestream_server() -> (FakeStreamServer, u16) {
+    let fake_stream_server = FakeStreamServer::new();
+    let server = ByteStreamServer::new(fake_stream_server.clone());
+    let free_port = free_local_port().unwrap();
+
+    background_spawn!("server", async move {
```
MarcusSorealheis, line 129: Overall, pretty good ideas. This is racy.

`port_check::free_local_port()` and then a bind later inside a detached task. That leaves a [TOCTOU ](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)window where another test/process can take the port. If bind fails, the panic happens in the background task while the test still tries to connect. Also `make_fake_bytestream_server()` returns before the server is known to be accepting connections, and `Retry::default()` has max_retries = 0, so a startup race can fail the test.
     

I would prefer binding first with TcpListener::bind("127.0.0.1:0").await, taking the port from the bound listener, then serving from that listener with `serve_with_incoming`

nativelink-store/tests/grpc_store_test.rs
```diff
@@ -43,3 +66,109 @@ async fn fast_find_missing_blobs() -> Result<(), Error> {
...
+    ) -> Result<Response<Self::ReadStream>, Status> {
+        unimplemented!();
+    }
+
+    async fn write(
+        &self,
+        grpc_request: Request<Streaming<WriteRequest>>,
+    ) -> Result<Response<WriteResponse>, Status> {
+        let write_request = match grpc_request.into_inner().next().await {
+            None => {
```
MarcusSorealheis, line 102: Does the fake ByteStream server accept incomplete writes?

MarcusSorealheis (changes requested):

nativelink-store/tests/grpc_store_test.rs
palfrey: It'll take `WriteRequest`'s with `finish_write` as `false`. It's not at all concerned with that, it just records whatever it gets. It pays no attention to things like bad values for `write_offset`, etc.

MarcusSorealheis (approved):

nativelink-store/tests/grpc_store_test.rs
palfrey: Done, but with `TcpIncoming` from tonic

PR #2288, opened 2026-04-29
Forward client headers and OTEL trace context to upstream gRPC stores

jsharpe:
# Description

Adds two complementary mechanisms to GrpcStore for propagating headers to upstream remote caches (e.g. Buildbarn):

1. `headers` (GrpcSpec): static key/value pairs attached to every outgoing request, useful for fixed auth tokens.

2. `forward_headers` (GrpcSpec): header names to forward from the inbound client request. OtlpMiddleware captures all ASCII-valued headers from the client into a ClientHeaders value stored in the task context; enrich_request reads them back and injects whichever names are listed here. This enables JWT pass-through so build clients can authenticate directly with upstream caches.

Additionally, every outgoing request now has the current W3C trace context (traceparent/tracestate) injected via the OpenTelemetry propagator, fixing distributed trace continuity across NativeLink instances and into upstream services.

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2288)

### DIFF
--- a/nativelink-store/src/grpc_store.rs
+++ b/nativelink-store/src/grpc_store.rs
@@ -88,7 +88,7 @@ fn enrich_request<T>(
         && let Some(client_headers) = Context::current().get::<ClientHeaders>()
     {
         for name in forward_headers {
-            if let Some(value) = client_headers.0.get(name)
+            if let Some(value) = client_headers.0.get(&name.to_lowercase())
                 && let (Ok(k), Ok(v)) = (
                     MetadataKey::from_bytes(name.as_bytes()),
                     MetadataValue::try_from(value.as_str()),
@@ -147,7 +147,8 @@ impl GrpcStore {
 
         let mut headers = Vec::with_capacity(spec.headers.len());
         for (name, value) in &spec.headers {
-            let key = MetadataKey::from_bytes(name.as_bytes()).map_err(|_| {
+            // We lowercase keys as HTTP headers are case-insensitive so we should match all cases
+            let key = MetadataKey::from_bytes(name.to_lowercase().as_bytes()).map_err(|_| {
                 make_err!(Code::InvalidArgument, "Invalid gRPC metadata key: {name}")
             })?;
             let val = MetadataValue::try_from(value.as_str()).map_err(|_| {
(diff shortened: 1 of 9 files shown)


nativelink-util/tests/telemetry_test.rs
```diff
@@ -1,44 +1,102 @@
...
+    }
+}
+
+async fn header_extract(
+    ExtractClientHeaders(client_headers): ExtractClientHeaders,
+    State(state): State<Arc<AppState>>,
+    request: axum::extract::Request,
+    next: Next,
+) -> axum::response::Response {
+    debug!(?client_headers, "Client headers");
```
MarcusSorealheis, line 57: Umm, I may be mistaken.... Does this mean we will log the headers? Don't we need to exclude the values here or altogether redact? Maybe on the keys.

nativelink-store/src/grpc_store.rs
```diff
@@ -47,15 +47,63 @@ use nativelink_util::proto_stream_utils::{
...
 
+struct TonicMetadataInjector<'a>(&'a mut tonic::metadata::MetadataMap);
+
+impl Injector for TonicMetadataInjector<'_> {
+    fn set(&mut self, key: &str, value: String) {
+        if let (Ok(k), Ok(v)) = (
+            MetadataKey::from_bytes(key.as_bytes()),
+            MetadataValue::try_from(&value),
+        ) {
+            self.0.insert(k, v);
```
MarcusSorealheis, line 71: Unless something changed, i think both tonic and hyper with lowercase `HeaderName`.

MarcusSorealheis: Unless something changed, i think both tonic and hyper _will_ lowercase `HeaderName`. Is that still true?

nativelink-util/tests/telemetry_test.rs
palfrey: Yes, but that's test code only

nativelink-store/src/grpc_store.rs
palfrey: Both https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 and https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names say that HTTP headers are case-insensitive. I've added therefore added changes to explicitly lowercase both any headers coming in and the values going out to enable case-insensitive matching.

MarcusSorealheis (approved):
This LGTM pending checks. Thank you for cleaning this one up.

PR #2294, opened 2026-04-30
Detect changes as a result of Bazel runs

palfrey:
# Description

Main item here was picking up changes to `MODULE.bazel.lock`. See example run at https://github.com/TraceMachina/nativelink/actions/runs/25171818162/job/73793030412?pr=2294

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2294)

### DIFF
--- a/.github/workflows/lre.yaml
+++ b/.github/workflows/lre.yaml
@@ -48,6 +48,7 @@ jobs:
         run: >
           nix develop --impure --command
           bash -c "bazel run \
+           --lockfile_mode=error \
            --verbose_failures \
            @local-remote-execution//examples:${TOOLCHAIN}"
 
--- a/.github/workflows/native-bazel.yaml
+++ b/.github/workflows/native-bazel.yaml
@@ -48,10 +48,12 @@ jobs:
         run: |
           if [ "$RUNNER_OS" == "Linux" ] || [ "$RUNNER_OS" == "macOS" ]; then
             bazel test //... \
+              --lockfile_mode=error \
               --extra_toolchains=@rust_toolchains//:all \
               --verbose_failures
           elif [ "$RUNNER_OS" == "Windows" ]; then
             bazel \
+              --lockfile_mode=error \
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -25,3 +25,4 @@ user.bazelrc
 *.log
 buck-out/
 nativelink_config.schema.json
+bc
```
palfrey: I have no idea where this comes from, but it turns up in the diff. See https://github.com/TraceMachina/nativelink/actions/runs/25172330601/job/73794994184?pr=2294

MarcusSorealheis: We need to remove it before merging.

MarcusSorealheis (changes requested):
The errant line is very strange, leading me to look further into the lock file itself.

MarcusSorealheis:
Maybe try running this:

```bash
bazel clean
bazel build //... 2>&1
ls -la bc  # check timestamp and type
```

jsharpe:
You can use https://bazel.build/reference/command-line-reference#common_options-flag--lockfile_mode specifically for MODULE.bazel.lock file changes - pass `--lockfile_mode=error` in CI.

.gitignore
palfrey: I've dropped this from the gitignore

palfrey:
> Maybe try running this:
> 
> ```shell
> bazel clean
> bazel build //... 2>&1
> ls -la bc  # check timestamp and type
> ```

This gets nothing locally. I've seen this before in Github actions runs, I think it's Github-specific.

palfrey:
> You can use https://bazel.build/reference/command-line-reference#common_options-flag--lockfile_mode specifically for MODULE.bazel.lock file changes - pass `--lockfile_mode=error` in CI.

Nice! Didn't know about that one. I've switched out the `git diff` work for adding this to all the Bazel calls in the workflows.

MarcusSorealheis (approved):

PR #2298, opened 2026-05-02
Bound CAS leader-wait + per-blob batch deadline; tolerate empty FT.AGGREGATE

amankrx:
# Description

Two regression fixes to the read-dedup timeout introduced in the previous CAS-hang patch.                                                                           

1. The leader's populate was being cancelled by tokio::time::timeout(60s, ...), failing every concurrent reader of multi-GB blobs with DEADLINE_EXCEEDED: populate cancelled after 60 s leader-wait. Tagged is_leader on LoaderGuard; the leader now waits unbounded (its populate is the actual work), only followers honour the 60 s wait.                                                                           
2. A follower whose closure was promoted to leader by tokio's OnceCell then lost its caller's gRPC writer to that promoted-leader closure; if the timeout subsequently cancelled it, the writer was dropped without send_eof, surfacing as INTERNAL: Sender dropped before sending EOF. Followers now pass None for the writer; if their closure is ever promoted, it populates only the fast cache and the follower reads from it via the existing recursion.     
                                                                                                                                                                                                               Restores correct behaviour for Container Push Digest input fetches on layers ≳ 1 GB while keeping the original goal of preventing wedged-leader fan-out into client-side DEADLINE_EXCEEDED storms. 

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Tested this in the actual environment where we were facing this issue. 
## Checklist

- [X] Tests added/amended

<!-- Reviewable:start -->
- - -
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/TraceMachina/nativelink/2298)
<!-- Reviewable:end -->

### DIFF
--- a/nativelink-service/src/cas_server.rs
+++ b/nativelink-service/src/cas_server.rs
@@ -14,13 +14,14 @@
 
 use core::convert::Into;
 use core::pin::Pin;
+use core::time::Duration;
 use std::collections::{HashMap, VecDeque};
 
 use bytes::Bytes;
 use futures::stream::{FuturesUnordered, Stream};
 use futures::{StreamExt, TryStreamExt};
 use nativelink_config::cas_server::{CasStoreConfig, WithInstanceName};
-use nativelink_error::{Code, Error, ResultExt, error_if, make_input_err};
+use nativelink_error::{Code, Error, ResultExt, error_if, make_err, make_input_err};
 use nativelink_proto::build::bazel::remote::execution::v2::content_addressable_storage_server::{
     ContentAddressableStorage, ContentAddressableStorageServer as Server,
 };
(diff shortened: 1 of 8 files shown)


MarcusSorealheis (approved):

nativelink-store/src/redis_store.rs
```diff
@@ -1755,6 +1755,10 @@ where
                 }
             };
 
+            if matches!(raw_redis_map, Value::Int(_)) {
+                return None;
+            }
```
palfrey, line 1811: What's this for?

palfrey (changes requested):
Not seeing any new/changed test code here?

nativelink-store/src/redis_store.rs
amankrx: Without this, every empty `FT.AGGREGATE` response surfaced as a hard `Code::Internal` error. Now, empty responses produce an empty Stream instead.

palfrey: Ok, great. Can we get a test for that case please?

nativelink-service/src/cas_server.rs
palfrey, line 1: Not seeing any tests for the changes here?

palfrey (changes requested):

palfrey (approved):

PR #2300, opened 2026-05-03
Remove cascading eviction map from `EvictingMap::get`

MarcusSorealheis:
# Description

**Before**: `get()` acquired the lock 3 times, ran a fast-path `should_evict` peek, then if true ran the global `evict_items` loop (popping every expired/over-budget LRU entry) and awaited every victim's `unref()` and remove-callbacks before returning. A read of a fresh key paid the cost of all expired neighbors.

**After**: `get()` acquires the lock once. If the queried entry doesn't exist, return `None`. If it exists and is itself expired (TTL/count predicates only — `sum_store_size=0, max_bytes=u64::MAX`), reap that one entry and drain its callbacks/unref outside the lock. Otherwise touch `atime` and return the value. Never cascades.

## Type of change

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2300)

### DIFF
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -434,42 +434,50 @@ where
         while callbacks.next().await.is_some() {}
     }
 
+    /// Returns the value for `key` if present and not expired, refreshing
+    /// its LRU/atime position. If the entry is present but TTL- or
+    /// count-expired, it is reaped and `None` is returned.
+    ///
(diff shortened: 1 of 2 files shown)


palfrey (approved):

palfrey:
Tested this against the Helm chart workflow, all good there. Stable run through the Step 7 tests of a full nativelink build with the image from here.

PR #2309, opened 2026-05-07
MacOS-specific free disk

palfrey:
# Description

We need to free up disk to run things like Nix, and the existing disk free action only works on Linux. This PR disables that step on non-Linux and runs a new Mac-specific one instead

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2309)


MarcusSorealheis (approved):
thanks

PR #2315, opened 2026-05-08
Add expiry to completed redis actions

palfrey:
# Description

The scheduler has a config option `retain_completed_for_s` that the Redis scheduler store doesn't actually pay attention to, and so you can end up over time with many (we've observed millions) of old completed entries in Redis that then slow everything down.

This PR explicitly sets a TTL with `EXPIRE` for completed actions, using the config option value.

## Type of change

  not work as expected)

This is a breaking change as there will be some cases where an action would have previously stuck around for longer and that would have been needed in some cases, but the user hadn't set the timeout to as high as they actually needed. In this case, they'll need to actually configure the timeout v.s. getting away without it.

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2315)

### DIFF
--- a/nativelink-config/src/schedulers.rs
+++ b/nativelink-config/src/schedulers.rs
@@ -112,7 +112,7 @@ pub struct SimpleSpec {
     /// config.
     pub supported_platform_properties: Option<HashMap<String, PropertyType>>,
 
-    /// The amount of time to retain completed actions in memory for in case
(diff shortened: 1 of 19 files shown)


MarcusSorealheis (approved):
LGTM

PR #2319, opened 2026-05-10
Add Cloudflare R2 support

### LINKED ISSUE #2293: Add Cloudflare R2 object store support
NativeLink should support Cloudflare R2 as a first-class CAS/AC object store backend.

R2 is S3-compatible, so we should reuse as much of the existing S3/cloud object store implementation as possible, while exposing R2-specific configuration that makes setup straightforward for users.

ashishsonamm:
# Description

Adds Cloudflare R2 as a first-class CAS/AC backend by generalizing `S3Store`

Introduced 3 new optional fields in `ExperimentalAwsSpec`
- endpoint
- access_key_id
- secret_access_key

If all the above three fields are unset, `S3Store::new` behaves exactly as before

Fixes #2293 

## Type of change

## How Has This Been Tested?

- Unit Test
- Bazel Test
- E2E against Real R2 store

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2319)


ashishsonamm (commented):
+@MarcusSorealheis

> @ashishsonamm made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2319#-:-OsHRe232BJuljdoLYCG:b2iyrkc)* status: 0 of 1 LGTMs obtained, and 0 of 8 files reviewed.

MarcusSorealheis:
Hey @ashishsonamm thanks for coming through and opening a PR. We'd love to support you in landing your first PR here. 

A couple initial thoughts:

1. This feature is user-facing so we'd like to add documentation in the web/ dir where the user docs mostly live. 

2. This feature is more similar to https://github.com/TraceMachina/nativelink/blob/main/nativelink-store/src/ontap_s3_existence_cache_store.rs in its function in the sense that while Cloudflare does implement a lot of the s3 API, it should not affect the s3 store. I'm curious about your thoughts there. How can we minimize the surface area of the s3 store to support the addition of Cloudflare?

Really awesome work!!!

ashishsonamm:
> 2\. This feature is more similar to https://github.com/TraceMachina/nativelink/blob/main/nativelink-store/src/ontap_s3_existence_cache_store.rs in its function in the sense that while Cloudflare does implement a lot of the s3 API, it should not affect the s3 store. I'm curious about your thoughts there. How can we minimize the surface area of the s3 store to support the addition of Cloudflare?

I looked at the `ontap_s3_existence_cache_store.rs`, it composes the inner store and `OntapS3Store` stays untouched. My PR does this partially, the R2 store composes S3Store, but it also modifies `ExperimentalAwsSpec` and `S3Store::new` to inject the endpoint and creds, which is the surface area you're talking about. 

The fix would be to rework R2Store so it builds its own aws S3 client with the endpoint and creds, and passes it to `S3Store::new_with_client_and_jitter`, this way S3Store stays untouched while still reusing most of the protocol code via the inner store.

I'll make this change and will also add the documentation

ashishsonamm:
Pushed the change for R2Store to build its own S3 client without touching any of the existing S3 store code.
For docs, the files under `web` are auto-generated from `nativelink-config/README.md`, so I have added the R2 section there.

palfrey:
> Pushed the change for R2Store to build its own S3 client without touching any of the existing S3 store code.

Nice, liking this. If it's meant to be "S3 with another endpoint/region config" then this is the right way to go with this.

palfrey (approved):

PR #2321, opened 2026-05-11
Test actual LRE in CI

palfrey:
# Description

We previously had LRE testing, but it was a mix of trivial examples and buggy k8s-in-CI setups long commented out. This PR spins up a docker compose setup with the current branch's nix images and tries to build them, which will make it much easier to make improvements to those.

## Type of change

## How Has This Been Tested?

CI, bunch of local `docker compose` and `bazel test` similar to the CI steps.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2321)


deployment-examples/docker-compose/Dockerfile
```diff
@@ -66,4 +66,4 @@ RUN apt-get update \
     && mkdir -p /root/.cache/nativelink
 
 EXPOSE 50051/tcp 50052/tcp
-CMD ["nativelink"]
+ENTRYPOINT ["nativelink"]
```
palfrey, line 69: Making it behave like the nix images

deployment-examples/docker-compose/local-storage-cas.json5
palfrey, line 1: Using the newer config so I don't get the warnings every time I boot things

deployment-examples/docker-compose/worker.json5
```diff
@@ -63,7 +63,7 @@
         work_directory: "/root/.cache/nativelink/work",
         platform_properties: {
           cpu_count: {
-            query_cmd: "nproc",
+            values: [],
```
palfrey, line 66: Path isn't present on the nix images

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -1416,6 +1418,7 @@ impl RunningActionImpl {
                 exit_code = ?execution_result.exit_code,
                 stdout = ?stdout[..min(stdout.len(), 1000)],
                 stderr = ?stderr[..min(stderr.len(), 1000)],
+                command = ?command_proto.arguments,
```
palfrey, line 1971: The error was just `Command returned non-zero exit code, exit_code: 1, stdout: "", stderr: "bootstrap_process_wrapper: execvp: No such file or directory\n"` or similar, now we get the full command

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -959,7 +959,9 @@ impl RunningActionImpl {
         //                    level more effectively and adjust this.
         info!(?args, "Executing command");
 
-        let program = self.canonicalise_path(args[0], &command_proto.working_directory)?;
+        let program = self
+            .canonicalise_path(args[0], &command_proto.working_directory)
+            .err_tip(|| format!("Canonicalisation failure. Command={args:#?}"))?;
```
palfrey, line 1268: `Error executing action, err: Error { code: NotFound, messages: ["No such file or directory (os error 2)", "Could not canonicalize path for command root /bin/bash."] }` isn't very useful without knowing what was trying to invoke bash

.github/workflows/lre.yaml
```diff
@@ -58,211 +58,57 @@ jobs:
...
+        nativelink_attic_token: ${{ secrets.NATIVELINK_ATTIC_TOKEN }}
+
+    - name: Build Nativelink images
+      run: |
+        nix run .#image.copyTo docker-daemon:local-nativelink:latest
+        nix run .#nativelink-worker-lre-rs.copyTo docker-daemon:local-worker-rs:latest
+        nix run .#nativelink-worker-init.copyTo docker-daemon:local-worker-init:latest
+
+    - name: Compile NativeLink with NativeLink
+      timeout-minutes: 5 # FIXME: Reset to 60 before merge
```
palfrey, line 96: FIXME

nativelink-store/Cargo.toml
```diff
@@ -124,10 +124,6 @@ wincode = { version = "0.5.4", default-features = false, features = [
 [dev-dependencies]
 nativelink-macro = { path = "../nativelink-macro" }
 
-aws-sdk-s3 = { version = "1.82.0", features = [
-  "http-1x",
-  "rt-tokio",
-], default-features = false }
```
palfrey, line 143: Duplicate of what's in non-dev

nativelink-store/Cargo.toml
```diff
@@ -17,7 +17,7 @@ nativelink-util = { path = "../nativelink-util" }
 async-lock = { version = "3.4.0", features = ["std"], default-features = false }
 async-trait = { version = "0.1.88", default-features = false }
 aws-config = { version = "1.6.1", default-features = false, features = ["sso"] }
-aws-sdk-s3 = { version = "1.82.0", features = [
+aws-sdk-s3 = { version = "1.134.0", features = [
```
palfrey: Mostly here to upgrade `crc-fast` to 1.10.0, which has some issues with musl

palfrey:
I think what's missing is libz, and something incorporating https://github.com/hermeticbuild/rules_rs/pull/139 might be able to fix this?

palfrey:
> I think what's missing is libz, and something incorporating [hermeticbuild/rules_rs#139](https://github.com/hermeticbuild/rules_rs/pull/139) might be able to fix this?

I'm now more uncertain about this. So, the reason I flagged libz is that the glibc version of rustc was trying to be run under the Nix LRE container, and it was failing due to a lack of libz. I had a fix for adding libz, but was still seeing the same problem. Now, ideally what we actually want is the _musl_ version of rustc run under the container, but getting `rules_rs` to do that seems tricky, and I haven't figured that out yet. I think also that a lot of the problems here were caused by moving to the hermetic rust setup, which then pulled in it's own copy of Rust v.s. having previously the Nix version. I think it's fine to use the Bazel hermeticity for getting Rust itself, we just need to make sure it's compatible with what Nix wants to do...

PR #2050, opened 2025-11-15
[v1.2.0-rc1 Feature] introduce safe warm workers

MarcusSorealheis:
# Description

Implements Copy-on-Write (COW) isolation for warm worker pools to prevent cross-tenant state contamination in multi-tenant RBE deployments.

Fixes #2049 

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

Locally. Needs to be tested by the community. 

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2050)

PR #2323, opened 2026-05-12
Implement Remote Persistent Workers

MarcusSorealheis:
# Description

Adds remote persistent worker support for NativeLink worker execution, including Bazel worker protocol handling, worker pooling, lifecycle management, and focused scheduler/worker tests.

Also adds persistent worker deployment examples and website docs, plus a Bazel wrapper for checking the website docs build.

If you need something that we don't have here, please let us know.

## Not Implemented In v1

  - Multiplex workers: no concurrent `WorkRequest`s against one process.
  - Per-request worker sandboxing: no Bazel `--worker_sandboxing` semantics yet.
  - Dynamic strategy: no local-plus-remote racing.
  - Cross-host or cross-worker migration of in-memory tool state.
  - Global live-worker memory cap across all `WorkerKey`s.
  - Helm memory-sizing guidance is called out in the design, but not implemented here.

Fixes #2050 

## Type of change

  not work as expected)

## How Has This Been Tested?

`bazel test //nativelink-worker:unit_test //nativelink-worker:integration //nativelink-scheduler:integration --test_output=errors`

AND (newly because this PR will put in place systems that make TS much faster)

bazel test //web/platform:check_test --test_output=errors

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2323)


MarcusSorealheis:
Oddly, the only failure we saw was on Windows but it should've failed on all platforms. Something to think about @palfrey and maybe a consequence of the work to integrate the new rulesets.

MarcusSorealheis:
FYI: Current CI failure is related to our upstream Nix cache not the code in the PR. It passed prior to the merge commit and has been stable in my testing.

nativelink-worker/src/persistent_worker/pool.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+
+        // Lock released; spawn outside the critical section.
+        let spawn_result = LiveWorker::spawn(
+            executable_path,
+            &key.startup_args,
+            key.wire_format,
+            working_dir,
+        )
+        .await;
```
erneestoc, line 339: Worker is spawned with `current_dir = action_directory`. That directory gets `remove_dir_all`'d at `running_actions_manager.rs:726`. The second action to reuse this worker runs with cwd pointing at a deleted inode — anything resolving relative paths (`javac -d .`, `tsc --rootDir`, every `clang -I.`) breaks silently or with confusing ENOENTs.

The integration test passes because the test worker ignores cwd and uses `sandbox_dir` directly.

nativelink-worker/src/persistent_worker/pool.rs
```diff
@@ -0,0 +1,425 @@
...
+    pub async fn release(mut self, healthy: bool) {
+        let Some(inner) = self.inner.take() else {
+            return;
+        };
+        inner
+            .pool
+            .return_worker(inner.key, inner.worker, healthy)
+            .await;
+    }
+}
```
erneestoc, line 185: `take()` is synchronous, then we await. If the caller's future is dropped mid-await (any `select!` wrapping this) the worker drops (process killed by `kill_on_drop`) but `decrement_count` inside `return_worker` (`:218`) is past `shutdown.await` and never runs. `total_count` is permanently inflated.

Same shape in `return_worker` itself, and same shape in `acquire` (see separate comment).

RAII slot guard that decrements on Drop unless explicitly committed closes all three sites with one helper. It should have a test that drops `release(...).await` mid-await and asserts the next `acquire` succeeds.

nativelink-worker/src/persistent_worker/pool.rs
```diff
@@ -0,0 +1,425 @@
...
+        )
+        .await;
+        let worker = match spawn_result {
+            Ok(w) => w,
+            Err(err) => {
+                // Roll back our reservation on spawn failure.
+                self.inner.decrement_count(&key);
+                return Err(err).err_tip(|| format!("Spawning persistent worker for {key:?}"));
+            }
+        };
```
erneestoc, line 345: Same cancellation-leak pattern as `release`. Slot is reserved under the lock (`total_count += 1`), lock released, then `LiveWorker::spawn(...).await`. Drop the future mid-spawn and `decrement_count` at `:305` never runs.

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -960,6 +1006,90 @@ impl RunningActionImpl {
...
+                    }
+                }
+                Err(err) => {
+                    info!(
+                        ?err,
+                        "Falling back to one-shot execution; unsupported persistent worker protocol"
+                    );
+                }
+            }
+        }
```
erneestoc, line 1154: `kill_channel_rx` is taken from state at `:980-984` and never selected on in the persistent-worker branch. The one-shot path correctly selects against it at `:1371`. So scheduler kills, drains, and client cancels are dropped silently — the dispatch eventually completes (or hits `self.timeout`, which defaults to `max_action_timeout`) and the response is discarded by the dropped future.

The practical impact: a NativeLink worker pod asked to drain cannot actually drain until every in-flight persistent dispatch hits its individual timeout.

Wrap the dispatch in a `select!` against `kill_channel_rx`. On kill, release the lease as unhealthy (the partial write makes the worker unsafe to reuse) and return `Code::Cancelled`.

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -960,6 +1006,90 @@ impl RunningActionImpl {
...
+                            } else {
+                                self.metrics().child_process_failure_error_code.inc();
+                            }
+                            info!(?args, ?key, "Persistent worker command complete");
+                            {
+                                let mut state = self.state.lock();
+                                state.command_proto = Some(command_proto);
+                                state.execution_result = Some(RunningActionImplExecutionResult {
+                                    stdout: Bytes::from(response.output),
+                                    stderr: Bytes::new(),
```
erneestoc: `Stdio::inherit()` on stderr sends worker diagnostics into NativeLink's own log. `stderr: Bytes::new()` on the response means the `ActionResult` always carries empty stderr.

nativelink-worker/src/persistent_worker/pool.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    #[test]
+    fn worker_key_distinguishes_wire_formats() {
+        let a =
+            WorkerKey::from_argv(&["javac", "@a"].map(String::from), WireFormat::Proto).unwrap();
+        let b = WorkerKey::from_argv(&["javac", "@a"].map(String::from), WireFormat::Json).unwrap();
+        assert_ne!(a, b);
+    }
+}
```
erneestoc, line 542: maybe add concurrency tests

erneestoc (commented):
Thank you for tackling this!

nativelink-worker/src/persistent_worker/pool.rs
MarcusSorealheis: See `CountSlotGuard`. 

```rust
let count_slot = CountSlotGuard::new(pool.clone(), key.clone());
                warn!(
                    ?key,
                    "Lease dropped without release; treating worker as unhealthy"
                );
                worker.shutdown(grace).await;
                count_slot.decrement_now();
```

nativelink-worker/src/running_actions_manager.rs
MarcusSorealheis: I've addressed this as well.

erneestoc (approved):

MarcusSorealheis:
Locally, another thing I tried was I set up and verified a local persistent-worker cluster on the MacBook.

```markdown

  What’s running:

  - NativeLink local scheduler/CAS/workers on grpc://127.0.0.1:50051
  - Worker API on 127.0.0.1:50061
  - Config: /tmp/nl-pw-smoke/nativelink-pw.json5 ~ actual link to the config at the bottom of this comment.
  - Smoke workspace: /tmp/nl-pw-smoke

  Verification passed:

  - Built a tiny Bazel workspace through local NativeLink remote execution.
  - Actions hit the persistent-worker path: logs show Spawned new
    persistent worker and Persistent worker command complete.
  - Repeated clean/build cycles advanced worker request counts from 1 to 2
    to 3, proving the persistent workers stayed alive and were reused
    across Bazel invocations.

  Latest output:

  one count=3 cwd=/private/tmp/nativelink-multi-test/worker2/work
  two count=3 cwd=/private/tmp/nativelink-multi-test/worker3/work
  three count=3 cwd=/private/tmp/nativelink-multi-test/worker1/work
  ```
  
Everything seems to be working just fine. After this is merged, I will upgrade the version to `v1.2.1`. [Here is the config I used to smoke test](https://gist.github.com/MarcusSorealheis/086993bf0a5d9d6fb8a3c60331f83cbf).

**Review by amankrx — APPROVED** [review]
Tested with the helm chart and Persistent Workers are working as expectec.

File: nativelink-worker/src/persistent_worker/mod.rs [hunk 7]
```diff
@@ -0,0 +1,36 @@
+// Copyright 2024 Trace Machina, Inc. All rights reserved.
+//
+// Licensed under the Business Source License, Version 1.1 (the "License");
+// you may not use this file except in compliance with the License.
+// You may requested a copy of the License by emailing contact@nativelink.com.
+//
+// Use of this module requires an enterprise license agreement, which can be
+// attained by emailing contact@nativelink.com or signing up for Nativelink
```
erneestoc, line 8: 🥲

MarcusSorealheis: Not you ser! You have contributed tremendously. It doesn't apply to you.

PR #2328, opened 2026-05-13
Astro 6

palfrey:
# Description

Upgrades Astro to v6 and related package changes. Continues https://github.com/TraceMachina/nativelink/pull/2226 because merging that branch broke Deno in some way, so https://nativelink.com/ started doing 504's.

## Type of change

## How Has This Been Tested?

Manually, CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2328)


thejackshelton:
Hey @palfrey, as a heads up I've been working on a rewrite of Qwik's bundler (in a relatively good state), but this requires Astro to release vite 8 and rolldown support for it to be used. I do believe this will lead to native deno support if it is not supported already or if you are having any current incompatibility issues.

palfrey:
> Hey @palfrey, as a heads up I've been working on a rewrite of Qwik's bundler (in a relatively good state), but this requires Astro to release vite 8 and rolldown support for it to be used. I do believe this will lead to native deno support if it is not supported already or if you are having any current incompatibility issues.

I'm still seeing
```[WARN] [vite] node_modules/@deno/astro-adapter/src/server.ts (108:12): "start" is not exported by "virtual:@deno/astro-adapter:config", imported by "node_modules/@deno/astro-adapter/src/server.ts".```
in the logs, which I suspect is the source of the broken deploy issues. Tried upgrading `@deno/astro-adapter` to 0.5.1, still no luck there.

Also `Qwik core bundle not found, is Qwik actually used in this project?` which is a fun one.

MarcusSorealheis:
this is a time suck. We are going to drop this website in favor of a new version.

thejackshelton:
> Also Qwik core bundle not found, is Qwik actually used in this project? which is a fun one.

This is not actually true. It's an assumption from the current qwik vite plugin that both builds happen in an expected order, which in the Astro integration the order is a bit different than expected due to the way that Astro currently expects the build.

As for the bundler rewrite, I have released its first version under qwik-bundler

https://github.com/thejackshelton/qwik-bundler/tree/main/fixtures

Here is the list of available fixtures. This even includes native deno support without any node compatibility using vite environments!

Unfortunately, I think the blocker here is on the Astro side. I cannot update the Qwik integration for Astro until the next major release where rolldown and vite 8 is supported. Unsure of the Deno adapter support for Astro.

MarcusSorealheis:
@thejackshelton we went in a different direction but we enjoyed using the project. We appreciate you for all the hard work and attentiveness.

thejackshelton:
Thank you! I look forward to seeing NativeLink evolve and happy I could be a part of the journey (even if it was short 🙂 )

MarcusSorealheis:
It was a couple years. We could be back.


Marcus Eagan

PR #2330, opened 2026-05-13
Split rbe-toolchain into multiple tests

palfrey:
# Description

It's the slowest test by about a factor of 2-3x now, mostly because it's got a lot of subtests. This PR gives the command a `list` subcommand as well as the ability to run individual subcommands. The overall end-to-end time for a RBE test goes from ~24m to ~6m max, even though there's a lot more of them. This does increase our needed test runner count a bunch, but only on Ubuntu machines which we have a lot of headroom for and tend to get allocated very fast.

It also removes the "remove package" step of the free disk action, as that wasn't freeing up much space, but it was very unreliable in terms of runtime, and especially with more separate jobs it was radically increasing the runtime sometimes.

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2330)

### DIFF
--- a/.github/actions/free-disk/action.yaml
+++ b/.github/actions/free-disk/action.yaml
@@ -4,6 +4,9 @@ description: "Free up disk space on workers"
 runs:
   using: "composite"
   steps:
+    - name: Check disk space
+      run: df -h /
+      shell: bash
     - if: runner.os == 'Linux' # Only works on Linux
       name: Free disk space
       uses: >- # v3.2.2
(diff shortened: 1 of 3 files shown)


.github/actions/free-disk/action.yaml
```diff
@@ -15,17 +18,6 @@ runs:
...
-          azure-cli
-          microsoft-edge-stable
-          google-chrome-stable
-          firefox
-          postgresql*
-          temurin-*
-          *llvm*
-          mysql*
-          dotnet-sdk-*
-        remove_packages_one_command: true
```
palfrey, line 28: This is incredibly slow, sometimes minutes, and doesn't save us _that_ much vs the raw removing of folders.

amankrx (approved):

PR #2333, opened 2026-05-14
Generate PreconditionFailure

palfrey:
# Description

`nativelink-util/src/precondition_failure.rs` had a hardcoded copy of `PreconditionFailure` when we've done all the other gRPC messages via generation from `.proto` files. This PR redoes it with that mechanism, and then uses the existing `TypeUrl` trait to be able to put the `TYPE_URL` constant in a slightly more organised location.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2333)


nativelink-proto/BUILD.bazel
```diff
@@ -186,9 +186,3 @@ rust_doc(
     crate = ":nativelink-proto",
     visibility = ["//visibility:public"],
 )
-
-rust_doc_test(
-    name = "doc_test",
-    timeout = "short",
-    crate = ":nativelink-proto",
-)
```
palfrey, line 194: For some annoying reason, the new proto makes some things that look like doc tests, but fail. Given all the code is generic generated code, it shouldn't really have any doc tests anyways.

MarcusSorealheis (approved):
We should have done this but iteration is most important.

PR #2243, opened 2026-03-24
Production hardening: scheduling, P2P transfers, directory cache, TLS/QUIC

rejuvenile:
## Summary

Comprehensive production hardening from running NativeLink at scale with 10 Mac Mini workers and a ZFS-backed server. 156 commits across the following areas:

### Store reliability & I/O performance
- Fix store pipeline races causing Bazel "Lost inputs" errors
- Fix ExistenceCacheStore false positives and LRU promotion race
- Fix BatchUpdateBlobs missing results for duplicate digests
- Retry hardlinks on cache eviction, hold file path read-lock during hardlink
- Fire-and-forget eviction unrefs + CAS dedup in FilesystemStore
- POSIX_FADV_SEQUENTIAL for kernel readahead, rayon-based blake3 hashing
- Fix LRU eviction ordering at startup (sort by atime, not directory order)
- Fix zero-digest file handling in FilesystemStore and worker pipeline

### MokaEvictingMap — lock-free TinyLFU cache (NEW)
- Replace Mutex+LRU EvictingMap with moka::sync::Cache (lock-free reads)
- TinyLFU admission with frequency bump on insert (prevents single-access rejection)
- Pinning via DashMap with 120s auto-timeout, 25% cap
- Bounded eviction channel (4096) with inline fallback
- Startup atime ordering preserved via insert_startup (no frequency bump)
- KB-scaled weigher for items up to 4TB
- All 4 stores migrated: ExistenceCacheStore, MemoryStore, FilesystemStore, MemoryAwaitedActionDb
- Old EvictingMap/ShardedEvictingMap removed (-2,177 lines)
- 18 integration tests covering full API
- **Production result: lock contention 391 (406ms worst) → 0**

### Redis pipelining (NEW)
- has_with_results: batch STRLEN+EXISTS into single pipeline (N keys → 1 round-trip)
- batch_get_part_unchunked: pipelined GETRANGE+EXISTS across store chain
- BatchReadBlobs: delegates to batch pipeline instead of N individual reads
- RENAME+PUBLISH pipelined (saves 1 RTT per write with pub_sub)
- Pipeline chunking (5000 max) prevents unbounded response buffering
- CrossSlot fallback for Redis cluster mode
- **Production result: Redis slow commands 10,400 → 0**

### Batch store reads (NEW)
- batch_get_part_unchunked trait method across full store chain
- MemoryStore override: direct get_many() + BytesWrapper::to_contiguous(), zero buf_channel
[...]

Issue #2034, opened 2025-11-10
Store level zero-digest fast-path


palfrey:
Currently we have some work in stores for fast-path'ing zero-digests because we can just assume they always exist and return a default item for them. This issue covers implementing them for all stores. We can't just do this in `StoreLike` as sometimes we use a specific store directly and need to actually handle this properly there. 

We should do this at least for all of the base stores (ones that store stuff themselves, not in a child store) as those have most of the overhead this helps, but it should be an improvement for the higher-level stores as well.

* [ ] CompletenessChecking
* [ ] Compression
* [ ] Dedup
* [ ] ExistenceCache
* [ ] FastSlow
* [ ] Filesystem https://github.com/TraceMachina/nativelink/pull/2033
* [x] GCS https://github.com/TraceMachina/nativelink/pull/1995
* [ ] gRPC
* [ ] Memory
* [ ] Mongo
* [x] NoOp (not needed)
* [ ] Ontap
* [ ] Redis
* [x] Ref (not needed)
* [ ] S3
* [ ] Shard (maybe just passthrough)
* [ ] SizePartitioning (again, maybe just passthrough)
* [ ] Verify

QuiteDiam0nd:
Hey @palfrey! 

I've been reading and investigating through the store layer and I ended up looking at this issue. The checklist is a bit stale as I see: Filesystem, Memory, S3, GCS, Redis, Mongo and OntapS3 already have at least partial zero-digest fast-paths via `is_zero_digest`. The gap that's left is that the wrapper stores don't.

ExistenceCacheStore and FastSlowStore still walk their full path for the empty digest before bottoming out at a base store that short-circuits. GrpcStore already has a partial shortcut, but it only checks size, not hash...

 Mind if I open it as a PR with changes for that? Happy to split per store if you'd prefer. I also want to measure somehow improvements from that as a follow-up, but don't have all possible setups to test.

palfrey:
> I've been reading and investigating through the store layer and I ended up looking at this issue. The checklist is a bit stale as I see: Filesystem, Memory, S3, GCS, Redis, Mongo and OntapS3 already have at least partial zero-digest fast-paths via `is_zero_digest`. The gap that's left is that the wrapper stores don't.

Filesystem and GCS are both done already, as noted in the ticked off list in the top of this. `is_zero_digest` is the core of things, but it does need to actually be called! For example, Memory's `update` method doesn't call it, and similarly for the others.
 
[...]

PR #2338, opened 2026-05-16
macOS: APFS clonefile fast path + concurrency cap + zero-byte fix for Bazel input materialization

erneestoc:
## Summary

Four small, atomic commits that make macOS RBE workers materialize Bazel
action input trees dramatically faster on APFS. This is a focused
re-extraction of the safest, most independently-verifiable subset of
#2243 — split out so each change can be reviewed, benched, and rolled
back on its own.

| Commit | Scope |
|---|---|
| `b3b0cd3f` | `fs_util`: APFS `clonefile(2)` fast path in `hardlink_directory_tree` (with `set_readwrite_recursive` so cloned trees stay writable). Falls through to per-file `hard_link` on Linux/Windows. |
| `ab49f162` | `DirectoryCache`: `clonefile_hits` / `hardlink_hits` AtomicU64 telemetry so operators can confirm the fast path is firing in production. |
| `1ddce0fc` | `running_actions_manager::download_to_directory`: cap concurrent `hard_link` calls at 64 via `stream::buffer_unordered` (was unbounded `FuturesUnordered`). |
| `8051ca9e` | `DirectoryCache`: handle the case where the CAS has no entry for the zero-byte digest — short-circuit and create an empty file rather than panic. |

## Motivation

Bazel exec-log analysis on our macOS RBE worker fleet shows that input
materialization dominates wall time:

| Metric | Value |
|---|---|
| Sum of action durations (process wall time) | 4,947 s |
| Sum of file fetching durations | **7,313 s** |
| Critical path | 1,074 s |
| Total wall time | 1,213 s |

File-fetching wall time exceeds *all* process wall time combined — the
worker spends more time materializing input trees than running actions.
The dominant action shape is SwiftCompile (635 input files / 183 MB
mean, 1,978 files / 466 MB at p95), which is exactly the workload APFS
\`clonefile(2)\` was designed for.

#2243 is large and bundles several unrelated changes. This PR isolates
the four pieces that (a) directly attack the file-fetching bottleneck
on macOS and (b) can be A/B tested independently via the telemetry in
\`ab49f162\`.

## Benchmark proof

See https://github.com/erneestoc/nativelink/pull/1 — adds two criterion
benches under \`nativelink-util/benches/\` that reproduce the wins on a
real APFS volume.

Headline numbers on Apple M4 Max / APFS:

- **\`hardlink_directory_tree\` (b3b0cd3f)** — 3.6×–4.0× faster across
  all five SwiftCompile-shaped trees. p95 shape (1,978 files / 466 MB)
  drops from **590 ms → 150 ms per action**.
- **\`download_to_directory\` 64-cap (1ddce0fc)** — performance-neutral
  on a single-process microbench (1.01× vs unbounded at 1,978 files).
  Confirms the cap is not a regression; any multi-action contention
  wins will show up in production telemetry.

### Caveat: 4× not 10×

#2243 reported ≥10× wins on this path. Our benches show a stable ~4×,
bounded by the O(N) \`set_readwrite_recursive\` chmod walk that runs
after the O(1) \`clonefile(2)\`. Diagnosis and follow-up (single
top-level chmod + lazy per-file) documented in the bench PR. 4× on
the dominant shape already moves the needle hard — proposing this
lands now, with the walk optimization as a separate follow-up.

## Telemetry & rollback

\`ab49f162\` adds \`DirectoryCache::stats()\` with \`clonefile_hits\` and
\`hardlink_hits\` counters so operators can confirm in production:

- macOS workers should show \`clonefile_hits\` climbing and
  \`hardlink_hits\` near zero
- Linux workers should show the inverse (clonefile path is
  macOS-gated at compile time)

The four commits revert independently — each is its own atomic change
with its own test coverage.

## Tests

10 \`fs_util\` tests cover the clonefile path (COW isolation, symlink
handling under \`CLONE_NOFOLLOW\`, error paths, permission transitions
from \`0o555\` → \`0o755\`). 2 \`DirectoryCache\` tests cover the telemetry
counters and the zero-byte short-circuit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2338)


nativelink-util/src/fs_util.rs
```diff
@@ -55,6 +63,44 @@ pub async fn hardlink_directory_tree(src_dir: &Path, dst_dir: &Path) -> Result<(
...
+    // clonefile(2) requires dst's parent to exist but dst itself must NOT
+    // exist. Make sure the parent is present without creating dst.
+    if let Some(parent) = dst_dir.parent() {
+        fs::create_dir_all(parent).await.err_tip(|| {
+            format!(
+                "Failed to create parent of destination: {}",
+                parent.display()
+            )
+        })?;
+    }
```
palfrey: Shouldn't this be folded into the `target_os = "macos"` block if it's only needed for clonefile?

palfrey:
> `Caveat: 4× not 10×`

This PR is a lot more readable and reviewable than that one, and I'll take 4x I can review over 10x with harder to evaluate side-effects!

Thanks for this, very happy with it. Will also be able to tick off DirectoryCache from https://github.com/TraceMachina/nativelink/issues/2034 after this merges.

palfrey (changes requested):
One minor item

nativelink-util/src/fs_util.rs
erneestoc: Good catch — you're right. Done!

palfrey (approved):

PR #2342, opened 2026-05-17
worker: prefer macOS P-cores via pthread QoS USER_INITIATED

erneestoc:
## Summary

On macOS, set `pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0)` on the main thread, the tokio runtime's worker threads (via `Builder::on_thread_start`), and at the top of `LocalWorker::run`. The QoS class is inherited by `pthread_create`, so tokio worker and blocking-pool threads pick it up; the explicit hook in `on_thread_start` is belt-and-suspenders for late-attached blocking threads.

No behavior change on Linux/Windows — `qos::set_user_initiated()` is a `const fn` no-op there, and the `libc` dep is `cfg(target_os = \"macos\")`-gated.

## Why

On Apple Silicon (M-series), XNU routes work between performance (P) and efficiency (E) cores partly based on QoS. The default QoS for a build daemon is `UTILITY` or below, which the scheduler can park on E-cores. Single-thread-bursty workloads — SwiftCompile is the canonical example — can run 2–3× slower on an E-core than a P-core.

`USER_INITIATED` is a **scheduler hint**, not a hard P-core pin (there is no public API for hard pinning, and `USER_INTERACTIVE` is reserved for UI work). Under saturation the scheduler may still spill to E-cores, but when capacity exists the bias clearly improves wall time for compile-heavy actions on macOS RBE workers.

## Files touched

- \`nativelink-worker/src/qos.rs\` (new, 152 lines) — macOS impl via \`libc::pthread_set_qos_class_self_np\`, non-macOS const-fn no-op, two macOS tests + one cross-platform compile-time test
- \`nativelink-worker/src/lib.rs\` — \`pub mod qos\`
- \`nativelink-worker/src/local_worker.rs\` — call at top of \`LocalWorker::run\`
- \`nativelink-worker/Cargo.toml\` — \`[target.'cfg(target_os = \"macos\")'.dependencies] libc = \"0.2\"\`
- \`nativelink-worker/BUILD.bazel\` — register \`src/qos.rs\` and macOS-only libc dep via \`select()\`
- \`src/bin/nativelink.rs\` — main-thread call before \`tokio::runtime::Builder::build\`, plus \`on_thread_start\` hook on the runtime builder

## Test plan

- [x] \`cargo test -p nativelink-worker --lib\` — 3 passed
  - \`qos::macos_tests::sets_user_initiated_on_current_thread\` — round-trips via \`pthread_get_qos_class_np\` to confirm the kernel accepted the class (not just that the syscall returned 0)
  - \`qos::macos_tests::tokio_worker_threads_inherit_user_initiated_via_on_thread_start\` — builds a fresh \`new_multi_thread\` runtime with the same hook used in \`main\`, forces scheduling onto a worker thread, reads back the class. This verifies the load-bearing claim that on_thread_start actually delivers USER_INITIATED to worker threads.
  - Cross-platform: \`#[cfg(not(target_os = \"macos\"))]\` no-op compile-time anchor
- [x] \`cargo build -p nativelink-worker --bin nativelink\` clean
- [x] \`cargo clippy -p nativelink-worker --lib -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean

## Provenance

Equivalent to commit \`0fce813bc\` from #2243, ported atomically. The 156-commit parent PR bundles many unrelated changes; this isolates a single change that (a) is macOS-only and (b) attacks the dominant SwiftCompile wall-time bottleneck on Apple Silicon RBE fleets.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2342)

### DIFF
--- a/nativelink-worker/BUILD.bazel
+++ b/nativelink-worker/BUILD.bazel
@@ -13,6 +13,7 @@ rust_library(
         "src/directory_cache.rs",
         "src/lib.rs",
         "src/local_worker.rs",
+        "src/qos.rs",
         "src/running_actions_manager.rs",
         "src/worker_api_client_wrapper.rs",
         "src/worker_utils.rs",
@@ -51,6 +52,7 @@ rust_library(
         "@crates//:uuid",
     ] + select({
         "@platforms//os:linux": ["@crates//:libc"],
+        "@platforms//os:macos": ["@crates//:libc"],
         "//conditions:default": [],
     }),
 )
@@ -98,6 +100,7 @@ rust_test_suite(
         "@crates//:which",
     ] + select({
         "@platforms//os:linux": ["@crates//:libc"],
+        "@platforms//os:macos": ["@crates//:libc"],
         "//conditions:default": [],
     }),
 )
--- a/nativelink-worker/Cargo.toml
+++ b/nativelink-worker/Cargo.toml
@@ -58,6 +58,9 @@ uuid = { version = "1.16.0", default-features = false, features = [
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = { version = "0.2.183", default-features = false }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+libc = { version = "0.2.183", default-features = false }
+
 [dev-dependencies]
 nativelink-macro = { path = "../nativelink-macro" }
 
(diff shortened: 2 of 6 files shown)


MarcusSorealheis (approved):
LGTM

PR #2344, opened 2026-05-18
Add cache metrics

MarcusSorealheis:
# Description

Re-adds cache metrics to the metrics implementation via a store wrapper that does not exist unless enabled.

## Type of change

## How Has This Been Tested?

```//nativelink-store:integration_tests/cache_metrics_store_test_test```

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2344)

### CODE REVIEW

File: nativelink-store/src/cache_metrics_store.rs [hunk 1]
```diff
@@ -0,0 +1,317 @@
...
+    ) -> Result<u64, Error> {
+        self.backend
+            .list(
+                (
+                    range.0.map(StoreKey::into_owned),
+                    range.1.map(StoreKey::into_owned),
+                ),
+                handler,
+            )
+            .await
```
palfrey:
```suggestion
        self.backend.list(range, handler).await
```
is the equivalent, but simpler

palfrey (changes requested):
One minor change, and I'm also seeing a pile of https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#single_match_else warnings which we should probably either fix or explicitly ignore

palfrey (approved):

PR #2345, opened 2026-05-18
Replace bincode with wincode

palfrey:
# Description

Fixes https://rustsec.org/advisories/RUSTSEC-2025-0141.html as bincode is no longer maintained

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2345)


nativelink-macro/Cargo.toml
```diff
@@ -12,4 +12,4 @@ proc-macro = true
 [dependencies]
 proc-macro2 = { version = "1.0.94", default-features = false }
 quote = { version = "1.0.40", default-features = false }
-syn = { version = "2.0.100", default-features = false }
+syn = { version = "^2.0.117", default-features = false }
```
MarcusSorealheis: Why not pin to a specific version here?

nativelink-util/Cargo.toml
```diff
@@ -93,6 +93,7 @@ uuid = { version = "1.16.0", default-features = false, features = [
   "v6",
 ] }
 walkdir = { version = "2.5.0", default-features = false }
+wincode = { version = "0.5.4", default-features = false, features = ["derive"] }
```
MarcusSorealheis, line 97: If this is pinned, which is good, can't we pin the deps?

nativelink-macro/Cargo.toml
palfrey: Not sure why I did this one, I think I thought for a moment that was the default pin format. I've fixed this now.

MarcusSorealheis (approved):
LGTM

PR #2346, opened 2026-05-18
worker: fix two silent output-corruption bugs (absolute symlinks + zero-digest files)

erneestoc:
# Description

Two correctness fixes on the worker's output-upload and input-materialization paths. Both are silent-corruption bugs — actions complete "successfully" but the output artifacts the client receives are wrong or missing.

### Absolute symlinks in output upload (`running_actions_manager.rs::inner_upload_results`)

When a Bazel action emits an absolute symlink as one of its declared outputs (e.g. `output_link → /usr/lib/libfoo.so` or `→ /tmp/nativelink/cache/...`), the pre-fix worker uploads the **literal target path string** to the AC. The client (or a different worker that downloads the AC result) follows that symlink on its own filesystem and finds nothing — the worker-local path doesn't exist on the requester.

The fix branches on `Path::is_absolute()` in the symlink handler. Absolute targets are resolved — directory targets get uploaded as Tree protos via `upload_directory` + `serialize_and_upload_message`, file targets via `upload_file`. Relative symlinks fall through to the existing `upload_symlink` codepath unchanged (relative symlinks travel correctly with the tree).

This aligns the actual worker behavior with our advertised `symlink_absolute_path_strategy: Disallowed` in CacheCapabilities.

### Zero-digest files missing from worker exec dirs (`FilesystemStore::get_file_entry_for_digest`)

The pre-fix function returned a synthetic `FileEntry` for the zero digest pointing at a content path that never exists on disk (zero-byte files aren't stored — nothing to store). Downstream hardlink calls failed silently — the zero-byte file went missing from the action's output tree. Swift's incremental compile produces empty `.o` files routinely; same for `.swiftdoc` placeholders.

Fix: return `Code::NotFound`. Callers must take the explicit empty-file path (`fs::create_file` + `write_all(&[])`). Companion change in `download_to_directory` wraps the zero-digest branch with `err_tip` so any future failures are diagnosable rather than silent.

## Provenance

Equivalent to upstream commits `08071533b` and `19d7e2052` from #2243, ported atomically to current main. A third upstream commit (`cf801c2b`, fix `spawn_upload_to_remote` missing output dir blobs) was intentionally not ported because it patches a function that doesn't exist on main; that bug is latent until the deferred fast/slow upload split lands.

## Non-overlap with #2338

PR #2338 (`8051ca9e`) fixed zero-byte handling in the **DirectoryCache** (input-fetch materialization path). This PR fixes the related bug in **FilesystemStore + running_actions_manager** (output-materialization + worker exec-dir path). Different stores, different code paths, different test files.

## Note on backward compatibility

Not a breaking change in the semver sense (no API surface change, no config surface change). However, the wire shape of `ActionResult` changes for actions that produce absolute symlinks: those entries now appear in `output_files` / `output_folders` (resolved content) rather than `output_file_symlinks` (worker-local target). This brings actual behavior into compliance with our advertised `symlink_absolute_path_strategy: Disallowed`. Clients depending on the prior (contract-violating) behavior will see different output entries; the byte semantics are equivalent.

## Type of change

## How Has This Been Tested?

- `cargo build -p nativelink-store -p nativelink-worker` clean
- `cargo test -p nativelink-store --test filesystem_store_test` — 22 passed
- `cargo test -p nativelink-worker --test running_actions_manager_test` — 33 passed
- `cargo clippy -p nativelink-store --lib --tests -- -D warnings` clean (2 pre-existing clippy errors on main in newly-merged `r2_store.rs` confirmed reproducible on plain `origin/main`)
- `cargo clippy -p nativelink-worker --lib -- -D warnings` clean
- `cargo fmt --check` clean for both packages

New tests:
- `upload_absolute_symlink_resolves_contents` — symlink-to-file resolution branch
- `upload_absolute_symlink_to_directory_uploads_tree` — symlink-to-directory branch, asserts Tree proto upload + nested blob in CAS
- `download_to_directory_zero_digest_empty_file_test` — extended with sibling + nested-in-subdir zero-digest files to exercise the recursive caller path
- `get_file_entry_for_zero_digest_returns_not_found` — FilesystemStore API contract

Updated `upload_dir_and_symlink_test` — retains the `ln -s /dev/null empty_sym` fixture; assertion updated to expect the now-resolved-to-empty-file behavior (canonical empty digest `e3b0c44...`/0).

## Checklist

- [ ] Updated documentation if needed (no public API change; in-code doc comments updated)
- [x] Tests added/amended
- [ ] `bazel test //...` passes locally (verified via `cargo` only; bazel-aspect clippy clean for changed files per grep against `clippy.toml` disallowed-methods)
- [ ] PR is contained in a single commit (multiple atomic commits + review-feedback follow-ups; happy to squash before merge)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2346)

### DIFF
--- a/nativelink-worker/tests/running_actions_manager_test.rs
+++ b/nativelink-worker/tests/running_actions_manager_test.rs
@@ -1180,15 +1180,6 @@ mod tests {
         let root_action_directory = make_temp_path("root_action_directory");
         fs::create_dir_all(&root_action_directory).await?;
 
-        // Out-of-tree payload that the action will reference via an
-        // *absolute* symlink. Exercises the "resolve absolute symlink and
-        // upload contents" code path alongside the relative-symlink
-        // preservation path below.
-        let external_root = make_temp_path("upload_dir_and_symlink_external");
-        fs::create_dir_all(&external_root).await?;
-        let external_file = format!("{external_root}/empty_payload");
-        tokio::fs::write(&external_file, b"").await?;
-
         let running_actions_manager = Arc::new(RunningActionsManagerImpl::new_with_callbacks(
             RunningActionsManagerArgs {
                 root_action_directory,
@@ -1218,22 +1209,24 @@ mod tests {
                 arguments: vec![
                     "sh".to_string(),
                     "-c".to_string(),
-                    format!(
-                        "mkdir -p dir1/dir2 && \
+                    "mkdir -p dir1/dir2 && \
                          echo foo > dir1/file && \
                          touch dir1/file2 && \
                          ln -s ../file dir1/dir2/sym && \
                          ln -s dir1/file rel_sym && \
-                         ln -s {external_file} empty_sym",
-                    ),
+                         ln -s /dev/null empty_sym"
+                        .to_string(),
                 ],
                 // `dir1` exercises the directory upload path,
                 // `rel_sym` exercises the relative-symlink-preserved path,
                 // `empty_sym` exercises the absolute-symlink-resolved path
-                // (previously this test asserted `empty_sym` was kept as a
-                // `SymlinkInfo` with target `/dev/null`; that behavior is
-                // now incorrect because absolute symlinks are worker-local
-                // and must be resolved before upload).
+                // against `/dev/null`. Pre-fix this test asserted `empty_sym`
+                // was kept as a `SymlinkInfo` with target `/dev/null`; that
+                // behavior is now incorrect because absolute symlinks are
+                // worker-local and must be resolved before upload. Reading
+                // `/dev/null` returns 0 bytes immediately by its character-
+                // device contract, so the worker produces an empty-file
+                // output with the canonical sha256 empty digest.
                 output_paths: vec![
                     "dir1".to_string(),
                     "empty_sym".to_string(),
--- a/nativelink-worker/tests/running_actions_manager_test.rs
+++ b/nativelink-worker/tests/running_actions_manager_test.rs
@@ -1087,11 +1087,13 @@ mod tests {
                         "echo foo > dir1/file && ",
                         "touch dir1/file2 && ",
                         "ln -s ../file dir1/dir2/sym &&",
-                        "ln -s /dev/null empty_sym",
+                        // Relative symlink — preserved as a symlink in the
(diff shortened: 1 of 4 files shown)


nativelink-worker/tests/running_actions_manager_test.rs
```diff
@@ -1475,6 +1526,161 @@ mod tests {
         Ok(())
     }
 
+    // Windows does not support symlinks.
```
palfrey: Not true. See https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_file.html / https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_dir.html

nativelink-worker/tests/running_actions_manager_test.rs
```diff
@@ -1087,11 +1087,13 @@ mod tests {
                         "echo foo > dir1/file && ",
                         "touch dir1/file2 && ",
                         "ln -s ../file dir1/dir2/sym &&",
-                        "ln -s /dev/null empty_sym",
```
palfrey, line 1090: Why remove the `empty_sym` case?

nativelink-worker/tests/running_actions_manager_test.rs
```diff
@@ -1249,6 +1251,154 @@ mod tests {
         Ok(())
     }
 
+    // Windows does not support symlinks.
```
palfrey: Again, not true.

palfrey (changes requested):

nativelink-worker/tests/running_actions_manager_test.rs
erneestoc: I forget to verify this part, wasn't entirely sure to post it - fixed

nativelink-worker/tests/running_actions_manager_test.rs
erneestoc: added back in here https://github.com/TraceMachina/nativelink/pull/2346/commits/2d368283abb378a942f5977473aafdd800c6bae5

palfrey (approved):

PR #2347, opened 2026-05-18
directory_cache: fix CAS inode corruption from chmod-during-eviction

erneestoc:
# Description

`DirectoryCache::evict_lru` calls `set_readwrite_recursive` to make the cache entry writable before `remove_dir_all`. That helper chmods both directories AND files. **The files in the cache entry are hardlinked from the CAS** — so chmod follows the hardlink and mutates the CAS file's mode mid-flight.

End result: a cached executable like `cc_wrapper.sh` with CAS mode `0o555` gets silently clobbered to `0o644` when an unrelated DirectoryCache eviction runs. The next action that materializes that CAS file hits `EACCES` on exec. The symptom looks like a totally unrelated bug because the eviction and the failing action are in different code paths.

Fix: introduce `set_dir_writable_recursive` in `nativelink-util/src/fs_util.rs` that chmods **only directories** (which is all you need to enable `unlink` on the files inside), and swap `evict_lru` to use it.

## Why this matters

This is **silent data corruption** that may be happening in any deployment running the DirectoryCache today. The symptom (EACCES from a cached file) presents as "my toolchain wrapper script suddenly stopped working" — operators correlate it with whatever they last touched, not with the eviction path of an unrelated action.

## Relationship to in-flight PRs

The two existing callers of `set_readwrite_recursive` are being replaced by safer variants across two PRs:

- **This PR (#2347)**: replaces the `DirectoryCache::evict_lru` caller with `set_dir_writable_recursive`
- **#2349 (skip chmod walk)**: replaces the `hardlink_directory_tree` caller (introduced by #2338) with `chmod_dir_writable`

After both land, `set_readwrite_recursive` and its private helper `set_readwrite_one_path` will have zero callers and can be removed in a small follow-up PR.

## Provenance

Equivalent to upstream commit `a47774d544` from #2243. Adapted for our codebase: the original commit was ~32 LOC removing a shell-exec `chmod -R u+rw`; our codebase uses an in-Rust helper, so the port had to be reshaped to chmod only directories (preserving `remove_dir_all`'s ability to unlink files inside) rather than skipping chmod entirely.

## Type of change

## How Has This Been Tested?

Added `test_eviction_cleanup_preserves_hardlinked_file_mode` in `nativelink-worker/src/directory_cache.rs`. The test creates a file with mode `0o444`, hardlinks it into a cache entry (simulating CAS sharing), runs the eviction path, and asserts the original file's mode is unchanged.

**Empirical FAIL-at-HEAD~1 / PASS-at-HEAD proof:** locally swapped the fix back to call `set_readwrite_recursive` and re-ran — test fails with `assertion left == right failed: eviction cleanup mutated the inode mode (was 0o444, now 0o644)`. Restored the fix → passes.

- `cargo build -p nativelink-util -p nativelink-worker` clean
- `cargo test -p nativelink-worker --lib directory_cache::` 2/2 pass (3/3 post-rebase, includes upstream's `test_directory_cache_zero_byte_file`)
- `cargo clippy -p nativelink-worker --lib --tests -- -D warnings` clean for changed files
- `cargo fmt --check` clean
- Disallowed-methods grep against `clippy.toml` — no matches in diff

## Checklist

- [ ] Updated documentation if needed (in-code doc comments updated; no public API change)
- [x] Tests added/amended
- [ ] `bazel test //...` passes locally (verified via `cargo` only)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2347)

### DIFF
--- a/nativelink-util/src/fs_util.rs
+++ b/nativelink-util/src/fs_util.rs
@@ -286,6 +286,28 @@ pub async fn set_readwrite_recursive(dir: &Path) -> Result<(), Error> {
     set_perms_recursive_impl(dir.to_path_buf(), set_readwrite_one_path).await
 }
 
+/// Sets only the **directories** in a tree to writable for the current user,
+/// leaving files untouched. This is the safe variant for cleanup paths that
+/// need to delete a tree containing CAS-hardlinked files.
+///
+/// On unix, write permission on the parent directory is sufficient to unlink
+/// files inside it — the files' own modes are irrelevant for unlinking. Chmoding
+/// a CAS-hardlinked file would silently mutate the shared inode's permissions
+/// for every other in-flight action that has hardlinked the same blob, leading
+/// to EACCES on exec or EPERM on open in unrelated actions.
+///
+/// # Arguments
+/// * `dir` - Directory whose directories should be made writable
+///
+/// # Platform Notes
+/// - Unix: Sets directory permissions to 0o755 (rwxr-xr-x); files are NOT touched.
+/// - Windows: Clears `FILE_ATTRIBUTE_READONLY` on directories only; files are NOT touched.
+pub async fn set_dir_writable_recursive(dir: &Path) -> Result<(), Error> {
+    error_if!(!dir.exists(), "Directory does not exist: {}", dir.display());
+
+    set_perms_recursive_impl(dir.to_path_buf(), set_dir_writable_one_path).await
+}
+
 fn set_readonly_one_path(
     path: PathBuf,
     metadata: Metadata,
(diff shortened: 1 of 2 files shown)


nativelink-util/src/fs_util.rs
palfrey, line 1: `set_readwrite_recursive` and `set_readwrite_one_path` can be dropped as part of this, as the only user of them is removed in this PR

palfrey (dismissed):
Ouch, nasty one, good catch! One minor item, otherwise good.

MarcusSorealheis:
🔥🔥🔥

MarcusSorealheis:
@erneestoc the generated descriptions from Claude did not specify whether or not this is a breaking change which is important for our Changelog generation. I believe this is only a bugfix but it would be good to include the PR template and wrap all of your Claude info in the Description

MarcusSorealheis:
Nevertheless, fantastic PR and than you.

nativelink-util/src/fs_util.rs
erneestoc: after the rebase onto current main (which now includes https://github.com/TraceMachina/nativelink/pull/2338), set_readwrite_recursive has a new caller at fs_util.rs:104 (the post-clonefile chmod walk). https://github.com/TraceMachina/nativelink/pull/2349 will replace that caller with a different helper, and then both helpers genuinely become dead code

MarcusSorealheis: I think we can resolve here

MarcusSorealheis (approved):
LGTM

nativelink-util/src/fs_util.rs
erneestoc: cleaned up in here https://github.com/TraceMachina/nativelink/pull/2349

PR #2352, opened 2026-05-20
simple_scheduler_state_manager: count worker disconnects toward max_job_retries

amankrx:
# Description

Avoid worker disconnects that keeps on running the build forever. 

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Added tests and tested it in a working build environment. 

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2352)

### DIFF
--- a/nativelink-scheduler/src/simple_scheduler_state_manager.rs
+++ b/nativelink-scheduler/src/simple_scheduler_state_manager.rs
@@ -785,7 +785,30 @@ where
                         ActionStage::Queued
                     }
                 }
-                UpdateOperationType::UpdateWithDisconnect => ActionStage::Queued,
+                UpdateOperationType::UpdateWithDisconnect => {
+                    awaited_action.attempts += 1;
+                    if awaited_action.attempts > self.max_job_retries {
+                        ActionStage::Completed(ActionResult {
+                            execution_metadata: ExecutionMetadata {
+                                worker: maybe_worker_id
+                                    .map_or_else(String::default, ToString::to_string),
+                                ..ExecutionMetadata::default()
+                            },
+                            error: Some(make_err!(
+                                Code::Aborted,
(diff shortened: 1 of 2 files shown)


MarcusSorealheis (approved):
LGTM

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -785,7 +785,30 @@ where
...
+                    awaited_action.attempts += 1;
+                    if awaited_action.attempts > self.max_job_retries {
+                        ActionStage::Completed(ActionResult {
+                            execution_metadata: ExecutionMetadata {
+                                worker: maybe_worker_id
+                                    .map_or_else(String::default, ToString::to_string),
+                                ..ExecutionMetadata::default()
+                            },
+                            error: Some(make_err!(
+                                Code::Aborted,
```
MarcusSorealheis, line 798: I was re-reading this one. This is the wrong error code. `Code::Aborted` is for concurrency/transaction conflicts. For an action that failed because the worker repeatedly died running it, the right code is `INTERNAL` (the safe generic).

If we could verify what the failure mode was, like OOM or something else, we could get more specific like `RESOURCE_EXHAUSTED`.

MarcusSorealheis (changes requested):

nativelink-scheduler/src/simple_scheduler_state_manager.rs
MarcusSorealheis: Can check here for reference: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

amankrx:
Rmote Persistent Workers PR included this change. So, closing this PR now.

MarcusSorealheis:
True good call

PR #2354, opened 2026-05-20
redis_store: raise check_health PING ceiling from 2s to 4s

amankrx:
# Description

The previous 2-second PING ceiling was tight enough that a routine Redis BGSAVE fork would push every RedisStore indicator over the line simultaneously: on an 11 GB production master under load we observe fork-induced pauses around 3 seconds, and with three RedisStore indicators (AC, CAS, scheduler) all PING-ing through the same connection pool, all three return Failed in lockstep — surfacing as a 503 on /status and a kubelet probe-failure event even though the Redis service is otherwise healthy.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Tested in an actual working configuration

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2354)

### DIFF
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -87,6 +87,10 @@ const DEFAULT_CONNECTION_TIMEOUT_MS: u64 = 3000;
 /// Note: If this changes it should be updated in the config documentation.
 const DEFAULT_COMMAND_TIMEOUT_MS: u64 = 10_000;
 
+/// The default `check_health` PING ceiling in milliseconds if not specified.
+/// Note: If this changes it should be updated in the config documentation.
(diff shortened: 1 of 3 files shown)


nativelink-store/src/redis_store.rs
```diff
@@ -1118,7 +1118,7 @@ where
         /// under `HealthServer`'s default per-indicator budget;
         /// loose enough to absorb a normally-slow PING during a
         /// `BGSAVE` fork or sentinel rebalance.
-        const PING_TIMEOUT: Duration = Duration::from_secs(2);
+        const PING_TIMEOUT: Duration = Duration::from_secs(4);
```
palfrey: This should be a config value

MarcusSorealheis (approved):
LGTM

PR #2357, opened 2026-05-22
worker: make directory-cache entries already-writable

erneestoc:
## Problem

`DirectoryCache` locks each cache entry down with `set_readonly_recursive` after
construction. That helper previously made the entire entry tree mode `0o555` —
directories included — so every materialization had to follow up with a separate
`set_dir_writable_recursive` recursive `chmod` walk in `prepare_action_inputs` to
re-add write permission to directories (Bazel actions declare outputs at paths
nested inside input subdirectories).

That per-materialization post-walk is redundant work.

## Fix

Directories are not hardlink-shared between cache entries — only file *content*
inodes are — so directory mode can safely be set once, at the cache entry,
instead of on every materialization.

`set_readonly_recursive` now makes only **files** read-only (`0o555`) and leaves
**directories** writable (`0o755`). Both materialization paths then produce a
directly-usable tree:

- macOS `clonefile(2)` copies the source's modes verbatim → writable dirs,
  read-only files.
- The Linux per-file hardlink walk creates fresh (writable) directories and
  hardlinks files (which keep the source inode's read-only mode).

Files stay read-only on both paths, so the hermeticity contract and the
CAS-hardlink shared-inode invariant (#2347) are preserved. The now-redundant
`set_dir_writable_recursive` call is removed from `prepare_action_inputs`;
`set_dir_writable_recursive` itself is unchanged and still used by the cache
eviction cleanup path.

## Testing

- `fs_util`: `test_set_readonly_recursive` now also asserts directories stay
  writable; the macOS clonefile tests assert cloned subdirs are writable and
  that a nested output file can be created with no `set_dir_writable_recursive`
  walk.
- `directory_cache`: new `test_materialized_tree_dirs_writable_files_readonly`
  asserts that after `get_or_create` on both the fresh-materialize and
  cache-hit paths, every directory is writable and every file read-only.
- `bazel test //nativelink-worker/... //nativelink-util/...` — 27/27 pass;
  clippy + rustfmt aspects clean.

## Note

Companion to #2358 (`util: make permission walks symlink-safe`). Both PRs touch
`nativelink-util/src/fs_util.rs` (`hardlink_directory_tree_recursive`);
whichever lands second needs a trivial rebase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2357)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2828,7 +2828,7 @@ checksum = "1d87ecb2933e8aeadb3e3a02b828fed80a7528047e68b4f424523a0981a3a084"
 
 [[package]]
 name = "nativelink"
-version = "1.3.0"
+version = "1.3.1"
 dependencies = [
  "async-lock",
  "axum",
@@ -2859,7 +2859,7 @@ dependencies = [
 
 [[package]]
 name = "nativelink-config"
-version = "1.3.0"
+version = "1.3.1"
 dependencies = [
  "byte-unit",
  "humantime",
@@ -2877,7 +2877,7 @@ dependencies = [
 
 [[package]]
 name = "nativelink-error"
-version = "1.3.0"
+version = "1.3.1"
 dependencies = [
  "mongodb",
  "nativelink-metric",
@@ -2899,7 +2899,7 @@ dependencies = [
 
 [[package]]
 name = "nativelink-macro"
-version = "1.3.0"
+version = "1.3.1"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2908,7 +2908,7 @@ dependencies = [
 
 [[package]]
 name = "nativelink-metric"
-version = "1.3.0"
+version = "1.3.1"
 dependencies = [
  "async-lock",
  "nativelink-metric-macro-derive",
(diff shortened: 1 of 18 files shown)


MarcusSorealheis:
@erneestoc I need time to test it. Also, could you increment the versions of all the top-level Cargo.toml's in line with the instructions in [CONTRIBUTING.md#creating-releases](https://github.com/TraceMachina/nativelink/blob/ce7bbe7beab42f784977bb3190d1f0770757e21d/CONTRIBUTING.md#creating-releases). We are basically going to release patch versions a lot more frequently to tighten the feedback loop.

MarcusSorealheis:
Also, for the past two versions Claude Code has been able to follow the instructions quite well so you could hand it off there.

erneestoc:
@MarcusSorealheis bumped the workspace version `1.3.0` → `1.3.1` in `MODULE.bazel`, the root `Cargo.toml`, and every `nativelink-*` crate manifest, per [CONTRIBUTING.md#creating-releases](https://github.com/TraceMachina/nativelink/blob/main/CONTRIBUTING.md#creating-releases) step 1; `Cargo.lock` refreshed for the 13 workspace members. The branch had already picked up the `1.3.0` release via the `main` merge, so this is the patch bump on top. Bazel build is green.

MarcusSorealheis:
this one was merged by @2362, or will be.

MarcusSorealheis:
merged as a part of #2366

PR #2358, opened 2026-05-22
util: make permission walks symlink-safe

erneestoc:
## Problem

The recursive permission walk `set_perms_recursive_impl` (which drives both
`set_readonly_recursive` and `set_dir_writable_recursive`) used `fs::metadata`
(`stat`), which **follows symlinks**. On input trees containing symlinks — e.g.
`.venv/bin/python3` produced by `rules_python` / `rules_apple` venv tooling —
this has two failure modes:

- A symlink to a directory reports `is_dir() == true`, so the walk recurses
  *through* the link, escaping the materialized tree or descending into an
  unrelated directory.
- A symlink is passed to `set_permissions`; `chmod` follows symlinks, so it
  mutates the link's *target*. When the target does not exist (a dangling link
  — common when a venv points outside the action's input set), the `chmod`
  returns `ENOENT` and fails the entire walk.

That `ENOENT` failure surfaces as `set_readonly_recursive` erroring inside
`DirectoryCache::get_or_create`, which makes `prepare_action_inputs` log
`"Directory cache failed, falling back to traditional download"` and take the
slow `download_to_directory` path.

## Fix

`set_perms_recursive_impl` now uses `symlink_metadata` (`lstat`) and returns
early on symlink entries — it never `chmod`s a symlink and never recurses
through one. Regular files keep their existing read-only (`0o555`) treatment,
so the CAS-hardlinked-inode hermeticity contract (#2347) is unchanged.

`hardlink_directory_tree_recursive` already recreated symlinks as symlinks; its
symlink branch is reordered ahead of the `is_dir()` / `is_file()` branches to
make the symlink-first intent explicit and robust.

## Testing

Adds regression tests covering the set-readonly, set-dir-writable, and
hardlink/clone walks over a tree containing a symlink to an in-tree file, a
dangling relative symlink, and a symlink to an in-tree directory — asserting
each walk succeeds and the symlinks are preserved with their targets intact.

`bazel test //nativelink-util/... //nativelink-worker/...` — 27/27 pass;
clippy + rustfmt aspects clean.

## Note

Companion to #2357 (`worker: make directory-cache entries already-writable`).
Both PRs touch `nativelink-util/src/fs_util.rs`
(`hardlink_directory_tree_recursive`); whichever lands second needs a trivial
rebase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2358)

### DIFF
--- a/nativelink-util/src/fs_util.rs
+++ b/nativelink-util/src/fs_util.rs
@@ -226,31 +226,21 @@ fn hardlink_directory_tree_recursive<'a>(
             })?;
 
             let dst_path = dst.join(&file_name);
+            // `DirEntry::metadata` does NOT traverse symlinks (it has
+            // `symlink_metadata`/lstat semantics), so `is_symlink()` below
+            // correctly identifies symlink entries and the symlink branch
+            // recreates them as symlinks rather than dereferencing them.
             let metadata = entry
                 .metadata()
                 .await
                 .err_tip(|| format!("Failed to get metadata for: {}", entry_path.display()))?;
 
-            if metadata.is_dir() {
-                // Create subdirectory and recurse
-                fs::create_dir(&dst_path)
-                    .await
-                    .err_tip(|| format!("Failed to create directory: {}", dst_path.display()))?;
-
-                hardlink_directory_tree_recursive(&entry_path, &dst_path).await?;
-            } else if metadata.is_file() {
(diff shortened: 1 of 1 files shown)


MarcusSorealheis (approved):
LGTM by itself

PR #2359, opened 2026-05-22
worker: faster directory-cache materialization — hardlink CAS blobs, drop redundant walks, narrow lock

erneestoc:
# Description

Three atomic commits that cut the cost of `DirectoryCache` materialization on
the worker — the path that dominates input-tree materialization wall time.
Part of the input-materialization performance work (related to #2243), kept as
a small standalone change.

On a `DirectoryCache` miss, `construct_directory` rebuilds the entire input
tree: for every file it reads the whole blob into memory (`get_part_unchunked`)
and writes a full byte copy (`fs::write`), then walks the finished tree twice
more (`set_readonly_recursive`, `calculate_directory_size`). The worker's
fallback path, `download_to_directory`, already materializes files by
hardlinking the CAS blob — this brings the directory-cache path in line and
removes the redundant walks.

**`f23dffa7` — hardlink CAS blobs.** `create_file` hardlinks the
`FilesystemStore` CAS blob directly into the cache entry
(`FileEntry::get_file_path_locked` + `fs::hard_link`), mirroring
`download_to_directory`. Eliminates the read-into-memory and the byte copy for
every non-executable file present in the local fast tier. Executable files get
a private inode (fetch + write + chmod) — a hardlinked CAS blob must never be
`chmod`'d, that mutates the shared inode (the invariant fixed in #2347). Blobs
not present in the filesystem tier fall back to fetch + write.
`DirectoryCache::new` now takes `Arc<FastSlowStore>` (was a generic `Store`) to
reach the typed `FilesystemStore` path APIs.

**`9425d4c8` — drop two redundant walks.** Entry size is accumulated from
`FileNode.digest.size_bytes` during construction (removes the
`calculate_directory_size` filesystem walk); file/directory modes are set at
creation time (removes the separate `set_readonly_recursive` post-walk).

**`33d5b1ba` — narrow the lock + single-flight.** The cache `RwLock` write
guard is held only for in-memory map mutations; the clone to the destination
runs unlocked, with the entry pinned via `ref_count` so a concurrent caller
cannot evict it mid-materialize. Eviction's filesystem I/O is moved off the
lock onto a background task. Concurrent callers for the same digest now
construct it once and share the result (single-flight).

Materialized output is byte-identical to before.

## Type of change

  not work as expected)

## How Has This Been Tested?

`bazel test //nativelink-worker/... //nativelink-util/... //nativelink-store/...`
— 51/51 pass (the build also runs the clippy and rustfmt aspects).

New tests in `directory_cache.rs` / `tests/directory_cache_test.rs` cover: a
hardlinked file shares the CAS inode; an executable gets a private inode and
leaves the CAS blob's mode untouched; materialized content round-trips
byte-identical; size accounting from digest sums, including nested trees;
concurrent single-flight (16 simultaneous callers for one digest → exactly 1
construction, 15 hits); cache-entry directory modes.

Beyond the test suite, this was exercised in a real-world application build:
input-materialization time dropped roughly 2x on a clean build with a cold
cache. That is a real-workload measurement rather than a synthetic
microbenchmark — measured clean-build/cold-cache; it should carry over to
real-world incremental builds as well.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2359)

### DIFF
--- a/nativelink-worker/src/directory_cache.rs
+++ b/nativelink-worker/src/directory_cache.rs
@@ -28,6 +28,7 @@ use nativelink_store::ac_utils::get_and_decode_digest;
 use nativelink_store::cas_utils::is_zero_digest;
 use nativelink_store::fast_slow_store::FastSlowStore;
 use nativelink_store::filesystem_store::{FileEntry, FilesystemStore};
+use nativelink_util::background_spawn;
 use nativelink_util::common::DigestInfo;
 use nativelink_util::fs_util::{CloneMethod, hardlink_directory_tree, set_dir_writable_recursive};
 use nativelink_util::store_trait::{StoreKey, StoreLike};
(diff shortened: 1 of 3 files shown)


MarcusSorealheis (approved):

MarcusSorealheis:
Merged as a part of #2362

PR #2362, opened 2026-05-23
Combine 2357 2358 2359

MarcusSorealheis:
# Description
Combines #2357 #2358 #2359 for testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2362)


MarcusSorealheis:
Finally testing after three hours with Mac OS apps compiling on Macs: 

image

MarcusSorealheis:
One test fails for me locally— and it's a genuine PR interaction, not a merge slip: 2359 deletes 2357's set_readonly_recursive post-walk and hardlinks CAS blobs instead, but nativelink-store's FilesystemStore writes blobs at 0o644 (writable) and create_file only chmods to read-only for executables. So combined, materialized input files are writable, violating 2357's hermeticity contract (test_materialized_tree_dirs_writable_files_readonly). Each PR passes its own tests; only combining surfaces it. This needs a design call (make CAS blobs read-only at rest / private read-only inodes / relax the contract) — I did not silently pick one.

I am in favor of CAS blobs read only because the main complaints are hermeticity and the time spent waiting on fetch. I've made that change and a change to a test that now fails because of an attempt to execute stat on an inode after a file had been changed to read only.

MarcusSorealheis:
We have observed some performance gains when building an iOS app, as have members of the community like the co-author of this PR. 

- Before: download_to_directory materialized an executable input by hard_link (then
  chmod). A hardlink opens no writable fd.
  - After this change: executables take the private-copy path — `std::fs::copy(src,
  dst)` in download_to_directory (and `tokio::fs::write in the cache's copy_file_to`). A
  copy opens the destination for writing.

So the change introduces a transient writable fd to executables into the worker's hot, highly-concurrent materialization path. That is a classic `ETXTBSY` [trigger from the early days of Bazel](https://github.com/bazelbuild/bazel/issues/614#issuecomment-168585011), because the worker runs many actions concurrently and spawns subprocesses with fork+execve `tokio::process … .spawn()` ~ line 1331.

I asked the native GitHub clanker to break it down, and it missed the plot hard. So I went to Claude for the breakdown of their own handy work and got this:

  1. Action A is mid-std::fs::copy of builder → a writable fd to that inode is open.
  2. Action B forks for its own spawn during that window → B's child inherits A's
  still-open writable fd (fork copies all fds; O_CLOEXEC only closes it at the child's
  own execve, not before).
  3. Action A (or anyone hardlinked to that inode) execves builder before B's child
  execs → ETXTBSY.

  The old hardlink path never opened a writer fd for executables, so it couldn't hit
  this. Your change traded the #2347 shared-inode-chmod problem for a new
  write-fd-in-the-hot-path problem. That's why it's intermittent and why it shows up
  under the heavy parallelism of bazel test //... with rules_go's builder.

  Caveats (so you can confirm rather than assume):
  - ETXTBSY with the rules_go builder is also a well-known flake in high-parallelism
  Bazel builds independent of NativeLink. So: does it reproduce on re-run? Is it
  deterministic? Does it appear on the combine branch but not on plain main? If it's
  branch-specific and recurs, that's strong confirmation it's the change.
  - The directory-cache path is less suspect than download_to_directory: there the copy
  happens once during construction (fd closed before the entry is inserted) and
  materialization is hardlink-only. The fresh per-action std::fs::copy in
  download_to_directory is the prime suspect.

I'll continue pursue a patch after releasing this version because it is also working well.

MarcusSorealheis:
closing in favor of #2366 , which improves upon the optimizations here with some stability improvements.

PR #2373, opened 2026-05-29
Fix fast-slow store optimisation

chrisstaite-menlo:
# Description

The PR #2343 updated the fast-slow store such that it ignored files missing in the slow store for the `has` calls and also said that a file was in the slow store while it was still being uploaded to it.

This causes significant regressions to the function of the fast-slow store such that files missing in the slow store do not get re-uploaded when they have been purged from an upstream cache and also starts actions downloading files from the slow store before they've been committed.

Here we fix both of those by reverting the logic and putting back the regression test for the former issue and also adding in a async Mutex which only reports the slow store upload after it's completed.  In order to perform this correctly the store trait is updated to return the uploaded size in the successful state.

## Type of change

## How Has This Been Tested?

Updated the tests.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2373)

### DIFF
--- a/nativelink-service/src/bytestream_server.rs
+++ b/nativelink-service/src/bytestream_server.rs
@@ -564,6 +564,7 @@ impl ByteStreamServer {
                 // Bytestream always uses digest size as the actual byte size.
                 .update(digest, rx, UploadSizeInfo::ExactSize(digest.size_bytes()))
                 .await
+                .map(|_| ())
         });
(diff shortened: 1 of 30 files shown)


MarcusSorealheis (approved):
LGTM

PR #2390, opened 2026-06-02
health_utils: run indicator checks in parallel, not serially

amankrx:
# Description

Each indicator's check_health does real I/O against its store (write/has/read roundtrip per StoreDriver::check_health), so iterating indicators serially makes the total response time N * timeout. Under load that easily exceeds Kubernetes liveness probe budgets, causing kubelet to kill an otherwise-healthy pod whose only sin was that one congested store made the probe handler queue behind it. Parallel execution caps the total at ~timeout regardless of how many stores are registered.

The order of the resulting descriptions is not part of the API contract; collect-into-Vec callers already ignore order.

Two new tests pin the behavior:

- indicator_checks_run_in_parallel: all indicators rendezvous on a Barrier inside check_health, which can only complete if every check is in flight at the same time. Serial execution would block at the barrier until the per-indicator timeout fires.
- stuck_indicators_time_out_concurrently: with paused tokio time, a report over three never-completing indicators must finish in ~one timeout of virtual time, not three back to back.

The second test needs tokio's test-util feature (start_paused), added as a dev-dependency the same way nativelink-service already declares it for cas_server_test.

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)]

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2390)


MarcusSorealheis (approved):
I could not find any issues here. I tried to find them. 

I would modify the description to say exactly what the fix is.. Something to the effect of:

Replace `stream::iter(...).then(...)` (serial await) with `buffer_unordered(...) (concurrent await)`, or something to that effect. Really nice job on this PR. We should cut a release ASAP. 

LGTM

MarcusSorealheis:
What is going on with this Vercel check @amankrx. For whatever reason, the web deployment is not working for one environment. I'm trying to debug so i can get this merged. I will cut a release today.

PR #2393, opened 2026-06-02
ft_aggregate: pass explicit TIMEOUT to absorb RediSearch slow scans

amankrx:
# Description

RediSearch's module default (~500 ms) is far too tight for the scheduler's awaited-action index under any meaningful load: queries time out, NativeLink surfaces them as parse errors, and the dedup lookup fails. When dedup fails the scheduler creates a duplicate operation for an action that is already in flight -- observed as "two same actions running on different PRs" with each running the full maxActionExecutingTimeoutS window before completing. Pass an explicit value generous enough to absorb 1M+ document scans on a busy RediSearch instance.

FakeRedisBackend's FT.AGGREGATE handler learns to tolerate the new explicit TIMEOUT clause (skipping it before validating the LOAD args), so the fake keeps working for both the new caller shape and any older one.

Test coverage, all in the existing separate test suites:

  * nativelink-store/tests/redis_store_test.rs pins the exact FT.AGGREGATE argument list (now including TIMEOUT 10000) in test_search_by_index, test_search_by_index_with_sort_key, test_search_by_index_resp3, test_search_by_index_skips_int_from_cursor_read, and the error-path log assertion in test_search_by_index_failure.

  * nativelink-scheduler/tests/redis_store_awaited_action_db_test.rs exercises the fake's TIMEOUT-skipping path end to end through the real ft_aggregate caller; it fails without the FakeRedisBackend change.

Fixes # (issue)

## Type of change

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2393)

### DIFF
--- a/nativelink-store/src/redis_utils/ft_aggregate.rs
+++ b/nativelink-store/src/redis_utils/ft_aggregate.rs
@@ -43,7 +43,7 @@ pub(crate) struct FtAggregateOptions {
 /// lookup fails. When dedup fails the scheduler creates a duplicate
 /// operation for an action that is already in flight — observed as
 /// "two same actions running on different PRs" with each running the
-/// full `maxActionExecutingTimeoutS` window before completing. Pass an
+/// full `max_action_executing_timeout_s` window before completing. Pass an
 /// explicit value generous enough to absorb 1M+ document scans on a
 /// busy `RediSearch` instance.
 const FT_AGGREGATE_TIMEOUT_MS: u64 = 10_000;
--- a/nativelink-store/src/redis_utils/ft_aggregate.rs
+++ b/nativelink-store/src/redis_utils/ft_aggregate.rs
@@ -35,6 +35,19 @@ pub(crate) struct FtAggregateOptions {
     pub sort_by: Vec<String>,
 }
 
+/// Per-query `FT.AGGREGATE` timeout in milliseconds.
+///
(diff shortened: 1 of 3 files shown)


nativelink-store/src/redis_utils/ft_aggregate.rs
```diff
@@ -35,6 +35,19 @@ pub(crate) struct FtAggregateOptions {
...
 
+/// Per-query `FT.AGGREGATE` timeout in milliseconds.
+///
+/// `RediSearch`'s module default (≈500 ms) is far too tight for the
+/// scheduler's awaited-action index under any meaningful load: queries
+/// time out, `NativeLink` surfaces them as parse errors, and the dedup
+/// lookup fails. When dedup fails the scheduler creates a duplicate
+/// operation for an action that is already in flight — observed as
+/// "two same actions running on different PRs" with each running the
+/// full `maxActionExecutingTimeoutS` window before completing. Pass an
```
palfrey: I assume you mean `max_action_executing_timeout_s`?
```suggestion
/// full `max_action_executing_timeout_s` window before completing. Pass an
```

palfrey (changes requested):
Tiny item, otherwise good

palfrey (approved):

PR #2396, opened 2026-06-03
De-duplicate filesystem emplace

palfrey:
# Description

We've seen intermittent hardlink failures on the buildstream runs (e.g. https://github.com/TraceMachina/nativelink/actions/runs/26881965504/job/79284041143?pr=2381) but the failure case was unclear. After some experimentation, I've been able to figure it out, both by adding logging here, and then debugging based on that.

The hardlink code first gets a file entry from the filesystem cache, then hardlinks to it. Most of the time, this succeeds, and the primary previous failure case was "your filesystem cache isn't big enough for a single build run, so it's evicting between getting and linking", which was a config fix. I added a logging helper that dumps the cache status, and established this wasn't the case for the buildstream case (was hitting ~1% of cache usage).

The problem for buildstream is that for some reason a single file (usually a Python library file in the case of buildstream) is being uploaded multiple times, and so we evict the previous instance and put a new one in. Sometimes, if we're very unlucky (~1/3rd of the time on buildstream runs right now), we're doing hardlinking in the middle of that.

This PR checks for duplicate file entries i.e. we're trying to emplace a file with identical contents to the current one, and so therefore we can throw away the new entry. This means the file is always correctly in place on the filesystem, and so the hardlink should no longer fail in the previous way.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2396)


integration_tests/buildstream/buildstream-with-nativelink-test.nix
```diff
@@ -30,7 +30,7 @@ writeShellScriptBin "buildstream-with-nativelink-test" ''
   nativelink_output=$(cat integration_tests/buildstream/nativelink.log)
 
   case $nativelink_output in
-    *"ERROR"* )
+    *\"ERROR\"* )
```
palfrey, line 33: Because the new logs now contain the word error in some file paths, which is fine. We're looking for a log message error, which has quoted `ERROR`. Bash case uses non-escaped quotes to handle the case for matches with spaces, which is not what we want.

amankrx (approved):

MarcusSorealheis (approved):

PR #2381, opened 2026-05-30
nix: replace deprecated `pkgs.system` with `stdenv.hostPlatform.system`

amaanq:
# Description

This pr fixes a deprecation warning that nix emits, as well as updating a package's name as it was renamed

Fixes # (issue)

N/A

## Type of change

Please delete options that aren't relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

I tested and eval'd this locally and it still works

## Checklist

- [ ] Updated documentation if needed
- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [ ] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2381)

PR #2379, opened 2026-05-29
nix: migrate off deprecated crane `stdenv` arg to `stdenvSelector`

amaanq:
# Description

[Crane](https://github.com/ipetkov/crane) deprecated passing `stdenv` into `mkCargoDerivation`, however, since the input of crane here is a bit behind, no warning is emitted yet. I use Nativelink between my own machines, but I set the crane input here to follow a more up-to-date crane, and as such, I get hit with the following warning:

```
evaluation warning: passing in an `stdenv` override into `mkCargoDerivation` is now deprecated.
                    Changing the default `stdenv` must be done at the `craneLib` level like so:

                      craneLib = (inputs.crane.mkLib pkgs).overrideScope (final: prev: {
                        # Set this to the chosen stdenv, e.g. `p.clangStdenv`
                        stdenvSelector = p: p.stdenv;
```

This PR's change is defensive in nature, but I went ahead and updated the crane input as well. If that's undesireable, please let me know and I can revert that, the fix works without bumping crane.

Thanks for the really cool project :)

Fixes # (issue)

(No issue, just something I ran into)

## Type of change

A defensive future-bug fix.

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

I ran this locally, overriding my input with the local path.

## Checklist

- [ ] Updated documentation if needed
- [ ] Tests added/amended
- [ ] `bazel test //...`  passes locally
- [ ] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2379)

PR #2397, opened 2026-06-03
ci(nix): retry transient Bazel external-fetch failures

amaanq:
# Description

GitHub's release CDN intermittently returns 502/403 mid-fetch and Bazel
does not retry HTTP 5xx (upstream issue bazelbuild/bazel#17223), so a
single hiccup aborts the build. This commit wraps the Nix-shell bazel
invocation in a 3-attempt loop that retries only when output matches a
transient-fetch signature.

I'd noticed this failure in both my prior PRs (#2379, #2381), so I figured
adding some kind of retry mechanism would be helpful.

Fixes # (issue)

n/a

## Type of change

## How Has This Been Tested?

I tested the logic against what the gh runner spewed before when a 502 was returned (`ERROR: … GET returned 502 Bad Gateway`). I'd done so by spinning up a local server that returned 502 and pointed Bazel's `http_archive` at it.

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2397)


palfrey:
Thanks. I'd been pinging upstream on this (see https://github.com/bazelbuild/bazel/issues/23334#issuecomment-4343296095), but wasn't expecting anything to get actually fixed so this is a good workaround

palfrey (approved):

PR #2399, opened 2026-06-04
listener: support `IP_FREEBIND` to bind not-yet-assigned addresses

amaanq:
# Description

Binding an address that another daemon brings up after start races interface
setup and fails with `EADDRNOTAVAIL`, crash-looping until it appears.

This commit adds an opt-in `freebind` flag on the HTTP listener that
sets `IP_FREEBIND` before bind, so an address brought up after start
(such as a tailscale IP) can be bound immediately.

Fixes # (issue)

n/a, but my use case is for hosting a nativelink instance over tailscale so any of my machines can pull from the cache. I noticed it crash looping before tailscale came online when rebooting the machine with the instance.

## Type of change

  not work as expected)

## How Has This Been Tested?

I've added a unit test, as well as tested this on my instance. It works as expected without crashing :)

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2399)

### DIFF
--- a/nativelink-config/src/cas_server.rs
+++ b/nativelink-config/src/cas_server.rs
@@ -536,6 +536,12 @@ pub struct HttpListener {
     #[serde(deserialize_with = "convert_string_with_shellexpand")]
     pub socket_address: String,
 
+    /// Allow binding `socket_address` before it is assigned locally.
+    ///
+    /// Default: false
+    #[serde(default)]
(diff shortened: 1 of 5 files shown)


palfrey (approved):

amaanq:
Thank you!! :heart:

PR #2402, opened 2026-06-05
filesystem_store: surface deferred write errors before emplacing files

### LINKED ISSUE #2279: Filesystem Store leaves partially written files at final digest paths when builds are interrupted by ENOSPC
NativeLink's filesystemCAS can leave partially-written or zero-byte files at their final digest paths when a write is interrupted by disk full error (ENOSPC). Once present, these ghosts are never cleaned up and permanently poison all future reads of those digests resulting in continuous build failures.

amaanq:
# Description

Tokio's `File` defers write errors to the next write or flush call, so
`sync_all` never reports them. A final-chunk failure (e.g. `ENOSPC`) was
therefore silently swallowed and the truncated temp file renamed into
the content path, which permanently poisoned that digest.

This commit `flush`es before calling `sync_all` so that the error propagates
and the file never becomes visible.

- Fixes #2279

## Type of change

## How Has This Been Tested?

I've added a unit test, as well as simulated the scenario that this was first reported in. I created a 1 MiB tmpfs (can be done yourself with `unshare -rm` + `mount -t tmpfs -o size=1M`), and observed a 1.1MiB upload leaving a truncated file at the final digest path. With this commit though, that same procedure now properly errors with `No space left on device` and does not write a truncated file.

It's hard to simulate this exact scenario in a unit test, so I decided to instead surface a similar kind of error that is replicable, that being intentionally overflowing the `RLIMIT_FSIZE`. Similar to the `ENOSPC` issue, writing past this number returns `EFBIG`, and also results in a truncated file being written. Naturally, it becomes easy to unit test this :grin:

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2402)


palfrey (approved):
Nice! Tricky one this, good to see a fix.

PR #2408, opened 2026-06-08
bytestream_server: single-acquire active_uploads in create_or_join_upload_stream

gergelytraveltime:
# Description

The Entry::Occupied (UUID-collision) arm acquires
`instance.active_uploads.lock()` twice in the same `let` statement: once as the match scrutinee
(`match instance.active_uploads.lock().entry(uuid_key)`) and once inside the arm
(`let mut active_uploads = instance.active_uploads.lock()`). The inline comment claims "Entry goes out of scope here, releasing the lock" between the two, but Rust's temporary-lifetime rules keep the scrutinee's `MutexGuard` alive until the enclosing `let` statement ends.
`parking_lot::Mutex` is non-reentrant, so the second `.lock()` call parks the calling thread on the same mutex it still holds.

Reproduction: Bazel CI cohort against a single-tier `filesystem` CAS topology (no memory or slow store), single in-cluster runner issuing ~120 concurrent local upload actions, many of them 100 MiB+ blobs. Under that load CAS emits a burst of `Unexpected EOF`, `Stream error at byte …`, and
`Expected WriteRequest (got None)` errors interleaved with the `UUID collision detected` warn from this function, then the process goes silent for several minutes — CPU drops to single- digit milli-cores while the gRPC listener keeps accepting and the pod stays alive. The HTTP `/metrics` endpoint hangs (default `StoreDriver::check_health` writes a probe blob through the same store path) and Bazel clients hit `DEADLINE_EXCEEDED`. The wedge sometimes self-recovers after ~5-10 min and sometimes persists, consistent with a contention-bound async-runtime stall on a hot synchronous mutex rather than a clean every-time deadlock; either way the restructure removes the pattern.

Note: for debugging the liveness probe in the helm chart was changed to that the hanging /metrics endpoint is not restarting CAS.

The collision path has no automated test; the manual-verification note in `bytestream_server_test.rs` exercised it at trivial concurrency, which is why the bug slipped through.

Restructure: bind `active_uploads` once at the top of the block, match on `active_uploads.entry(uuid_key)`, drop the `Entry::Occupied` entry's borrow explicitly before calling `insert` under the unique key, and let the guard drop on scope exit.

Empirical impact under the reproduction config:
- Unpatched: ~18 ByteStream error bursts in <30 s, in-flight upload buffer climbs to ~11 GiB, test wall regression / failure.
- Patched (same config): 0 ByteStream errors, in-flight buffer stays <1 MiB, faster CI wall (CAS CPU shifts from parking on the lock to real work).

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- NativeLink Bazel tests (added a new test case)
- Tested the built image in our installation and it fixed the deadlock/contention scenario

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2408)

### DIFF
--- a/nativelink-service/src/bytestream_server.rs
+++ b/nativelink-service/src/bytestream_server.rs
@@ -494,8 +494,9 @@ impl ByteStreamServer {
         // Parse UUID string to u128 key for efficient HashMap operations
         let uuid_key = parse_uuid_to_key(uuid_str);
 
-        let (uuid, bytes_received, is_collision) =
-            match instance.active_uploads.lock().entry(uuid_key) {
+        let (uuid, bytes_received, is_collision) = {
+            let mut active_uploads = instance.active_uploads.lock();
+            match active_uploads.entry(uuid_key) {
                 Entry::Occupied(mut entry) => {
                     let maybe_idle_stream = entry.get_mut();
                     if let Some(idle_stream) = maybe_idle_stream.1.take() {
@@ -512,8 +513,8 @@ impl ByteStreamServer {
                             .fetch_add(1, Ordering::Relaxed);
                         return idle_stream.into_active_stream(bytes_received, instance);
                     }
-                    // Case 3: Stream is active - generate a unique UUID to avoid collision
-                    // Using nanosecond timestamp makes collision probability essentially zero
+                    // Case 3: Stream is active - generate a unique UUID to avoid collision.
+                    // Using nanosecond timestamp makes collision probability essentially zero.
                     let original_key = *entry.key();
                     let unique_key = Self::generate_unique_uuid_key(original_key);
                     warn!(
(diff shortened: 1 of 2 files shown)


nativelink-service/src/bytestream_server.rs
```diff
@@ -494,8 +494,9 @@ impl ByteStreamServer {
         // Parse UUID string to u128 key for efficient HashMap operations
         let uuid_key = parse_uuid_to_key(uuid_str);
 
-        let (uuid, bytes_received, is_collision) =
-            match instance.active_uploads.lock().entry(uuid_key) {
+        let (uuid, bytes_received, is_collision) = {
+            let mut active_uploads = instance.active_uploads.lock();
+            match active_uploads.entry(uuid_key) {
```
MarcusSorealheis, line 499: Ahh, interesting. In introduced this lock but your change is an improvement.

MarcusSorealheis (commented):
We need the CLA and I need an hour to review further.

gergelytraveltime:
> We need the CLA and I need an hour to review further.

CLA signed

MarcusSorealheis (approved):
LGTM

palfrey:
@gergelytraveltime If you can bring this back up to date with current `main` we can get this one merged. Thanks!

MarcusSorealheis:
We have already merged this PR in #2429

PR #2409, opened 2026-06-09
fix(worker): sample action peak memory by process group

amankrx:
# Description

The action-resource-usage sampler walked the spawned child's process subtree (/proc/<pid>/task/*/children) to sum RSS. When an intermediate shell exits and reparents the real workload to the worker (PID 1), the spawned child is left a childless zombie, so the walk reports 0. The worker then omits ActionResourceUsage (gated on peak_memory_kb > 0), no ResponseEvent.ActionResourceUsage origin event is published, and execution_tasks.observed_worker_peak_memory_mib stays NULL.

Spawn the action as its own process-group leader and sum RSS of every process whose pgrp == pgid instead. Process-group membership is inherited and survives reparenting, and excludes the worker's own group.

## Type of change

## How Has This Been Tested?

Tested with local on prem test

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2409)

### DIFF
--- a/nativelink-worker/src/running_actions_manager.rs
+++ b/nativelink-worker/src/running_actions_manager.rs
@@ -109,11 +109,11 @@ struct ActionResourceUsageSampler {
 }
 
 #[cfg(target_os = "linux")]
-fn start_action_resource_usage_sampler(root_pid: u32) -> ActionResourceUsageSampler {
+fn start_action_resource_usage_sampler(pgid: u32) -> ActionResourceUsageSampler {
     let (stop_tx, stop_rx) = watch::channel(false);
     let handle = background_spawn!(
         "action_resource_usage_sampler",
-        sample_action_peak_memory_kb(root_pid, stop_rx)
+        sample_action_peak_memory_kb(pgid, stop_rx)
     );
     ActionResourceUsageSampler { stop_tx, handle }
 }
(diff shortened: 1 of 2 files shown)


nativelink-worker/src/running_actions_manager.rs
```diff
@@ -153,24 +155,36 @@ async fn sample_action_peak_memory_kb(root_pid: u32, mut stop_rx: watch::Receive
...
+/// reparents the real workload to the worker (PID 1); the tree walk then sees
+/// only a childless zombie and reports zero. Process-group membership is
+/// inherited and survives reparenting, and excludes the worker's own group.
+/// Returns `None` when no member process can be read (the group is empty).
 #[cfg(target_os = "linux")]
-fn sample_process_tree_memory_kb(root_pid: u32) -> Option<u64> {
-    let mut stack = vec![root_pid];
-    let mut seen = HashSet::new();
+fn sample_process_group_memory_kb(pgid: u32) -> Option<u64> {
+    let Ok(entries) = std::fs::read_dir("/proc") else {
```
palfrey, line 171: Can we add a log message here, as not being able to read `/proc` is unusual!

nativelink-worker/src/running_actions_manager.rs
```diff
@@ -185,25 +199,16 @@ fn read_process_rss_kb(pid: u32) -> Option<u64> {
     })
 }
 
+/// Reads a process's group id (`pgrp`, field 5 of `/proc/<pid>/stat`).
+///
+/// `comm` (field 2) can contain spaces and parentheses, so fields are parsed
+/// relative to the final `)` to avoid miscounting.
```
palfrey: Especially given this, please add tests around this function.

palfrey (changes requested):

palfrey (approved):

PR #2415, opened 2026-06-10
fast_slow_store: bypass leader/follower dedup for huge blobs

amankrx:
# Description

The populating-digests dedup is a net loss for multi-GB blobs: the leader's pull from the slow store reliably exceeds LEADER_WAIT_TIMEOUT, every concurrent follower falls through to its own slow-store read anyway, and populating the fast tier with the huge blob evicts a large number of smaller, more-useful entries.

Add a configurable size threshold (FastSlowSpec::bypass_dedup_threshold_bytes, default 256 MiB when unset) at and above which get_part streams straight from the slow store and skips the populate. Below the threshold the existing leader/follower path is unchanged. A huge_blob_dedup_bypasses metric counts the bypasses.

## Type of change

## How Has This Been Tested?

Tests (in the existing nativelink-store/tests/fast_slow_store_test.rs suite, via a CountingSlowStore that counts get_part calls):

- huge_blob_bypasses_dedup_and_skips_populate: a blob above the threshold drives one slow-store call per concurrent reader and leaves the fast tier empty.
- small_blob_still_dedups_and_populates: a blob below the threshold collapses to a single slow-store call and populates the fast tier.
- bypass_threshold_is_inclusive_at_exact_size: size == threshold takes the bypass path (the condition is >=).

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2415)

### DIFF
--- a/nativelink-store/src/fast_slow_store.rs
+++ b/nativelink-store/src/fast_slow_store.rs
@@ -59,6 +59,8 @@ pub struct FastSlowStore {
     #[metric(group = "slow_store")]
     slow_store: Store,
     slow_direction: StoreDirection,
+    /// See [`FastSlowSpec::bypass_dedup_threshold_bytes`].
+    bypass_dedup_threshold_bytes: u64,
     weak_self: Weak<Self>,
     #[metric]
     metrics: FastSlowStoreMetrics,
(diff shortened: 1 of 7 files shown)


nativelink-store/src/fast_slow_store.rs
```diff
@@ -154,11 +156,17 @@ impl Drop for LoaderGuard<'_> {
 
 impl FastSlowStore {
     pub fn new(spec: &FastSlowSpec, fast_store: Store, slow_store: Store) -> Arc<Self> {
+        let bypass_dedup_threshold_bytes = if spec.bypass_dedup_threshold_bytes == 0 {
+            DEFAULT_BYPASS_DEDUP_THRESHOLD_BYTES
+        } else {
+            spec.bypass_dedup_threshold_bytes
+        };
```
palfrey: I'm inclined to have the default be the old behaviour i.e. no bypass, especially as 256MB feels like the sort of threshold where it's good for some systems, and terrible for others. WDYT?

palfrey (approved):

PR #2424, opened 2026-06-12
filesystem_store: make unref ENOENT idempotent and demote map/disk divergence to warn

amankrx:
# Description

Two robustness fixes for the eviction map vs on-disk divergence that shows up under heavy write+evict concurrency:

  * unref: when the content file is already gone (rename returns ENOENT), mark the entry Temp and adopt the temp key, exactly as a successful rename would. A repeat unref then hits the `path_type == Temp` early-return instead of racing the vanished path again, and EncodedFilePath::drop stops re-claiming the content path.

  * get_part: a map entry whose file is missing on disk is recoverable (the entry is removed and a fast/slow caller re-populates from the slow store), not catastrophic. Demote the log from error! to warn! and drop the stale "process probably need restarted" wording.

## Type of change

## How Has This Been Tested?

Tests (in the existing nativelink-store/tests/filesystem_store_test.rs suite):

- unref_is_idempotent_when_file_already_gone: deletes the backing file, calls unref twice, and asserts the second call hits the Temp early-return. Fails on the pre-fix path (no path_type flip).

- get_part_on_map_disk_divergence_warns_and_removes_entry: deletes the backing file, reads, and asserts NotFound surfaces, the warn fires (not the old error), and the stale entry is removed from the map.

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2424)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -152,13 +152,15 @@ impl Drop for EncodedFilePath {
             "Spawned a filesystem_delete_file"
         );
         background_spawn!("filesystem_delete_file", async move {
-            let result = fs::remove_file(&file_path)
-                .await
-                .err_tip(|| format!("Failed to remove file {}", file_path.display()));
-            if let Err(err) = result {
-                error!(?file_path, ?err, "Failed to delete file",);
-            } else {
-                debug!(?file_path, "File deleted",);
+            match fs::remove_file(&file_path).await {
+                Ok(()) => debug!(?file_path, "File deleted"),
+                // The file already being gone is the desired end state of a
+                // delete, not a failure — e.g. an entry marked Temp after an
+                // already-gone unref points at a path that was never created.
+                Err(err) if err.code == Code::NotFound => {
+                    debug!(?file_path, "File already gone, nothing to delete");
+                }
+                Err(err) => error!(?file_path, ?err, "Failed to delete file"),
             }
             // .fetch_sub returns previous value, so we subtract one to get approximate current value
             let current_active_drop_spawns = shared_context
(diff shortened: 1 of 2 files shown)


MarcusSorealheis (changes requested):
In general this is the right direction. However, I have two thoughts..

`ENOENT` on rename is ambiguous.` rename(2)` returns `ENOENT` if the source is gone, but also if a directory component of the destination is missing. If the temp directory has been removed (operator cleanup, misconfig, tmpfs reaper), this branch will treat a perfectly intact content file as "already gone," flip the entry to `Temp`, and orphan the real file on disk . IT's a silent leak and the map now lies about what exists. 

I recommend on  `ENOENT` to do a `fs::metadata(&from_path)` check; only take the benign path if the source genuinely doesn't exist, otherwise fall through to the `warn!` branch. One extra syscall on an already-cold error path.

Also, adopting `new_key` means `drop` now deletes a file that doesn't exist. When this entry is eventually dropped, `EncodedFilePath::drop` will spawn the temp-file removal, hit `ENOENT`, and (last I looked at that path) log it as a failure. So the patch trades one noisy log for another, just relocated to drop. Either teach the drop path that `ENOENT-on-temp-delete` is benign (probably worth doing anyway), or consider whether there's a path type / sentinel that means "nothing on disk to reclaim" instead of pointing at a phantom temp file. If `PathType::Custom` already skips deletion in drop, check whether its other semantics make it usable here; if not, an explicit state is cleaner than a fake temp key.

MarcusSorealheis (approved):
I owe you a correction, and it cuts against my own suggestion: this is not a cold path. Your retained comment says the benign ENOENT case "dominates the log volume... fast enough to drown the runtime under sustained pressure" — which means I was wrong to call the metadata check "one extra syscall on an already-cold error path." Under a write+evict storm you're now doing an fs::metadata per vanished file, and if NativeLink's fs wrapper routes that through the open-file semaphore (check — I believe some of the wrapper's ops do acquire it), the unref path starts contending with readers for fd permits exactly when the system is busiest. I still think the check is right — correctness over a stat — but verify the semaphore question, and if it does acquire, consider whether tokio::fs::symlink_metadata directly (bypassing the wrapper) is acceptable here since no fd is held open.

Use symlink_metadata, not metadata, regardless of #1. metadata follows symlinks, and this store handles symlink entries. The edge: rename fails ENOENT because the temp dir is missing, and the source is a dangling symlink — metadata follows it, reports NotFound, you flip to Temp, and the symlink itself is orphaned on disk. That's precisely the bug class the check exists to prevent, just for link entries. symlink_metadata stats the link itself and closes it.

PR #2425, opened 2026-06-12
Drop vercel.json

palfrey:
# Description

We're seeing requests to authorize vercel even for PRs that don't edit anything under `web`. I suspect there's a clash between what Vercel has configured in the web UI and the static values in the source code. Of particular note is that the root directory is defined in the web ui as `web` where as the `vercel.json`'s are meant to [implicitly define it by where they're are located](https://vercel.com/docs/project-configuration/vercel-json). My guess (although it's hard to tell from outside Vercel itself) is that deleting these _might_ fix things, or at least let us fix the Web UI config.

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2425)


MarcusSorealheis (approved):
Did you compile the app locally as well?

Lgtm

PR #2433, opened 2026-06-16
Autogenerate docs again

MarcusSorealheis:
# Description

Autogenerated docs was broken. It's fixed now. 

## Type of change

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2433)


palfrey:
* `NamedConfig`/`NamedConfig2` seems weird? Similarly `WithInstanceName2`-7
* This doesn't auto-regenerate on `main`/tag changes, but we can do that in a future PR.
* `Example JSON Config` - actually JSON5 configs
* There's a `string?` (and other types) syntax being used that's familiar for Typescript devs, but not for others. We should probably just drop it
* If the default is `null` it doesn't feel necessary to list it?
* `ExperimentalSimpleSchedulerBackend` - not just `object` but explicitly `ExperimentalRedisSchedulerBackend` which isn't linked.
* `DirectoryCacheConfig` - `cache_root` - table says one thing for default, comments say another. Similarly `EndpointConfig.timeout`, vairous on `GrpcEndpoint` and probably a few others
* `RedisMode` - no descriptions
* `ExperimentalOntapS3Spec` - weird table wrapping due to large `retry` default
* `Retry` - comments need reformatting

palfrey (approved):
These still need addressing, but this is an improvement over what we've got already, so approving and we'll follow-up.

PR #2435, opened 2026-06-16
fix(redis_store): survive Redis failover across writes and reads

amankrx:
## Problem
With Redis replication, a failover *during* an operation surfaced transient errors. The write path failed hard (`Data length mismatch … got 0 bytes` → lost BEP events on a live cluster); the read paths (`has`/`get_part`/`list`) leaned only on the driver's single retry and could surface a transient error on a slow failover.

## Cause
Operations that span a master change ran their verify/read on a connection that became a demoted replica:
- `update`: chunk `SETRANGE` reconnects on `ReadOnly` (data lands on the new master), but the `STRLEN` verify + `RENAME` used the pre-write connection → `STRLEN` reads 0 → whole write rejected.
- `has_with_results` / `get_part` / `list`: no app-level re-resolve+retry (the `has` pipeline even had a `TODO(palfrey)` about it).

## Fix
A shared retryable-error check (`is_retryable_redis_error`: dropped/refused connection, IO error, or `ReadOnly`) plus, on a transient error, re-resolve the master (`reconnect`) and retry up to `MAX_REDIS_RETRY_ATTEMPTS`:
- `update`: retry the `STRLEN` verify; reconnect once on `ReadOnly` for `RENAME`. Genuine double-append still errors; a value that never appears still errors after retries.
- `has_with_results`, `get_part` (getrange + exists): retry the idempotent read.
- `list`: restart the SCAN from scratch (SCAN already may re-emit keys, so callers tolerate duplicates).
(`update_data` already reconnects; pub/sub re-subscribes on reconnect.)

## Verification
- Unit tests (MockRedisConnection): update retry-succeeds + give-up; `has` retry-after-transient-error. Full `redis_store` suite **40 passed**; `cargo check`/`rustfmt` clean.
- Live: an image built from this branch ran **8 builds through a forced master kill — all EXIT 0, 0 `Data length mismatch`, complete BEP data** (the old image failed the same scenario).

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2435)

### DIFF
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -79,6 +79,12 @@ const DEFAULT_CONNECTION_POOL_SIZE: usize = 3;
 /// Note: If this changes it should be updated in the config documentation.
 const DEFAULT_RETRY_DELAY: f32 = 0.1;
 
+/// Maximum attempts to verify a freshly written value (re-resolving the master
+/// each time) in `RedisStore::update`. Covers a Redis failover or brief replica
+/// lag between the chunk writes and the length check, so a transient topology
+/// change doesn't fail an otherwise healthy write.
+const MAX_UPDATE_VERIFY_ATTEMPTS: u32 = 5;
+
 /// The default connection timeout in milliseconds if not specified.
 /// Note: If this changes it should be updated in the config documentation.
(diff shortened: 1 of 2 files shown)


MarcusSorealheis (approved):

MarcusSorealheis:
Thank you

PR #2438, opened 2026-06-16
fix(redis_store): recover from Sentinel master failover without freezing the store

amankrx:
## Problem

A Redis Sentinel master failover stalls the CAS hard enough that builds fail and (with a redis-coupled liveness probe) the pod restarts — and tuning the store timeouts never fixed it, because the stall isn't a timeout-value problem:

1. `reconnect()` re-resolves the master via Sentinel (seconds during a failover) **while holding the `connection_manager` write lock**. Every `get_connection()` is a reader on that lock, so the whole store freezes behind one slow reconnect.
2. The scheduler index path (`search_by_index_prefix`) only re-resolves on `READONLY`. A dead/old master that drops the connection or times the command out is **not** `READONLY`, so the matching loop spins on the stale handle and never follows the failover.
3. `HealthStatusIndicator::check_health` PINGs but never reconnects on failure. After a failover a store whose connection is wedged — especially an idle one with no other traffic — fails the PING forever, so `/status` returns 503 and readiness sheds traffic from an otherwise-recovered pod until it's restarted.

## Fix

- Move the slow connect off the `connection_manager` write lock onto a dedicated `reconnect_lock`. Readers keep using the old handle and fail fast on their own command timeout; only the brief handle swap takes the write lock. A generation re-check keeps a failover storm to a single re-resolution.
- `search_by_index_prefix` reconnects on any retryable error (dropped / refused / timed-out / `READONLY`). A missing index stays non-retryable and falls through to the existing create path.
- `check_health` re-resolves the master and PINGs again on a failed PING, so an idle store self-heals within one readiness-probe cycle.

Builds on `is_retryable_redis_error` from #2437 (merged). Paired with the chart-side fix (TCP liveness + faster Sentinel failover) in enterprise-nativelink.

## Verified

Live GKE master failover (kill master mid-build → Sentinel promotes a replica): build completes EXIT 0, CAS 0 restarts, readiness stays healthy, every store connection re-resolves within ~18s, `/metrics` back to 200 in ~5ms. Unit: `test_search_by_index_retries_on_failover`, `check_health_recovers_after_failed_ping`; full `redis_store` suite (42) passes, clippy clean.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2438)

### DIFF
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -217,6 +217,14 @@ where
     /// for this connection in order to avoid multiple reconnection attempts.
     connection_manager: tokio::sync::RwLock<(C, Uuid)>,
 
+    /// Serializes reconnect attempts so a Sentinel master failover triggers a
+    /// single re-resolution instead of a thundering herd. Kept separate from
+    /// `connection_manager` on purpose: the (potentially multi-second) connect
+    /// runs while holding only this lock, so in-flight `get_connection` readers
+    /// keep using the existing handle and are never frozen behind a slow
+    /// reconnect. See [`Self::reconnect`].
+    reconnect_lock: tokio::sync::Mutex<()>,
+
     /// A list of subscription that should be performed on reconnect.
     subscriptions: Mutex<HashSet<String>>,
 }
@@ -254,6 +262,7 @@ where
             connect_func,
             update_if_version_matches_script,
             connection_manager: tokio::sync::RwLock::new((connection_manager, Uuid::new_v4())),
+            reconnect_lock: tokio::sync::Mutex::new(()),
             subscriptions: Mutex::new(HashSet::new()),
         };
         {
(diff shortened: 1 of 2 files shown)


MarcusSorealheis:
> Paired with the chart-side fix (TCP liveness + faster Sentinel failover) in enterprise-nativelink.

This reads like it's broken in free, open source, but it's not. You are saying that you fixed the chart to support this change. Is that correct?

Marcus

amankrx:
> > Paired with the chart-side fix (TCP liveness + faster Sentinel failover) in enterprise-nativelink.
> 
> This reads like it's broken in free, open source, but it's not. You are saying that you fixed the chart to support this change. Is that correct?
> 
> Marcus

It should be broken everywhere. We have adapted the enterprise nativelink from the open-source helm chart.

MarcusSorealheis (approved):
The fast-path/re-check uses !=, which means a second failover during a reconnect gets swallowed. Walk it: caller A holds the reconnect lock and is mid-connect for generation uuid_1. The master fails again; the handle is now generation uuid_2 but still broken. Caller B arrives with uuid_2, takes the read lock, sees guard.1 (uuid_2) != uuid_1? No — B's uuid is uuid_2, it matches, so B correctly proceeds to reconnect. Fine. But now the inverted case: caller B arrives still holding uuid_1 (it captured the old generation before A published), waits on the reconnect lock, A finishes and publishes uuid_2, B's re-check sees uuid_2 != uuid_1 → B returns A's fresh handle without reconnecting. That's the intended dedup and it's correct. So the != logic is actually right — but it hinges entirely on every caller passing the generation it observed as failing, and the check_health path (see #2) and the search_by_index_prefix path pass connect_id captured at different times. I talked myself back from "bug" to "verify": please confirm both call sites capture connect_id from the same get_connection whose command failed, not a re-fetched one. If any site re-fetches the connection (and thus a possibly-newer uuid) before calling reconnect, the != dedup hands back a handle that was never validated. Worth a comment at the reconnect signature stating the contract: "pass the uuid of the generation whose command just failed."

I will approve because this has been validated in the lab with customer workload and then you need to let me know if this is solved.

PR #2451, opened 2026-06-18
ref_store now resolves on boot, not first query

palfrey:
# Description

Previously, if you had a broken `ref_store` config i.e. it pointed to a non-existent store, Nativelink would still boot fine. It would then have errors on every request to that store, but still keep going with anything else. This is a bad behaviour as if we have a permanently broken config (i.e. not a transient error, but one that requires a config change to fix) then we should fail ASAP so the end user can fix it.

## Type of change

  not work as expected)

This will break any configs with bad ref stores. They shouldn't have been regarded as valid to begin with, but it's still a breaking change.

## How Has This Been Tested?

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2451)

### DIFF
--- a/nativelink-service/tests/bep_server_test.rs
+++ b/nativelink-service/tests/bep_server_test.rs
@@ -101,6 +101,10 @@ struct FlakyStore {
 #[async_trait]
 #[allow(clippy::todo)]
 impl StoreDriver for FlakyStore {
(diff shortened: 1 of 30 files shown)


MarcusSorealheis (approved):
I honestly could not find anything wrong or anything to nit about here. Nice!

PR #2454, opened 2026-06-19
Multi arch image publish

palfrey:
# Description

Fixes #2401. Makes our image publishing multiarch (x86_64 + aarch64).

This does two things:
* Unbreaks ARM image building which I think I broke fixing `mimalloc` on Macs at some point
* Makes it possible to build/push multi-arch Nativelink images.

Note this still leaves the other remaining image types (nativelink-worker-init, nativelink-worker-lre-cc, nativelink-worker-lre-rs) in x86-only setup, but this is a good step forwards. I think lre-cc/lre-rs mostly need Mac-specific ones, but worker-init should be made multi-arch in a future PR.

Annoyingly, it's not able to do things in a pure-Nix way, so we end up currently with a non-nix2container setup where we need to specify the set of images to bolt together, but it does at least work which is good enough for now. As a side-effect of that, I've added in date-based image naming as that makes it a lot easier to go "how old is this random Nativelink tag".

## Type of change

Please delete options that aren't relevant.

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

`nix run .#create-multi-arch-image nativelinkImageForX64 nativelinkImageForAarch64`

## Checklist

- [ ] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2454)

PR #2452, opened 2026-06-18
Improve security scorecard

MarcusSorealheis:
# Description

add fuzzing, and some other best practices to improve our rating in the [OpenSSF](https://openssf.org/) framework.

Fixes # (issue)

## Type of change

Please delete options that aren't relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

invoking the tests, bazel test //...

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2452)

PR #2457, opened 2026-06-22
Fix various OCI image upload issues

palfrey:
# Description

I broke some things in https://github.com/TraceMachina/nativelink/pull/2454, partially fixed in https://github.com/TraceMachina/nativelink/pull/2455 and then there was further problems added in https://github.com/TraceMachina/nativelink/pull/2452.

This PR should _hopefully_ tidy-up all of that, hence the extra work in here to deliberately also run everything on this branch before we merge. Along the way, I've also unified the three different flows (on main, on tag, on demand) for making images to all support multi-arch in a unified way.

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2457)

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -16,11 +16,25 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
+  get-images-matrix:
+    runs-on: ubuntu-24.04
+    outputs:
+       matrix: ${{ steps.setmatrix.outputs.matrix }}
+    steps:
+      - name: Checkout
+        uses: >- # v6.0.2
+          actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - name: Set Images matrix
+        id: setmatrix
+        run: |
+          MATRIX_STRINGIFIED=$(jq -c . < .github/images-matrix.json)
+          echo "matrix=$MATRIX_STRINGIFIED" >> $GITHUB_OUTPUT
+
   publish-image:
+    needs: get-images-matrix
     strategy:
       fail-fast: false
-      matrix:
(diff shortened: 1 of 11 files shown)


.github/workflows/image.yaml
```diff
@@ -47,13 +47,15 @@ jobs:
...
       # FIXME: merge the multi-arch and single-arch paths
       - name: Test multi-arch image
-        if: ${{ matrix.image == 'image' && github.ref == 'refs/heads/main' }}
+        if: ${{ matrix.image == 'image' }}
         run: |
           nix run --fallback .#create-multi-arch-image localhost:5000 nativelink:${{ steps.version.outputs.version-string }} nativelink-image-for-x64 nativelink-image-for-aarch64
           nix run --fallback .#local-image-test localhost:5000/nativelink:${{ steps.version.outputs.version-string }}
 
+      # FIXME: after we've merged this and made sure it all works _then_ remove the 2457 ref
+      # See https://github.com/TraceMachina/nativelink/pull/2457
```
palfrey: This is a bit odd, but given this all came about because I thought I got it all right in https://github.com/TraceMachina/nativelink/pull/2454 before reverting prior to merge, I explicitly want to merge this and fix it only after we know it's all fine.

flake.nix
```diff
@@ -162,7 +162,6 @@
           })
           // (pkgs.lib.optionalAttrs isLinuxTarget {
             CARGO_BUILD_RUSTFLAGS = "-C target-feature=+crt-static";
-            TARGET_CC = "${pkgs.lre.clang}/bin/customClang"; # So mimalloc gets the right compiler not defaulting to gcc
```
palfrey, line 165: This breaks cross-compiling, and hence the block just above which sets this only for non-cross compile builds.

MarcusSorealheis (approved):
Ooops

PR #2467, opened 2026-06-23
memory_store: skip writes larger than max_bytes instead of buffering them

corcillo:
# Description

`MemoryStore::update` reads the **entire** payload into memory (`reader.consume(None)`, plus a defensive full copy) before inserting into the evicting map and running eviction.

For a write whose size already exceeds the store's `max_bytes`, that buffering is wasted: `EvictingMap::evict_items` pops LRU entries `while sum_store_size >= max_bytes` with no "keep the newest" guard, so an entry that alone exceeds the budget is evicted the instant it's inserted. The blob was never retainable — yet we materialized its whole payload (and a second copy) in RAM to get there.

Under **concurrent large writes that's an OOM vector.** A `memory` fast tier inside a `fast_slow` store fronting CAS receives every blob regardless of size (`fast_slow` fans writes to both tiers), so a burst of oversized ByteStream uploads each materializes its full payload before eviction runs — enough parallel uploads can blow the pod's memory limit even though the steady-state cache is tiny. (Surfaced from a production CAS pod getting `OOMKilled` during a build's upload burst.)

**Fix:** when the eviction policy caps size (`max_bytes != 0`) and the write's `ExactSize` exceeds it, drain the stream and return without buffering:

```rust
if self.max_bytes != 0
    && let UploadSizeInfo::ExactSize(sz) = size_info
    && sz > self.max_bytes
{
    let drained = reader.drain().await.err_tip(|| /* ... */)?;
    return Ok(drained);
}
```

This has **no observable behavior change** — an over-budget blob is not retained either way (today it's inserted then immediately evicted, so `has`/`get` already return not-found); the change just skips the pointless materialization in between. Only `ExactSize` is trusted (a `MaxSize` upper bound could over-estimate and wrongly skip a blob that fits), and it's guarded on `max_bytes != 0` so count/time-only and unbounded stores are untouched.

**Scope:** this targets the single-write-exceeds-budget case. It does **not** address many *concurrent sub-budget* writes collectively exhausting memory — that needs a global in-flight-bytes / write-concurrency budget in the memory (or `fast_slow`) store, which is out of scope here.

Fixes # _(no tracked issue — surfaced from a production CAS OOM incident; see above)_

## Type of change

## How Has This Been Tested?

Added `nativelink-store/tests/memory_store_test.rs::skips_writes_larger_than_max_bytes`: with a small `max_bytes`, a within-budget write is stored (`has()` returns its size), and a write larger than `max_bytes` succeeds but is not cached (`has()` returns `None`). Existing `memory_store_test` cases continue to cover within-budget reads/writes and the full-copy guarantee.

## Checklist

- [x] Updated documentation if needed (none required — internal store behavior)
- [x] Tests added/amended
- [ ] `bazel test //...` passes locally — please run before merge (not run in this environment)
- [x] PR is contained in a single commit, using `git amend`

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2467)

### DIFF
--- a/nativelink-store/src/memory_store.rs
+++ b/nativelink-store/src/memory_store.rs
@@ -68,6 +68,11 @@ pub struct MemoryStore {
         SystemTime,
         RemoveItemCallbackHolder,
     >,
+    /// The eviction policy's `max_bytes` (0 = unbounded). Cached here so `update`
+    /// can skip writes larger than the entire store budget without buffering
+    /// them — see the note in `update`.
+    #[metric(help = "Maximum bytes this store will hold before eviction (0 = unbounded)")]
+    max_bytes: u64,
 }
 
 impl MemoryStore {
@@ -75,6 +80,7 @@ impl MemoryStore {
         let empty_policy = nativelink_config::stores::EvictionPolicy::default();
         let eviction_policy = spec.eviction_policy.as_ref().unwrap_or(&empty_policy);
         Arc::new(Self {
+            max_bytes: eviction_policy.max_bytes as u64,
             evicting_map: EvictingMap::new(eviction_policy, SystemTime::now()),
         })
     }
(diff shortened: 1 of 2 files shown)


MarcusSorealheis:
after the tests are passing, I would like to have a warning that tells customers to consider the size partitioning store.

corcillo:
> after the tests are passing, I would like to have a warning that tells customers to consider the size partitioning store.

@MarcusSorealheis what is our current mechanism to warn customers?

palfrey:
> > after the tests are passing, I would like to have a warning that tells customers to consider the size partitioning store.
> 
> @MarcusSorealheis what is our current mechanism to warn customers?

Our usual mechanism would just be `warn!`

nativelink-store/src/memory_store.rs
```diff
@@ -138,8 +144,27 @@ impl StoreDriver for MemoryStore {
...
+        // skip instead. Only `ExactSize` is trusted here; a `MaxSize` upper bound
+        // could over-estimate and wrongly skip a blob that would actually fit.
+        if self.max_bytes != 0
+            && let UploadSizeInfo::ExactSize(sz) = size_info
+            && sz > self.max_bytes
+        {
+            let drained = reader
+                .drain()
+                .await
+                .err_tip(|| "Failed to drain oversized write in memory_store::update")?;
```
palfrey, line 165: Please add a `warn!` log here so customers are aware of the problem

PR #2472, opened 2026-06-23
Migrate Azure Blob store to azure_storage_blob 1.0

amankrx:
# Description

Moves the experimental Azure Blob store off the archived community crate (`azure_storage_blobs` 0.21) and onto the stable GA SDK: `azure_core` / `azure_storage_blob` / `azure_identity` 1.0.

- **Auth**: Entra ID via `WorkloadIdentityCredential` (keyless) by default, with a pre-formed SAS URL as a fallback. Shared-key auth is dropped — the GA SDK no longer supports it.
- **Transport**: removed the custom hyper transport bridge; the store uses the SDK's reqwest transport, pinned to **ring** TLS (`rustls-tls-webpki-roots`) so `aws-lc-sys` stays out of the build graph.
- **Config**: `ExperimentalAzureSpec` now takes `account_name`, `container`, `endpoint` (Azurite override) and `sas_url`.
- **Tests**: rewritten against a 1.0 fake-transport `HttpClient` mock (14 tests).
- Updated `Cargo.toml`, `BUILD.bazel`, `Cargo.lock` and repinned `MODULE.bazel.lock`.

## Notes

This is a breaking change for Azure users: shared-key (`AZURE_STORAGE_KEY`) is replaced by Workload Identity or a SAS URL. Helm chart changes for the new auth model are handled separately.

## Type of change

- [X] Breaking Change

## How Has This Been Tested?

Full `bazel build //:nativelink` with a real Azure environment. 

## Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2472)


nativelink-config/src/stores.rs
palfrey, line 1: Can you regenerate the docs so we make sure the common bits do the right thing here?

nativelink-config/src/stores.rs
```diff
@@ -1191,27 +1191,34 @@ pub struct ExperimentalGcsSpec {
...
+    /// is not provided.
     #[serde(default, deserialize_with = "convert_string_with_shellexpand")]
     pub account_name: String,
 
     /// The container name to use as the backend.
     #[serde(default, deserialize_with = "convert_string_with_shellexpand")]
     pub container: String,
 
+    /// Optional blob endpoint host override (for example an Azurite emulator host
+    /// such as `http://127.0.0.1:10000/devstoreaccount1`). When set, this replaces
```
palfrey, line 1205: Not for this PR, but just pondering: can we run this against an Azurite emulator in CI?

palfrey (changes requested):
LGTM other than those small items

nativelink-config/src/stores.rs
amankrx: Yes. We can run Azurite in CI as a service container, same pattern as the existing `mongo_runner`. We can point the store at it via endpoint and a SAS URL.

palfrey: https://github.com/TraceMachina/nativelink/issues/2511

palfrey (approved):

PR #2473, opened 2026-06-23
memory_store: skip writes larger than max_bytes instead of buffering them

corcillo:
# Description
This PR makes the `MemoryStore` **early-reject** writes whose size meets or exceeds the store's `max_bytes`— draining the stream and removing it from the existence cache store + logging a warning. Before, it would buffer the whole payload into RAM only to immediately evict it. If max_bytes is not configured, it will have the same behavior as before. PR https://github.com/TraceMachina/cloud-platform/pull/1092 will add this to the helm charts and set a default `max_bytes` to 16Mb.

What it does NOT do: 
Technically, we'll still hit the OOM issue if we have X small (<`max_bytes`) blobs being ingested.  This pr shifts that oom to only happen if there are a large number of small blobs (like 700 x 15.9Mb blobs).  The tradeoff is we only cache small-medium files.  With the helm chart pr, companies can decide where this line is for them.

Originally opened as #2467, but I closed and accidentally orphaned that commit; this is the clean re-open.

## Background / root cause

`MemoryStore::update` reads the **entire** payload into memory (`reader.consume(None)`, plus a defensive full copy) before inserting into the evicting map and running eviction. For a write whose size already meets the store's `max_bytes`, that buffering is pure waste: `EvictingMap::evict_items` pops LRU entries `while sum_store_size >= max_bytes` with no "keep the newest" guard, so an entry that alone meets the budget is evicted the instant it's inserted. The blob was never retainable — we paid full RAM for it and dropped it immediately.

**Eviction was not broken.** It correctly bounds the *committed cache* (`sum_store_size`, the total of everything already inserted). The problem is what it *can't* see: the per-write `consume()` buffer exists in RAM **before** the blob is ever inserted into the map, so eviction has no knowledge of it and no power to reclaim it. Eviction bounds the cache; it does nothing about the **transient memory of in-flight writes**.

That gap becomes an **OOM vector under concurrency.** A `memory` fast tier inside a `fast_slow` store fronting CAS receives every blob regardless of size (`fast_slow` fans writes to both tiers). A burst of oversized ByteStream uploads each materializes its full payload at the same time — none of them inserted yet, so eviction is powerless — and the sum of those in-flight buffers blows the cgroup limit. This surfaced from a production CAS pod `OOMKilled` (exit 137) at its 12Gi limit during a build that uploaded ~66 large blobs (≈280–555MB each) concurrently.

## Fix

When the eviction policy caps size (`max_bytes != 0`) and the write's `ExactSize` is `>= max_bytes`, drain the stream and return without buffering. The `>=` deliberately matches the eviction comparator (`sum_store_size >= max_bytes`), so a blob of exactly `max_bytes` — which would also be evicted on arrival — is skipped rather than buffered-then-evicted. Only `ExactSize` is trusted: a `MaxSize` upper bound could over-estimate and wrongly skip a blob that actually fits, so `MaxSize` writes take the normal path.

Because the rejected write never enters the map, the insert-then-evict removal callbacks don't fire on their own. A wrapping `ExistenceCacheStore` relies on those callbacks to drop a just-written-then-evicted key (its "copes with dropped items" guarantee), so the skip path fires them explicitly via a new `EvictingMap::fire_remove_callbacks` — a no-op when no callbacks are registered. This keeps observable behavior identical to the old store-then-evict path: the existence cache never advertises a digest the memory tier didn't keep, and the blob remains durably stored on the `fast_slow` slow tier either way.

## Before / after

```mermaid
flowchart LR
  subgraph B["BEFORE — buffer, then evict (OOM under load)"]
    direction TB
    b1["update(ExactSize = 500MB)"] --> b2["reader.consume(None)<br/>buffer the WHOLE blob in RAM"]
    b2 --> b3["evicting_map.insert<br/>sum_store_size += 500MB"]
    b3 --> b4{"sum_store_size >= max_bytes ?"}
    b4 -- yes --> b5["pop_lru() evicts the blob + neighbors<br/>fires remove callbacks"]
    b5 --> b6["but RAM already spiked<br/>× N concurrent uploads → OOMKilled (137)"]
  end
  subgraph A["AFTER — size-check before buffering (this PR)"]
    direction TB
    a1["update(ExactSize = 500MB)"] --> a2{"max_bytes != 0 AND ExactSize(sz)<br/>AND sz >= max_bytes ?"}
    a2 -- "yes — skip" --> a3["reader.drain()<br/>discard, O(1) memory"]
    a3 --> a4["warn!(skipping oversized write)"]
    a4 --> a5["evicting_map.fire_remove_callbacks(key)<br/>un-marks ExistenceCacheStore"]
    a5 --> a6["return Ok — durable on slow tier, no OOM"]
    a2 -- no --> a7["consume → insert → evict if over budget<br/>(unchanged path)"]
  end
  B ~~~ A
```

## Scope, and the follow-up this does *not* do

This closes the **single-blob-exceeds-the-cache** vector. Together with a complementary chart-level mitigation (a `size_partitioning` store that routes blobs above a configurable threshold — default 16mb — to a `noop` store, keeping large blobs out of the RAM tier entirely on already-deployed binaries), it resolves the production incident.

It does **not** bound *aggregate* in-flight memory. Many concurrent writes each **under** `max_bytes` still each buffer their full payload, and their sum is unbounded — so a large enough fan-out of sub-threshold blobs could still OOM. This fix raises the cliff (the incident died at ~66 large blobs; you'd now need many hundreds of small ones) but does not remove it.

The complete fix is **admission control / backpressure**: cap the total bytes allowed in-flight across all concurrent writes (e.g. a byte-budget semaphore acquired by declared size *before* buffering, released after insert/eviction), so a write waits rather than allocating when the budget is exhausted. That makes the guarantee **count- and size-independent** — peak in-flight RAM ≤ budget, regardless of how many uploads arrive or how big they are.

That's intentionally **out of scope here**:

- **Blast radius.** This PR only *declines to cache* things that would have been evicted anyway — zero throughput cost, obviously safe. Backpressure sits on the hot write path of every upload and deliberately *slows producers down*; it's a global throttle, not a no-op.
- **It's a perf/product tradeoff, not a pure bug fix.** The budget value trades memory safety against throughput, needs load testing and per-deployment configurability, and warrants its own design/RFC rather than riding in on an incident hotfix.
- **Correctness traps.** Byte-semaphores on write paths invite deadlock (a single blob larger than the whole budget — handled here by the size-skip, so the two compose), fairness/starvation, and double-counting in `fast_slow` (which fans to two tiers).
- **Open design question of where it lives** — the memory store, the `fast_slow` store, a new wrapping store, or the ByteStream server layer — each with different tradeoffs.

Happy to open a design proposal / tracking issue for the admission-control work as a follow-up.

## Type of change

## How Has This Been Tested?

Rewrote `nativelink-store/tests/memory_store_test.rs` to drive the streaming `update` path. (`update_oneshot` takes already-in-memory `Bytes` and hits `MemoryStore`'s optimized direct-insert path, which **bypasses** the new skip — so the tests stream via `make_buf_channel_pair` to actually exercise it.) Added:

- `skips_writes_larger_than_max_bytes` — with a small `max_bytes`: a within-budget write is stored (`has()` returns its size), an oversized write succeeds but is not cached (`has()` returns `None`), and the within-budget entry **survives** the oversized write (the old buffer-then-evict path would have evicted it trying to make room).
- `oversized_skip_fires_remove_callbacks` — registers a counting `RemoveItemCallback` and asserts the skip path fires it exactly once.
- `max_size_over_budget_with_small_actual_is_stored` — a `MaxSize` over `max_bytes` whose actual content fits is still cached (the upper bound must not trigger the skip).

The existing `existence_store_test::copes_with_dropped_items` continues to pass, exercising the removal-callback path end to end.

Validated locally in the Nix dev shell: `cargo test -p nativelink-store --test memory_store_test` (11/11 pass), `existence_store_test` (5/5 pass), and `cargo clippy` clean on `nativelink-store --tests` and `nativelink-util --lib`.

## Checklist

- [x] Updated documentation if needed (none required — internal store behavior)
- [x] Tests added/amended
- [x] `cargo test` + `cargo clippy` pass locally (Nix dev shell); full `bazel test //...` validated in CI

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2473)


nativelink-store/src/memory_store.rs
```diff
@@ -138,8 +145,43 @@ impl StoreDriver for MemoryStore {
...
+        // could over-estimate and wrongly skip a blob that would actually fit.
+        if self.max_bytes != 0
+            && let UploadSizeInfo::ExactSize(sz) = size_info
+            && sz > self.max_bytes
+        {
+            let drained = reader
+                .drain()
+                .await
+                .err_tip(|| "Failed to drain oversized write in memory_store::update")?;
+            warn!(
```
corcillo, line 175: This addresses original comments from @MarcusSorealheis and @palfrey 
`after the tests are passing, I would like to have a warning that tells customers to consider the size partitioning store.` from @MarcusSorealheis and 
`Our usual mechanism would just be warn` and `Please add a warn! log here so customers are aware of the problem` from @palfrey on the `.err_tip` line

https://github.com/TraceMachina/nativelink/pull/2467

nativelink-util/src/evicting_map.rs
```diff
@@ -490,6 +490,23 @@ where
         while callbacks.next().await.is_some() {}
     }
 
+    /// Fires the registered remove callbacks for `key` without touching the map.
+    /// A store uses this to invalidate downstream listeners (e.g. an
+    /// `ExistenceCacheStore`) when a write is rejected outright and never
+    /// inserted — mirroring the callbacks an insert-then-immediate-evict would
+    /// otherwise have fired. A no-op when no callbacks are registered.
+    pub async fn run_remove_callbacks(&self, key: &Q) {
```
corcillo: tests were originally failing because this part was missing, this pr stopped adding oversized files to ram, but never fired the removal callback the ExistenceCacheStore relies on to know a write was dropped

nativelink-store/src/memory_store.rs
corcillo: @palfrey `warn!` is not too loud right? it will create a warn for every item above the max threshold.

MarcusSorealheis:
I think this PR has come a long way. I'd like for the iteration to happen in public but this is great. One important note is that if you don't set `maxBytes`, this PR is inert.

nativelink-store/src/memory_store.rs
palfrey: It's a bit noisy, but that's kinda the point. This is very much IMHO a "we can make it work but you really need to change your config" situation. Also, if it's only happening once or twice, it'll probably get lost in the noise (which is ok) v.s. if they're doing it all the time then they _definitely_ need to change their config.

palfrey (approved):

PR #2479, opened 2026-06-24
Dep bump to improve OSSF score card: bump quinn-proto to 0.11.15 and memmap2 to 0.9.11 (RustSec)

corcillo:
## Summary

Patch-level, **lockfile-only** bumps clearing two recently published RustSec advisories against transitive dependencies:

| Crate | From → To | Advisory | Issue |
|---|---|---|---|
| `quinn-proto` | 0.11.14 → 0.11.15 | [RUSTSEC-2026-0185](https://rustsec.org/advisories/RUSTSEC-2026-0185.html) | Remote memory exhaustion via unbounded out-of-order QUIC stream reassembly |
| `memmap2` | 0.9.9 → 0.9.11 | [RUSTSEC-2026-0186](https://rustsec.org/advisories/RUSTSEC-2026-0186.html) | Unchecked pointer offset in `advise_range` / `flush_range` |

Both are semver-compatible patch releases and `cargo check --workspace` passes. This restores the OpenSSF Scorecard **Vulnerabilities** check, which these two advisories dropped from 5 → 3 (overall **9.4 → 9.2**).

> **TL;DR on risk:** both fixes are legitimate, but **neither closed an exploitable hole in NativeLink as built** — `quinn-proto` isn't compiled into our binary at all, and `memmap2` is, but the vulnerable functions are never called. This is dependency hygiene + score restoration, not an incident. Details below.

---

### `memmap2` 0.9.9 → 0.9.11 — [RUSTSEC-2026-0186](https://rustsec.org/advisories/RUSTSEC-2026-0186.html) ([fix](https://github.com/RazrFalcon/memmap2-rs/pull/170))

**What changed** (the security-relevant hunk; the rest of the diff is `unsafe`-block hygiene and an MSRV bump to 1.65):

```rust
// added to the inner advise/flush range impls, before the pointer math:
if offset > self.len || len > self.len - offset {
    return Err(io::ErrorKind::InvalidInput.into());
}
```

Previously `advise_range` / `unchecked_advise_range` / `flush_range` / `flush_async_range` computed `ptr.offset(offset)` with **no bounds check** and passed the result to the `madvise` / `msync` syscalls. The fix rejects out-of-bounds `offset` / `len` up front. Per the advisory, the bad pointer was **never dereferenced** — only handed to a syscall (→ `EINVAL`) — so this is UB-in-pointer-arithmetic, not memory corruption. Low severity (no CVSS assigned).

**Reachable in NativeLink?** The crate *is* compiled in: `memmap2 → blake3 → nativelink-store` / `nativelink-util` (blake3's `mmap` feature, for file hashing). **But** blake3 only calls `Mmap::map()` and hashes the whole slice — it never calls the `*_range` advise/flush APIs with caller-controlled offsets. So: reachable crate, **unreachable vulnerable path**, and the new error branch only fires on already-invalid arguments → zero behavior change for normal use.

---

### `quinn-proto` 0.11.14 → 0.11.15 — [RUSTSEC-2026-0185](https://rustsec.org/advisories/RUSTSEC-2026-0185.html) ([fix](https://github.com/quinn-rs/quinn/pull/2694))

**What changed** (`quinn-proto/src/connection/assembler.rs`): `Assembler::insert` now returns `Result<(), TooManyChunks>` and **caps the number of buffered out-of-order chunks**. After attempting `defragment()`, if a peer has stuffed the buffer with too many small / gapped fragments, it returns the new propagates up (`recv.rs` / `connection/mod.rs`) to fail the stream instead of bufferingunboundedly. The vuln (`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`) is a **remote, unauthenticated memory-exhaustion DoS** against anyone running a quinn QUIC server.

**Reachable in NativeLink?** **No.** `cargo tree -i quinn-proto` returns *"nothing to print"* — it's pulled by `quinn`, a target-/feature-gated transitive dependency that **isn't activated in NativeLink's
build**. The vulnerable Assembler code isn't compiled into the binary, and NativeLink doesn **not reachable** — pure Scorecard noise.

---

### Bottom line

| Crate | Fix is | Compiled in? | Vuln path reachable? | Real risk to NativeLink |
|---|---|---|---|---|
| `memmap2` | real + correct | yes (via blake3) | no (range APIs unused) | ~nil |
| `quinn-proto` | real + correct (DoS hardening) | no (target-gated) | no (not in binary) | nil |

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — dependency security bump

## How Has This Been Tested?

- `cargo check --workspace` passes in the Nix dev shell.
- Lockfile-only change — no source touched (diff is `Cargo.lock`: 4 lines, the two version
- Reachability verified with `cargo tree -i quinn-proto` (not in the build graph) and `cargo tree -i memmap2` (in, via blake3).
- Full `bazel test //...` runs in CI.

## Checklist

- [x] Updated documentation if needed (none required — dependency bump)
- [x] Tests added/amended (n/a — lockfile-only dependency bump)
- [x] `cargo check --workspace` passes locally (Nix dev shell); full `bazel test //...` run

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2479)


MarcusSorealheis (approved):
LGTM

PR #2482, opened 2026-06-25
Add trace logging for bytestream read

palfrey:
# Description

Adds trace logging to bytestream server to allow tracking down various cases of stalling

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2482)

### DIFF
--- a/nativelink-service/src/bytestream_server.rs
+++ b/nativelink-service/src/bytestream_server.rs
@@ -1363,13 +1363,19 @@ impl ByteStream for ByteStreamServer {
         let start_time = Instant::now();
 
         let read_request = grpc_request.into_inner();
-        let resource_info = ResourceInfo::new(&read_request.resource_name, false)?;
+        let resource_name = read_request.resource_name.clone();
+        let resource_info = ResourceInfo::new(&resource_name, false)?;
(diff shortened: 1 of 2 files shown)


amankrx (approved):

nativelink-service/src/bytestream_server.rs
```diff
@@ -1363,14 +1363,20 @@ impl ByteStream for ByteStreamServer {
         let start_time = Instant::now();
 
         let read_request = grpc_request.into_inner();
-        let resource_info = ResourceInfo::new(&read_request.resource_name, false)?;
+        let resource_name = read_request.resource_name.clone();
```
MarcusSorealheis, line 1366: is this `.clone()` expensive? It's the entire stream right?

palfrey: No, it's just cloning the name of the resource, which should be fairly cheap. The demo example one (`foo_instance_name/blobs/0123456789abcdef000000000000000000000000000000000123456789abcdef/19`) is 91 characters, which is probably in line with what I'd expect.

PR #2489, opened 2026-06-26
Drop rsa in favour of aws-lc-rs

palfrey:
# Description

This PR resolves https://rustsec.org/advisories/RUSTSEC-2023-0071.html by dropping rsa in favour of aws-lc-rs, but aws-lc is a pain to get working.

This PR includes upgrades to `rules_rs` and patches related to https://github.com/hermeticbuild/hermetic-llvm/issues/405 to get `aws-lc` working under non-Nix bazel. However, this then breaks Bazel under Nix :( To work around this, `tools/nix/flake-module.nix` rewrites the `tools/aws-lc.MODULE.bazel` file under Nix to be a form that works under that. This then promptly both causes issues with git seeing changes _and_ the `MODULE.bazel.lock` being incorrect. There are then workarounds to both of those, in the form of `git update-index --assume-unchanged` magic and ignoring Bazel lockfile changes on Nix builds.

The nix build is also not great, as it's using the `aws-lc` package files which then requires dynamic linking of the libraries there. It seems to work so far, but not entirely happy with that.

In other words, this is a bit of a hack! It's primary benefit is that everything still builds and we're able to resolve the security issue, but other options that achieve those goals in a less terrible way would be good.

## Type of change

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2489)


palfrey:
Obsoleted by https://github.com/TraceMachina/nativelink/pull/2734

PR #2497, opened 2026-07-02
Add REAPI content-defined chunking (SplitBlob/SpliceBlob) support

### LINKED ISSUE #2496: Support REAPI content-defined chunking (SplitBlob/SpliceBlob) for Bazel's --experimental_remote_cache_chunking
## Summary

Bazel 8.7.0 / 9.1.0+ ship `--experimental_remote_cache_chunking` ([bazelbuild/bazel#28437](https://github.com/bazelbuild/bazel/pull/28437)), which uses content-defined chunking (FastCDC 2020) to upload/download only the chunks of a large blob that changed, instead of re-transferring the whole blob whenever its digest changes. The protocol surface — `SplitBlob` / `SpliceBlob` RPCs on `ContentAddressableStorage`, plus `ChunkingFunction` negotiation and capability advertisement — is standardized upstream in remote-apis ([bazelbuild/remote-apis#282](https://github.com/bazelbuild/remote-apis/pull/282), follow-up in #357). BuildBuddy is currently the only server implementation (see their [write-up](https://www.buildbuddy.io/blog/content-defined-chunking/)); they report ~85% of written bytes deduplicated for eligible files, and Bazel's benchmarks showed ~40% less data uploaded and ~40% smaller local disk cache.

This issue proposes implementing the server side in NativeLink.

## Why NativeLink is well positioned

NativeLink already has production CDC machinery at the storage layer:

- `DedupStore` (`nativelink-store/src/dedup_store.rs`) chunks blobs with a vendored FastCDC implementation (`nativelink-util/src/fastcdc.rs`, 64k/256k/512k min/avg/max defaults), stores chunks individually, and persists a serialized chunk-digest index keyed by the original blob digest.

That is almost exactly the data model `SplitBlob`/`SpliceBlob` expose over the wire — but today it is invisible to REAPI, so it saves storage without saving network transfer. Wire-level chunking is where most of the user-visible win is (large linked binaries, archives, container layers re-transferred after one-line source changes).

## Current gaps

- The vendored `nativelink-proto/build/bazel/remote/execution/v2/remote_execution.proto` predates the extension: no `SplitBlob`/`SpliceBlob` RPCs, no `ChunkingFunction`, no `split_blob_support`/`splice_blob_support` in `CacheCapabilities`.
- `cas_server.rs` implements only the classic four CAS RPCs; `capabilities_server.rs` advertises nothing chunking-related.
- `DedupStore`'s chunk digests are internal BLAKE3 hashes regardless of the outer digest function, so its index can't be returned to clients as-is — the protocol requires chunks to be ordinary, client-addressable CAS blobs under the negotiated digest function.

## Proposed phased plan

1. **Proto + capabilities.** Vendor the updated `remote_execution.proto`; add `split_blob_support` / `splice_blob_support` and supported `ChunkingFunction` values to `GetCapabilities`, gated behind config and off by default. Bazel only activates chunking when the server advertises it, so this is safe to ship incrementally.
2. **`SpliceBlob` (write path).** Accept an ordered chunk-digest list for a blob, verify all chunks exist, verify the concatenation actually hashes to `blob_digest` before registering the mapping (otherwise this is a cache-poisoning vector), and store the blob→chunks index.
3. **`SplitBlob` (read path).** Return the stored layout when a splice was registered; optionally chunk on demand server-side using the existing FastCDC util for blobs that were uploaded whole — this benefits mixed fleets where some clients upload without chunking.
4. **(Optional) `DedupStore` unification.** Evaluate sharing chunking/index machinery between `DedupStore` and the wire protocol so a chunk-uploaded blob is stored deduplicated without re-chunking.

## Open design questions

- **Chunk lifetime / eviction coupling.** Chunks must not be evicted while an index referencing them is alive. This needs a design that works across NativeLink's composable store topologies (evicting memory stores, S3 backends, sharded stores) — similar in spirit to the ActionResult→blob existence dependency, but inside the CAS.
- **Where the blob→chunks index lives.** A dedicated keyspace/store vs. a CAS blob keyed off the original digest; interaction with `FindMissingBlobs` semantics (does a registered splice make the blob "exist"?).
- **Chunking parameters.** Server-advertised FastCDC 2020 parameters must match what Bazel uses, or cross-client dedup silently degrades; `DedupStore`'s current defaults may differ from Bazel's.
- **Digest functions.** Chunk digests must use the request's digest function (SHA256/BLAKE3), unlike `DedupStore`'s internal fixed-BLAKE3 hashing.

## References

- Bazel client implementation: https://github.com/bazelbuild/bazel/pull/28437
- Protocol: https://github.com/bazelbuild/remote-apis/pull/282
- BuildBuddy design/results: https://www.buildbuddy.io/blog/content-defined-chunking/

erneestoc:
Implements the server side of the REAPI blob split/splice extension ([remote-apis#282](https://github.com/bazelbuild/remote-apis/pull/282)) used by Bazel's `--experimental_remote_cache_chunking` (Bazel 8.7.0+ / 9.1.0+).

Fixes #2496.

## What this adds

- **Vendored protocol**: `SplitBlob`/`SpliceBlob` RPCs, `ChunkingFunction`, `FastCdc2020Params`, `RepMaxCdcParams`, and `CacheCapabilities` fields 8–12, verbatim from upstream remote-apis.
- **`SpliceBlob`** re-assembles chunked uploads: validates chunk sizes/counts, checks all chunks exist (touching them), streams them through an incremental hasher into the CAS with reads pipelined 10-at-a-time — the digest is verified *before* EOF so a mismatch aborts the upload uncommitted — then persists the chunk layout in a configurable index store. The blob is fully materialized so ByteStream/`FindMissingBlobs` stay correct for non-chunking clients.
- **`SplitBlob`** serves stored layouts (validated to sum to the blob size, so corrupt or truncated index entries are never served), and otherwise **chunks blobs on demand** with FastCDC 2020 — this is the path that matters for RBE, since worker-uploaded action outputs are never spliced by a client. Unusable layouts (evicted chunks, decode failures, transient existence-check errors) fall back to re-chunking.
- **Native proxy forwarding**: grpc-store-backed instances forward both RPCs verbatim to the backend (instance-name rewriting + retries), making NativeLink relays transparent for chunking.
- **Capabilities** advertise `split_blob_support`/`splice_blob_support` and FastCDC 2020 params per instance (collected across all server blocks, so split-listener deployments work), gated behind the new opt-in `experimental_chunking` CAS service config.
- **Observability**: `ChunkingMetrics` — splice/split request totals, split hit/miss/on-demand counters, byte totals, and digest verification failures (tracked precisely via an explicit flag, not error-code inference).

## Conformance

The chunker is the `fastcdc` crate at **3.2.1** — the exact version the REAPI spec names as a compliant implementation — used with the spec-mandated normalization level 2 (the crate default level 1 does **not** match). `fastcdc_conformance_test.rs` validates chunk offsets, lengths, SHA-256s, and gear-hash fingerprints against the official `fastcdc2020_test_vectors.txt` from remote-apis, for both seed 0 and seed 666, in both in-memory and streaming form. This is the guarantee that server-produced chunks dedupe byte-for-byte against Bazel-produced chunks. The canonical fixture image is the one already vendored at `nativelink-util/tests/data/SekienAkashita.jpg` for the existing FastCDC tests (its SHA-256 is asserted in the test).

Note: the vendored `nativelink-util/src/fastcdc.rs` used by `DedupStore` is *not* FastCDC-2020-compliant and is deliberately untouched — deployed dedup indexes depend on its exact boundaries; the two implementations coexist.

## Safety / hardening

- Digests are never trusted: splice verifies the re-assembled digest before commit; per-chunk content lengths are validated against their digests.
- `index_store == cas_store` is rejected at startup (layouts stored under blob digests would overwrite blob content).
- Setting `index_store` on a grpc-store instance is rejected (the backend owns layouts there).
- Per-chunk size cap (16 MiB) and a configurable per-instance `max_chunk_count` (default 50k) bound memory, layout size, and response size; layout reads are capped consistently with the configured count and truncation is detected by the size-consistency check.
- Off by default; with `experimental_chunking` unset, behavior is byte-for-byte identical to before (new RPCs return `Unimplemented`, capabilities advertise `false`).

## Design decisions worth reviewer attention

1. **Materialize-on-splice**: the spliced blob is written whole into the CAS rather than serving reads from the layout. This keeps every existing read path correct and lets storage-level dedup compose naturally when `cas_store` is a `DedupStore`; the cost is chunks + blob both stored. A `ChunkingStore` that serves reads from layouts is a possible future direction.
2. **Chunk/index lifetime coupling is best-effort**: existence checks touch chunks, but stores answering from existence caches may not promote backing entries, and layouts can outlive chunks. Degradation is graceful (fallback to re-chunking or full download), never incorrect. A `CompletenessCheckingStore`-style wrapper is the analogous future fix.
3. **`ChunkingMetrics` follows the `ByteStreamMetrics` pattern** (not yet wired into a metrics root — same as existing service metrics).

## Testing

- 23 service tests: splice/split round trip, digest & size mismatch rejection, missing-chunk `NotFound`, on-demand chunking (single- and multi-chunk with reassembly verification), layout reuse, unusable-layout fallback, absent-blob `NotFound`, disabled-instance `Unimplemented`, config rejection cases, and metric assertions throughout.
- 2 conformance tests against upstream vectors.
- GrpcStore forwarding tested against a fake CAS backend over a real gRPC round trip (passthrough + instance-name rewriting).
- `bazel test` green across `nativelink-service`, `nativelink-store`, `nativelink-config`, `nativelink-util` (63 tests) with clippy + rustfmt aspects; `cargo fmt --check` clean; `bazel build //:nativelink` succeeds.

## Notes

- **Verified end-to-end against a real Bazel client** (see PR comment for full numbers): incremental uploads of a ~30 MB artifact drop 99.4% for a single-point edit (183 KB) and **83.9% for six scattered edits (4.96 MB)** — in line with published production CDC figures — with fetched bytes SHA-verified identical through both the ByteStream and `SplitBlob` paths.
- Blobs exceeding the configurable `max_chunk_count` (default 50k, ~25 GiB at the default 512 KiB average) are served without chunking: `SplitBlob` returns `NOT_FOUND` and clients fall back to a regular download. Raise the knob for larger blobs, keeping client gRPC message limits in mind.
- The chunking flag requires Bazel 8.7+/9.1+ on the client; the repo's current Bazel pin (9.0.2) predates it, so the LRE E2E jobs will start exercising this path once the pin moves past 9.1. Recommend **9.1.1+** specifically: 9.1.0 has an upstream client bug that corrupts outputs when `--experimental_remote_cache_chunking` is combined with `--disk_cache` (fixed in 9.1.1; reproducible against any chunking server).

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2497)


MarcusSorealheis (commented):
One comment, really.

For new configuration options, we'd like to clarify that this is optional and include a configuration example. See [./nativelink-config](https://github.com/TraceMachina/nativelink/tree/main/nativelink-config) and lots of examples.

Otherwise, this one is definitely on the correct path.

erneestoc:
Thanks for the review! Addressed in ed980638:

- Added a complete configuration example at `nativelink-config/examples/chunking_cas.json5` (validated by the existing `test_example_parsing` config test), with each field annotated as optional/required and its default.
- The `experimental_chunking` doc comment now states explicitly that the option is optional and that behavior is byte-for-byte unchanged when it is unset, and points to the example.

erneestoc:
**End-to-end verification against a real Bazel client** (NativeLink from this branch + `--experimental_remote_cache_chunking`, wire bytes measured at the TCP socket with a byte-counting proxy). Each row: incremental re-upload after a small change, chunking off vs on.

| Artifact (change) | Upload, chunking off | Upload, chunking on | savings |
|---|---|---|---|
| 30 MB text, 1 edit at front (CDC best case) | 30.7 MB | 183 KB | 99.4% |
| 30 MB text, 6 scattered edits | 30.7 MB | 4.96 MB | 83.9% |
| **16.8 MB tar layer, 1 file changed** | 16.8 MB | 2.1 MB | **87.5%** |
| **3.4 MB clang-linked binary, 1 source file changed** | 3.4 MB | 653 KB | **80.8%** |
| 7.2 MB gzip of the tar (change near end of stream) | 7.2 MB | 3.3 MB | 54.2% |

Representative artifacts (tar layers, linked binaries, scattered edits) cluster at **~80–88%**, matching BuildBuddy's published production figure of ~85% deduped bytes. Compressed artifacts are position-dependent: everything after the first changed byte re-uploads (a front-of-stream change would approach 0% savings) — inherent to CDC, not this implementation. Download side mirrors upload: a fresh client re-fetching the edited text artifact downloaded 178 KB instead of 30.7 MB.

Correctness verified end-to-end: fetched artifacts are byte-identical to expected (SHA-256) via both the plain ByteStream path (validates splice materialization) and the chunked `SplitBlob` path. Server-side stores show the expected mechanics (shared chunks + layout entries, no errors). Capability gating verified: with the server not advertising chunking, a flag-enabled client falls back to full uploads.

One upstream finding: **Bazel 9.1.0's** `--experimental_remote_cache_chunking` combined with `--disk_cache` reassembles corrupt outputs (reproducible against any server; the RPCs the server serves are identical and correct). **Bazel 9.1.1 fixes it** — worth recommending 9.1.1+/8.7+ in docs.

MarcusSorealheis (changes requested):
We need the `web/` documentation as well, as that's where most customers live as opposed to the repo.

corcillo:
Cool addition and impressive savings! @MarcusSorealheis ill leave it to you to approve since you had comments but i think theyve been addressed, namely via `web/apps/docs/content/docs/configuration/chunking.mdx`, and specifically highlights the 8.7/9.1.1 Bazel req as well.

MarcusSorealheis (approved):

PR #2503, opened 2026-07-04
evict .exec variant when its digest is evicted

### LINKED ISSUE #2474: Executable-variant cache (.exec) is not tracked by max_bytes and never evicted at runtime
## Summary

`FilesystemStore` maintains a per-digest on-disk executable-variant cache at `{content_path}.exec`, but those variants are never inserted into `evicting_map`. As a result they are not counted against `max_bytes` and are never reclaimed at runtime — the only cleanup is a `remove_dir_all` of the directory on store startup. On a worker doing executable-heavy input materialization, .exec grows ≈1:1 with the real CAS, roughly **doubling** the on-disk footprint and driving the host past `max_bytes` until it runs out of disk.

## Details

In `nativelink-store/src/filesystem_store.rs`:
- `get_executable_hardlink_source` → `create_executable_variant` does a `std::fs::copy` of the `0o444` CAS blob into `{content_path}.exec/d/{digest}` and chmods it `0o555`.
- The variant is never added to `evicting_map`, so it is invisible to `max_bytes` accounting.
- The only reclamation is `fs::remove_dir_all({content_path}.exec`) at store startup. There is no runtime eviction, no size cap, and no config flag to gate the behavior.

## Impact

On long-lived workers, disk usage climbs monotonically past `max_bytes` until the host hits a disk-full / disk-pressure condition (e.g. pod eviction in Kubernetes). 

## Origin

Introduced with the executable-variant materialization path added to fix the ETXTBSY / "Text file busy" race (#2366, addressing #2362).

MarcusSorealheis:
# Description

Evict .exec variant when the digest is evicted. Pretty simple..

Fixes #2474 

## Type of change

## How Has This Been Tested?

`bazel test`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2503)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -786,6 +786,50 @@ where
     Ok(false)
 }
 
+/// Deletes a digest's `.exec` variant (see
+/// [`FilesystemStore::get_executable_hardlink_source`]) when that digest is
+/// evicted or replaced in the primary CAS `evicting_map`. Without this, the
+/// `.exec` directory is invisible to `max_bytes` and is only ever cleared by
+/// the startup `remove_dir_all`, so it grows without bound at runtime (#2474).
+/// Tying its lifetime to the primary entry instead bounds total disk use to
+/// roughly `2 * max_bytes` in the worst case (every blob also executable).
+#[cfg(unix)]
+#[derive(Debug)]
+struct ExecutableVariantRemover {
+    content_path: String,
+}
+
+#[cfg(unix)]
+impl RemoveItemCallback for ExecutableVariantRemover {
+    fn callback<'a>(
+        &'a self,
+        store_key: StoreKey<'a>,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + 'a>> {
+        Box::pin(async move {
+            let StoreKey::Digest(digest) = store_key else {
+                return;
+            };
+            let variant_path = format!(
+                "{}{EXECUTABLE_DIR_SUFFIX}/{DIGEST_FOLDER}/{digest}",
+                self.content_path
+            );
+            match fs::remove_file(&variant_path).await {
+                Ok(()) => debug!(
+                    ?variant_path,
+                    "Deleted executable variant for evicted digest"
+                ),
+                // Common case: no variant was ever materialized for this digest.
+                Err(err) if err.code == Code::NotFound => {}
+                Err(err) => warn!(
+                    ?variant_path,
+                    ?err,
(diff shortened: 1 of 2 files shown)


corcillo (approved):

corcillo:
@MarcusSorealheis wondering thinking between your approach (deleting associated exec files on the call back post deletion, thus not counting in overall running size count), vs keeping track of the exec files in the size count? Basically if you wait to do the exec eviction like this the size count wont technically be accurate as it will be missing those exec sizes.

corcillo:
Also, if a digest gets evicted while its .exec copy is still being created, the eviction's cleanup runs too early before there's anything to delete? Then the copy would get published with no owner and still leak into memory until restart.  Not sure if this will happen a lot based on timing or if its an edge case?

MarcusSorealheis:
It's a good question.. I did consider giving .exec its own tracked size in evicting_map and went with the callback/tether approach instead, for a few reasons. 

The first is, `EvictingMap` calls `len()` once at insert to add to `sum_store_size`, and again at removal to subtract it ([evicting_map.rs:150](https://github.com/TraceMachina/nativelink/blob/6dca947a135bea52585ebaa4a627510656e02b62/nativelink-util/src/evicting_map.rs#L150), [:645-656](https://github.com/TraceMachina/nativelink/blob/6dca947a135bea52585ebaa4a627510656e02b62/nativelink-util/src/evicting_map.rs#L645-L656)). Sothere's no resize hook in between. If a digest's `len()` grew after insert because an executable variant got materialized later, removal would subtract a bigger number than was ever added, underflowing `sum_store_size`. This is the core of the issue as I understand it today. @dkostyrev let me know if you have thoughts here. 

The actual failure that @dkostyrev pointed out is unbounded disk growt:

> roughly doubling the on-disk footprint and driving the host past max_bytes until it runs out of disk.

and here, deleting a digest's .exec sibling via the existing remove-callback hook whenever that digest is evicted from the primary map bounds total disk to ~2x `max_bytes` absolute worst-case  instead of unbounded.

Also, I did not want to add any additional config options or accounting if not absolutely necessary.

corcillo:
@MarcusSorealheis Ok so for the race condition I mentioned, can you do something to recheck membership after create_executable_variant publishes the variant, so if the digest was evicted mid-copy and the exec file is now orphaned, we follow up by deleting it and returning NotFound?

MarcusSorealheis:
> can you do something to recheck membership after create_executable_variant publishes the variant

yes, good thinking. hold my beer!

PR #2521, opened 2026-07-07
Redis eviction events support

### LINKED ISSUE #2436: Existence Cache serves stale existence after Redis maxmemory eviction
```
Remote Execution Failure:
Failed Precondition: status: NotFound, message: "Object xxxx not found in either fast or slow store. If using multiple workers, ensure all workers share the same CAS storage path
```

When using an Existence Cache that wraps a RedisStore and Redis is configured to use an evicting maxmemory-policy (e.g. allkeys-lru), the existence cache keeps reporting blobs present after Redis evicts the keys causing build failures as shown above. 

Seems like after the change in #1947 the remove-callbacks in redis_store.rs makes it a silent no-op, refer comment [here](https://github.com/TraceMachina/nativelink/pull/1947/changes#diff-eb7e886d565c46adc8ede5a01a3b309e6d543a8343147eddffabfe672acc8613R665).

Seems like that comment may not be true as already noted [here](https://github.com/TraceMachina/nativelink/pull/1981/changes#r2433657261).

palfrey:
# Description

Subscribes to Redis eviction events so we can expire keys and solve problems with existence cache on top of Redis.

As a result of this PR, if you are using Redis together with any stores above it e.g. existence cache, you will need to configure `notify-keyspace-events` to `KA` as per https://redis.io/docs/latest/develop/pubsub/keyspace-notifications/#configuration in order for us to get eviction events. Failing to do so will get you log messages complaining about it, as well as errors like https://github.com/TraceMachina/nativelink/issues/2436. Not doing so gets you the identical behaviour as to prior to this work, so it's not a breaking change as such, but it will complain about the situation :)

Fixes #2436

## Type of change

## How Has This Been Tested?

`bazel run :redis_store_tester` / `bazel test //...`

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2521)


MarcusSorealheis:
@palfrey i feel this change explicitly needs a documentation update. Could you give it one?

palfrey:
> @palfrey i feel this change explicitly needs a documentation update. Could you give it one?

There's an update in `web/apps/docs/content/docs/reference/nativelink-config/main.mdx`. Are you looking for more than this?

nativelink-store/src/redis_store.rs
```diff
@@ -1293,11 +1305,38 @@ where
...
-        // As redis doesn't drop stuff, we can just ignore this
+    fn register_remove_callback(self: Arc<Self>, callback: RemoveCallback) -> Result<(), Error> {
+        debug!(?callback, "New callback");
+        let local_self = self.clone();
+        background_spawn!("remove_callback_subscribe", async move {
+            self.remove_callbacks.lock().await.push(callback);
+            if let Err(err) = local_self.clone().has_remove_callback_subscribe
+                .get_or_try_init(|| async move {
+                    let mut client = local_self.get_client().await?;
+                    let cfg = redis::cmd("CONFIG").arg("GET").arg("notify-keyspace-events").to_owned().query_async::<Vec<(String,String)>>(&mut client.connection_manager).await.expect("Parsing notify-keyspace-events");
```
MarcusSorealheis: Please don't `expect` on `CONFIG GET` here, and don't make the config probe required for subscribing. `register_remove_callback` returns `Ok(())`, but this work runs in a background task. If `CONFIG GET notify-keyspace-events` is denied or unsupported, the task panics before `PSUBSCRIBE`, callbacks never start, and the caller cannot observe the failure

MarcusSorealheis: Treat this probe as best-effort diagnostics. Log the Redis error, still attempt `PSUBSCRIBE`, and avoid `expect` while parsing the response. If subscription failure should block startup, move the initialization into a path that can return an error, such as `post_init`.

Redis keyspace notifications can be enabled through redis.conf or `CONFIG SET`, but Pub/Sub delivery is fire-and-forget: https://redis.io/docs/latest/develop/pubsub/keyspace-notifications/.

And, async command execution returns a Redis result and [should be handled instead of panicking](https://docs.rs/redis/1.0.0/redis/struct.Cmd.html#method.query_async).

nativelink-store/src/redis_store.rs
```diff
@@ -1693,7 +1736,36 @@ impl RedisSubscriptionManager {
...
+                                                error!(?push_info, "No key in eviction event");
+                                                continue;
+                                            };
+                                            trace!(?eviction_key, "Eviction key");
+                                            let Some((_prefix, internal_key)) = eviction_key.split_once(':') else {
+                                                error!(?eviction_key, "Eviction key doesn't contain a colon");
+                                                continue;
+                                            };
+
+                                            let store_key = StoreKey::new_str(internal_key).into_owned();
```
MarcusSorealheis: This needs to decode the Redis physical key back to the original logical `StoreKey` before invoking remove callbacks. Redis keyspace notifications put the affected Redis key in the keyspace channel, e.g. `__keyspace@0__:<key>`, with the event name (`evicted`) as the message. This code strips only the Redis channel prefix, then always constructs `StoreKey::Str(internal_key)`.

 That is wrong for existence-cache-over-CAS keys. `RedisStore::encode_key` writes digest keys as `key_prefix + key.as_str()`, where a `StoreKey::Digest` becomes the textual `<hash>-<size>` form. But `ExistenceCacheStore::callback` calls `store_key.into_digest()`, and for `StoreKey::Str` that hashes the string instead of parsing `<hash>-<size>` back into the original `DigestInfo`. So even an unprefixed evicted digest key will not remove the cached digest entry; with `key_prefix: "cas:"`, it is worse because the callback sees `cas:<digest>`.

  Please add an inverse key decode path here: strip/validate `self.key_prefix`, then reconstruct `StoreKey::Digest` for digest-shaped keys used by CAS/existence-cache paths, or otherwise preserve the original key kind. Add a regression test with an existence cache above Redis for both an unprefixed digest key and a non-empty `RedisSpec.key_prefix`.

MarcusSorealheis (changes requested):

MarcusSorealheis:
> There's an update in \web/apps/docs/content/docs/reference/nativelink-config/main.mdx`. Are you looking for more than this?

nope. sufficient.

nativelink-store/src/redis_store.rs
palfrey: Fixed

nativelink-store/src/redis_store.rs
palfrey: Fixed all of that.

MarcusSorealheis (approved):

PR #2526, opened 2026-07-07
Materialize directory-cache nodes concurrently on cold misses

erneestoc:
## Summary

`DirectoryCache::construct_directory` materializes every node of a directory **sequentially** — each file, subdirectory recursion, and symlink is awaited one at a time. Every file that is not in the fast tier pays a full slow-store round trip, so cold-cache construction of a large input tree serializes N round trips. On a production workload (iOS Bazel build, ~1400 input files, remote slow store at ~50ms) this showed up as **~70 seconds of `prepare` time for an action with kilobytes of actual output** — the time is round trips, not bytes.

This PR does two things:

1. **Concurrent materialization, same shape as the peer.** All nodes of a directory level (files, subdirectory recursions, symlinks) share one `buffer_unordered(64)` budget — the exact pattern and bound `download_to_directory` already uses, for the same two reasons: overlap slow-store round trips, and keep hardlink/copy syscalls from fighting the filesystem's metadata locks.
2. **A cache-wide bound on aggregate slow-store load.** The per-level budget compounds multiplicatively across tree levels and concurrent actions — a stress run measured **>1800 concurrent slow-store fetches** for 6 concurrent ~500-file trees (`download_to_directory` has the same multiplicative property today). A cache-wide semaphore caps total in-flight slow-store fetches at 64. Permits are held only across leaf I/O (directory-proto fetches, blob fetches/populates) and **never across recursion**, so recursion depth cannot deadlock the semaphore.

## Measurements

A/B benchmark against a latency-injecting slow store (`MemoryStore` wrapper adding a fixed delay per `get_part`), real `DirectoryCache`, cold construction of a 3-level tree with 256 files + 21 directory protos:

| Simulated slow-store RTT | Sequential (before) | Concurrent + bounded (after) | Speedup |
|---|---|---|---|
| 5 ms | 3857 ms | 495 ms | **7.8x** |
| 25 ms | 9554 ms | 803 ms | **11.9x** |

The ratio grows with RTT (the serialized term is N x RTT), so remote slow stores benefit the most. Warm cache hits are unaffected (1–3 ms in both versions; they never enter this path). Extrapolating with the measured per-fetch fixed cost reproduces the ~70 s observed in production. Worst-case slow-store concurrency during these runs: 64 (capped), versus ~500 uncapped for a single tree of this shape.

## Testing

- `cold_construct_fetches_concurrently`: regression test asserting cold construction overlaps slow-store fetches; verified it **fails against the sequential implementation**.
- `stress_concurrent_construction`: 24 concurrent materializations over 4 distinct trees (6x duplicate requests per tree → single-flight stampede) with eviction churn (`max_entries: 2`), digests shared across trees (hardlink inode sharing), executables (copy path), symlinks, and empty files (zero-digest path), against a jittered-latency slow store. Byte-verifies every materialized tree (contents, +x bits, symlink targets) and asserts the aggregate in-flight fetch bound holds. Runs in ~5s.
- Full `//nativelink-worker:integration` suite passes (eviction accounting, single-flight construction, hardlink semantics unchanged); `cargo check --tests` clean.
- Ordering note: per-node work was already order-independent — files are independent blobs, subdirectory sizes are summed, and `total_size` accounting is digest-derived — so unordered completion does not change any observable result.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2526)


erneestoc:
Force-pushed an update after stress-testing this at higher concurrency — one finding worth reviewer attention:

**Finding:** the 64 budget applies per directory level, so it compounds multiplicatively across tree levels and concurrent constructions. A stress run (36 concurrent materializations over 6 ~500-file trees) measured **>1800 concurrent slow-store fetches**. `download_to_directory` has the same multiplicative property today, so this is not a new risk class — but extending the pattern to a second path was a good moment to bound it.

**Change:** a cache-wide semaphore now caps total in-flight slow-store fetches at 64 across all constructions. Permits are held only across leaf I/O (directory-proto fetches, blob fetches/populates) and never across subdirectory recursion, so recursion depth cannot deadlock it. Measured peak in-flight is exactly 64 with the headline speedup intact (9554ms → 803ms at 25ms simulated RTT, still 11.9x over sequential).

**New test:** `stress_concurrent_construction` — 24 concurrent materializations over 4 trees with duplicate requests (single-flight stampede), eviction churn, digests shared across trees (hardlink inode sharing), executables, symlinks, and empty files. Byte-verifies every materialized tree (contents, +x bits, symlink targets) and asserts the aggregate in-flight bound holds. ~5s runtime.

Rationale for bounding client-side rather than letting the backing store queue: at saturation the server's throughput ceiling sets completion time either way — queueing in the worker avoids connection churn and 503/retry amplification against a shared backing store.

palfrey (approved):
Nice. LGTM.

PR #2533, opened 2026-07-08
Reject duplicate store names in StoreManager

### LINKED ISSUE #1998: Multiple stores with the same name allowed in config
I accidentally added two stores with the same name in to the stores config and it didn't throw an error and everything broke.  This should panic.

heyitsalec:
# Description

`StoreManager::add_store` registered stores with `HashMap::insert`, which
silently overwrites any existing entry. A config declaring two stores with the
same `name` dropped the first store with no error.

`add_store` now checks whether the name is already registered
and returns `Err(Code::AlreadyExists)` instead of overwriting.   I opted for enforcing uniqueness on `StoreManager` (rather than only in the config loader) to protect every caller and composes with the existing parse-time`check_store_conflict`. The startup wiring loop propagates the error with `err_tip`, so a duplicate-name config now fails fast at startup with a clear message instead of silently discarding a store.

Fixes #1998

## Type of change

## How Has This Been Tested?

Added an inline unit test in `store_manager.rs`
(`add_store_rejects_duplicate_name`) that registers a store twice under the same
name and asserts the second call returns `Code::AlreadyExists`. All existing
`add_store` call sites were updated to the fallible signature.

relying on this PR's CI (`bazel test //...`, which also runs `rustfmt`) as the verification gate
rather than a local run. Happy to iterate quickly if CI flags anything.

## Checklist

- [x] Tests added/amended
- [x] `bazel test //...` passes locally

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2533)


nativelink-store/tests/ref_store_test.rs
```diff
@@ -32,15 +32,17 @@ async fn setup_stores() -> (Arc<StoreManager>, Store, Store) {
     let store_manager = Arc::new(StoreManager::new());
 
     let memory_store = Store::new(MemoryStore::new(&MemorySpec::default()));
-    store_manager.add_store("foo", memory_store.clone());
+    store_manager
+        .add_store("foo", memory_store.clone())
+        .unwrap();
```
chrisstaite-menlo: Would be better to use `expect` with a reason here, but otherwise an awesome patch, thanks.

nativelink-store/src/store_manager.rs
```diff
@@ -58,3 +65,25 @@ impl StoreManager {
 }
 
 impl RootMetricsComponent for StoreManager {}
+
+#[cfg(test)]
+mod tests {
```
chrisstaite-menlo: Sorry, the structure of the repo is to split tests out.  Could this be moved to `nativelink-store/tests/store_manager_test.rs` please?

heyitsalec: Done — moved the test to `nativelink-store/tests/store_manager_test.rs` and registered it in the `integration` `rust_test_suite` in `nativelink-store/BUILD.bazel`. Also updated the commit message to match. Thanks for the pointer!

nativelink-store/tests/ref_store_test.rs
heyitsalec: Done — both `add_store` calls in `setup_stores` now use `.expect(...)` with a reason (e.g. `.expect("adding memory store 'foo' should succeed")`). Thanks!

chrisstaite-menlo (approved):
Thanks.

PR #2539, opened 2026-07-09
Coalesce FilesystemStore durability flushes on macOS

erneestoc:
# Coalesce FilesystemStore durability flushes on macOS (userspace group commit)

## Problem

Every `FilesystemStore` upload makes its blob durable with `File::sync_all`
before publishing it into the content directory. On macOS, `sync_all` issues
`fcntl(F_FULLFSYNC)` — a full device-cache flush — once per blob. The flush
serializes at the device and costs multiple milliseconds, so workloads that
upload thousands of small blobs (a worker's fast tier materializing a
many-tiny-file action input tree) spend nearly all of their time in flushes:
a 4,400-file / 2 MB input tree takes 14.5 s cold (21.0 s under contention),
of which ~12 s is `F_FULLFSYNC`, while `tar -x` of the same tree takes ~1.4 s.

Linux does not suffer from this because concurrent `fsync` calls coalesce
inside the filesystem journal's group commit. macOS has no journal-level
coalescing and no `syncfs(2)`, and `F_BARRIERFSYNC` measures just as slow
per-file here (~15.6 s).

## Change: group commit in userspace (macOS only, no config)

`F_FULLFSYNC` has an exploitable asymmetry: writing one file's pages to the
storage device is per-file work (plain `fsync(2)` — measured effectively
free), but the expensive device-cache drain is device-wide by nature. So a
`FlushCoalescer` inside `FilesystemStore` applies the ext4 trick one layer
up:

1. Each writer pushes its own blob to the device with `fsync(2)` (cheap).
2. It then joins the current commit round. One round leader issues a
   single `F_FULLFSYNC` (on a dedicated sentinel file in the content root)
   that makes every previously-`fsync`ed blob durable at once.
3. Rounds self-batch exactly like a journal: while one flush runs, later
   writers accumulate into the next round. A ~3 ms accumulation window
   (about the cost of one flush) fattens rounds further.
4. Only after the covering flush completes does a writer rename its blob
   into the content directory.

**Durability is unchanged**: a blob is only published after a device-cache
flush that started after its own data reached the device — the same
guarantee as per-file `F_FULLFSYNC`, at ~1/10th the flush count. A solo
(unbatched) write pays one flush plus the 3 ms window, roughly what it pays
today. Linux and Windows keep plain `sync_all` (their journals already
group-commit; behavior byte-identical).

## Measurements

Real-stack benchmark (DirectoryCache → FastSlowStore(FilesystemStore fast) →
GrpcStore → in-process ByteStream/CAS servers over TCP), 4,400-file / ~450 B
tree, Apple Silicon / APFS, output trees byte-verified every run:

| configuration | alone | contended (5×800-file neighbors) |
|---|---|---|
| per-file `F_FULLFSYNC` (today) | 14.5 s | 21.0 s |
| **coalesced (this PR)** | **~3.0 s** | **~5.6 s** |
| no flush at all (floor) | 2.2 s | 4.1 s |
| `tar -x` reference | ~1.3–1.6 s | — |

4.8× / 3.7× with full durability, within ~35% of the physically possible
no-flush floor (contended runs ranged 4.9–6.3 s across iterations).

## Relationship to a possible sync_policy knob

A companion branch (not yet a PR) adds `FilesystemSpec.sync_policy`
("full" | "data" | "fsync" | "none", default "full") for deployments that
want to trade durability for the last ~25% on re-fetchable cache tiers.
The two compose — this coalescer would become the implementation of `full`
on macOS — and neither depends on the other; happy to open it as a
follow-up if there's interest.

## Tests

- `concurrent_upload_burst_round_trip_test`: 200 concurrent uploads through
  the coalescer, all round-trip and survive a store restart.
- Existing store suite unchanged and green (the coalescer sits strictly
  behind the old `sync_all` call sites), including the #2051 open-file
  semaphore-exhaustion regression test, which caught (and now guards) the
  permit-deadlock hazard described below.
- Also validated merged together with the read-coalescing (#2540) and
  subtree-caching (#2541) PRs: zero conflicts, all 40 bazel tests across
  config/store/worker green on the combined branch.

## Post-review hardening (multi-angle review applied)

- Sentinel moved to a SIBLING of the content path (`{content_path}.flush_sentinel`,
  the `.exec`-directory precedent): the startup scan's legacy root sweep
  would otherwise migrate and delete a sentinel inside the content root on
  every restart.
- Sentinel creation is non-fatal and off the async thread: a read-only
  content volume degrades to per-file `sync_all` instead of failing store
  construction (macOS-only regression otherwise).
- The blocking fsync task owns a dup'd fd (`try_clone`), so caller
  cancellation can never race a reused descriptor. It deliberately does
  NOT take an open-file permit: the caller's `FileSlot` already holds one,
  and requiring a second deadlocks under a drained semaphore (caught by
  the #2051 regression test, which pins the semaphore at one permit).
  Concurrency is still bounded by the callers' own permits.
- `EINTR` retried exactly as std's `sync_all` (`cvt_r`) does, for both
  `fsync(2)` and `F_FULLFSYNC`.
- A send-on-drop guard makes a cancelled flusher round error its waiters
  instead of stranding them; the sentinel dirty-write is best-effort so an
  ENOSPC on the 8-byte write cannot fail a whole round.
- The 3 ms accumulation window only applies to rounds with >1 waiter — a
  solo write on an idle store pays roughly today's latency.
- Leader-election was replaced by a single long-lived flusher task per
  store (simpler audit surface; serialization is inherent). Sharing one
  flusher per *device* (`st_dev`) would amortize across stores on one
  volume but requires a runtime-agnostic flusher (dedicated OS thread) —
  multiple tokio runtimes per process (tests) break a task-based global —
  left as a documented follow-up.
- The executable-variant copy path drops its flush entirely: the `.exec`
  directory is cleared on every startup, so durability there bought
  nothing while costing one `F_FULLFSYNC` per executable digest.

## Notes for reviewers

- The sentinel file lives at `{content_path}.flush_sentinel` (a sibling of
  the content root), and is re-dirtied (best-effort) each round so the
  flush is never elided.
- The flusher is one detached long-lived task per store, woken by a
  `Notify`; writers notify strictly after subscribing to their round, so
  lost wakeups are impossible, and `FlushCoalescer::drop` wakes the task
  so it exits instead of parking forever.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2539)

### DIFF
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -378,6 +378,227 @@ pub fn make_temp_key(key: &StoreKey) -> StoreKey<'static> {
     StoreKey::Digest(make_temp_digest(key.borrow().into_digest()))
 }
 
+/// Group-commit flush coalescer (macOS only).
+///
+/// On macOS, `File::sync_all` issues `fcntl(F_FULLFSYNC)` — a full
+/// device-cache flush — once per call. The flush is serialized at the
+/// device, costs multiple milliseconds, and dominates uploads of many
+/// small blobs (a 4,400-tiny-file action input tree spends ~12s of its
+/// ~14.5s materialization in these flushes). Linux does not have this
+/// problem because concurrent `fsync` calls coalesce inside the
+/// filesystem journal's group commit.
+///
+/// This applies the same group-commit idea in userspace, exploiting an
+/// asymmetry in `F_FULLFSYNC`: writing a file's pages to the storage
+/// device is per-file work (plain `fsync(2)`, cheap), while the expensive
+/// device-cache drain is device-wide by nature. Each writer first pushes
+/// its own data to the device with `fsync(2)`, then joins the current
+/// commit round; a single long-lived flusher task per DEVICE issues one
+/// `F_FULLFSYNC` (on a dedicated sentinel file) per round, covering every
+/// previously-`fsync`ed blob at once. Rounds self-batch exactly like a
+/// journal: while one flush is running, later writers accumulate into
+/// the next round.
+///
+/// The coalescer (and its flusher task) is per store instance. Sharing
+/// one coalescer across all stores on a device would amortize further —
+/// the flush is device-wide — but a process-global flusher task cannot
+/// safely outlive the tokio runtime that spawned it (multiple runtimes
+/// coexist in one process, e.g. one per test); doing this correctly
+/// needs a runtime-agnostic flusher (a dedicated OS thread) and is left
+/// as a follow-up.
+///
+/// Durability is identical to per-file `F_FULLFSYNC`: a writer only
+/// proceeds (and only renames its blob into the content directory) after
+/// a device-cache flush that started after its own `fsync(2)` completed.
+#[cfg(target_os = "macos")]
+#[derive(Debug)]
+struct FlushCoalescer {
+    /// The round currently accepting waiters. Swapped out by the flusher
+    /// right before it flushes, so writers whose `fsync(2)` finished
+    /// after the flush began land in the next round.
+    current_round: parking_lot::Mutex<Arc<FlushRound>>,
+    /// Wakes the flusher task. Writers notify after subscribing to their
+    /// round, so a wakeup can never be observed before its waiter.
+    wake: Arc<tokio::sync::Notify>,
+    /// Dedicated file the `F_FULLFSYNC` is issued on. A SIBLING of the
+    /// store's content path (like the `.exec` variant directory), so
+    /// neither the startup scan nor `move_old_cache`'s legacy root sweep
+    /// ever sees it.
+    sentinel: Arc<std::fs::File>,
+}
+
+#[cfg(target_os = "macos")]
+#[derive(Debug)]
+struct FlushRound {
+    /// Broadcasts the flush outcome to every writer in the round.
+    result_tx: tokio::sync::watch::Sender<Option<Result<(), Error>>>,
+}
+
+#[cfg(target_os = "macos")]
+impl FlushRound {
+    fn new() -> Arc<Self> {
+        let (result_tx, _) = tokio::sync::watch::channel(None);
+        Arc::new(Self { result_tx })
+    }
+}
+
+/// Guarantees a [`FlushRound`]'s waiters always receive a result: if the
+/// flusher task is cancelled at an await point (runtime shutdown), waiters
+/// must get an error rather than pend forever — their own `Arc<FlushRound>`
+/// keeps the channel alive, so a dropped-sender wakeup can never happen.
+#[cfg(target_os = "macos")]
+struct SendOnDrop(Option<Arc<FlushRound>>);
+
+#[cfg(target_os = "macos")]
+impl SendOnDrop {
+    fn finish(mut self, result: Result<(), Error>) {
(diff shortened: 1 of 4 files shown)


nativelink-store/src/filesystem_store.rs
```diff
@@ -1097,12 +1341,11 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
...
-                let f = std::fs::File::open(&temp_owned)
-                    .map_err(|e| make_err!(Code::Internal, "executable-variant reopen: {e:?}"))?;
-                f.sync_all()
-                    .map_err(|e| make_err!(Code::Internal, "executable-variant fsync: {e:?}"))?;
-                drop(f);
+                // No flush: the `.exec` directory is cleared on every
+                // startup (see `new_with_timeout_and_rename_fn`), so a
+                // variant is never trusted across a crash — durability
+                // would buy nothing, and on macOS `sync_all` is a
+                // multi-ms full device-cache flush per executable.
```
palfrey: Shouldn't this still get done on non-Mac unix systems?

palfrey (changes requested):

nativelink-store/src/filesystem_store.rs
erneestoc: I don't think it buys anything on any platform: the `.exec` variant directory is unconditionally wiped on every startup (`drop(fs::remove_dir_all(&executable_dir).await)` in `new_with_timeout_and_rename_fn`, filesystem_store.rs:893), so a variant can never survive a crash to be trusted afterwards — the flush was guarding a file that's deleted before it could ever be read again. Process crashes don't need it either (page cache survives the process), and the ETXTBSY protocol only requires the writer fd to be *closed* before the rename, not flushed.

The fsync was only load-bearing if the variants were meant to persist across restarts — and they deliberately aren't (they're regenerable, and the startup clear exists so a stale variant can't leak across runs).

That said, it's cheap on Linux — if you'd prefer belt-and-suspenders on non-macOS I'm happy to restore it behind `#[cfg(not(target_os = "macos"))]`. Just flagging that it would be dead code by the above reasoning rather than a platform behavior difference.

palfrey: I think I'd prefer belt-and-suspenders, just to preserve prior behaviour

erneestoc: Done in the latest push (also rebased onto main): the pre-publish flush is restored — `sync_all` on non-macOS exactly as before, and plain `fsync(2)` on macOS (keeps the variant valid across kernel panics without the multi-ms `F_FULLFSYNC` device-cache drain; the startup wipe covers power loss).

While re-checking this I also noticed the startup wipe itself swallowed errors (`drop(fs::remove_dir_all(..))`), which was the one path where a stale/torn variant could still have been served by the hardlink fast path — it now aborts store construction on any failure other than NotFound. So belt, suspenders, and the belt-loop the suspenders clip to.

MarcusSorealheis: lol

MarcusSorealheis: LGTM now

MarcusSorealheis (approved):
LGTM

palfrey (approved):

PR #2545, opened 2026-07-10
Remove StoreSpec docs

palfrey:
# Description

We generated two copies of the store docs, one called `NamedStoreConfig` with the `name` field spelt out, and the second called `StoreSpec` which did not. This PR drops the `StoreSpec` doc in favour of the `NamedStoreConfig` doc.

## Type of change

## How Has This Been Tested?

`bun --filter @nativelink/docs gen:config-reference main`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2545)


web/apps/docs/scripts/lib/schema-to-mdx.mjs
```diff
@@ -44,6 +44,7 @@ function escapeTableCell(text) {
 
 const displayNames = new Map([
   ["NamedConfig", "NamedStoreConfig"],
+  ["StoreSpec", "NamedStoreConfig"],
```
MarcusSorealheis, line 47: looks like NameStoreConfig is duplicated. Also look at this map. It feels confusing. It may not be but take a look. Two `NamedStoreConfig`s and a `NamedConfig2`, and `GrpcSpec2`.

MarcusSorealheis (approved):
ahh, it's confusing but we can iterate and improve later.

web/apps/docs/scripts/lib/schema-to-mdx.mjs
palfrey: Yes. It's meant to be duplicated, because it renames `StoreSpec` refs to `NamedStoreConfig`

PR #2546, opened 2026-07-10
Prefetch directory-cache tree protos with one GetTree stream

erneestoc:
# Prefetch directory-cache tree protos with one GetTree stream

> Previously stacked on #2541 (subtree caching); now that #2541 has merged,
> this is rebased onto main as a single standalone commit.

## Problem

On a directory-cache miss, the worker discovers `Directory` protos level by
level: a child's digest is only known once its parent's proto has arrived,
so a cold tree pays **one worker→CAS round trip per tree depth** before
file fetching can even begin. Real input trees are commonly 10+ levels
deep; at millisecond RTTs this serialization costs tens to hundreds of
milliseconds per cold action. REAPI's `GetTree` exists precisely for this:
one RPC streams every directory proto of a tree.

## Change

Opt-in `DirectoryCacheConfig.experimental_get_tree_prefetch`: on a
**root-level cold miss**, issue a single `GetTree` stream (one round trip,
one fetch permit) and key the returned protos by their re-computed digest —
the stream carries protos, not digests, so each is hashed with the caller's
digest function. Construction consults the map and falls back to the
existing per-level fetch for any missing proto, any stream failure, or a
non-gRPC slow tier: the feature can only degrade to today's path, never
fail because of it.

Interaction with subtree caching (#2541): prefetch triggers only when a
full tree construction begins. Subtree cache hits never construct and never
prefetch; subtree misses reached during a construction reuse the map
already fetched for the root.

```json5
"directory_cache": {
  "experimental_subtree_caching": true,
  "experimental_get_tree_prefetch": true,
}
```

**Default off**, for an honest reason: the serving CAS answers `GetTree` by
walking the tree against its own backing store, so deployments should
enable this only when directory protos are served from a fast tier
(memory/filesystem), which is the common scheduler-side setup.

## Measurements

RPC-shape A/B (real `CasServer` over TCP, client-injected RTT, identical
proto counts on both paths):

| tree shape | RTT | per-level walk (today) | one GetTree | speedup |
|---|---|---|---|---|
| deep chain, depth 32 | 25 ms | 877 ms | 26 ms | **33.7×** |
| deep chain, depth 32 | 5 ms | 221 ms | 7 ms | 31.6× |
| wide 221-proto, depth 3 | 25 ms | 164 ms | 30 ms | 5.5× |
| wide 221-proto, depth 3 | 5 ms | 43 ms | 9 ms | 4.8× |

End-to-end validated against a real server: a 4,400-file / 221-proto tree
constructs entirely from one prefetched stream (zero per-level fetches),
byte-identical output.

## Tests

- Flag off: default behavior unchanged (existing suite).
- Flag on with a non-gRPC slow tier: prefetch returns `None`, construction
  falls back and produces correct results (new test).
- Full worker + config suites green under bazel (clippy + rustfmt aspects)
  and cargo.

## Notes for reviewers

- The prefetch future is `Box::pin`'d at call sites deliberately: inlined,
  it enlarges the callers' futures enough to overflow the stack in
  small-stack (fastbuild) test builds.
- `GetTree` pagination is honored via the streaming client; an incomplete
  map is tolerated per-proto by the fallback.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2546)


MarcusSorealheis (approved):
LGTM

MarcusSorealheis:
@erneestoc I found a small issue in your PR. It ignores the `next_page_token`. I'm pushing a small fix.

PR #2548, opened 2026-07-13
Handle zero digests in CompressionStore

### LINKED ISSUE #2542: CompressionStore -> FilesystemStore fails on zero byte upload
I think #2034 (zero byte digest optimisation) is perhaps misunderstood as "we should optimise this everywhere but no rush" when it is also a source of bugs. Some stores do zero digest optimisation, some don't. In the specific case of CompressionStore, it does not do it, and also errors when the underlying store DOES do it:

> Error was returned on the stream by RE: Failed to write to data, receiver
> disconnected : Failed to write footer to inner store in compression store :
> Uploading stdout : Error while uploading results : --- : Job cancelled because
> it attempted to execute too many times 4 > 3 times for operation_id: ...

Basically CompressionStore is trying and failing to write data but (in this case) FilesystemStore has just returned Ok(0) and has hung up its reader half. So CompressionStore fails to write. You'll note that in general `compress("")` is not `""`, it's a compression header.

So, CompressionStore needs to be updated to do the zero digest optimisation as well. Then I think you need to go through the other stores and audit if there are other such problems.

Nativelink 1.6.0.

aetosdios27:
# Description

Fix `CompressionStore` handling of canonical zero-byte digests.

`CompressionStore` previously forwarded zero-byte uploads to its inner store after producing compression metadata. This breaks when the inner store already optimizes canonical zero digests and returns immediately without consuming the stream.

For example, when `CompressionStore` wraps `FilesystemStore`, `FilesystemStore` recognizes the zero digest, returns `Ok(0)`, and closes its reader. `CompressionStore` then attempts to write the compressed stream footer to that closed reader and fails with a disconnected-receiver error.

This change makes `CompressionStore` consistently handle canonical zero-byte digests before delegating to its backend.

### Uploads

For both canonical zero-byte digest algorithms supported by NativeLink:

- SHA-256
- BLAKE3

`CompressionStore::update` now:

- returns `Ok(0)` for an empty upload;
- rejects non-empty data supplied under a canonical zero digest;
- does not invoke the inner store in either case.

This prevents the compression stream from being created when there is no content to compress and avoids the disconnected-reader failure described in
#2542.

### Existence checks

`CompressionStore::has_with_results` now:

- returns `Some(0)` for canonical zero digests without consulting the inner
  store;
- avoids calling the inner store entirely when a batch contains only zero
  digests;
- forwards only non-zero keys when a batch contains both zero and non-zero
  keys;
- forwards mixed-batch non-zero keys in a single inner-store call;
- reconstructs results in the same order as the original request;
- retains the existing direct delegation path for batches containing no zero
  digests.

## Tests

Regression coverage verifies that:

- empty SHA-256 and BLAKE3 zero-digest uploads return successfully;
- successful zero-digest uploads never reach the inner store;
- non-empty data is rejected for both canonical zero digests;
- rejected uploads also never reach the inner store;
- zero-only existence batches return `Some(0)` without an inner-store call;
- mixed batches preserve result ordering;
- mixed batches forward the exact expected non-zero keys;
- non-zero keys are forwarded in one batch rather than as individual calls;
- all-nonzero batches continue using a single direct delegation.

## Scope

This PR addresses the `CompressionStore` portion of the broader zero-byte digest audit in #2034. It does not modify zero-digest behavior in unrelated store implementations.

Fixes #2542
Part of #2034

## Type of change

## How Has This Been Tested?

The following checks pass locally:

- `cargo test -p nativelink-store --test compression_store_test`
- `cargo test -p nativelink-store`
- `cargo clippy -p nativelink-store --test compression_store_test -- -D warnings`
- `bazel test //... --lockfile_mode=error --extra_toolchains=@rust_toolchains//:all --verbose_failures`
- `git diff --check`

The focused CompressionStore suite passes all 13 tests. The complete `nativelink-store` Cargo suite passes, and all 99 Bazel test targets pass, including the repository's rustfmt and Clippy aspects.

## Checklist

- [x] Updated documentation if needed (no documentation changes were required)
- [x] Tests added/amended
- [x] `bazel test //...` passes locally
- [x] PR is contained in a single signed commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2548)

### DIFF
--- a/nativelink-store/src/compression_store.rs
+++ b/nativelink-store/src/compression_store.rs
@@ -288,7 +288,38 @@ impl StoreDriver for CompressionStore {
         digests: &[StoreKey<'_>],
         results: &mut [Option<u64>],
     ) -> Result<(), Error> {
-        self.inner_store.has_with_results(digests, results).await
+        if !digests.iter().any(|digest| is_zero_digest(digest.borrow())) {
+            return self.inner_store.has_with_results(digests, results).await;
+        }
+
+        for (digest, result) in digests.iter().zip(results.iter_mut()) {
+            if is_zero_digest(digest.borrow()) {
+                *result = Some(0);
+            }
+        }
+
+        let nonzero_digests = digests
+            .iter()
+            .filter(|digest| !is_zero_digest(digest.borrow()))
+            .map(StoreKey::borrow)
+            .collect::<Vec<_>>();
+        if nonzero_digests.is_empty() {
+            return Ok(());
+        }
+
+        let mut nonzero_results = vec![None; nonzero_digests.len()];
+        self.inner_store
+            .has_with_results(&nonzero_digests, &mut nonzero_results)
+            .await?;
+        for (result, inner_result) in digests
+            .iter()
+            .zip(results.iter_mut())
+            .filter_map(|(digest, result)| (!is_zero_digest(digest.borrow())).then_some(result))
+            .zip(nonzero_results)
+        {
+            *result = inner_result;
+        }
+        Ok(())
     }
 
     async fn update(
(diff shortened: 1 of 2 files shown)


palfrey (approved):
LGTM. Thanks for this!

PR #2552, opened 2026-07-16
Enforce Vale settings for nativelink-config as well as markdown

palfrey:
# Description

The primary motivation behind this is fixing issues like https://github.com/TraceMachina/nativelink/actions/runs/29435732168/job/87421795992 where we generate the config reference and then it fails because we're not checking the comments that the markdown is being generated from. OTOH, this also results in some more consistent wording/formatting for our config files as well! There are also some wider spelling/capitalisation fixes, as I was initially looking at applying this to everything, but only went part way through that as it's a much larger piece of work.

## Type of change

I'll be using the [config reference regenerate action](https://github.com/TraceMachina/nativelink/actions/workflows/config-reference.yaml) after this gets merged anyways, so will pull in the changes to the config docs then.

## How Has This Been Tested?

`pre-commit run -a vale`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2552)


MarcusSorealheis (approved):
LGTM

PR #2556, opened 2026-07-16
Remove quoting around NetApp

palfrey:
# Description

The NetApp part of the docs have weird quoting because the [default lint config considers them identifiers](https://doc.rust-lang.org/clippy/lint_configuration.html#doc-valid-idents), which isn't the case for the other cloud providers. See
image

This PR fixes that by fixing the lint.

## Type of change

## How Has This Been Tested?

`bun dev:docs`

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2556)


tools/generate-bazel-rc/src/main.rs
```diff
@@ -76,7 +76,7 @@ fn get_lints_from_key(lints_table: &Map<String, Value>, key: &str) -> BTreeSet<L
                 .expect("Expected small integer for priority"),
             });
         } else {
-            panic!("{}", value);
+            panic!("Odd lint value: {}", value);
```
palfrey, line 79: Hit this slightly confusing error while experimenting with ways to fix this. Not directly relevant for the end fix, but a good improvement anyways.

amankrx (approved):

PR #2557, opened 2026-07-16
Add fallback match interval to simple scheduler

### LINKED ISSUE #2550: Race causing single Redis operation to get stuck at queued
We're seeing sometimes when there's an OOM on a worker a single task can end up stuck in QUEUED state.  After some time pondering this, I don't think there's a code ordering or wake up issue here. I think this is due to Redis updating the FT.SEARCH index in the background.  So the flow is:

- Worker dies (OOM) and is detected
- Immediate evict worker is called which updates the Redis entry to say it's queued
- Notify is called to wake up do try match
- Do try match calls FT.SEARCH for all Queued entries, finds none
- Redis completes index update
- Job stuck in Queued state forever until the next trigger of do try match

Not sure of the best way to resolve this.  Perhaps we should have a polling interval for do try match as well as simply a notification period?

MarcusSorealheis:
# Description

When a worker dies (e.g. OOM-killed), its operations are re-queued and the matching engine is notified immediately. With an eventually consistent scheduler backend (Redis with RediSearch), the re-queued operation may not yet be visible to the search that its own notification triggered — the matching pass finds nothing and, with no further notifications, the operation stays `QUEUED` forever.

This adds `fallback_match_interval_s` to the simple scheduler (default 5s; `-1`/`0` disables): the matching pass re-runs periodically even when no task or worker change notification arrives. Under load, notifications win the `select` and the timer never fires; when quiet, it is one cheap pass per interval. Preferred over a retry counter because the staleness window (e.g. replica lag) is unbounded, and exhausted retries could only fail an action that is actually fine.

Fixes #2550

## Type of change

## How Has This Been Tested?

Two new paused-time tests using a wrapper `AwaitedActionDb` that hides queued actions from `get_range_of_actions`, simulating the stale search: with the fallback enabled the hidden action is rescued; with it disabled the action stays stuck until an unrelated event triggers a pass (the reported bug). Also ran `cargo test -p nativelink-scheduler -p nativelink-config --profile=smol`.

## Checklist

- [x] Updated documentation if needed (config field doc comment; reference docs regen is automated)
- [x] Tests added/amended
- [x] `bazel test //...` passes locally (100/100)

https://claude.ai/code/session_017UuGYox73nv9NRDM1knzyb

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2557)


nativelink-config/src/schedulers.rs
```diff
@@ -167,6 +172,20 @@ pub struct SimpleSpec {
...
     pub worker_match_logging_interval_s: i64,
+
+    /// Every N seconds, run a worker matching pass even if no task or worker
+    /// change notification arrived. This is a safety net for missed
+    /// notifications and for scheduler backends with eventually consistent
+    /// searches (for example Redis), where an operation that was re-queued
+    /// may not be visible to the search triggered by its own notification.
+    /// Without this, such an operation can stay queued until an unrelated
+    /// event triggers another matching pass.
+    /// Defaults to 5s. Can be set to -1 or 0 to disable.
```
chrisstaite-menlo: Maybe have any value < 0 disable since from_secs(-2) is going to cause issues.

chrisstaite-menlo: In fact, it does disable it, but with a warning...  Might as well just avoid the difficulty of logging and document it.

chrisstaite-menlo (approved):
Thanks for this, even if it's not the root cause I think it's a useful option to have available.

nativelink-config/src/schedulers.rs
MarcusSorealheis: Good thought

MarcusSorealheis: Are you still using Fred? We migrated to `redis-rs` a while ago. We tried to reproduce this failure and did not. I am sure it is still happening, though. It's a real bug.

chrisstaite-menlo: Yes, not done the migration work because it does the job.  Might do it at some point though.  It's unrelated to the issue though.  The issue is a logic flaw for the queue calculation.

MarcusSorealheis: Done — any value <= 0 now disables it, no logging path, and the doc comment says so. (`..=0 => None`, so `from_secs` only ever sees a positive value.)

chrisstaite-menlo: Thanks!

PR #2584, opened 2026-07-21
Let GCS get_part defer NotFound to the retrier

### LINKED ISSUE #2582: gcs_store::get_part treats a read 404 (NotFound) as terminal: transient/finalizing objects never retried, forcing RBE scheduler re-dispatch
## Summary

When a `get_part` read against a GCS-backed CAS (`experimental_cloud_object_store`, `provider: gcs`) returns HTTP 404, NativeLink treats it as a **permanent, non-retryable** error. For a blob the caller expects to exist, such as an RBE action input or an output referenced under Build-without-the-Bytes, a 404 is frequently **transient**: the object's upload is still finalizing, or it was just evicted from a fast tier and is being repopulated. Because the read is terminal, the store's configured `retry` budget never applies, the action fails, and under Remote Execution the only recovery is a full scheduler re-dispatch. Under concurrent cold-build load those re-dispatches can be exhausted and the operation is cancelled.

## Environment

- NativeLink current `main` (behavior also present in released 1.x)
- CAS backed by GCS via `experimental_cloud_object_store` with a `retry` block configured (`max_retries`, `delay`, `jitter`)
- Remote Build Execution enabled, co-located scheduler
- Multiple concurrent clients driving cold builds (high write+read churn against the shared bucket)

## Observed

A remote action fails while the worker stages inputs; a store read surfaces as:

```
Remote Execution Failure: NotFound: HTTP status client error (404 Not Found)
for url (https://storage.googleapis.com/storage/v1/b/<bucket>/o/cas%2F...?alt=media)
...
Job cancelled because it attempted to execute too many times N > M times
```

The object is present moments later (the upload finalizes / the producing action repopulates it), so a delayed retry would succeed, but no store-level retry happens, and the scheduler's per-op re-dispatch budget (`max_job_retries`) is the only thing between the race and a hard build failure.

## Root cause

Two independent places make a gcs read-404 terminal:

1. `gcs_store::get_part` returns `RetryResult::Err` (not `RetryResult::Retry`) when `read_object_content` yields `Code::NotFound`, so the retrier never gets a chance:
   https://github.com/TraceMachina/nativelink/blob/d1e49ac5f547006c3ecba9e6a94b16ab82897dce/nativelink-store/src/gcs_store.rs#L432

   (Contrast `has()`, which intentionally maps NotFound to `Ok(None)`, correct for an existence probe: https://github.com/TraceMachina/nativelink/blob/d1e49ac5f547006c3ecba9e6a94b16ab82897dce/nativelink-store/src/gcs_store.rs#L185)

2. Even if `get_part` did emit `RetryResult::Retry`, the shared `Retrier` classifies `Code::NotFound` as non-retryable and short-circuits:
   https://github.com/TraceMachina/nativelink/blob/d1e49ac5f547006c3ecba9e6a94b16ab82897dce/nativelink-util/src/retry.rs#L115 (NotFound in the `=> false` set) and https://github.com/TraceMachina/nativelink/blob/d1e49ac5f547006c3ecba9e6a94b16ab82897dce/nativelink-util/src/retry.rs#L166-L167 ("Not found, not retrying").

So `retry.max_retries` on the gcs store has no effect on a read-404, by design at both layers.

## Why it matters for RBE

For content-addressed reads of blobs the caller already believes exist (validated via `FindMissingBlobs`, or referenced as an action input / BwoB output), a 404 is far more likely to be a read-after-write / eviction-repopulation race than a genuinely absent object. Turning that transient into a terminal read pushes **all** recovery onto the scheduler's `max_job_retries`, which:

- is coarse: it re-dispatches the whole action rather than re-reading one blob,
- is a shared, finite budget that concurrent load can exhaust → `attempted to execute too many times`,
- turns a self-healing race into a red build.

## Proposed fix

Roughly in order of preference:

1. **Bounded read-after-write retry for `get_part` NotFound.** Treat a NotFound in `get_part` as retryable for a small, separately-bounded number of attempts with backoff, independent of the global `should_retry` classification, so `has()` and legitimate absent-object semantics elsewhere are unaffected. This matches the finalization/eviction window directly.
2. **Config knob** (e.g. `retry_on_not_found` / `consistency_retries`) on `experimental_cloud_object_store`, default off, letting operators opt reads into NotFound-retry where the backing store is eventually/finalization-consistent under load.
3. At minimum, have `get_part` emit `RetryResult::Retry` and let the retrier decide, with NotFound retryable for read paths only.

Happy to send a PR for option 1 if that direction is agreeable.

j-j-m:
# Description

`gcs_store::get_part` returned `RetryResult::Err` when
`read_object_content` failed with `Code::NotFound`, which bypasses the
`Retrier` entirely. That makes a read-404 terminal at the store layer,
even though `retry.retry_on_errors` already exists as an opt-in
mechanism for classifying `NotFound` as retryable. For content-addressed
reads of blobs the caller already believes exist (validated via
`FindMissingBlobs`, referenced as an action input, or a BwoB output), a
404 is frequently a read-after-write race — the object is still
finalizing or being repopulated — and turning it terminal pushes all
recovery onto the scheduler's `max_job_retries`, which concurrent cold
builds can exhaust (`Job cancelled because it attempted to execute too
many times`).

This change removes the `NotFound` special case in `get_part` so the
error is emitted as `RetryResult::Retry` and the `Retrier` classifies
it:

- **Default behavior is unchanged.** The retrier does not retry
  `NotFound` unless it's listed in `retry_on_errors`, so a missing
  object still fails immediately without consuming the retry budget.
- **Operators can now opt in** to read-404 retries on
  `experimental_cloud_object_store` (GCS) with the existing config
  surface, no new knob needed:

  ```json5
  "retry": {
    "max_retries": 6,
    "delay": 0.3,
    "jitter": 0.5,
    // Default retryable codes plus NotFound. Note this list
    // replaces the defaults, so include them explicitly.
    "retry_on_errors": [
      "Unknown", "Cancelled", "DeadlineExceeded", "ResourceExhausted",
      "Aborted", "Internal", "Unavailable", "DataLoss", "NotFound"
    ]
  }
  ```
- `has()` is untouched: it maps `NotFound` to `Ok(None)` before the
  retrier, which is correct for an existence probe.

This is options 2+3 from the issue via existing config rather than a
new store-level knob (option 1, a separately-bounded consistency-retry
budget). Happy to build that instead/on top if maintainers prefer a
default-on bounded retry for the finalization window.

Note for a possible follow-up: `s3_store::get_part` has the same
terminal `RetryResult::Err` pattern for `NoSuchKey`, so the same
treatment would apply there. Kept out of this PR to stay single-purpose.

Fixes #2582

## Type of change

## How Has This Been Tested?

- `cargo test -p nativelink-store --test gcs_store_test` — 20 passed,
  including two new tests:
  - `get_part_not_found_not_retried_by_default`: with `max_retries: 2`
    and default classification, a missing object yields `NotFound`
    after exactly 1 read call (no behavior change).
  - `get_part_retries_not_found_when_opted_in`: with
    `retry_on_errors: [NotFound]` and `max_retries: 2`, the read is
    attempted 3 times before surfacing `NotFound`.
  - The pre-existing `get_part_handles_not_found_error` still passes
    unmodified.
- `cargo clippy -p nativelink-store --tests` — clean.
- `rustfmt --check` on touched files — clean. (No local nix/bazel env,
  so relying on CI for `bazel test //...`.)

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2584)

### DIFF
--- a/nativelink-store/src/gcs_store.rs
+++ b/nativelink-store/src/gcs_store.rs
@@ -429,9 +429,11 @@ where
                         .await
                     {
                         Ok(stream) => stream,
-                        Err(e) if e.code == Code::NotFound => {
-                            return Some((RetryResult::Err(e), (offset, writer)));
-                        }
+                        // NotFound is intentionally not special-cased here:
+                        // the retrier doesn't retry NotFound by default, but
+                        // emitting `Retry` lets `retry.retry_on_errors` opt
+                        // reads into retrying read-after-write races where
+                        // an object is still finalizing or being repopulated.
                         Err(e) => return Some((RetryResult::Retry(e), (offset, writer))),
                     };
 
--- a/nativelink-store/tests/gcs_store_test.rs
+++ b/nativelink-store/tests/gcs_store_test.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 
 use bytes::{BufMut, Bytes, BytesMut};
 use mock_instant::thread_local::MockClock;
-use nativelink_config::stores::{CommonObjectSpec, ExperimentalGcsSpec};
+use nativelink_config::stores::{CommonObjectSpec, ErrorCode, ExperimentalGcsSpec, Retry};
 use nativelink_error::{Code, Error, make_err};
 use nativelink_macro::nativelink_test;
 use nativelink_store::cas_utils::ZERO_BYTE_DIGESTS;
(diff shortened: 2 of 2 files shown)


palfrey (approved):
Nice! This looks to have been there [since the original work on GCS over a year ago](https://github.com/TraceMachina/nativelink/pull/1645) so we probably just didn't spot this case at the time

PR #2594, opened 2026-07-22
Fix zstd upload decoder rejecting exact-buffer-multiple blobs

erneestoc:
## Problem

The streaming zstd decoder for REAPI `compressed-blobs` uploads
(`stream_decode_compressed_upload`) misreports a fully-decoded stream as
truncated whenever the blob's decompressed size is an exact multiple of the
decoder output buffer size (`DCtx::out_size()`, 128KiB-class). The final
`run` fills the output buffer exactly with `hint == 0` (frame complete and
flushed), but the drain loop treated a full output buffer as "may have more
to flush" and polled the finished decoder once more. A finished decoder
returns the input-size hint for a *new* frame header, so the post-EOF
`frame_input_hint != 0` truncation check fires and the upload fails with
`InvalidArgument` ("stream ended in the middle of a zstd frame").

This is deterministic and hits spec-conforming clients today: a Bazel
upload with `--remote_cache_compression` fails for any blob whose
uncompressed size lands on the buffer-multiple boundary.

## Fix

Exit the drain loop when the decoder reports the frame complete and fully
flushed (`hint == 0`), regardless of whether the output buffer was filled
exactly. One condition change plus an explanatory comment.

## Verification

- New regression test `decode_accepts_exact_output_buffer_multiple`
  (1MiB blob, exact multiple of the output buffer): **fails without the
  fix** (misreported as truncated), **passes with it**.
- Suites: bazel `//nativelink-service/...` 14/14 including clippy pedantic
  + nightly rustfmt aspects; `cargo check --tests --workspace` clean;
  wire-compression cargo suite 10/10.

Found while building compressed-upload support for NativeLink's own
`GrpcStore` transfers, where round-trip tests hit the boundary case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2594)

### DIFF
--- a/nativelink-service/src/wire_compression.rs
+++ b/nativelink-service/src/wire_compression.rs
@@ -322,7 +322,16 @@ pub async fn stream_decode_compressed_upload(
                 hasher.update(&produced);
                 tx.send(produced).await?;
             }
-            if in_buffer.pos() == in_buffer.src.len() && !output_was_full {
+            // `hint == 0` means the frame is completely decoded AND fully
+            // flushed. It must terminate the loop even when the output
+            // buffer was filled exactly: polling the decoder again after
+            // frame end would return the input-size hint for a NEW frame
+            // header, and the post-EOF `frame_input_hint != 0` check would
+            // then misreport a fully-decoded stream as truncated. This is
+            // deterministic for blobs whose decompressed size is an exact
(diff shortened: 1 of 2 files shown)


erneestoc:
closed as it's in https://github.com/TraceMachina/nativelink/pull/2596

PR #2527, opened 2026-07-07
Marcussorealheis extend zstd

MarcusSorealheis:
# Description

A continuation of #2416 with merge conflicts resolves and shell expand sorted.

Fixes #2416

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

`bazel test //...`

## Checklist

- [x] Updated documentation if needed
- [x] Tests added/amended
- [x] `bazel test //...`  passes locally
- [x] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2527)

PR #2596, opened 2026-07-22
Add opt-in zstd wire compression to GrpcStore transfers

erneestoc:
## Problem

NativeLink accepts and advertises REAPI `compressed-blobs/zstd` (#2527),
but never *sends* it: every `GrpcStore` resource name is plain `blobs/`,
so worker→CAS output uploads and worker←CAS input downloads ship raw bytes
in both directions regardless of any compression setting. On
bandwidth-limited links this leaves the full compression ratio on the
table for NativeLink's own traffic.

## Mechanism

Opt-in `GrpcSpec.experimental_remote_cache_compression`. Uploads and
full-blob downloads of blobs ≥64KiB use `compressed-blobs/zstd` (client
encodes at zstd level 1 — measured 3.7GB/s single-core, never the
bottleneck at ≤1Gbps; level 3 bought nothing on the test corpus). Ranged
reads and small blobs stay identity. Downloads digest-verify at EOF;
decoder-detected corruption is terminal (`InvalidArgument`), while
retryable mid-stream transport failures resume via an identity read at the
forwarded offset. Compressed uploads are explicitly **non-resumable**
mid-stream (the server contract): a transient failure fails fast with the
transport error after one attempt instead of burning the retry budget on
replays the server must reject. The streaming codec moved
`nativelink-service` → `nativelink-util` (service re-exports; server keeps
level 3).

## Measured (token-bucket throttled real TCP, through this feature path, byte-verified both directions)

| link | payload | wall (off → on) | wire bytes |
|---|---|---|---|
| 1Gbps | compressible 96MiB (3.7:1) | 712ms → 205ms up, 706ms → 205ms down | 101.0MB → 26.9MB |
| 100Mbps | compressible 24MiB | 1975ms → 492ms | 25.2MB → 6.7MB |
| both | incompressible random | unchanged | unchanged |

3.5–4.4× wall on compressible artifacts at both link speeds and in both
directions; zero wall or size penalty on incompressible data.

## Notes

- Contains the decoder exact-multiple fix also extracted standalone as
  #2594 (in the relocated util file here); whichever merges second
  resolves trivially.
- Will conflict textually with #2591 in `GrpcStore::update`; this branch
  rebases after #2591 and inherits its early-completion drain in the
  compressed path.
- No capabilities probe: the flag trusts the operator; against an upstream
  without compression enabled the first RPC fails fast with
  `InvalidArgument` (documented + tested).
- Precedence: `experimental_chunked_uploads` (#2595) wins for blobs above
  its threshold; combined-flags test lands in whichever merges second.

## Tests

9 feature tests incl. fails-without-fix regressions for non-resumable
uploads (old code: 4 attempts then terminal `InvalidArgument`; new: 1
attempt, prompt transport error) and early-completion return, plus a
deterministic corrupt-download-is-terminal property test. Suites:
`cargo check --tests --workspace` clean; bazel store+config+service+util
68/68 with clippy pedantic + nightly rustfmt aspects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2596)

## Review update (2026-07-22)

The wire-compression path was hardened after adversarial review:

- Compressed download feed/decode/pump tasks now cancel together on decoder failure, avoiding hangs when an upstream stream remains open.
- Compressed upload early-completion paths drain the caller's remaining input without blocking the RPC completion; this preserves proxy-chain producer behavior.
- Boolean compression configuration now supports the same expansion/parsing behavior as the other config booleans.
- Added and registered service integration coverage for corrupt-download cancellation and early-completion uploads.

Focused Cargo and Bazel wire-compression suites, formatting checks, and `git diff --check` pass. The feature remains opt-in; compressed uploads remain non-resumable by design.


MarcusSorealheis:
This has been tested and is looking good for our customers.

MarcusSorealheis:
@amankrx I think we want to try and get it in a `1.6.4` release.

MarcusSorealheis:
@amankrx

MarcusSorealheis (approved):

MarcusSorealheis (approved):
LGTM

PR #2599, opened 2026-07-22
Record the execution metrics that were declared but never emitted

### LINKED ISSUE #2583: Multiple Execution Metrics Missing in v1.5.0 — Only active_count, completed_count_total, and stage_transitions_total are present
We are running NativeLink worker (v1.5.0) with OpenTelemetry metrics enabled. The metrics pipeline (Worker → OpenTelemetry Collector → Prometheus) is confirmed to be working, as we can successfully scrape several metrics, including cache-related histograms.

However, according to the [official Metrics Catalog](https://docs.nativelink.com/deployment/metrics#execution-metrics), the following Execution metrics are documented but never appear in Prometheus:

nativelink_execution_stage_duration (Histogram)

nativelink_execution_total_duration (Histogram)

nativelink_execution_queue_time (Histogram)

nativelink_execution_output_size (Histogram)

nativelink_execution_retry_count_total (Counter)

Only these three Execution metrics are actually visible:

nativelink_execution_active_count (Gauge) ✅

nativelink_execution_completed_count_total (Counter) ✅

nativelink_execution_stage_transitions_total (Counter) ✅

We have confirmed that the worker is actively executing tasks (not just hitting cache), as nativelink_execution_active_count shows non-zero values and completed counts increase. Yet none of the histogram metrics or the retry counter appear.

Environment:

NativeLink version: v1.5.0 (Docker image nativelink:v1.5.0)

Deployment: Docker Compose

Metrics exporter: OpenTelemetry OTLP/gRPC → OpenTelemetry Collector → Prometheus

Environment variables set on worker:

text
OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>:4317
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_SERVICE_NAME=nativelink_worker
OTEL_EXPORTER_OTLP_COMPRESSION=zstd
OTEL_RESOURCE_ATTRIBUTES=...
Worker configuration file (JSON5) includes cache_metrics wrapper for cache stores (cache histograms are visible, so the OTLP pipeline works).

Steps to reproduce:

Deploy NativeLink worker with the above configuration.

Submit build actions that cause actual execution (cache misses).

Check Prometheus for any of the missing metrics.

Expected behavior:

All eight documented Execution metrics should appear after tasks are executed.

Actual behavior:

Only three metrics (active_count, completed_count_total, stage_transitions_total) are present. The five metrics listed above are completely absent.

Additional investigation:

A similar observation was reported in [Issue #2096](https://github.com/TraceMachina/nativelink/issues/2096), where users also saw only a subset of execution metrics.

It appears that only the Gauge and Counter metrics were implemented, while all Histogram-type execution metrics (stage_duration, total_duration, queue_time, output_size) and the retry counter are missing in v1.5.0.

Request:
Could you confirm whether these metrics are intentionally not yet implemented in v1.5.0, or if there is a specific configuration flag required to enable them? If they are planned for a future release, any timeline guidance would be appreciated.

Thank you for your excellent work on NativeLink!

amankrx:
# Description

nativelink-util defined execution_stage_duration, execution_total_duration, execution_queue_time, execution_output_size, and execution_retry_count, but nothing ever recorded them, so they never reached OTLP/Prometheus even though the metrics catalog documented them.

Fixes #2583 

## Type of change

## How Has This Been Tested?

- [X] Added tests 
- [x] Tested on a real environment. 

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2599)

### DIFF
--- a/nativelink-util/src/metrics.rs
+++ b/nativelink-util/src/metrics.rs
@@ -15,10 +15,11 @@
 // limitations under the License.
 
 use std::sync::LazyLock;
+use std::time::SystemTime;
 
 use opentelemetry::{InstrumentationScope, KeyValue, Value, global, metrics};
 
-use crate::action_messages::ActionStage;
+use crate::action_messages::{ActionResult, ActionStage};
 
 // Metric attribute keys for cache operations.
 pub const CACHE_TYPE: &str = "cache.type";
@@ -566,39 +567,6 @@ pub static EXECUTION_METRICS: LazyLock<ExecutionMetrics> = LazyLock::new(|| {
             ])
             .build(),
 
-        execution_cpu_time: meter
-            .f64_histogram("execution.cpu.time")
-            .with_description("CPU time consumed by action execution in seconds")
-            .with_unit("s")
-            .with_boundaries(vec![
-                0.01,   // 10ms
-                0.1,    // 100ms
-                1.0,    // 1s
-                10.0,   // 10s
-                60.0,   // 1 minute
-                300.0,  // 5 minutes
-                600.0,  // 10 minutes
-                1800.0, // 30 minutes
-                3600.0, // 1 hour
-            ])
-            .build(),
-
-        execution_memory_usage: meter
-            .u64_histogram("execution.memory.usage")
-            .with_description("Peak memory usage during execution in bytes")
-            .with_unit("By")
-            .with_boundaries(vec![
-                1_048_576.0,      // 1MB
-                10_485_760.0,     // 10MB
-                104_857_600.0,    // 100MB
-                524_288_000.0,    // 500MB
-                1_073_741_824.0,  // 1GB
-                5_368_709_120.0,  // 5GB
-                10_737_418_240.0, // 10GB
-                53_687_091_200.0, // 50GB
-            ])
-            .build(),
-
         execution_retry_count: meter
             .u64_counter("execution.retry.count")
             .with_description("Number of execution retries")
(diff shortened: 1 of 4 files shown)


nativelink-util/src/metrics.rs
```diff
@@ -624,10 +592,6 @@ pub struct ExecutionMetrics {
     pub execution_stage_transitions: metrics::Counter<u64>,
     /// Histogram of output sizes in bytes
     pub execution_output_size: metrics::Histogram<u64>,
-    /// Histogram of CPU time in seconds
-    pub execution_cpu_time: metrics::Histogram<f64>,
-    /// Histogram of peak memory usage in bytes
-    pub execution_memory_usage: metrics::Histogram<u64>,
```
palfrey, line 630: Are these two now not outputted?

amankrx: The same issue as others, but with these removed metrics, there's no data source present either. Like the other five documented metrics were atleast declared, but these haven't even been declared yet. So, it's better to remove them. 
Or we can comment them out for now, until we add the instrumentation later on, and should create an issue to add the instrumentation for these missing metrics including any other instrumentation that should be helpful.

palfrey: Can you add an issue for these, as these two are definitely very useful!

palfrey (approved):
CPU/memory needs flagging as an issue, otherwise good

nativelink-util/src/metrics.rs
palfrey: https://github.com/TraceMachina/nativelink/issues/2614

PR #2608, opened 2026-07-22
Make filesystem store posix_fadvise(DONTNEED) opt-in

### LINKED ISSUE #2607: per-blob posix_fadvise(POSIX_FADV_DONTNEED): a globally-serialized all-CPU stall on real filesystems
The filesystem store calls `advise_dontneed()` on every blob file it writes or reads —
four call sites in `nativelink-store/src/filesystem_store.rs` (~lines 1117/1377/1416/1480 at
v1.5.2), backed by `nativelink-util/src/fs.rs`:

```rust
pub fn advise_dontneed(&self) {
    let fd = self.inner.as_raw_fd();
    let ret = unsafe { libc::posix_fadvise(fd, 0, 0, libc::POSIX_FADV_DONTNEED) };
    ...
}
```

On a real filesystem (xfs here; ext4 behaves the same), `POSIX_FADV_DONTNEED` takes the
kernel's `generic_fadvise()` path, which calls **`lru_add_drain_all()`**: it schedules a
workqueue item on **every online CPU and blocks until all of them have run it**, and
concurrent callers serialize on a global mutex. The cost is therefore O(nr_cpus) per call
and system-wide serialized — on a 255-vCPU guest each call is a many-millisecond stall, and
the whole machine can only retire a bounded number of fadvise calls per second **no matter
how many worker processes or actions are in flight**.

A busy worker issues these calls per blob: materializing one action's input tree touches
thousands of files. The result on our host:

- Steady state **1.2 actions/s** with 173 actions in the "executing" stage, a standing queue,
  **~7% CPU** and **~1% I/O pressure (PSI)** — the machine was idle by every conventional
  metric while builds crawled.
- Spawned commands themselves completed in 0.3–2 s (from the worker's own
  `Executing command` / `Command complete` log pairs); ~98% of each action's ~148 s wall time
  was input/output handling around them.
- Worker D-state threads sampled via `/proc/<pid>/task/<tid>/stack`:

  ```
  [<0>] __lru_add_drain_all+0x3a/0x2d0
  [<0>] lru_add_drain_all+0x10/0x20
  [<0>] generic_fadvise+0x249/0x2e0
  [<0>] xfs_file_fadvise+0x29/0xa0 [xfs]
  [<0>] ksys_fadvise64_64+0x55/0xb0
  [<0>] __x64_sys_fadvise64+0x1c/0x30
  ```

  (plus companion threads parked in `futex_wait` behind locks held by the fadvise callers).

Two properties make this easy to miss:

- **It scales with CPU count.** The same workload ran fine on a 16-vCPU cloud VM (each drain
  is 16× cheaper and the serialized budget 16× less contended) and crawls on many-core
  hardware — the opposite of what one buys big machines for.
- **It's invisible on tmpfs** (fadvise is a no-op there) and invisible to CPU/disk metrics —
  the time is spent in a kernel wait that PSI-IO does not attribute.

There is also a second-order cost: `DONTNEED` actively evicts the page cache. Deployments
that keep the store on a fast local disk and rely on the page cache as the hot tier (a
reasonable design — it's ours) lose exactly that cache after every write and read.

amankrx:
# Description

Made the fadvise opt-in: added an `evict_page_cache` config flag (default false) and guarded all four call sites on it. So by default NativeLink no longer issues the advice and operators who genuinely want page-cache eviction can turn it back on.

Fixes #2607 

## Type of change

## How Has This Been Tested?

Unit Test and tested on linux

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2608)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -754,6 +754,16 @@ pub struct FilesystemSpec {
     /// Default: unlimited
     #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
     pub max_concurrent_writes: usize,
+
+    /// When true, advise the kernel to drop the page cache for each blob after
+    /// it is written or read (`posix_fadvise` with `POSIX_FADV_DONTNEED`). On
+    /// real filesystems this takes a globally serialized, all-CPU kernel path
+    /// that stalls on many-core hosts, and it evicts the page-cache tier that
+    /// fast-disk deployments rely on. Leave off unless you specifically want to
+    /// keep this store's I/O out of the page cache.
+    /// Default: false
+    #[serde(default)]
+    pub evict_page_cache: bool,
 }
 
 // NetApp ONTAP S3 Spec
--- a/nativelink-store/src/filesystem_store.rs
+++ b/nativelink-store/src/filesystem_store.rs
@@ -1110,6 +1110,9 @@ pub struct FilesystemStore<Fe: FileEntry = FileEntryImpl> {
     block_size: u64,
     #[metric(help = "Size of the configured read buffer size")]
     read_buffer_size: usize,
+    /// See [`FilesystemSpec::evict_page_cache`]. When false (the default) the
+    /// per-blob `posix_fadvise(DONTNEED)` calls are skipped.
+    evict_page_cache: bool,
     weak_self: Weak<Self>,
     rename_fn: fn(&OsStr, &OsStr) -> Result<(), std::io::Error>,
     /// Limits concurrent write operations to prevent disk I/O saturation.
@@ -1243,6 +1246,7 @@ impl<Fe: FileEntry> FilesystemStore<Fe> {
             evicting_map,
             block_size,
             read_buffer_size,
+            evict_page_cache: spec.evict_page_cache,
             weak_self: weak_self.clone(),
             rename_fn,
             write_semaphore,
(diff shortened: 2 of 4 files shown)


MarcusSorealheis:
This is a crazy bug. I'm really shocked it's even here. @amankrx you must've had some reason for adding this feature [here](https://github.com/TraceMachina/nativelink/commit/727760d1e208ca8be7bc134f432baf5dc5bf5928). So please try to think back as to why it was added because now we will need to be called out very explicitly. Something to the effect of, "This is source-compatible, but it materially changes memory and I/O behavior,"  with a little bit of the history or backstory.

Given that this was merged in February I believe it was caused by one of our bigger filesystem only deployments. I cannot imagine why someone using an object store and multiple CAS nodes (most enterprise deployments) would ever enable this behavior after this PR. 

Thank you for fixing it and my apologies for not thinking about this issue back then.

One thing to think about in the documentation, which will be the only piece missing from the PR is the fact that a global per-CPU drain will usually cost more on larger machines with more cores. Adding workers or concurrent actions cannot overcome the bottleneck because callers converge on the same serialized kernel path.

MarcusSorealheis (changes requested):
Please add documentation about this feature such that customers can understand that if they enable it on large machines (many cores), there is an additional cost to consider.

amankrx:
That was added because Workers were getting `OOMKilled` on every occasion. I mean the workers are still not perfect in that sense, and we still observe some OOM kills on different occasions. But, this was one of the effort to reduce the possibility of workers getting OOM Killed. 
I think we need to put in some effort to re-work on the workers such that it is resilient to OOM Kills. With the right configuration, we have managed to reduce the frequency of these errors, but still a larger effort is needed to get the workers to a more stable state.

MarcusSorealheis:
_more stable_ state

MarcusSorealheis:
once the doc is up I will approve

MarcusSorealheis (approved):
LGTM. Thanks.

PR #2611, opened 2026-07-23
Batch worker fs tree walks into one spawn_blocking per directory

### LINKED ISSUE #2609: per-file spawn_blocking + per-process global mutexes: lock-handoff storms on many-core hosts
The worker's input-materialization paths (`download_to_directory`,
`fs_util::hardlink_directory_tree_recursive`, `fs_util::set_perms_recursive`, the directory
cache's clone/construct paths) issue a `tokio::spawn_blocking` and touch per-process shared
state **per file**. An action whose input tree carries a hermetic toolchain is thousands of
files; at ~200 concurrent actions this produces millions of acquisitions of two per-process
global locks:

- tokio's blocking-pool spawner mutex (every `spawn_blocking` takes it, plus a condvar wake), and
- the store's `EvictingMap` state mutex (`nativelink-util/src/evicting_map.rs` —
  `parking_lot::Mutex` around the whole LRU; taken on every blob get/insert/touch).

The critical sections are short, but on a large multi-socket guest the cost is the **handoff**:
each contended lock transfer is a cross-socket cache-line migration, and with hundreds of
waiters the queue drains at cache-coherency speed. Userspace backtraces of the worker under
load (`eu-stack`, ~900 threads, v1.5.2 + the fadvise shim from Finding 1 already active):

```
  450  parking_lot::raw_mutex::RawMutex::lock_slow
  389  parking_lot::condvar::Condvar::wait_until_internal   (blocking-pool waiters)
  153  tokio::runtime::blocking::pool::Spawner::spawn_task  (holds/queues on the pool lock)
  113  nativelink_util::fs_util::hardlink_directory_tree_recursive::{closure}
   54  nativelink_util::fs_util::set_perms_recursive_impl::{closure}
   26  nativelink_worker::directory_cache::DirectoryCache::copy_file_to::{closure}
```

Steady state at that moment: 1.0–1.5 actions/s with 200 actions in "executing", ~7% CPU on
255 vCPUs, ~1% PSI-IO — spawned commands themselves finished in 0.3–2 s.

amankrx:
# Description

The worker's input-materialization walks issued a tokio::fs call (each a spawn_blocking on tokio's blocking pool) per file. Materializing an action's input tree touches thousands of files, and at high action concurrency the per-file blocking-pool handoff dominates on many-core hosts while CPU and I/O sit near idle.

Batch both recursive walks to one blocking task per directory:

- hardlink_directory_tree_recursive: read the directory, hardlink files and recreate symlinks inline, create subdirectories, and return them for the async layer to recurse into.
- set_perms_recursive (set_readonly_recursive / set_dir_writable_recursive): convert perms_fn to a synchronous closure and chmod each directory's entries plus the directory itself in a single blocking task, recursing per subdirectory.

Fixes #2609 

## Type of change

## How Has This Been Tested?

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2611)

### DIFF
--- a/nativelink-util/src/fs_util.rs
+++ b/nativelink-util/src/fs_util.rs
@@ -239,83 +239,101 @@ fn hardlink_directory_tree_recursive<'a>(
     dst: &'a Path,
 ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send + 'a>> {
     Box::pin(async move {
-        let mut entries = fs::read_dir(src)
-            .await
-            .err_tip(|| format!("Failed to read directory: {}", src.display()))?;
+        let src_buf = src.to_path_buf();
+        let dst_buf = dst.to_path_buf();
 
-        while let Some(entry) = entries
-            .next_entry()
-            .await
-            .err_tip(|| format!("Failed to get next entry in: {}", src.display()))?
-        {
-            let entry_path = entry.path();
-            let file_name = entry.file_name().into_string().map_err(|os_str| {
-                make_err!(
-                    Code::InvalidArgument,
-                    "Invalid UTF-8 in filename: {:?}",
-                    os_str
-                )
-            })?;
+        let subdirs = tokio::task::spawn_blocking(
+            move || -> Result<Vec<(PathBuf, PathBuf)>, Error> {
+                let mut subdirs = Vec::new();
+                let read_dir = std::fs::read_dir(&src_buf)
+                    .err_tip(|| format!("Failed to read directory: {}", src_buf.display()))?;
+                for entry in read_dir {
+                    let entry = entry
+                        .err_tip(|| format!("Failed to get next entry in: {}", src_buf.display()))?;
+                    let entry_path = entry.path();
+                    let file_name = entry.file_name().into_string().map_err(|os_str| {
+                        make_err!(Code::InvalidArgument, "Invalid UTF-8 in filename: {:?}", os_str)
+                    })?;
+                    let dst_path = dst_buf.join(&file_name);
 
-            let dst_path = dst.join(&file_name);
-            // `DirEntry::metadata` does NOT traverse symlinks (it has
-            // `symlink_metadata`/lstat semantics), so `is_symlink()` below
-            // correctly identifies symlink entries and the symlink branch
-            // recreates them as symlinks rather than dereferencing them.
-            let metadata = entry
-                .metadata()
-                .await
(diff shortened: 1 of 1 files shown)


palfrey (changes requested):
This is failing due to uses of `tokio::task::spawn_blocking` and should be using `spawn_blocking!` instead

MarcusSorealheis:
@amankrx I'm picking this up so #2609 can move forward. I'll push three commits on top of your branch, without a force-push, that address @palfrey's review and a few gaps I found reading it against current `main`:

- **Merge of `upstream/main`.** There were no conflicts, since `main` hasn't touched `fs_util.rs` since this branch's base.
- **Use `spawn_blocking!` instead of raw `tokio::task::spawn_blocking`.** That was the requested change and the clippy failure.
- **Walk the whole tree in one blocking task, not one per directory.**
  - Sequential recursion still cost one blocking-pool handoff per directory. Trees with many small directories, such as `site-packages`, only gained about 3 to 5 times.
  - The walk now uses an explicit stack, and the `exists()` checks and the macOS `clonefile` attempt run inside the same task, off the async thread.
- **Use `DirEntry::file_type()` instead of `metadata()`.** Only the type is needed, and on most filesystems it comes from `d_type` with no `stat` per entry. The lstat semantics are unchanged.
- **Stop a walk when its future is dropped.**
  - A blocking task can't be aborted once it starts. Before, an abandoned walk kept creating entries while `DirectoryCache` cleaned up the same directory.
  - A drop guard now raises a flag that the walk checks before each entry. An abandoned walk finishes at most the entry it's on and returns `Cancelled`.

**What stays the same:** symlinks are handled before directories and never followed, directories end at 0o755, files are hardlinked rather than copied, and the error messages are unchanged.

**New tests:**
- A nested tree is reproduced exactly: files, empty and nested-empty directories, a 0o700 directory, and symlinks to a file, to a directory and dangling. The check covers the same inodes and 0o755 directories.
- A walk that starts already cancelled creates nothing.
- A future dropped partway through a walk of about 10,000 entries stops within one entry. This test is Linux-only, because macOS clones in a single call.

**Test results:**
- `cargo test -p nativelink-util -p nativelink-worker` passes on macOS and on Linux (Docker).
- Clippy with `-D warnings` is clean, except the pre-existing `connection_manager_test.rs:145`, which is identical on `main`.
- The pre-commit hooks pass.

**Not included:** batching the `EvictingMap` lookups, the other half of #2609. That's being handled separately.

PR #2620, opened 2026-07-23
Skip empty chunks in GCS get_part read stream

j-j-m:
# Description

`gcs_store::get_part` forwards every chunk from the `read_object_content` stream into `buf_channel::send()`. `send()` rejects zero-length buffers by contract (`Cannot send EOF in send(). Instead use send_eof()`), but HTTP/2 peers are allowed to end a response body with a zero-length DATA frame, and reqwest surfaces that as an empty `Bytes` chunk. When it happens, a perfectly healthy 200 OK read fails with a terminal `Internal` error on the first attempt (the send failure is returned as `RetryResult::Err`, so the retrier never engages).

We hit this in production against a GCS-backed CAS: Google's storage front end emits the trailing empty DATA frame intermittently under real load, and every affected read failed hard with

```
Cannot send EOF in send(). Instead use send_eof() : On attempt 1
```

Wire captures show clean 200 responses with complete bodies, so this is purely client-side frame handling. The read loop is identical on current main and v1.6.2, so both are affected. Go-based test servers (e.g. fake-gcs-server) never reproduce it because of their h1/h2 framing choices, which is probably why it has gone unnoticed.

The fix skips empty chunks in the read loop. Behavior for non-empty chunks is unchanged, and offset accounting is unaffected since an empty chunk contributes zero bytes.

## Reproduction

Putting a reverse proxy that always terminates streams with a bare END_STREAM DATA frame (mitmproxy in `--mode reverse:https://storage.googleapis.com`) in front of GCS makes the failure deterministic: unpatched, 0/10 ByteStream `Read` calls through the CAS succeed; with this change, 10/10. Direct against GCS it reproduces intermittently, dependent on Google's framing at that moment.

## Type of change

## How Has This Been Tested?

- New regression test `get_part_ignores_empty_stream_chunks` wraps `MockGcsOperations` so the content stream interleaves zero-length chunks the way an HTTP/2 peer that emits empty DATA frames does. It fails without the fix (same `Cannot send EOF` error) and passes with it.
- Full `gcs_store_test` suite passes (21/21).
- The deterministic mitmproxy reproduction above, against real GCS.
- Running in our RBE farm (patched build of `0e7580f`) since 2026-07-23 with the failure gone.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2620)

### DIFF
--- a/nativelink-store/src/gcs_store.rs
+++ b/nativelink-store/src/gcs_store.rs
@@ -440,6 +440,13 @@ where
                     while let Some(next_chunk) = stream.next().await {
                         match next_chunk {
                             Ok(bytes) => {
+                                // HTTP/2 peers may emit zero-length DATA frames
+                                // (e.g. a bare END_STREAM frame), which surface
+                                // here as empty chunks. `send()` rejects empty
+                                // buffers by contract, so skip them.
+                                if bytes.is_empty() {
+                                    continue;
+                                }
                                 offset += bytes.len() as u64;
                                 if let Err(err) = writer.send(bytes).await {
                                     return Some((RetryResult::Err(err), (offset, writer)));
--- a/nativelink-store/tests/gcs_store_test.rs
+++ b/nativelink-store/tests/gcs_store_test.rs
@@ -875,3 +875,148 @@ async fn check_health_failed_on_object_exists_error() -> Result<(), Error> {
         other => panic!("expected HealthStatus::Failed, got {other:?}"),
     }
 }
+
+#[nativelink_test]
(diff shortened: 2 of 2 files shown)


MarcusSorealheis:
that is a bug. /thanks for filing and fixing. We will cut a new release tomorrow in light of this fix after a 12 hour soak test we'll start.

MarcusSorealheis (approved):
Perfect PR

j-j-m:
> Perfect PR

🙏

PR #2625, opened 2026-07-24
Caching the attic client

palfrey:
# Description

Ideally we'd like to cache a static version of the attic client for fastest builds, but [that doesn't currently build](https://github.com/zhaofengli/attic/issues/354) so instead this PR tries caching around Nix for just that bit. We [removed the `magic-nix-cache` before](https://github.com/TraceMachina/nativelink/pull/2059) due to size/speed issues, so I'm hoping restricting this to just attic should get us the benefits without the drawbacks....

## Type of change

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2625)

### DIFF
--- a/.github/actions/prepare-nix/action.yaml
+++ b/.github/actions/prepare-nix/action.yaml
@@ -20,6 +20,14 @@ runs:
           log-lines = 100 # Necessary in CI to get enough failure context
           keep-outputs = true # Don't delete non-root outputs, because might be needed for other builds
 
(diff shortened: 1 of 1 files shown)


palfrey:
This didn't shave as much off as I hoped. We were at ~1m30 give or take for both Ubuntu and MacOS for the "Prepare worker" step, and this is showing approximately the same speeds give or take. Not enough to take a risk on this yet.

PR #2633, opened 2026-07-26
Log client server mismatch

### LINKED ISSUE #2632: Client and Server Hash Algorithm Mismatch Fails Silently
When a client and server rely on different hashing algorithms, nothing happens other than slow builds. The server should log this mismatch to guide the user to a more optimal setup.

MarcusSorealheis:
# Description

This PR ensures ByteStream uploads are verified using the client-declared algorithm, and warns when client and server configurations differ because mixed algorithms cannot share cache hits.
 
Fixes #2632 

## Type of change

## How Has This Been Tested?

`bazel test //...` with added regression coverage.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2633)


MarcusSorealheis:
This was going to be a breaking change.


| Benchmark phase            | SHA-256 mean | BLAKE3 mean | Result                  |
|----------------------------|-------------:|------------:|-------------------------|
| Compile and populate cache |      180.04s |     183.57s | SHA-256 was 1.9% faster |
| Cached rebuild             |        6.16s |       6.58s | SHA-256 was 6.3% faster |

It's clear for our use case that SHA-256 is the better digest function but that's not always the case.

heyitsalec (approved):
lgtm

palfrey (approved):

PR #2634, opened 2026-07-27
Kill full process groups and report resource usage on macOS workers

erneestoc:
## Problem

On macOS, NativeLink kills only the direct child of an action on timeout or
scheduler kill. Anything the action spawned below that — compiler daemons,
simulator processes, test runners — survives as an orphan, retains CPU/RAM,
and accumulates until the worker degrades. macOS workers also report no
per-action resource usage, while Linux samples peak memory (#2409).

## Mechanism

- The action is now its own process-group leader on all unix platforms. On
  macOS this maps to posix_spawn's `POSIX_SPAWN_SETPGROUP` attribute — the
  standard library only falls back to fork/exec for uid/gid/`pre_exec`/
  groups/chroot, so the fast spawn path is preserved (`pre_exec` is only
  used under Linux namespaces).
- Every kill path (timeout, scheduler kill, drop-guard) additionally
  signals the entire group via `kill(-pgid, SIGKILL)` on macOS. Linux
  behavior is unchanged (namespaces already contain descendants there;
  non-namespaced Linux remains issue #225).
- The resource-usage sampler gains a macOS implementation with parity to
  the Linux `/proc` sampler: `proc_listpids(PROC_PGRP_ONLY)` +
  `proc_pid_rusage` sum resident memory per tick and keep each pid's
  latest cumulative CPU (converted from mach absolute time units via
  `mach_timebase_info`), so members that exit between ticks retain their
  CPU in the totals. `ActionResourceUsage` gains additive `cpu_user_ns`/
  `cpu_system_ns` fields (populated on macOS; zero elsewhere).

## Verification

The group-kill regression test backgrounds a grandchild with a pgrep-able
marker, kills the action, and asserts the grandchild dies — it fails when
the group signal is disabled and passes with it. The rusage test runs a
CPU-burning helper through the full action pipeline and asserts sampled
peak memory and plausible CPU time. Suites: worker integration 37/37;
`cargo check --tests --workspace` clean; bazel worker+util+proto+scheduler
41/41 with clippy pedantic + nightly rustfmt aspects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2634)

### DIFF
--- a/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
+++ b/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
@@ -125,7 +125,17 @@ message ActionResourceUsage {
     /// The worker ID that observed the resource usage.
     string worker_id = 4;
 
-    reserved 5; // NextId.
+    /// Total user-mode CPU time consumed by the action's process group in
+    /// nanoseconds, when the platform sampler collects it. Zero when
+    /// unavailable.
+    uint64 cpu_user_ns = 5;
+
+    /// Total kernel-mode CPU time consumed by the action's process group in
+    /// nanoseconds, when the platform sampler collects it. Zero when
+    /// unavailable.
+    uint64 cpu_system_ns = 6;
+
+    reserved 7; // NextId.
 }
 
 /// Result sent back from the server when a node connects.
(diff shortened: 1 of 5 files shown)


MarcusSorealheis (approved):

MarcusSorealheis:
@erneestoc we need to resolve the conflicts and then we can merge this one

PR #2638, opened 2026-07-28
fix(worker): prevent upload_results deadlock when file count exceeds semaphore permits (#2636)

### LINKED ISSUE #2636: upload_results deadlocks when file count exceeds OPEN_FILE_SEMAPHORE capacity
## Summary

A worker's `upload_results` phase deadlocks when an action's output directory has enough files to exhaust `OPEN_FILE_SEMAPHORE` relative to `RLIMIT_NOFILE` — including at `RLIMIT_NOFILE=24576`, nativelink's own recommended value, given enough files.
Unlike #2051 (fixed before 1.6.2), this doesn't involve `hard_link`/`symlink`.
Triggered with just plain file writes. `max_upload_timeout_s` fires and logs `DeadlineExceeded`.
The scheduler immediately retries the same operation into the same deadlock.
The worker never actually recovers — closer to #2403 than to #2051/#907.

## Environment

- nativelink v1.6.2
- Linux x86_64

## Repro

No RE client dependency required — reproduces with any action whose output directory has enough files relative to the worker's `RLIMIT_NOFILE`.

1. Start a worker with `RLIMIT_NOFILE` capped at 24576 (soft and hard) — nativelink's own recommended value:
   ```
   bash -c 'ulimit -n 24576; ulimit -Hn 24576; exec nativelink config.json5'
   ```
   Log confirms the cap took effect: `set_open_file_limit() assigns new open file limit 24576`.
2. Submit an action whose command writes ~52,000 files (200 dirs × 260 files, 5KB each) to its declared output directory, nothing else — no hardlink, no symlink, just plain writes:
   ```sh
   for d in $(seq 1 200); do
     mkdir -p "$out/dir$d"
     for f in $(seq 1 260); do
       head -c 5000 /dev/urandom > "$out/dir$d/file$f.bin"
     done
   done
   ```
3. Command completes successfully (`Command complete`, no error).
4. `upload_results: starting with timeout` logs immediately after.
5. At `max_upload_timeout_s`, `Error { code: DeadlineExceeded, ... }` fires correctly.
6. The scheduler retries the same operation within ~1.5s. It stalls identically. Repeats indefinitely — worker never makes progress on this operation again.

A smaller/easier-to-run variant: `RLIMIT_NOFILE=1024` with only ~5,200 files (20 dirs × 260 files, 35KB each) reproduces the identical symptom. The same ~5,200-file action does *not* reproduce at `RLIMIT_NOFILE=24576` — the file count needed to trigger it scales with the limit, it isn't a fixed threshold.

## Isolation

Confirmed the trigger is file-count-relative-to-`RLIMIT_NOFILE`, nothing else:
- Reproduces at `RLIMIT_NOFILE=1024` with ~5,200 files, and at `RLIMIT_NOFILE=24576` with ~52,000 files.
- At the larger (24576/52,000) scale, a second, more specific symptom is visible alongside the timeout — a rename-into-CAS race:
  ```
  Failed to rename file, err: Error { code: NotFound, messages: ["No such file or directory (os error 2)",
  "Failed to rename temp file to final path .../content_path-worker/d/<hash>-5000"] }
  ```
  i.e. a temp file selected for emplacement into CAS is gone by the time the rename runs — consistent with something else (a permit-timeout cleanup?) racing the emplace path under permit pressure.

## Expected vs actual

Expected: either the action fails cleanly once and reports `DeadlineExceeded` to the client, or the semaphore backpressure prevents exhaustion in the first place (per #2051/#907's fixes for other code paths).

Actual: repeated deadlock + timeout + retry loop, worker effectively loses capacity on that operation slot permanently.

## Related

- #2051 / #2052 — same `OPEN_FILE_SEMAPHORE`, different trigger (`hard_link`/`symlink` holding a `FileEntry` lock). Fixed, included in v1.6.2, doesn't cover this case (no hardlink/symlink here).
- #907 / #908 — same semaphore, older trigger (move-optimization holding a permit while a reader stalls). Fixed 2024, doesn't cover this case.
- #2403 — open, closely matches the retry-without-recovery symptom, but that report doesn't see a `DeadlineExceeded` error at all; here it fires correctly but the operation still doesn't recover.

namdamdoi68-oss:
Awaits directory traversal futures before awaiting file and symlink upload futures in upload_directory to prevent circular wait deadlocks on OPEN_FILE_SEMAPHORE. Fixes #2636.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2638)


nativelink-worker/src/running_actions_manager.rs
palfrey, line 1: Can you write some test code for this please?

namdamdoi68-oss: Added `upload_with_single_permit_nested_dirs` — creates a 3-level deep directory tree (`out/a/b/`) with files at each level under a single semaphore permit. This is the exact scenario that deadlocked before the fix.

MarcusSorealheis: Can you articulate what happened and how you encountered this issue? Maybe file an issue?

MarcusSorealheis: Also thank you for the contribution

namdamdoi68-oss: The issue is already filed as #2636 by @rbtcollins with a detailed repro. The root cause is a circular wait in `upload_directory`: `try_join3` polled `dir_futures`, `file_futures`, and `symlink_futures` concurrently, so recursive directory traversals held `OPEN_FILE_SEMAPHORE` permits while file uploads in the same join were blocked waiting for those same permits. Under permit starvation (file count approaching `RLIMIT_NOFILE`) this deadlocked reliably. The fix awaits directory futures first so their permits are released before file uploads begin.

MarcusSorealheis: Awesome. Thank you for the contribution!

palfrey (approved):

PR #2640, opened 2026-07-30
Harden the compressed GrpcStore download path

walter-zeromatter:
# Description

Follow-ups to the opt-in zstd wire compression added to `GrpcStore` in #2596.
These are correctness fixes in the compressed **download** path, plus one guard
on the compressed upload path, found while reviewing that work.

**1. Pump-stage failures were misclassified as retryable.**
`get_part_compressed` classified pump-stage failures with `is_retryable_code`,
but pump errors come from our own writer, and `buf_channel` reports a dropped
receiver as `Internal` — which counts as retryable. A consumer that hung up
mid-download therefore returned `Ok(Some(forwarded))`, and `get_part` responded
by issuing a second, full-size identity `Read`, streaming the remainder of the
blob into a channel nobody was reading. Pump failures are now always terminal:
they mean either the consumer went away or the decoder aborted, and the
decoder's own verdict is already surfaced by the arm above it, since `try_join!`
polls decode first.

**2. Resume could fire after the download had already succeeded.**
The fallback treated a late failure from another stage as a reason to resume
even when the blob had already been fully delivered and verified. The pump now
records that it forwarded the decoder's EOF — which the decoder only emits after
the blob passes its size and digest checks — and a late error after that point
resolves as success instead of resuming into an already-closed writer.

**3. A genuine resume over-requested.**
The identity retry now asks for exactly the undelivered tail. It previously
reused the caller's original `length`, which the entry guard only permits when
it covers the whole blob, so the resumed request ran past the end of the blob
and relied on the server clamping it.

**4. Compressed upload is now restricted to real digest keys.**
A string key's `into_digest()` hashes the key itself, which the remote's
mandatory digest verification would reject. Also skips the background drain
spawn when the encoder finished cleanly and the reader is already at EOF.

## Type of change

## How Has This Been Tested?

New cases in `nativelink-service/tests/grpc_wire_compression_test.rs` covering
the aborted-consumer path and the resume path. The resume path had no coverage
at all before this change:

- The fake ByteStream now honors `read_offset` / `read_limit`, so the resumed
  range can be asserted byte-exactly rather than just "it eventually finished".
- Added a mid-frame abort mode and a mixed-entropy payload helper, since
  `make_content` compresses far enough that multi-megabyte blobs otherwise fit
  in a single wire chunk and never exercise the partial paths.
- Resume is asserted both with `length: None` and with an explicit `length`
  covering the blob. Only the latter over-requested before this change, because
  `read_limit: 0` already means "to the end" on the wire.

Verified locally on this branch:

```
bazel test //...
# Executed 73 out of 102 tests: 102 tests pass.
# (rustfmt + clippy aspects are enabled by default in .bazelrc, so lint runs with the build)
```

## Checklist

- [x] Updated documentation if needed (no user-facing config or behavior change)
- [x] Tests added/amended
- [x] `bazel test //...` passes locally (ran the store + service wire-compression
      targets; the aspects cover clippy/rustfmt for the touched crates)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2640)


palfrey (approved):

palfrey:
[Seeing a failure in `corrupt_compressed_download_is_terminal`](https://github.com/TraceMachina/nativelink/actions/runs/30807540413/job/91666280034?pr=2640) and I'm not quite sure if this is a flaky test, or something with this PR (or both)? @walter-zeromatter Any thoughts there?

walter-zeromatter:
Looks like there was a nondeterministic test. Found & fixed, along with another unrelated bit of hardening

PR #2647, opened 2026-07-31
Fix a blocking-pool deadlock in `nativelink_util::fs::read_dir` [1.6-patch-5]

cormacrelf:
# Description

The previous implementation acquired a blocking-pool permit and then, from inside that blocking-pool thread, called `Handle::current().block_on(tokio::fs::read_dir(path))`. But `tokio::fs::read_dir` is itself implemented as a `spawn_blocking`, so this nested `block_on` required a *second* blocking-pool thread to be available to make progress — every call to `read_dir` needed two pool threads at once instead of one.

Under enough concurrent callers (e.g. `upload_directory`'s unbounded directory-tree fan-out on an action with many inputs), all blocking-pool threads could end up parked waiting on inner tasks that could never get a thread to run on, freezing every `fs::` operation in the process. I believe this may have caused deadlocks on actions with a lot of inputs (700+).

Fix: acquire the permit, then simply `.await` `tokio::fs::read_dir` directly instead of `block_on`-ing it — no nested blocking-pool thread is required.

## Type of change

## How Has This Been Tested?

New regression test `read_dir_needs_only_one_blocking_thread` in `nativelink-util/tests/fs_test.rs`: builds a runtime with `max_blocking_threads(1)` and asserts `fs::read_dir` completes within a 5s timeout. The pre-fix implementation deadlocks in this configuration since it requires two blocking-pool threads at once.

## Checklist

- [x] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2647)

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -4,6 +4,7 @@ disallowed-methods = [
   { path = "tokio::runtime::Builder::new_current_thread", reason = "use one of the `nativelink-util::task` functions instead" },
   { path = "tokio::runtime::Builder::new_multi_thread", reason = "use one of the `nativelink-util::task` functions instead" },
   { path = "tokio::runtime::Builder::new_multi_thread_alt", reason = "use one of the `nativelink-util::task` functions instead", allow-invalid = true }, # Can be invalid because depends on feature flags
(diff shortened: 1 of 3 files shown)


amankrx:
Hi @cormacrelf, Can you please fix the pre-commit hook test that's failing right now?

PR #2653, opened 2026-08-02
Add support for multi arch container on docker compose examples

matdexir:
# Description

Based on the current docker compose examples, the bazelisk binary are hardcoded to use `amd64`/`x86_64`. Now that nativelink support both arm64 and x86 containers, there is no longer use for hardcoding to a specific architecture. 

## Type of change

  not work as expected)

## How Has This Been Tested?

## Checklist

- [ ] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2653)


matdexir:
I'm unsure if this needs a documentation update, please let me know and I will update accordingly.

palfrey:
Thanks for the fix!

> I'm unsure if this needs a documentation update, please let me know and I will update accordingly.

It does, because https://github.com/TraceMachina/nativelink/blob/4649c5034b0e04bce676b9825cb9bd15c52c5c69/deployment-examples/docker-compose/README.md?plain=1#L7 still says "This setup currently only works on x86_64/amd64 architectures"

matdexir:
Hi @palfrey,
I have updated the README accordingly. Hopefully this is clarifies things better for the end user.

deployment-examples/docker-compose/README.md
```diff
@@ -4,7 +4,7 @@ This guide explains how to run NativeLink with multiple workers using Docker Com
 
 ## Prerequisites
 
-- **Architecture**: This setup currently only works on **x86_64/amd64** architectures. ARM64/Apple Silicon isn't supported for the test client container.
+- **Architecture**: This setup supports **x86_64/amd64** and **ARM64** architectures, except for Apple Silicon, which is not supported by the test client container
```
palfrey, line 7: Why not Apple Silicon? I don't have one of them to hand, but I presume they can run the docker-compose setup just fine with an ARM64 image?

matdexir: In theory yes, since they're running lightweight linux machines but I have not tested it yet so I was a bit afraid to lie about it 🤣 .

palfrey: Fair! Will leave it in there for the moment, and will see if we get responses claiming otherwise :)

palfrey (approved):

PR #2659, opened 2026-08-03
fix(store): split gRPC write requests over the 4 MiB message limit

b7r6:
## What

`GrpcStore::update()` forwards each reader chunk as a single `WriteRequest`. A chunk larger than the receiving server's gRPC decoding limit (tonic's 4 MiB default) becomes one oversized message the peer rejects:

```
code: 'Operation was attempted past the valid range',
message: "Error, decoded message length too large: found 6291595 bytes, the limit is: 4194304 bytes"
```

`update_oneshot()` delivers a whole blob as one reader chunk, so any blob over 4 MiB proxied through `GrpcStore` fails this way (we hit it in production forwarding CAS blobs between NativeLink instances). `update_compressed()` has the same exposure with large encoder output chunks.

## Fix

Buffer the current chunk in `LocalState` and emit it across successive `WriteRequest`s capped at 2 MiB (headroom under the 4 MiB default), in both `update()` and `update_compressed()`. Behavior for chunks at or under the cap is unchanged, including the empty finishing request carrying `finish_write`.

## Tests

- `write_update_splits_chunks_over_grpc_message_limit`: a real tonic server with default limits + a 6 MiB blob. Fails against the uncapped client with the exact error above; with the fix, asserts multiple requests each under the limit, byte-identical reassembly, and `finish_write` on the final request.
- `write_stream_terminates_after_finish_write`: adds a draining `ByteStream` mock — the existing `FakeStreamServer` responds after the first message, which hides request-stream bugs — and locks `WriteRequestStreamWrapper`'s stream-termination invariant against a real drain.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2659)


nativelink-store/tests/grpc_store_test.rs
```diff
@@ -234,6 +234,164 @@ async fn write_update_works_core(
...
+}
+
+/// Regression test: a reader chunk larger than the gRPC maximum message size
+/// (tonic's 4 MiB decoding default) must be split across several `WriteRequest`s.
+/// Unsplit, the oversized request is rejected by the receiving server — this
+/// server runs tonic's defaults, so the test fails against an uncapped client.
+/// `update_oneshot` delivers a whole blob as one reader chunk, so any blob
+/// over 4 MiB took this path.
+#[nativelink_test]
+async fn write_update_splits_chunks_over_grpc_message_limit() -> Result<(), Error> {
```
palfrey, line 348: This is currently failing in CI

palfrey (changes requested):
Thanks for this. Can you sign the CLA please?

MarcusSorealheis:
@b7r6 You need to sign the CLA for compliance

Issue #2129, opened 2026-02-01
Lost inputs no longer available remotely


rejuvenile:
Randomy, in my Bazel build, I will often see a slew of errors like:

`Running Cargo build script crc32fast failed: Lost inputs no longer available remotely: external/rules_rs++crate+crates__crc32fast-1.5.0/crc32fast_build_script_ (47e96e44aa3acea97bd6ed4654277a29753a44863b525c23b540cf616ea75789/474144)`.

It's not clear to me what's causing this. The cache is warm, so I don't believe this is due to eviction, though it could be due to a race. It's possible the server or workers are not configured optimally; perhaps a consistency check is being missed.

rejuvenile:
I'm on the latest version of Nativelink (client and server): v0.8.0.

kkimdev:
@rejuvenile I believe that https://github.com/TraceMachina/nativelink/pull/2675 could help, but you need to enable the  `experimental_active_input_leases` flag. If you get a chance to try, let me know how it goes!

PR #2675, opened 2026-08-10
fix(worker): 🐛 lease active action inputs across eviction

kkimdev:
## 🚀 What

Keep every digest in an active action's input Merkle closure available in locally eviction-managed CAS tiers until the action has finished cleanup. This prevents `Lost inputs no longer available remotely` failures caused by local fast-tier eviction during input materialization.

## 🤔 Why

Under fast-tier pressure, an input could be evicted after an action began but before the worker materialized or hard-linked it. The resulting missing input caused remote execution failures such as `Lost inputs no longer available remotely` (see #2129).

## 🔧 How

- Add reference-counted, pre-admission leases to `EvictingMap`. A digest is leased before its value is inserted, so insertion cannot race with eviction. Leased keys are excluded from size, count, and TTL eviction.
- Keep a dedicated LRU of unleased entries synchronized with the resident LRU, so pressure eviction never selects leased inputs. If every resident entry is leased, defer eviction and emit diagnostic lease metrics/logging.
- Expose filesystem-backed tiers from `FastSlowStore`, including a filesystem store behind a `RefStore`, so the worker can protect every applicable local CAS tier.
- Lease the command and input-root digests immediately, then lease every file and directory digest before fetching or populating it. Child-directory leases are registered before their materialization futures are polled, closing the bounded-concurrency admission race.
- Make `DirectoryCache` lease-aware. Cache hits pin the materialized entry across hardlinking; cache misses reserve every CAS digest before construction. The directory-cache pin and CAS lease protect different resources and are released independently.
- Release leases after work-directory cleanup. Batch release performs one eviction pass for the whole Merkle closure, and detached release work keeps cleanup cancellation-safe. Duplicate operation IDs are rejected without deleting an active action directory.

## 💥 Breaking Changes

If every eviction candidate is leased, a local filesystem tier may temporarily exceed its configured capacity; normal eviction resumes after leases are released. Scheduler backpressure for this state is intentionally left to a follow-up. Lease discovery covers the filesystem-backed tiers used by the worker; remote and non-eviction stores remain unchanged.

## 🧪 Validation

- `nix develop -c cargo fmt --all -- --check`
- `nix develop -c cargo test -p nativelink-util --test evicting_map_test` (24 passed)
- `nix develop -c cargo test -p nativelink-worker --test running_actions_manager_test active_action_input_lease_survives_fast_store_pressure -- --nocapture` (1 passed)
- `nix develop -c cargo check -p nativelink-worker`

## 🔗 Related

Fixes the active-input eviction class described in #2129.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2675)

### DIFF
--- a/nativelink-util/src/evicting_map.rs
+++ b/nativelink-util/src/evicting_map.rs
@@ -145,10 +145,13 @@ impl<
     }
 
     /// Keep candidate recency aligned with a normal read from `lru`.
-    fn touch_evictable(&mut self, key: &Q) {
-        if !self.is_leased(key) {
-            self.evictable_lru.get(key);
+    /// Returns whether the key is leased so callers do not repeat the lookup.
+    fn touch_evictable(&mut self, key: &Q) -> bool {
+        if self.is_leased(key) {
+            return true;
         }
+        self.evictable_lru.get(key);
+        false
     }
 
     /// Add a newly written key to the candidate index unless it is reserved
@@ -423,8 +426,7 @@ where
     ) -> bool {
         let is_over_size = max_bytes != 0 && sum_store_size >= max_bytes;
 
-        let elapsed_seconds =
-            i32::try_from(self.anchor_time.elapsed().as_secs()).unwrap_or(i32::MAX);
+        let elapsed_seconds = self.elapsed_seconds();
         let evict_older_than_seconds = elapsed_seconds.saturating_sub(self.max_seconds);
         let old_item_exists =
             self.max_seconds != 0 && peek_entry.seconds_since_anchor < evict_older_than_seconds;
@@ -433,6 +435,10 @@ where
             self.max_count != 0 && u64::try_from(lru_len).unwrap_or(u64::MAX) > self.max_count;
 
         is_over_size || old_item_exists || is_over_count
+    }
+
+    fn elapsed_seconds(&self) -> i32 {
+        i32::try_from(self.anchor_time.elapsed().as_secs()).unwrap_or(i32::MAX)
     }
 
     // Gets a debugging snapshot of the state of the map. It's inevitably out of date by the time it gets sent to the user
@@ -547,10 +553,11 @@ where
             let mut data_to_unref = Vec::new();
             let mut removal_futures = Vec::new();
             for (key, result) in keys.into_iter().zip(results.iter_mut()) {
-                let is_leased = state.is_leased(key.borrow());
-                if !peek {
-                    state.touch_evictable(key.borrow());
-                }
+                let is_leased = if peek {
+                    state.is_leased(key.borrow())
+                } else {
+                    state.touch_evictable(key.borrow())
+                };
                 let maybe_entry = if peek {
                     state.lru.peek_mut(key.borrow())
                 } else {
@@ -573,9 +580,7 @@ where
                             }
                         } else {
                             if !peek {
-                                entry.seconds_since_anchor =
-                                    i32::try_from(self.anchor_time.elapsed().as_secs())
-                                        .unwrap_or(i32::MAX);
+                                entry.seconds_since_anchor = self.elapsed_seconds();
                             }
                             *result = Some(entry.data.len());
                         }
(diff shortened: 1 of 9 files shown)


MarcusSorealheis:
Thanks for the PR! We are going to review it as time permits.

kkimdev:
The only thing on my mind is whether this should be controlled by an optional (opt-in) flag. I'm fine either way, so please let me know!

kkimdev:
OK I've finished a few iterations and it's now ready for review I believe. It's a big change but if you exclude the tests, it's about ~500 new lines.

nativelink-util/src/evicting_map.rs
```diff
@@ -105,6 +105,14 @@ struct State<
...
     lru: LruCache<K, EvictionItem<T>>,
     btree: Option<BTreeSet<K>>,
+    /// A mirror of `lru` that contains only currently unleased keys.
+    /// Keeping this index in LRU order lets eviction visit candidates without
+    /// scanning entries that an active operation has pinned.
+    evictable_lru: LruCache<K, ()>,
+    /// Reference counts for keys that must not be evicted while an active
+    /// operation is using them. A key can be leased before it is inserted,
+    /// which closes the admission-to-insert race for cache populations.
+    leases: HashMap<K, u32>,
```
corcillo, line 115: So (as acknowledged elsewhere in the PR) if a bunch of big items are leased we can silently exceed max_bytes/max_count. Maybe we can add two new fields to the State struct (like leased_items and leased_bytes) with #[metric] annotations (see example in sum_store_size just below), so we know how many bytes are leased and not evictable? 

`leased_bytes` would also need a running sum updated in lease/release_lease/remove, since a key can be leased before its value exists. 

We could also use logging around this for when eviction is skipped because everything is leased, but will add that comment elsewhere.

nativelink-util/src/evicting_map.rs
```diff
@@ -374,15 +469,17 @@ where
 
     #[must_use]
     fn evict_items(&self, state: &mut State<K, Q, T, C>) -> (Vec<T>, Vec<RemoveFuture>) {
-        let Some((_, mut peek_entry)) = state.lru.peek_lru() else {
+        let Some(first_entry) = state.peek_evictable() else {
```
corcillo, line 494: Can we add debug logging here if lru is NOT empty, to say an eviction was requested but every entry is leased? along with info on states leases

nativelink-util/src/evicting_map.rs
```diff
@@ -147,6 +229,8 @@ impl<
         if let Some(btree) = &mut self.btree {
             btree.remove(key);
         }
+        // Keep every auxiliary resident index in sync with the main LRU.
+        self.evictable_lru.pop(key);
```
corcillo, line 245: Since an explicit remove deliberately leaves a leased key's lease alive with only the lease_key doc comment guarding it, could we add a unit test covering lease → insert → remove → insert → release, and finally check it's evictable again with no leftover lease?

corcillo:
Overall looks solid! just made a few comments regarding metrics/logging/tests that will help us debug in the future.

kkimdev (commented):
Thanks for the review! My only concern is that this can (temporarily) exceed the configured max_bytes / max_count limits. What do you think about gating this behind an opt-in flag?

nativelink-util/src/evicting_map.rs
kkimdev: Done!

nativelink-util/src/evicting_map.rs
kkimdev: Done!

nativelink-util/src/evicting_map.rs
kkimdev: Done!

MarcusSorealheis:
@kkimdev Please gate behind flag but default to off

corcillo:
I added a default off flag for this in a new commit

MarcusSorealheis (approved):
Approve

PR #2680, opened 2026-08-12
Don't let one scheduler instance time out another's in-flight actions

amankrx:
# Description

This issue is wrt to the HPA where one scheduler instance times out the other's in-flight actions, resulting in CAS pod crashes or builds stalling. 

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Added unit tests. 

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2680)

### DIFF
--- a/nativelink-scheduler/src/simple_scheduler_state_manager.rs
+++ b/nativelink-scheduler/src/simple_scheduler_state_manager.rs
@@ -41,7 +41,7 @@ use tracing::{debug, info, trace, warn};
 use super::awaited_action_db::{
     AwaitedAction, AwaitedActionDb, AwaitedActionSubscriber, SortedAwaitedActionState,
 };
-use crate::worker_registry::SharedWorkerRegistry;
+use crate::worker_registry::{SharedWorkerRegistry, WorkerLiveness};
 
 /// Maximum number of times an update to the database
 /// can fail before giving up.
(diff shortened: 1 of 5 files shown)


MarcusSorealheis:
The tri-state liveness direction is good, but the fix is incomplete for HPA deployments.

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -367,34 +367,42 @@ where
...
-            if self.max_executing_timeout > Duration::ZERO {
-                let last_update = awaited_action.last_worker_updated_timestamp();
-                if let Ok(elapsed) = now.duration_since(last_update) {
-                    return elapsed > self.max_executing_timeout;
+        match liveness {
+            // Unknown means "not registered here", which with several
+            // schedulers on shared state is usually a peer's healthy worker.
+            // We never see its heartbeats, so only the stuck-but-alive ceiling
+            // can judge it; that also backstops a genuine orphan.
+            WorkerLiveness::Alive | WorkerLiveness::Unknown => {
```
MarcusSorealheis, line 386: You treat `Unknown` like `Alive`, relying exclusively on `max_action_executing_timeout_s`. That setting defaults to zero/disabled in nativelink-config/src/schedulers.rs:139, and the enterprise deployment also advertises zero.

MarcusSorealheis (commented):
[nativelink-scheduler/src/store_awaited_action_db.rs:734](https://github.com/TraceMachina/nativelink/blob/624420d63aa95d3e3976034cd9a0ca283bc0afd8/nativelink-scheduler/src/store_awaited_action_db.rs#L734) independently treats an executing action as abandoned when its shared `last_worker_updated_timestamp` + `worker_timeout_s` expires. See the comment above the line linked above. Worker heartbeats update only the owner’s process-local registry, not this shared timestamp.

Therefore, a retried/duplicate Execute arriving on another scheduler can still create a new queued operation while the original worker is healthy. That risks duplicate execution, increased CAS traffic, and broken/stalled subscribers—the production failure this PR is intended to solve.

MarcusSorealheis:
If you want to punt on solving these now and simply add a todo, I'm ok with it, especially regarding [this comment](https://github.com/TraceMachina/nativelink/pull/2680#discussion_r3763593085).

MarcusSorealheis:
In general, this is a major improvement. The flake is identified in #2682 and fixed in #2683. We should merge this PR immediately after that.

MarcusSorealheis (approved):

PR #2684, opened 2026-08-12
Add Azurite test support for AzureBlobStore

### LINKED ISSUE #2511: Test Azure work with Azurite
As discussed at https://github.com/TraceMachina/nativelink/pull/2472#discussion_r3530348152. @amankrx said "We can run Azurite in CI as a service container, same pattern as the existing mongo_runner. We can point the store at it via endpoint and a SAS URL", and that's what this issue covers.

Muktarsadiq:
## What and why

Adds Azurite (Azure Storage emulator) support to the test suite for
`AzureBlobStore`, so it's exercised against a real emulator rather than
only mocks. Follows the existing `mongo_runner` embedded-runner pattern.

Fixes #2511

## How was this verified?

Ran `cargo test -p nativelink-store --test azurite_store_test` locally:
`upload_and_get_data`, `upload_empty_data`, and
`zero_len_items_exist_check` all pass against a real `AzureBlobStore`
talking to a locally spawned Azurite instance no mocking. Two
`#[ignore]`d tests document manual verification of the SAS signer and
container-bootstrap logic, done independently before wiring them into
the automatic flow.

Also ran the full suite under Bazel/Nix per review feedback
(`bazel-retry test //nativelink-store:integration ...`): 25/27 pass.
The 2 known failures are `azurite_store_test_test` (Bazel's sandboxed
test execution has no network access to run `bun install`, so
`azurite-blob` isn't present at runtime, same root cause as the
`p.mongodb` symlink hack `mongo_runner` needs in `flake.nix`) and a
one-off `grpc_store_test_test` timeout that passes cleanly in
isolation, confirmed as resource pressure flakiness rather than a
real regression.

Fixed 11 Clippy findings surfaced by the Bazel/Nix run (import
ordering, `tokio::spawn` → `nativelink_util::background_spawn!`, doc
comment formatting, `#[tokio::test]` → `#[nativelink_test]`, etc.),
and a Windows-specific bug where the runner looked for `azurite-blob`
without the `.cmd` extension npm/bun generate on that platform.

Could not verify `bazel test //...` (full monorepo) or the Nix-driven
pre-commit hooks directly at first, due to macOS 12/Monterey being
unsupported by the current Nix installer. Later verified both via an
Ubuntu VM once set up.

## Risk

Low. This is purely additive test infrastructure, no production code
(`azure_blob_store.rs` itself) is touched beyond adapting one call site
to match its `async`/non-`async` signature after an unrelated upstream
Rust 1.97.1 toolchain bump changed it mid-review. New dev-only
dependencies (`chrono`, `hmac`) and a Bun-based install step affect
test builds and CI only, not the shipped `nativelink` binary. The known
gap is that the new Azurite tests don't yet execute under Bazel/Nix's
sandboxed test paths (see above), a coverage gap for this new suite,
not a regression, since the existing mocked `azure_blob_store_test.rs`
still runs everywhere unaffected.

## AI assistance

Used GitHub Copilot to help diagnose several CI failure logs during
review (Clippy findings, a stale commit false positive on an
`AzureBlobStore::new()` signature question, and the PR-template check
itself), verified each suggestion against the actual source/log
output before applying, and caught at least one incorrect diagnosis
from it in the process.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2684)


MarcusSorealheis:
Thank you for the PR. We will check it out soon!

.github/workflows/native-cargo.yaml
```diff
@@ -5,13 +5,13 @@ on:
...
-      - 'web/**'
+      - ".github/styles/**"
+      - "web/**"
   pull_request:
     branches: [main]
     paths-ignore:
-      - '.github/styles/**'
-      - 'web/**'
+      - ".github/styles/**"
+      - "web/**"
```
palfrey: Can you please reset the format changes here

.github/workflows/native-cargo.yaml
```diff
@@ -53,6 +53,18 @@ jobs:
...
+        uses: >- # v7.0.0
+          actions/setup-node@820762786026740c76f36085b0efc47a31fe5020
+        with:
+          node-version: "22"
+          cache: "npm"
+          cache-dependency-path: nativelink-store/tests/azurite_runner/package-lock.json
+
+      - name: Install Azurite for Azure store tests
+        run: npm ci
+        working-directory: nativelink-store/tests/azurite_runner
```
palfrey: Please use bun for package management, as per our existing other usage in https://github.com/TraceMachina/nativelink/blob/main/web/bun.lock

nativelink-store/Cargo.toml
```diff
@@ -135,7 +135,13 @@ wincode = { version = "0.5.4", default-features = false, features = [
 libc = { version = "0.2.177", default-features = false }
 
 [dev-dependencies]
+chrono = { version = "0.4.44", default-features = false, features = ["clock"] }
+hmac = { version = "0.12.1", default-features = false }
 nativelink-macro = { path = "../nativelink-macro" }
+tokio = { version = "1.52.2", features = [
+  "process",
+  "sync",
+], default-features = false }
```
palfrey: Please move these further down. Our standard formatting for dependencies is the nativelink internal ones at the top, then everything else in alphabetical order after that.

.github/workflows/native-cargo.yaml
palfrey: Also, please make these steps a new reusable action, as they'll need to get added to various other workflows e.g. `.github/workflows/nix.yaml` and `.github/workflows/coverage.yaml`

nativelink-store/tests/azurite_runner/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "nativelink-azurite-test-runner",
+  "private": true,
+  "dependencies": {
+    "azurite": "3.36.0"
+  }
+}
```
palfrey, line 7: This needs reformatting with the `pre-commit` rules. If you run the [Nix setup](https://docs.nativelink.com/contribute/nix) you'll have `pre-commit` setup correctly as part of that.

nativelink-store/tests/azurite_store_test.rs
palfrey, line 1: This needs adding to the test suite in `nativelink-store/BUILD.bazel`

palfrey (changes requested):
Thanks for doing this. There's a few issues though.

.github/workflows/native-cargo.yaml
Muktarsadiq: For nix.yaml's nix-cargo job, i added the composite action that works cleanly since it runs inside nix develop (a shell, not a sandbox). on the other hand coverage.yaml's job runs via nix build, a fully sandboxed derivation with no network access, the same reason mongo_runner needs the p.mongodb symlink hack in flake.nix. Azurite isn't a nixpkgs-native package the way MongoDB is, so the same trick isn't directly available. Happy to look into packaging Azurite for Nix if that's the preferred path, but wanted to flag this as a separate problem rather than silently skip it or fake a fix.

nativelink-store/tests/azurite_store_test.rs
Muktarsadiq: Added tests/azurite_store_test.rs to the suite, plus the needed compile_data/deps. Same sandboxing caveat as my note on the coverage.yaml comment. I suspect bazel test's sandbox has the same no network issue, so I can't verify this actually passes at runtime, only that it should compile. Happy to dig further if useful.

nativelink-store/BUILD.bazel
```diff
@@ -177,12 +180,15 @@ rust_test_suite(
         "@crates//:aws-smithy-types",
         "@crates//:azure_core",
         "@crates//:azure_storage_blob",
+        "@crates//:base64"
```
palfrey: Missing comma at the end of this line

palfrey:
I would suggest running all of this locally under Bazel with nix, so you can fix the variety of lint failures there as well as run `pre-commit run -a` to fix all the pre-commit issues currently being flagged in the build.

Muktarsadiq:
Ran bazel-retry test //nativelink-store:integration under Nix. 25/27 pass. The 2 failures were azurite_store_test_test fails exactly as flagged in my earlier comment (sandbox has no network to run bun install), and grpc_store_test_test failed on an unrelated pre existing test with a hardcoded 5s timeout, which I believe was resource pressure flakiness on my VM rather than a real regression

MarcusSorealheis:
@Muktarsadiq You are very close. Let me know if you need any help. The answer is in the logs here.

Muktarsadiq:
> @Muktarsadiq You are very close. Let me know if you need any help. The answer is in the logs here.

Muktarsadiq:
@MarcusSorealheis Thank you for the encouragement, i do indeed need some help, i'm struggling to fix a couple of the failed checks.

nativelink-store/tests/azurite_store_test.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    // Split out like `mongo_store_test`'s equivalent: separate from `new_spec`
+    // because some tests want to edit the spec after getting a live process
+    // but before constructing the store.
+    async fn new_with_spec_and_process(
+        spec: ExperimentalAzureSpec,
+        azurite_local: Option<AzuriteParts>,
+    ) -> Result<Self, Error> {
+        let now_fn: fn() -> SystemTime = SystemTime::now;
+        let store = AzureBlobStore::new(&spec, now_fn).await?;
```
palfrey, line 104:
```suggestion
        let store = AzureBlobStore::new(&spec, now_fn)?;
```
This will fix the main current build failures

nativelink-store/tests/azurite_runner/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+/// the whole suite. Requires `npm ci` to have been run in this directory
+/// first; see this crate's `tests/azurite_runner/package.json`.
+fn find_azurite_binary() -> Result<PathBuf, Error> {
+    let binary_name = if cfg!(windows) {
+        "azurite-blob.cmd"
+    } else {
+        "azurite-blob"
+    };
+
+    let path = Path::new(env!("CARGO_MANIFEST_DIR"))
```
palfrey: This now breaks in the Bazel builds
`Error: ProcessWrapperError("compiled Rust output bazel-out/k8-fastbuild/bin/nativelink-store/integration_tests/azurite_store_test_test embeds the absolute working directory /home/runner/.bazel/sandbox/processwrapper-sandbox/2461/execroot/_main. Do not retain env!(\"CARGO_MANIFEST_DIR\") or env!(\"OUT_DIR\") in compiled code; include_str!() and include_bytes!() may use those values only for compile-time file access")`

This is because [we upgraded rules_rs](https://github.com/TraceMachina/nativelink/pull/2724) and that pulled in https://github.com/hermeticbuild/rules_rust/pull/36. Replacing this with `std::env::var` i.e. runtime not build time and possibly also doing some checks to see if you're in bazel paths or not may well help here.

Muktarsadiq: i switched to std::env::var("CARGO_MANIFEST_DIR") at runtime instead of the env!() compile time macro. Verified with bazel-retry test //nativelink-store:integration_tests/azurite_store_test_test, the ProcessWrapperError about the embedded absolute path is gone, and the test now fails only with the already flagged "azurite-blob not found, run npm ci" sandbox network limitation.

PR #2685, opened 2026-08-13
Reap orphaned actions when the executing timeout is disabled

amankrx:
# Description

Follow-up to #2680. Two places still decide an action is abandoned by reading
a timestamp in the shared store that worker heartbeats never update.

One leaks: on the default config `max_action_executing_timeout_s` is 0, so an
action whose scheduler went away is never cleaned up. The other duplicates
work: `try_subscribe` recreates an action that is still running instead of
joining it, so the same build step executes twice.

Both now check worker liveness first, and an unrecognised worker gets a long
backstop instead of nothing.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Added unit tests. Each one fails if you revert the fix it covers.

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2685)


MarcusSorealheis (approved):
LGTM

PR #2679, opened 2026-08-11
Chunk ByteStream writes instead of sending a whole blob as one message

amankrx:
# Description

As the title suggests, this PR chunks ByteStream writes instead of sending a whole blob as one message. We recently observed a failure where the ByteStream message exceeded the limit, and resulted in a build failure. This PR should fix the issue. 

Fixes # (issue)

## Type of change

Please delete options that aren't relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Added unit tests

## Checklist

- [X] Updated documentation if needed
- [X] Tests added/amended
- [X] `bazel test //...`  passes locally
- [X] PR is contained in a single commit, using `git amend` see some [docs](https://www.atlassian.com/git/tutorials/rewriting-history)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2679)

PR #2686, opened 2026-08-13
Chunk oversized messages in the ByteStream write and batch update paths

amankrx:
# Description

Follow-up to #2679, which fixed one of three places that send a whole blob as
a single gRPC message. The other two were missed: the `ByteStream` write
pass-through, and `batch_update_blobs`.

Both are hit when NativeLink sits in front of another CAS. The receiver
rejects anything over its 4MiB default, and it fails the same way on every
retry.

Writes are now split into smaller messages. A batch is one message and cannot
be split that way, so an oversized batch is sent as several RPCs instead.

## Type of change

## How Has This Been Tested?

Added unit tests for both paths. Each one fails if you revert the fix it
covers.

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2686)


MarcusSorealheis (approved):
LGTM

PR #2687, opened 2026-08-13
Add OTLP metrics for workers, gRPC, the scheduler, store tiers and health

### LINKED ISSUE #2614: Complete the OTLP metrics coverage: instrument declared-but-missing metrics and add missing operational metrics
## Context
Follow-up to #2583 / PR #2599. Two execution instruments (execution_cpu_time, execution_memory_usage) were removed in the PR for lack of a data source. NativeLink has two metric systems: OTLP (Prometheus/Grafana, currently only cache_* + execution_*) and the older #[metric] component system (separate JSON endpoint). Lots of operational data lives only in the old component system, and in one case the OTLP docs reference a metric that's never emitted.
1. Removed execution metrics (need worker-side sampling, ExecutionMetadata has no CPU/mem): 
  - `execution_cpu_time`
  - `execution_memory_usage`                                                                                                    
  - `nativelink_worker_connected_count`                                                                                             
2. Missing OTLP categories (mostly exist as #[metric], need bridging):
  - Worker fleet: connected count, per-worker started/completed, paused/draining, keepalive/disconnect (worker.rs). Fleet health + autoscaling.                                                                                         
  - Scheduler queue: depth per priority + matching latency. Primary saturation signal.
  - gRPC/API RED: rate/errors/latency per service (CAS/AC/Execution/Capabilities/ByteStream). No OTLP coverage; one interceptor covers all.                                                                                             
  - Store/CAS I/O: bytes r/w per store, fast/slow hit ratio, error rate (fast_slow_store.rs).
  - Health/connection: health-check pass/fail, pool utilization, Redis conn state.

amankrx:
## What and why

Fixes #2614. NativeLink had OTLP metrics for cache and execution only. Everything else lived in the older `#[metric]` system or nowhere.

Adds the missing categories: workers (connected, connects, disconnects, keepalives, paused/draining), gRPC (one duration histogram by service, method and status, giving rate, errors and latency together), scheduler matching duration, fast/slow store tier hits and bytes, and health check results.

Also the two execution metrics dropped in #2599 for want of a data source. Memory comes from `ActionResourceUsage.peak_memory_kb`, which the worker already reported. CPU time is new: the worker samples `/proc/<pid>/stat` for the process group every 250ms already, so it now reads `utime`/`stime` from the same line and reports it as `cpu_time_ms`.

Connection pools are covered too: free slots at acquisition, whether a caller had to queue, and Redis reconnects. A pool that keeps reporting zero free slots is the bottleneck whatever the latency elsewhere says.

Nothing is labelled by worker id or action digest; both grow without bound.

## How this was verified

Unit tests for the metric helpers, the gRPC path splitter, the health status labels, and the `/proc/<pid>/stat` CPU parser.

## Risk

Moderate, mostly in accounting rather than behaviour.

CPU time is cumulative per process and a process that exits disappears from `/proc`, so the sampler keeps the last value per pid and totals them at the end rather than summing the live group. `cutime`/`cstime` are excluded, since they only cover reaped children and would double count.

Worker counts are per scheduler instance; with several schedulers the fleet total is the sum. They are exact by construction: unbounded LRU so nothing is silently evicted, one removal path, and re-adding an id does not double count.

gRPC status comes from response headers. Tonic puts it in trailers when a call fails mid-stream, and trailers are not reachable from that layer without consuming the body, so error rate undercounts mid-stream failures. Rate and latency are exact.

The proto gains `cpu_time_ms` at tag 5, a new optional field, so an old worker reports zero.

Nothing from #2614 is left out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2687)

### DIFF
--- a/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
+++ b/nativelink-proto/com/github/trace_machina/nativelink/remote_execution/worker_api.proto
@@ -125,7 +125,12 @@ message ActionResourceUsage {
     /// The worker ID that observed the resource usage.
     string worker_id = 4;
 
-    reserved 5; // NextId.
+    /// Total CPU time consumed by the action process tree, in milliseconds.
+    /// User and system time combined, summed across every process in the
+    /// action's process group.
+    uint64 cpu_time_ms = 5;
+
+    reserved 6; // NextId.
 }
 
 /// Result sent back from the server when a node connects.
--- a/nativelink-proto/genproto/com.github.trace_machina.nativelink.remote_execution.pb.rs
+++ b/nativelink-proto/genproto/com.github.trace_machina.nativelink.remote_execution.pb.rs
@@ -104,6 +104,11 @@ pub struct ActionResourceUsage {
     /// / The worker ID that observed the resource usage.
     #[prost(string, tag = "4")]
     pub worker_id: ::prost::alloc::string::String,
+    /// / Total CPU time consumed by the action process tree, in milliseconds.
+    /// / User and system time combined, summed across every process in the
+    /// / action's process group.
+    #[prost(uint64, tag = "5")]
+    pub cpu_time_ms: u64,
 }
 /// / Result sent back from the server when a node connects.
 #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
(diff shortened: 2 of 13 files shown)


MarcusSorealheis (approved):
LGTM

PR #2688, opened 2026-08-13
Add an optional TTL for keys a redis store writes

amankrx:
# Description

Keys a redis store writes never expire, so a store whose consumer stops keeps everything it ever wrote. One deployment reached 1.79M keys with `expires=0` after its BEP consumer stalled, and the Redis node ran out of memory.

The store already supports a TTL, but only on the versioned write path used by the scheduler. The plain `update()` path a BEP store takes has no expire call at all.

Adds `key_ttl_s` to the redis store config. When set, `update()` expires the key after the rename. Default is 0, meaning no expiry, so nothing changes for any existing store.

Deliberately per store, not global. The same store type backs the CAS fast tier and the scheduler, and expiring scheduler state would drop in-flight actions.

Note this trades data for a bound: anything not consumed inside the window is deleted, so the value has to exceed the longest consumer outage you want to survive. That is called out in the config docs.

## Type of change

## How Has This Been Tested?

Added unit tests, one for the TTL being set and one for the default staying
off. The first fails if you revert the fix.

## Checklist

- [X] Tests added/amended

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2688)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -1785,6 +1785,30 @@ pub struct RedisSpec {
     #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
     pub connection_pool_size: usize,
 
+    /// Expire keys written by this store after this many seconds.
+    ///
+    /// Redis does not expire these on its own, so a store whose consumer
+    /// stops keeps every key it ever wrote. A BEP store is the case this
+    /// exists for: if the ETL stops consuming, the backlog grows until the
+    /// Redis node runs out of memory.
+    ///
+    /// Set this per store, not globally. The same store type backs the CAS
+    /// fast tier and the scheduler, and expiring scheduler state would drop
+    /// in-flight actions.
+    ///
+    /// This trades data for a bound. Anything not consumed within the window
+    /// is deleted, so the value has to exceed the longest consumer outage you
+    /// intend to survive, and it must also exceed how long a single upload can
+    /// take: the temp key an upload builds carries this same TTL, so a value
+    /// below the upload duration would expire the write in flight.
+    ///
+    /// Zero disables expiry. One second is the smallest value that enables it,
+    /// and any value that small is almost certainly a mistake.
+    ///
+    /// Default: 0 (keys never expire)
+    #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
+    pub key_ttl_s: u64,
+
     /// The maximum number of upload chunks to allow per update.
     /// This is used to limit the amount of memory used when uploading
     /// large objects to the redis server. A good rule of thumb is to
--- a/nativelink-store/src/redis_store.rs
+++ b/nativelink-store/src/redis_store.rs
@@ -125,6 +125,16 @@ pub const DEFAULT_MAX_COUNT_PER_CURSOR: u64 = 1_500;
 
 const DEFAULT_CLIENT_PERMITS: usize = 500;
 
+/// Converts a TTL to the seconds `EXPIRE` takes.
+///
+/// Clamped to at least one second. Redis treats `EXPIRE` with a value of zero
+/// or less as "delete now", so a sub-second duration reaching here would
+/// destroy the value it was meant to protect. Config already rejects zero, so
+/// this only guards a future caller.
+fn ttl_seconds(key_ttl: Duration) -> i64 {
+    i64::try_from(key_ttl.as_secs()).unwrap_or(i64::MAX).max(1)
+}
+
 /// A wrapper around Redis to allow it to be reconnected.
 pub trait RedisManager<C>
 where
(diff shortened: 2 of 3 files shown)


nativelink-config/src/stores.rs
```diff
@@ -1785,6 +1785,25 @@ pub struct RedisSpec {
...
+    /// fast tier and the scheduler, and expiring scheduler state would drop
+    /// in-flight actions.
+    ///
+    /// This trades data for a bound. Anything not consumed within the window
+    /// is deleted, so the value has to exceed the longest consumer outage you
+    /// intend to survive. The default is deliberately off.
+    ///
+    /// Default: 0 (keys never expire)
+    #[serde(default, deserialize_with = "convert_numeric_with_shellexpand")]
+    pub key_ttl_s: u64,
```
corcillo, line 1810: this is probably an obvious answer but the right ttl minimum should always be >1 second right?

nativelink-store/src/redis_store.rs
```diff
@@ -1176,6 +1185,21 @@ where
...
         }
 
+        // Expire the final key, not the temp one: RENAME would carry a TTL
+        // over, but setting it here keeps the window where the key exists
+        // without one down to a single round trip and keeps the intent
+        // obvious. A failure here is a real error rather than something to
+        // swallow, since a key that silently never expires is the bug this
+        // option exists to prevent.
+        if let Some(key_ttl) = self.key_ttl {
+            let ttl_secs = i64::try_from(key_ttl.as_secs()).unwrap_or(i64::MAX);
```
corcillo, line 1228: so all of the other steps in update do this on error: 
1) check`is_retryable_redis_error` 
2) attempt to reconnect once
3) attempt to retry command once

Is there a reason you didnt have this one do that? or should we add that here too?

nativelink-store/src/redis_store.rs
```diff
@@ -1176,6 +1185,21 @@ where
...
+            client
+                .connection_manager
+                .expire::<_, ()>(final_key.as_ref(), ttl_secs)
```
corcillo, line 1231: So this will eventually publish an event of type `expired`?   Do we need to add a new type to the RedisSubscriptionManager, because right now we only catch `if value == "evicted"`?

Also I don't think that pathway is tested either?

nativelink-store/tests/redis_store_test.rs
corcillo, line 1: So these tests only test *setting* ttl right? Can we add tests to make sure eviction happens when keys expire? I believe that would also catch the comment i made on redis_store about expired types not actually being handled (or prove it wrong)

nativelink-store/src/redis_store.rs
```diff
@@ -1176,6 +1185,21 @@ where
             }
         }
 
+        // Expire the final key, not the temp one: RENAME would carry a TTL
```
corcillo, line 1221: wouldnt we also want to set a TTL on the temp key incase the copying over fails midway? This was the temp keys will get cleaned up even if final key is not added? 

So as is we do:
1. Write chunks → temp key
2. Verify length
3. RENAME temp → final
4. EXPIRE final ← new
5. Publish

If it fails between 1–3 then temp key lives forever. and if it fails at 4 until the data is stored, upload reports failure, and doesn't publish.

Instead:
1. Write first chunk → temp key
2. *set expiration on temp key*
3. Write rest, verify length
4. RENAME temp → final (TTL rides along)
5. Publish

This way if it fails anywhere the temp key self-destructs.

I see your comment `A failure here is a real error rather than something to
        // swallow, since a key that silently never expires is the bug this
        // option exists to prevent.` But not doing adding expiration to temp keys this will just accumulate more dead keys and not surface the error either right?  Also adding a retry like i say in the other comment should help too?

nativelink-config/src/stores.rs
amankrx: Redis treats EXPIRE with ≤0 as delete now

nativelink-store/src/redis_store.rs
amankrx: updated!

nativelink-store/tests/redis_store_test.rs
amankrx: done

nativelink-store/src/redis_store.rs
amankrx: updated!

nativelink-store/src/redis_store.rs
corcillo: This looks the same?

Here is an example from above, where they check the error is retryable, and try to reconnect once and retry.  Is there a reason not to also do an equivalent here?
From line 1195:
```match client
    .connection_manager
    .rename::<_, _, ()>(&temp_key, final_key.as_ref())
    .await
{
    Ok(()) => {}
    Err(err) if is_retryable_redis_error(&err) => {
        let (connection_manager, uuid) =
            self.connection_manager.reconnect(client.uuid).await?;
        client.connection_manager = connection_manager;
        client.uuid = uuid;
        client
            .connection_manager
            .rename::<_, _, ()>(&temp_key, final_key.as_ref())
            .await
            .err_tip(
                || "While queueing key rename (after reconnect) in RedisStore::update()",
            )?;
    }
    Err(err) => {
        return Err(
            Error::from(err).append("While queueing key rename in RedisStore::update()")
        );
    }

corcillo (approved):

nativelink-store/src/redis_store.rs
amankrx: Oops, Missed the retryable check on the final key. Will consolidate this change in the next PR.

PR #2689, opened 2026-08-13
Rework the PR template and check that it is filled in

amankrx:
## What and why

The PR template asks for things CI already proves and not much a reviewer actually needs, so it gets filled in as ceremony. Measured across the last 33 merged PRs that used it:

- 29/33 tick "`bazel test //...` passes locally". Nobody runs the full suite before every PR, so the box carries no information.
- 17/33 tick every box, including all four mutually exclusive "Type of change" options.
- 15/33 leave "Please delete options that aren't relevant" in the body.
- 14/33 have an effectively empty "How Has This Been Tested?", which is the one section a reviewer cannot get from the diff.
- 6/33 leave "Fixes # (issue)" verbatim.

Several of those are mine. This replaces it with three sections that each ask for something only the author knows: what problem this solves, how it was actually verified, and what breaks if it is wrong. All guidance moved into HTML comments so it cannot leak into the permanent record.

The checklist is gone. `bazel test //...`, formatting and lints are already CI jobs and pre-commit hooks, which enforce them properly rather than on the honour system.

## How this was verified

Ran the checker against real inputs: a filled-in body passes, the unedited template fails on all three sections, a body with no headings fails, an empty body fails, and a body with "n/a" under a heading fails on that heading only. Also ran it against #2683, an existing PR using the old template, and confirmed it reports the headings as missing rather than crashing.

Workflow YAML parses and the triggers resolve as intended. I could not exercise the workflow end to end, since that only happens once it is on the default branch.

## Risk

This is a process change, so it needs agreement rather than just review.

The check fails PRs whose description is thin, which is the point, but two details keep it from being hostile:

- It skips bots and honours a `skip-pr-template` label, which a maintainer will need to create.
- It comments saying exactly which section is short and re-runs on `edited`, so it goes green as soon as the author fixes it. A red X with no explanation is how you lose a drive-by contributor.

Open PRs written against the old template will fail this check the next time they are edited or pushed to. The label is the escape hatch, or hold the merge until the queue drains.

It runs on `pull_request_target` so it can comment on fork PRs, where a first-time contributor most needs the hint. Nothing from the PR is checked out or executed: the checkout is the base branch, and the description is only ever read from a file, never interpolated into a shell command.

No new third-party action; it reuses the `actions/checkout` pin already trusted in this repo and the preinstalled `gh` CLI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2689)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,7 +14,7 @@ that, and it is better at it than either of us. -->
 If there is an issue, link it as "Fixes #123" so it closes on merge. If there
 isn't, say what prompted the change. -->
 
-## How this was verified
+## How was this verified?
 
 <!-- What you ran, on what, and what you saw. Beyond CI.
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,27 +1,36 @@
-# Description
+<!-- Thanks for contributing to NativeLink.
 
-Please include a summary of the changes and the related issue. Please also
-include relevant motivation and context.
+Everything outside an HTML comment becomes part of the permanent record, so
+replace the prompts below with your own words. The comments themselves are
+invisible once posted; leaving them in is fine.
 
-Fixes # (issue)
+You do not need to confirm that tests, formatting or lints pass. CI checks
+that, and it is better at it than either of us. -->
 
-## Type of change
+## What and why
 
-Please delete options that aren't relevant.
+<!-- What this changes, and what problem it solves. Two or three sentences.
 
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to
-  not work as expected)
-- [ ] This change requires a documentation update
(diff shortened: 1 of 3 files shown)


palfrey (approved):

MarcusSorealheis (approved):
LGTM

corcillo (approved):

.github/pull_request_template.md
```diff
@@ -1,27 +1,36 @@
...
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to
-  not work as expected)
-- [ ] This change requires a documentation update
+If there is an issue, link it as "Fixes #123" so it closes on merge. If there
+isn't, say what prompted the change. -->
 
-## How Has This Been Tested?
+## How this was verified
```
modernmedici: How was this verified?

modernmedici (commented):
Looks good, just adjusted the wording in one place to make sense.

PR #2690, opened 2026-08-13
Sample CPU time for actions and report it with resource usage

amankrx:
## What and why

Part of #2614 (Separating the PR out, since this feature is getting quite big). `execution_cpu_time` was removed in #2599 because nothing reported CPU time.

The worker already samples the action's process group every 250ms for peak memory, reading `/proc/<pid>/stat` for the process group id. `utime` and `stime` are in that same line, so this reads them while it is already there and reports the total as `cpu_time_ms` on `ActionResourceUsage`.

That message already flows to the scheduler and out as an origin event, so the value reaches the BEP pipeline with no further plumbing.

## How this was verified

Added unit tests for the parser.

## Risk

Low, and confined to Linux workers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2690)


amankrx:
Better to handle this directly in #2687

PR #2696, opened 2026-08-18
Release NativeLink v1.6.5

amankrx:
## What and why

Releases Nativelink v1.6.5. This commit is the version bump and changelog only; every change in it landed and was reviewed in its own PR.

Focuses on multi-replica scheduler correctness and large-blob handling. Two scheduler fixes stop replicas timing out each other's in-flight actions and stop orphaned actions leaking on the default configuration, and the remaining ByteStream paths now chunk oversized messages instead of failing them outright. Also adds full OTLP metric coverage for workers, gRPC, the scheduler, store tiers and health, and an optional TTL for Redis store keys. Anyone running more than one CAS replica, or proxying to another CAS, should upgrade.

## How was this verified?

 Each change was verified in its own PR: unit tests throughout. And the environment was tested in a live cluster with positive results. 

## Risk

Moderate, and concentrated in two behaviour changes rather than in the release mechanics.                                                                                                                    

- Orphaned actions now get reaped. On a deployment leaving `max_action_executing_timeout_s` at its default of 0, an action whose scheduler disappeared previously hung until the client gave up; it is now cleaned up after an hour.
- Duplicate Execute now joins instead of recreating. 
- Wire framing changed for ByteStream pass-through. 
- The Redis key TTL defaults to 0 (no expiry), so it does nothing unless configured.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2696)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,47 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.6.5](https://github.com/TraceMachina/nativelink/compare/v1.6.4..v1.6.5) - 2026-08-18
+
+### ⛰️  Features
+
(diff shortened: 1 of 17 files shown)


MarcusSorealheis (approved):
LGTM

PR #2701, opened 2026-08-19
docs: generate the metrics reference and bring the config reference to v1.6.5

amankrx:
## What and why
Adds a generated metrics reference: `gen-metrics-reference.mjs` parses every OpenTelemetry instrument out of `nativelink-util/src/metrics.rs` and scans the workspace for its call sites, so the page states for each of the 30 instruments whether the binary actually emits it, under what condition, and which Prometheus series it becomes. Two small fixes to the existing generators ride along: the changelog page no longer puts a link inside each release heading (Fumadocs wraps headings in their own anchor, and the nested <a> broke hydration), and the config-reference generator stops listing index in the folder's `meta.json` (Fumadocs then uses the index page as the folder link instead of showing it twice) and skips manifest entries for tags that have no page. Also a Shiki alias so starlark code fences highlight.      
  
## How was this verified?
Ran all three generators against this checkout and built the docs app (`bun run build`, 58 pages). Spot-checked the metrics page's "emitted" verdicts against the call sites in the scheduler, worker-API, store-tier, health and connection code; the two "never emitted" instruments (`cache.size`, `cache.entries`) match the known gap in the shipped recording rules. Confirmed the changelog page renders without the hydration error. `schema-to-mdx.mjs` is untouched, and no committed generated page is changed in this PR: the next config-reference run should diff only the `meta.json` line. Vale and the pre-commit hooks pass  over the changed files.

## Risk
Low. New script and new page, plus behaviour changes in two generators that only show up on their next run (one `meta.json` line; a manifest that omits dead versions). The metrics page links to `/operate/observability`, which lands with the content PR; until then that link 404s. `gen:metrics-reference` is not yet wired into the config-reference workflow, so the page can drift from metrics.rs until someone re-runs it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2701)


.github/styles/config/vocabularies/TraceMachina/accept.txt
```diff
@@ -326,3 +324,7 @@ inlined
 protobufs
 arg
 rustfmt
+||||||| parent of a4c90ca9 (docs: generate the metrics reference and bring the config reference to v1.6.5)
```
MarcusSorealheis: Why is the is in the accept. Should just exclude the relevant section Vale right?

palfrey: This looks like it shouldn't be here. Merge issue?

web/apps/docs/scripts/gen-config-reference.mjs
palfrey, line 1: Have things been re-run with these changes?

web/apps/docs/source.config.ts
```diff
@@ -1,7 +1,20 @@
...
   dir: "content/docs",
 });
 
-export default defineConfig();
+export default defineConfig({
+  mdxOptions: {
+    rehypeCodeOptions: {
+      ...rehypeCodeDefaultOptions,
+      // Shiki has no `starlark` grammar. Starlark is a Python dialect, so load
+      // Python and highlight `starlark` blocks (Bazel BUILD/.bzl) with it.
```
palfrey: Where's the `starlark` block?

palfrey (changes requested):

.github/styles/config/vocabularies/TraceMachina/accept.txt
amankrx: Yes. Merge issues when separating out the commits. Fixed it now.

web/apps/docs/scripts/gen-config-reference.mjs
amankrx: This should automatically run on the next workflow dispatch.

web/apps/docs/source.config.ts
amankrx: I have removed it from here, and will update the docs rewrite commit that has the actual starlark block.

web/apps/docs/scripts/gen-config-reference.mjs
palfrey: What I'd like to see is the results of the changes here. When I ran the config generator locally, I saw `index` removed from `web/apps/docs/content/docs/reference/nativelink-config/meta.json` which for example isn't present in this PR, and so makes it harder to review the results. I've tested that locally and it looks ok, but I'd like to see the results changes in this PR as well.

palfrey (approved):

PR #2703, opened 2026-08-20
Collapse the PR template comment once the description is fixed

### LINKED ISSUE #123: Make custom binaries for different services

HackAttack:
## What and why

If a comment ever gets posted in response to an inadequately filled out PR template, it stays in that state even once fixed, unlike e.g. the CLA comment, which updates.

## How was this verified?

Ran the shell code against a stub `gh` to check each combination of comment state and check outcome

## Risk

Breaking the PR template check job; no functional risk.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2703)


amankrx (approved):
Thanks!

PR #2705, opened 2026-08-20
Match undeclared platform properties dynamically instead of rejecting them

amankrx:
## What and why

An action or worker carrying a platform property key the scheduler's `supported_platform_properties` does not declare was rejected outright: workers refused at registration, actions logged `Unknown platform property` and sat queued forever. Undeclared keys now become `Unknown` values that wildcard-match: the action still routes, matching on the declared keys, and an undeclared key on both sides matches by exact string. Declared property types keep their existing semantics untouched.

## How was this verified?

New tests in `platform_property_manager_test.rs`, `worker_capability_index_test.rs` and `platform_properties_tests.rs`: an undeclared key no longer errors (this failed before the change), exact-string matching between undeclared keys on action and worker, missing-key wildcard behaviour through the capability index, and the default property names (`cpu_count`, `memory_kb`, `gpu_count`, `container-image`) parsing under every declared type.

## Risk

Behaviour change: configs relying on the scheduler rejecting a typo'd property key now match instead, so a misspelled `exact` constraint routes more broadly rather than erroring. Declared keys behave exactly as before; the change only affects keys absent from `supported_platform_properties`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2705)


MarcusSorealheis (approved):
Lgtm

MarcusSorealheis:
We have needed this for years

Issue #2672, opened 2026-08-07
[Bug]: Worker pool permanently deadlocks in RunningActionsManagerImpl::cleanup_action after sustained load, requires manual restart


msshapira:
### Version

1.6.4

### Client

Bazel

### OS / Architecture

macOS 15.7.3 (24G419), Apple Silicon (ARM64)

### Deployment method

Nix

### Storage backend

S3

### Number of workers

1

### Configuration

```json5
{
  stores: [
    {
      name: "AC_MAIN_STORE",
      filesystem: {
        content_path: "/tmp/nativelink/data-worker-test/content_path-ac",
        temp_path: "/tmp/nativelink/data-worker-test/tmp_path-ac",
        eviction_policy: {
          // 10gb.
          max_bytes: 10000000000,
        },
      },
    },
    {
      name: "WORKER_FAST_SLOW_STORE",
      fast_slow: {
        // "fast" must be a "filesystem" store because the worker uses it to make
        // hardlinks on disk to a directory where the jobs are running.
        fast: {
          filesystem: {
            content_path: "/tmp/nativelink/data-worker-test/content_path-cas",
            temp_path: "/tmp/nativelink/data-worker-test/tmp_path-cas",
            eviction_policy: {
              // 100gb.
              max_bytes: 100000000000,
            },
          },
        },
        slow: {
          /// Discard data.
          /// This example usage has the CAS and the Worker live in the same place,
          /// so they share the same underlying CAS. Since workers require a fast_slow
          /// store, we use the fast store as our primary data store, and the slow store
          /// is just a noop, since there's no shared storage in this config.
          noop: {},
        },
      },
    },
  ],
  schedulers: [
    {
      name: "MAIN_SCHEDULER",
      simple: {
        supported_platform_properties: {
          cpu_count: "minimum",
          memory_kb: "minimum",
          network_kbps: "minimum",
          disk_read_iops: "minimum",
          disk_read_bps: "minimum",
          disk_write_iops: "minimum",
          disk_write_bps: "minimum",
          shm_size: "minimum",
          gpu_count: "minimum",
          gpu_model: "exact",
          cpu_vendor: "exact",
          cpu_arch: "exact",
          cpu_model: "exact",
          kernel_version: "exact",
          OSFamily: "priority",
          "container-image": "priority",
          "lre-rs": "priority",
          ISA: "exact",
          InputRootAbsolutePath: "ignore", // used by chromium builds, but we can drop it
        },
      },
    },
[...]

PR #2706, opened 2026-08-21
Never notify the action-done channel while holding the running-actions mutex

amankrx:
## What and why

`cleanup_action` notified the action-done watch channel while holding the running-actions mutex, and `kill_all` took that mutex inside a `wait_for` closure, which runs under the channel's internal lock. On a disconnect-triggered `kill_all` racing the killed actions' cleanups, the two parked each other's runtime threads forever (#2672). The worker kept keepalives up while doing nothing. The mutex is now released before any notify, and the drain wait checks the map outside the channel's lock.

## How was this verified?

New multi-thread regression test racing concurrent cleanups against `kill_all`: it deadlocks on the previous code and passes in under a second with the fix. All existing worker tests pass, including the one pinning kill_all's wait-for-drain semantics.

## Risk

Low: same semantics, narrower critical sections; the only behavioral surface is the drain wait, covered by existing tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2706)

### DIFF
--- a/nativelink-worker/src/running_actions_manager.rs
+++ b/nativelink-worker/src/running_actions_manager.rs
@@ -3331,10 +3331,20 @@ impl RunningActionsManagerImpl {
     }
 
     fn cleanup_action(&self, operation_id: &OperationId) -> Result<(), Error> {
-        let mut running_actions = self.running_actions.lock();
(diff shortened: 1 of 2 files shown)


MarcusSorealheis (approved):
Simple

PR #2707, opened 2026-08-21
Evict a worker that never acknowledges a kill

amankrx:
## What and why
Adds bounded recovery for workers that receive a kill request but never acknowledge it, preventing them from remaining stuck indefinitely.

## How was this verified?
Added a test for a live worker that never acknowledges a kill. All 33 scheduler, config, and worker API server tests pass, with clippy clean.

## Risk
Low. The change only affects workers with unacknowledged kill requests and uses a configurable timeout with a 60s default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2707)


MarcusSorealheis (approved):
Lgtm

PR #2712, opened 2026-08-23
Label execution resource samples with the action mnemonic

amankrx:
## What and why

Adds `execution.action_mnemonic` to `execution.peak.memory` and `execution.cpu.time`.  

## How was this verified?

Tested this locally with the unit tests and verified the results in a live environment.

## Risk

Low risk. Adds a metric so no change to the build performance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2712)

### DIFF
--- a/nativelink-util/tests/metrics_test.rs
+++ b/nativelink-util/tests/metrics_test.rs
@@ -348,8 +348,8 @@ fn test_new_metric_helpers_are_callable() {
     record_store_tier_read("slow", "hit");
     record_store_tier_io("fast", "read", 4096);
     record_health_check("store", "ok");
-    record_execution_peak_memory(512 * 1024, "main");
-    record_execution_cpu_time(90_000, "main");
+    record_execution_peak_memory(512 * 1024, "main", "CppCompile");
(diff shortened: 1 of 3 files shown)


nativelink-util/tests/metrics_test.rs
```diff
@@ -348,8 +348,8 @@ fn test_new_metric_helpers_are_callable() {
     record_store_tier_read("slow", "hit");
     record_store_tier_io("fast", "read", 4096);
     record_health_check("store", "ok");
-    record_execution_peak_memory(512 * 1024, "main");
-    record_execution_cpu_time(90_000, "main");
+    record_execution_peak_memory(512 * 1024, "main", "CppCompile");
```
corcillo, line 351: shouldnt you add some kind of test to see what the results of including are?

corcillo (approved):

nativelink-util/tests/metrics_test.rs
amankrx: Not really useful in this case, since we just added an extra parameter.

PR #2718, opened 2026-08-26
Upgrade to Rust 1.97.1

palfrey:
## What and why

Upgrades the flakes, to allow the following:
* Now that https://github.com/awslabs/aws-sdk-rust/issues/1451 is resolved, we can fix https://rustsec.org/advisories/RUSTSEC-2026-0253, but the newer AWS packages need a newer Rust, which this enables
* https://github.com/taiki-e/cargo-llvm-cov/pull/500, which is the reason we had a custom cargo-llvm-cov [got released as part of v0.9.0](https://github.com/taiki-e/cargo-llvm-cov/releases/tag/v0.9.0) which is now part of [nix unstable](https://github.com/NixOS/nixpkgs/pull/553435) so we can drop our custom version

There are a _lot_ of lint changes, and a number of errors that have new messages as well. I've made the minimum set of changes to be compatible with the new stable and nightly rust versions.

## How was this verified?

`bazel test //...` and `cargo build --tests`

## Risk

Medium. If it passes CI, it should be safe. The Rust version changes should be safe, but the AWS packages _could_ potentially have behaviour changes that we don't test for fully. `cargo-llvm-cov` is fine provided the Coverage tests are fine.

Biggest item is that this drops `x86_64-darwin` because [Nix has dropped it](https://nixos.org/manual/nixpkgs/unstable/release-notes#x86_64-darwin-26.05)

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2718)


.bazelrc
```diff
@@ -145,7 +145,7 @@ build --@rules_rust//:clippy_flag=-Dclippy::semicolon_if_nothing_returned
 build --@rules_rust//:clippy_flag=-Dclippy::similar_names
 build --@rules_rust//:clippy_flag=-Dclippy::single_char_pattern
 build --@rules_rust//:clippy_flag=-Dclippy::single_match_else
-build --@rules_rust//:clippy_flag=-Dclippy::std_instead_of_core
+build --@rules_rust//:clippy_flag=-Aclippy::std_instead_of_core
```
palfrey, line 148: See https://github.com/rust-lang/rust-clippy/pull/16964 ref in `Cargo.toml`. This will get fixed in 1.98

nativelink-store/Cargo.toml
```diff
@@ -137,10 +139,6 @@ libc = { version = "0.2.177", default-features = false }
 [dev-dependencies]
 nativelink-macro = { path = "../nativelink-macro" }
 
-aws-sdk-s3 = { version = "1.82.0", features = [
-  "http-1x",
-  "rt-tokio",
-], default-features = false }
```
palfrey, line 143: This was duplicating the non-dev requirement

nativelink-worker/tests/running_actions_manager_test.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
+
```
palfrey, line 2: Because otherwise there's a failure with one of the `nativelink_test` attributes

nativelink-util/tests/telemetry_test.rs
```diff
@@ -39,16 +39,21 @@ where
 {
     type Rejection = (StatusCode, &'static str);
 
-    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+    fn from_request_parts(
```
palfrey, line 42: `unused_async_trait_impl` (new in nightly) doesn't like this, so fixed

nativelink-config/Cargo.toml
```diff
@@ -32,7 +32,7 @@ tracing = { version = "0.1.41", default-features = false }
 pretty_assertions = { version = "1.4.1", features = [
   "std",
 ], default-features = false }
-serde_test = { version = "1.0", default-features = false }
+serde_assert = { version = "0.8", default-features = false }
```
palfrey, line 35: `serde_test` isn't maintained anymore and didn't have enough flexibility to check multiple error strings.

flake.nix
```diff
@@ -480,11 +480,11 @@
         };
         lre = {
           Env = with pkgs.lre;
-            if pkgs.stdenv.isDarwin
+            if pkgs.stdenv.hostPlatform.isDarwin
```
palfrey, line 483: The `stdenv.is*` aliases got deprecated in https://github.com/NixOS/nixpkgs/pull/518407

MarcusSorealheis (approved):
Lgtm

PR #2722, opened 2026-08-28
web: point Media kit Download at the shared Drive folder

modernmedici:
The company CTA was an incomplete folders/ URL and 404ed. Restore the folder id from the previous company page so Download opens the kit.

## What and why

URL was broken for the media kit. We need to bring back the pre-existing folder id so that viewers and agents can access our Media kit in the future without needing us to go retrieve them.

## How was this verified?

Tested at different viewports in cursor cloud agent

## Risk

It was already broken so this is an upgrade.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2722)

### DIFF
--- a/web/apps/web/app/company/page.tsx
+++ b/web/apps/web/app/company/page.tsx
@@ -35,7 +35,10 @@ const contactCards = [
   {
     title: "Media kit",
     body: "Logos, branding guidelines, and approved press resources.",
-    cta: { label: "Download", href: "https://drive.google.com/drive/folders/" },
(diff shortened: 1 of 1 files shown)


MarcusSorealheis (changes requested):
Not enough characters in the `What` of the PR Description. You need to add more details so that the artifact is searchable.

modernmedici:
the CLA block is now passing

MarcusSorealheis (approved):

PR #2723, opened 2026-09-02
docs: track Menlo case-study figures under public/assets

modernmedici:
## What and why

Updated to allow the images to render on the page. Moved the wordmark and figure paths off /case-studies, stop ignoring JPEGs on the marketing site, and allow these two figures through forbid-binary-files so they clone and render.

## How was this verified?

Tested locally and with the Netlify deploys, inspected visually.

## Risk

Allow these two figures through forbid-binary-files so they clone and render.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2723)


MarcusSorealheis (approved):
Thanks.

PR #2724, opened 2026-09-02
Upgrade rules_rs to 0.0.108 to fix zlib issues

palfrey:
## What and why

Upgrades rules_rs to 0.0.108. This is because earlier versions had a hardcoded zlib package URL that didn't work anymore. See also https://github.com/hermeticbuild/rules_rs/commit/81336041f1c092b94ba0d35384c4dfc03ce45a07 and https://github.com/TraceMachina/nativelink/actions/runs/33634963280/job/100266751805. I've also bumped rules_cc and llvm versions as they were complaining about different versions getting pulled in, presumably from the rules_rs bump.

## How was this verified?

`bazel test //...`. The changes are only Bazel rule changes, so if bazel builds, it's fine :)

## Risk

Low: rules_rs version upgrades are fairly safe assuming they still build fine which this one does.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2724)


toolchain-examples/MODULE.bazel
palfrey, line 1: Ran into a whole stack of issues trying to update the lockfile from here (`bazel mod deps --lockfile_mode=update`) and so there are various changes as a result of that.

toolchain-examples/MODULE.bazel
```diff
@@ -7,39 +7,30 @@ module(
...
 # Python
 bazel_dep(name = "rules_python", version = "2.0.0")
 
-pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
-pip.parse(
-    hub_name = "pip",
-    python_version = "3.12",
-    requirements_lock = "//:requirements_lock.txt",
-)
-
```
palfrey, line 27: We don't actually have a requirements file for python at all here, we actually don't have any dependencies :)

toolchain-examples/MODULE.bazel
```diff
@@ -57,7 +48,8 @@ rust_toolchains.toolchain(
 use_repo(rust_toolchains, "default_rust_toolchains")
 
 register_toolchains(
-    "@default_rust_toolchains//:all",
+    "@default_rust_toolchains//...",
+    "@llvm//toolchain:all",
```
palfrey, line 56: As per the `rules_rs` README. Also, because the old version didn't work.

toolchain-examples/MODULE.bazel
MarcusSorealheis: oof! I forgot why this was added. Oh, maybe for customers.

zbarsky-openai (commented):
Hey, apologies for grabbing that dep from the ubuntu pool and not the stable snapshots. Rookie mistake :/

As an aside, I think this rules_rs version bump should get you the new hermiticity checker (we now fail rlib compilation if you accidentally embed the execroot path) so hopefully it's not too bumpy!

MarcusSorealheis (approved):

MarcusSorealheis (approved):
Lgtm

MarcusSorealheis:
@zbarsky-openai thank you for all the stewardship here. Big boon to the ecosystem. Mistakes happen.

PR #2730, opened 2026-09-04
Bound the wait for a GCS connection permit

benelser:
## What and why

`GcsClient` serializes every request behind a `tokio::sync::Semaphore`
sized by `multipart_max_concurrent_uploads` (default 10). The permit is
acquired FIFO with no bound, so under a lookup storm the queue on one
pod grows past `cas_server`'s 30 s per-blob deadline and the caller sees
a bare `DeadlineExceeded` that says nothing about the cause. Measured on
a v1.6.3 deployment with a 256-permit client: 69 `GetActionResult` and
4 `BatchReadBlobs` at >= 29 s per storm minute, p99 30.0 s. Raising the
permit count to 10000 did not help; it turned the FIFO wait into
reqwest 3 s read timeouts and h2 `too_many_internal_resets` connection
kills. The queue needs a bound, not more depth.

This adds `permit_wait_timeout_s` to `ExperimentalGcsSpec`. The default
`0` keeps the current unbounded wait, so existing configs behave exactly
as before. When set, the acquire is wrapped in `tokio::time::timeout`
and expiry returns `FailedPrecondition` with a message naming the bound,
the permit count and the queue depth. Both `with_connection` (metadata,
uploads, delete) and the streaming `read_object_content` path go through
the same bounded acquire. The `Debug` output of `GcsClient`, which
printed the free permit count under the label `max_connections`, now
shows the limit, the free count and the configured bound.

`FailedPrecondition` is deliberate: `Retrier::should_retry` and the
retry loop in `upload_from_reader` both retry `ResourceExhausted` and
`Unavailable`, and retrying a permit expiry only re-queues the request
at the tail of the same saturated semaphore. `FailedPrecondition` is in
the retrier's never-retry set, so the caller gets a classified error on
the first expiry. Operators who want retries can still list it in
`retry.retry_on_errors`.

## How was this verified?

Four tests in `nativelink-store/tests/gcs_client_test.rs` use a real
`GcsClient` with one permit held by a request a local TCP server never
answers: the bounded wait fails within the bound with
`FailedPrecondition` and the documented message; the streaming read
path is bounded the same way; a permit freed inside the bound lets the
waiting request reach the server; `0` still queues indefinitely.
Without the client change the first test fails after 60 s with
`NotFound` (the waiter only proceeded once the hung request's read
timeout released the permit). Deployed with a 5 s bound alongside two
related changes, the same storm produced 0 requests at >= 29 s and a
`GetActionResult` p99 of 0.85 s.

## Risk

Default off, no request behavior changes unless `permit_wait_timeout_s`
is set. When set too low for a healthy but busy client, requests fail
with `FailedPrecondition` instead of waiting; the error message names
the permit count and queue depth so the misconfiguration is visible. The
extra `AtomicUsize` is two relaxed operations per bounded acquire.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2730)

### DIFF
--- a/nativelink-config/src/stores.rs
+++ b/nativelink-config/src/stores.rs
@@ -1281,6 +1281,14 @@ pub struct ExperimentalGcsSpec {
     /// Default: 3000
     #[serde(default, deserialize_with = "convert_duration_with_shellexpand")]
     pub read_timeout_s: u64,
+
+    /// Maximum time in seconds a request waits for one of the
+    /// `multipart_max_concurrent_uploads` connection permits before it fails
+    /// with `FailedPrecondition`. A value of `0` waits until a permit is free.
+    ///
+    /// Default: 0 (unbounded)
+    #[serde(default, deserialize_with = "convert_duration_with_shellexpand")]
+    pub permit_wait_timeout_s: u64,
 }
 
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
--- a/nativelink-store/src/gcs_client/client.rs
+++ b/nativelink-store/src/gcs_client/client.rs
@@ -14,6 +14,7 @@
 
 use core::fmt::{Debug, Formatter};
 use core::future::Future;
+use core::sync::atomic::{AtomicUsize, Ordering};
 use core::time::Duration;
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -33,7 +34,7 @@ use nativelink_error::{Code, Error, ResultExt, make_err};
 use nativelink_util::buf_channel::DropCloserReadHalf;
 use rand::Rng;
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
-use tokio::time::sleep;
+use tokio::time::{sleep, timeout};
 
 use crate::common_s3_utils::install_default_rustls_crypto_provider;
 use crate::gcs_client::types::{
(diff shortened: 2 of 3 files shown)


MarcusSorealheis:
Hi @benelser this is an interesting PR. Thank you for the pull request. We are going to test this on our test workload and then we will be able to merge it. At first read, it looks good. 

One other issue I need to look into is the behavior of the other 5 stores we support.

MarcusSorealheis:
I misspoke. There are 4 other object stores. We are designing a 5th. We need to ensure we can support those.

benelser:
Thanks Marcus. On the other stores, from reading main today: the GCS client is the only store that routes every request, including metadata and reads, through one shared permit pool, so it is the only place where a queue of small requests can outlast the CAS server's per-blob deadline. S3 and ONTAP S3 bound concurrent multipart parts with a channel per upload and do not queue metadata calls. The gRPC batcher uses `try_acquire`, so it never waits. Redis (`client_permits`) and Mongo (`request_permits`) do wait unbounded on their semaphores, the same shape as GCS before this change, but neither is on a per-request path with a 3-request write like the GCS resumable upload was for us.

If a bound on the Redis or Mongo permit wait would be useful, I can send that as a follow-up in the same form. The load recipe we used to reproduce the queueing on our own deployment is a warm-cache buck2 build start against a fast_slow{redis, gcs} CAS; happy to share the numbers or the config if it helps your test workload.

benelser:
@MarcusSorealheis how did the run on your test workload go? This has been in our production CAS since Sep 4, and hourly readings since then show no requests reaching the 30 s per-blob deadline. If a variant would help your test, I can share the config and the warm-cache buck2 recipe from my earlier comment.

PR #1785, opened 2025-05-22
Use native root certs for S3 stores

jaroeichler:
Switch from `webpki-roots` to `rustls-platform-verifier` for TLS certificate verification.

Similarly to #1782, it makes sense to use platform certificates. This makes it easier for users to update certificates.
See also https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/.

I've tested this change by running the `bazel` Nix template against a `nativelink-worker-lre-cc` running in Docker, configured with `s3_backend_with_local_fast_cas.json5` and an AWS S3 instance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/1785)

Issue #441, opened 2023-12-03
We unintentionally depend on existing CAs during testtime and runtime


aaronmondal:
We're using hyper-rustls with the `webpki-tokio` feature enabled. The intention behind this is that we vendor the CA via webpki and make this explicitly extensible with custom CAs in the future.

Running the testsuite in an airgapped environment without a host CA shows that a bunch of tests fail (log is incomplete, these tests interrupt on error).

<details>
  <summary>Logs</summary>

```bash
native-link-nextest>     Finished release [optimized] target(s) in 6m 28s
native-link-nextest>     Starting 237 tests across 38 binaries7:
native-link-nextest>         PASS [   0.008s] native-link-scheduler::action_messages_test action_messages_tests::action_state_any_url_test
native-link-nextest>         PASS [   0.008s] native-link-scheduler::action_messages_test action_messages_tests::equal_priority_earliest_first
native-link-nextest>         PASS [   0.005s] native-link-scheduler::property_modifier_scheduler_test property_modifier_scheduler_tests::add_action_overwrites_property
native-link-nextest>         PASS [   0.006s] native-link-scheduler::property_modifier_scheduler_test property_modifier_scheduler_tests::add_action_adds_property
native-link-nextest>         PASS [   0.007s] native-link-scheduler::cache_lookup_scheduler_test cache_lookup_scheduler_tests::add_action_handles_skip_cache
native-link-nextest>         PASS [   0.006s] native-link-scheduler::cache_lookup_scheduler_test cache_lookup_scheduler_tests::platform_property_manager_call_passed
native-link-nextest>         PASS [   0.008s] native-link-scheduler::action_messages_test action_messages_tests::highest_priority_action_first
native-link-nextest>         PASS [   0.006s] native-link-scheduler::cache_lookup_scheduler_test cache_lookup_scheduler_tests::find_existing_action_call_passed
native-link-nextest>         PASS [   0.007s] native-link-scheduler::cache_lookup_scheduler_test cache_lookup_scheduler_tests::add_action_validates_outputs
native-link-nextest>         PASS [   0.007s] native-link-scheduler::cache_lookup_scheduler_test cache_lookup_scheduler_tests::add_action_does_cache_lookup
native-link-nextest>         PASS [   0.008s] native-link-scheduler::action_messages_test action_messages_tests::execute_response_status_message_is_some_on_success_test
[...]

PR #2748, opened 2026-09-08
Return an error when TlsClient cannot load CA root certificates

### LINKED ISSUE #2747: [Bug]: S3/GCS/R2 stores fail with a panic when no system CA store is present (official image ships none) instead of returning a config error
### Version

v1.6.7 (`ghcr.io/tracemachina/nativelink:v1.6.7`) and `main` at 76392b5. `Cargo.lock`: hyper-rustls 0.27.7, rustls-platform-verifier 0.6.2, rustls-native-certs 0.8.2.

### Client

Buck2 (not client-specific; happens before any client connects)

### OS / Architecture

Linux x86_64 (container)

### Deployment method

Docker / Kubernetes (stock `ghcr.io/tracemachina/nativelink` image)

### Storage backend

S3 (also applies to `gcs` and `r2`, which share `TlsClient`)

### Configuration

```json5
{
  stores: [
    {
      name: "CAS_S3",
      experimental_cloud_object_store: {
        provider: "aws",
        region: "us-east-1",
        bucket: "some-bucket",
        key_prefix: "cas/",
      },
    },
  ],
  // ... servers using CAS_S3
}
```

### Expected behavior

If no CA roots can be loaded, `S3Store::new` (and the GCS/R2/OCI constructors) should return a `nativelink_error::Error` that names the store and the remedy (mount a CA bundle and point `SSL_CERT_FILE`/`SSL_CERT_DIR` at it), the same way the ONTAP store already does via `with_native_roots()?`. Ideally `CommonObjectSpec` should also accept a `root_certificates` path like `OntapS3Spec` does, so the bundle can be configured in-file rather than only through an undocumented environment variable.

### Actual behavior

The process aborts with a panic from a third-party `expect` during store construction, before serving anything.

`TlsClient::new_with_http_support` builds the connector with the panicking `HttpsConnectorBuilder::new().with_platform_verifier()`:

https://github.com/TraceMachina/nativelink/blob/76392b556171292fc300cc262614d427b7ab6dc1/nativelink-store/src/common_s3_utils.rs#L75-L79

In hyper-rustls 0.27.7 that method is `self.try_with_platform_verifier().expect("failure to initialize platform verifier")` (`src/connector/builder.rs#L76-L79`). On Linux `rustls-platform-verifier` loads roots with `rustls-native-certs` and returns `Err(rustls::Error::General("No CA certificates were loaded from the system"))` when the resulting store is empty (`rustls-platform-verifier/src/verification/others.rs#L94-L110`), so the `expect` fires.

The official image is built by `nativelinkImageFor` in `flake.nix` with `copyToRoot` containing only the nativelink `/bin` (`pkgs.cacert` appears only as a build-time `buildInputs` entry, not in the image):

https://github.com/TraceMachina/nativelink/blob/76392b556171292fc300cc262614d427b7ab6dc1/flake.nix#L272-L296

So any `experimental_cloud_object_store` with `provider: "aws" | "gcs" | "r2"` panics on first start in the stock image unless the operator already knows to mount a CA bundle and set `SSL_CERT_FILE`. The same `TlsClient::new` is used by `s3_store.rs#L112-L113`, `r2_store.rs#L49`, `oci_store.rs#L58` and `ontap_s3_existence_cache_store.rs#L418`, all of which are already inside `Result`-returning constructors.

Why this is a defect rather than "bring your own certs":

- #708 / #810 settled on not bundling CAs in the image, which is fine. But the failure mode should be a config error, not a panic. The ONTAP path already gets this right: `ontap_s3_store.rs#L133-L139` uses `ClientConfig::builder().with_native_roots()?` and propagates the error.
- `CommonObjectSpec` (`nativelink-config/src/stores.rs#L1322`) has no `root_certificates` field, while `OntapS3Spec` has one (`stores.rs#L861`). For `aws`/`gcs`/`r2` the only knob is the `SSL_CERT_FILE`/`SSL_CERT_DIR` env var honoured by `rustls-native-certs`, which is not mentioned anywhere in the repo (no hits for `SSL_CERT_FILE` in code, docs, or `kubernetes/`).
- `web/apps/docs/content/docs/how-to/stores/s3-and-compatible.mdx#L123-L125` says "Without it, the system roots are used" (in the ONTAP section) without noting that the published image has no system roots, and the `aws`/`gcs`/`r2` sections do not mention trust roots at all.

### Logs / error output

Expected form of the panic (message from hyper-rustls `expect`, error is Debug-formatted):

```shell
thread '...' panicked at .../hyper-rustls-0.27.7/src/connector/builder.rs:...:
failure to initialize platform verifier: General("No CA certificates were loaded from the system")
```

Reproduce with the stock image and any config containing an `aws` store (credentials via env; the panic happens before credentials are used):

```shell
docker run --rm -v "$PWD/cas.json5:/cfg.json5" ghcr.io/tracemachina/nativelink:v1.6.7 /cfg.json5
```

Mounting a PEM bundle into the container and setting `SSL_CERT_FILE=/path/to/ca-certificates.crt` makes it start.

### Additional context

Proposed fix (small):

1. `nativelink-store/src/common_s3_utils.rs`: replace `with_platform_verifier()` with `try_with_platform_verifier()` and make `TlsClient::new` / `new_for_credentials` / `new_with_http_support` return `Result<Self, Error>` (`Code::InvalidArgument`, message along the lines of "no CA certificates could be loaded for the TLS client; mount a CA bundle and set SSL_CERT_FILE, or set root_certificates"). Update the five call sites above plus `nativelink-store/tests/common_s3_utils_test.rs#L49,L65` (all already in `Result` contexts). Roughly <80 lines; unit-testable by constructing `TlsClient` with `SSL_CERT_FILE` pointing at an empty file and `SSL_CERT_DIR` at an empty directory.
2. Optionally add `root_certificates: Option<String>` to `CommonObjectSpec`, mirroring `OntapS3Spec`, and build the connector with `with_tls_config` + `with_root_certificates` when set (the ONTAP store already has `load_custom_certs` to reuse).
3. Docs: one sentence in `s3-and-compatible.mdx` and the container/Kubernetes deploy pages stating that the image carries no CA store and showing the `SSL_CERT_FILE` mount.

Related: #441 (open; "we unintentionally depend on existing CAs during testtime and runtime" — this issue is the concrete user-facing failure mode for the S3-family stores), #708 / #810 (decision not to bundle CAs), #1785 (switched S3 stores to the platform verifier).

Found while operating this in production; happy to send a PR if maintainers agree with the analysis.

KR-Ravindra:
## What and why

**Problem**

With the published `ghcr.io/tracemachina/nativelink` image (which carries no CA store) any `experimental_cloud_object_store` with `provider: "aws"` or `"r2"`, as well as the OCI store and the ONTAP S3 existence cache, aborts the whole process during store construction instead of reporting a configuration error:

```
thread '...' panicked at .../hyper-rustls-0.27.7/src/connector/builder.rs:78:14:
failure to initialize platform verifier: General("No CA certificates were loaded from the system")
```

Fixes #2747

**Root cause**

`nativelink-store/src/common_s3_utils.rs:78` built the connector with `HttpsConnectorBuilder::new().with_platform_verifier()`. In hyper-rustls 0.27.7 that method is `try_with_platform_verifier().expect("failure to initialize platform verifier")`, and `rustls-platform-verifier` returns `Err` on Linux when the root store it loads via `rustls-native-certs` ends up empty. `TlsClient::new`, `new_for_credentials` and `new_with_http_support` returned `Self`, so there was no way to surface that error even though every caller (`s3_store.rs:112-113`, `r2_store.rs:49`, `oci_store.rs:58`, `ontap_s3_existence_cache_store.rs:418`) already sits inside a `Result`-returning constructor. The ONTAP store itself already handles this correctly through `with_native_roots()?`.

**Fix**

- `TlsClient::new`, `TlsClient::new_for_credentials` and `TlsClient::new_with_http_support` now return `Result<Self, Error>` and use `try_with_platform_verifier()`, mapping the failure to `Code::InvalidArgument` with a message that names the remedy (mount a CA bundle and point `SSL_CERT_FILE` or `SSL_CERT_DIR` at it).
- The four production call sites propagate the error with `?`; the two existing `TlsClient` tests `unwrap()` the constructor.
- New test binary `nativelink-store/tests/tls_client_no_roots_test.rs` (registered in `BUILD.bazel`), gated to Linux because only the Linux platform verifier reads `SSL_CERT_FILE`/`SSL_CERT_DIR`. It lives in its own binary because it sets process-wide environment variables that the other `TlsClient` tests would otherwise observe.

Scope note: the GCS store builds its HTTP client through `reqwest` rather than `TlsClient`, so it is not on this panic path and is unchanged here. Adding a `root_certificates` option to `CommonObjectSpec` and documenting the CA requirement (items 2 and 3 in the issue) are left for follow-ups so this stays a small behavioural fix.

## How was this verified?

**How tested**

Before the fix, the new test body (with the assertion reduced to just constructing the client, since the old return type was not a `Result`) reproduces the report:

```
$ cargo test -p nativelink-store --test tls_client_no_roots_test
test tls_client_without_ca_roots_does_not_panic ... FAILED

thread 'tls_client_without_ca_roots_does_not_panic' panicked at .../hyper-rustls-0.27.7/src/connector/builder.rs:78:14:
failure to initialize platform verifier: General("No CA certificates were loaded from the system")
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

After the fix, on the same machine (Linux x86_64, Rust 1.97.1, `SSL_CERT_FILE` pointed at an empty file and `SSL_CERT_DIR` at an empty directory inside the test):

```
$ cargo test -p nativelink-store --test tls_client_no_roots_test
test tls_client_without_ca_roots_returns_error ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test -p nativelink-store --test common_s3_utils_test
test credential_client_allows_http ... ok
test s3_client_rejects_http_by_default ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

`cargo fmt --all -- --check` and `cargo clippy -p nativelink-store --tests` (workspace lints: `all`/`pedantic`/`nursery` denied) are clean. The Bazel target was added to the `integration` `rust_test_suite` but I did not run the Bazel build locally; CI will cover it.

## Risk

Low. The only behavioural change is on the failure path: a configuration that used to abort with a panic now fails store construction with an `InvalidArgument` error carrying an actionable message. Deployments that already have a CA store see no difference. The public signatures of `TlsClient::new` and `TlsClient::new_for_credentials` change from `Self` to `Result<Self, Error>`; all in-tree callers are updated.

## Links

- Issue: https://github.com/TraceMachina/nativelink/issues/2747
- Related: https://github.com/TraceMachina/nativelink/issues/441 (test-time/runtime dependency on system CAs), https://github.com/TraceMachina/nativelink/pull/1785 (switch to the platform verifier)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2748)

### DIFF
--- a/nativelink-store/BUILD.bazel
+++ b/nativelink-store/BUILD.bazel
@@ -150,6 +150,7 @@ rust_test_suite(
         "tests/shard_store_test.rs",
         "tests/size_partitioning_store_test.rs",
         "tests/store_manager_test.rs",
+        "tests/tls_client_no_roots_test.rs",
         "tests/verify_store_test.rs",
     ],
     compile_data = [
--- a/nativelink-store/src/common_s3_utils.rs
+++ b/nativelink-store/src/common_s3_utils.rs
@@ -41,7 +41,7 @@ use nativelink_config::stores::CommonObjectSpec;
 // when in a retryable wrapper. Always prefer Code::Aborted or another
 // retryable code over Code::InvalidArgument or make_input_err!().
 // ie: Don't import make_input_err!() to help prevent this.
-use nativelink_error::{Code, make_err};
+use nativelink_error::{Code, Error, make_err};
 use nativelink_util::buf_channel::DropCloserReadHalf;
 use nativelink_util::fs;
 use nativelink_util::retry::{Retrier, RetryResult};
@@ -58,8 +58,7 @@ pub struct TlsClient {
 }
 
 impl TlsClient {
-    #[must_use]
-    pub fn new(common: &CommonObjectSpec) -> Self {
+    pub fn new(common: &CommonObjectSpec) -> Result<Self, Error> {
         Self::new_with_http_support(common, common.insecure_allow_http)
     }
 
@@ -67,15 +66,23 @@ impl TlsClient {
     ///
     /// The default credential chain uses plain HTTP for link-local metadata
     /// services, including the EKS Pod Identity, ECS, and EC2 providers.
-    #[must_use]
-    pub fn new_for_credentials(common: &CommonObjectSpec) -> Self {
+    pub fn new_for_credentials(common: &CommonObjectSpec) -> Result<Self, Error> {
         Self::new_with_http_support(common, true)
     }
 
-    fn new_with_http_support(common: &CommonObjectSpec, allow_http: bool) -> Self {
+    fn new_with_http_support(common: &CommonObjectSpec, allow_http: bool) -> Result<Self, Error> {
         install_default_rustls_crypto_provider();
 
-        let connector_with_roots = HttpsConnectorBuilder::new().with_platform_verifier();
+        let connector_with_roots = HttpsConnectorBuilder::new()
+            .try_with_platform_verifier()
+            .map_err(|e| {
+                make_err!(
+                    Code::InvalidArgument,
+                    "Failed to load CA root certificates for the TLS client: {e}. \
+                     Mount a CA bundle into the container and point SSL_CERT_FILE \
+                     or SSL_CERT_DIR at it."
+                )
+            })?;
 
         let connector_with_schemes = if allow_http {
             connector_with_roots.https_or_http()
@@ -89,7 +96,7 @@ impl TlsClient {
             connector_with_schemes.enable_http1().enable_http2().build()
         };
 
-        Self::with_https_connector(common, connector)
+        Ok(Self::with_https_connector(common, connector))
     }
 
     pub fn with_https_connector(
--- a/nativelink-store/src/oci_store.rs
+++ b/nativelink-store/src/oci_store.rs
@@ -55,7 +55,7 @@ impl OciStore {
         I: InstantWrapper,
         NowFn: Fn() -> I + Send + Sync + Unpin + 'static,
     {
-        Self::new_with_http_client(spec, TlsClient::new(&spec.common.clone()), now_fn).await
+        Self::new_with_http_client(spec, TlsClient::new(&spec.common.clone())?, now_fn).await
     }
 
     /// Builds the store with a caller-supplied HTTP client. Production uses
--- a/nativelink-store/src/ontap_s3_existence_cache_store.rs
+++ b/nativelink-store/src/ontap_s3_existence_cache_store.rs
@@ -415,7 +415,7 @@ where
 }
 
 async fn create_s3_client(spec: &ExperimentalOntapS3Spec) -> Result<Client, Error> {
-    let http_client = TlsClient::new(&spec.common);
+    let http_client = TlsClient::new(&spec.common)?;
     let credentials_provider = DefaultCredentialsChain::builder()
         .configure(
             ProviderConfig::without_region()
--- a/nativelink-store/src/r2_store.rs
+++ b/nativelink-store/src/r2_store.rs
@@ -46,7 +46,7 @@ impl R2Store {
         let aws_spec = Self::build_aws_spec(spec);
         let jitter_fn = spec.common.retry.make_jitter_fn();
 
-        let http_client = TlsClient::new(&spec.common.clone());
+        let http_client = TlsClient::new(&spec.common.clone())?;
         let endpoint = Self::derive_endpoint(spec);
 
         let mut config_loader = aws_config::defaults(BehaviorVersion::latest())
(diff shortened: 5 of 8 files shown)


KR-Ravindra:
Round 1 self-review.

Checked against `main`: the three `TlsClient` constructors returned `Self` and used `with_platform_verifier()`, which in hyper-rustls 0.27.7 is `try_with_platform_verifier().expect(..)`, and all four production callers already return `Result`, so propagating with `?` is the minimal change. Cargo Dev, both Bazel jobs, pre-commit and the PR template check are green; the new test binary is in the `rust_test_suite` whose deps already include `tempfile`. Marking ready for review. I still need to sign the CLA.

palfrey (approved):
Nice fix, thanks!

PR #2750, opened 2026-09-08
worker: keep directory construction retries on the same lock

### LINKED ISSUE #123: Make custom binaries for different services

dumbmoron:
## What and why

When several build actions need the same input directory, the worker uses a shared lock so only one request prepares the cached directory at a time. If that request fails, the worker removes the lock even though another request may still be waiting for it; a new request can then get a separate lock, allowing two requests to prepare the same directory at once. This fix keeps requests sharing the same lock until the last one finishes.

## How was this verified?

The regression test sets up the state left by a failed request, with another request still holding the original lock, then asks the cache for the same directory again. Without the fix, the new request goes ahead immediately. With the fix, it waits until the original lock is released, then prepares the directory successfully. The test also checks that the worker removes the lock when it is no longer needed.

## Risk

Low. This changes how the worker coordinates requests for the same cached directory. If the lock is removed too early, those requests could still write to the directory at the same time; if it is kept too long, unused locks could accumulate in memory. The test checks both that requests wait and that the lock is eventually removed.

## AI assistance

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2750)


MarcusSorealheis (approved):
LGTM, thanks again!

PR #2752, opened 2026-09-08
Say stipend instead of salary in the RE API Fellow post

### LINKED ISSUE #123: Make custom binaries for different services

MarcusSorealheis:
## What and why

We need to ensure that we are compliant with all laws and legalistic frameworks. This role will explicitly be a fellowship and not a job. It has a start date and end date.

We changed the word salary to stipend to improve alignment with the position. We kept other verbiage around competitiveness. 

## How was this verified?

It was verified with the standard linters. We also built the site locally.

cd web/platform
bun setup
bun docs
rm -r dist && bun run build
bun preview

## Risk

There is basically no risk. This PR is literally a one word change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2752)


corcillo (approved):
nice

ryaan-md (commented):
> @ryaan-md reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2752)* status: 0 of 1 LGTMs obtained, and all files reviewed.

PR #2753, opened 2026-09-08
Add figures and Biology updates to the RE API Fellow post

modernmedici:
## What and why

Updates the RE API Blog with three key figures to illustrate the point and provide more visual context for readers.. Some people like pictures. 

## How was this verified?

Tested on my local machine for usability and accessibility. It also passed the pre-commit checks in place. We also have run the standard:

cd web/platform
bun setup
bun docs
rm -r dist && bun run build
bun preview

So let's see if this is till an effective way to verify changes.

## Risk

Images may not render correctly. If that's the case, there will be a follow up PR. In some cases the images have failed to render at all. If it happens in this case, I'll look to add a CI check to prevent it. Page load may also slow down. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2753)


MarcusSorealheis (dismissed):

MarcusSorealheis (approved):
LGTM

PR #2757, opened 2026-09-09
Give each action a private /tmp inside the mount namespace

MarcusSorealheis:
## What and why

Actions running with `use_mount_namespace` still shared the host's `/tmp`, so concurrent actions could collide on predictable paths there and state could leak from one action to the next through it. This gives each action a private, empty tmpfs `/tmp` inside its mount namespace, behind a new `isolate_tmp` worker option that defaults on when `use_mount_namespace` is on and `/tmp` exists. If the root action directory lives under `/tmp`, its path is recreated inside the new tmpfs and the action directory is restored from the saved file descriptor, so work directories under `/tmp` keep working. `perform_remount` now reports errors through `pre_exec` instead of panicking in the forked child.

## How was this verified?

- New tests in `namespace_utils_test.rs` check that the host's `/tmp` is invisible to an action and its writes never reach the host, that a root action directory under `/tmp` stays usable (inputs by relative and absolute path, siblings masked, outputs land on the host), that two concurrent actions writing the same `/tmp` path do not see each other, that `isolate_tmp: false` keeps the host `/tmp` visible, and that `namespaces_supported(false, true)` is rejected. Each fails on the old code, where the host `/tmp` is visible.
- New `isolate_tmp_gives_action_a_private_tmp` in `running_actions_manager_test.rs` runs a full action through `RunningActionsManagerImpl` with `YesAndMount { isolate_tmp: true }` and checks the output file is still uploaded from the work directory.
- Ran clippy and the `namespace_utils_test`, `running_actions_manager_test` and `local_worker_test` targets on Linux (aarch64, privileged Docker, both as root and as uid 65534 so user namespace setup matches an unprivileged runner), plus clippy and the `nativelink-config` tests on macOS. All green.
- Not verified: the Windows build and a real Kubernetes deployment.

## Risk

- Changed default on existing deployments: any worker with `use_mount_namespace: true` now gives actions an empty `/tmp`. An action that depended on a tool or file staged under the host's `/tmp` needs `isolate_tmp: false`. Files an action writes to `/tmp` are now memory backed and vanish with the action.
- Workers whose image has no `/tmp` (the flake's release image does not include one) keep the old behaviour with a startup warning rather than failing, so nothing that starts today stops starting.
- Hot path: one extra tmpfs mount and a few `mkdir` calls per action in the `pre_exec` hook, only when `isolate_tmp` is on. Nothing changes for workers without mount namespaces.

https://claude.ai/code/session_01KucPh5sXMeQ7WhdXDEJeTz

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2757)


MarcusSorealheis:
This AI generated PR has problems. I'll convert to draft for now.

PR #2760, opened 2026-09-14
fix: don't panic on a corrupt fetch-store entry in FetchServer

immanuwell:
## What and why

`FetchServer::fetch_blob` panics via `decode(bytes).unwrap()` if the stored entry isn't valid protobuf (corrupt data, or store reused across services). 
Returns an error instead now

## How was this verified?

Added a test writing garbage bytes at the digest, confirmed it panics on main (`InvalidVarint` at fetch_server.rs:101), passes after the fix. 
`cargo test -p nativelink-service --test fetch_server_test`, clippy and fmt all clean

## Risk

Low, failure path only, callers already handle `Result<.., Error>` from this fn


amankrx (approved):
Thank you for the contribution!

PR #2765, opened 2026-09-15
scheduler: fix active counts when a client disappears

### LINKED ISSUE #123: Make custom binaries for different services

dumbmoron:
## What and why

I saw the `executing` metric stay at 10 after the queue emptied and completions stopped. The memory scheduler updates `execution.active.count` when an action changes stages, but its last-client cleanup can remove an action without changing that count. If the client entry expires while the action is still marked `executing`, the metric keeps counting an operation that no longer exists. This subtracts it from its current stage when the final client goes away.

## How was this verified?

The new test moves an action into `executing`, advances the mock clock until its client entry expires, and inserts another action to trigger cleanup. The first operation is gone, but with the original code the recorded count stays at 1 and the test fails. With the fix it returns to 0.

## Risk

Low. This only changes the memory scheduler's stage metric. A wrong or repeated decrement could make a stage count negative, although this cleanup runs only when the final client leaves. The `completed` stage will now shrink as retained operations are removed, so consumers treating it as a lifetime total may see different values.

## AI assistance

Debugged the issue with Grafana MCP + Codex, reviewed the code and refined it myself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2765)


MarcusSorealheis (changes requested):
The fix is right, and it closes the only gap: `add_action` adds one to `queued`, stage changes move that one between stages, and last-client cleanup is the single place an operation leaves the map, so the counter now balances for every stage including `completed`. The test is sound too; `nativelink_test` runs on a current-thread runtime, so the yield loop deterministically lets the spawned event handler run.

One thing before merge: `nativelink-scheduler/BUILD.bazel` lists test sources explicitly in `rust_test_suite`, so the new file needs adding there or Bazel never runs it. The Bazel job passed here without executing it. The suite's deps already include `mock_instant` and `opentelemetry`, so it is just the one `srcs` line.

Also, the PR template gained a required "AI assistance" section today, and the check will ask for it on your next push.

MarcusSorealheis (approved):
thank you for this one. We are load testing now.

PR #2766, opened 2026-09-15
Empower AI systems

### LINKED ISSUE #123: Make custom binaries for different services

MarcusSorealheis:
## What and why

Makes NativeLink readable by AI agents the way it is by people. Every page of nativelink.com and docs.nativelink.com now ends with an "Ask AI" block: ChatGPT, Claude, Perplexity, Google AI Mode and Copilot, each opened with a prompt that points the assistant at our llms files first, plus links to `llms.txt`, `llms-small.txt` and `llms-full.txt`.

## How was this verified?

- `bun typecheck`, Biome on every changed file, and `bun run build` for both apps. Served the builds with `next start`: `/agents`, `/llms.txt`, `/llms-small.txt`, `/llms-full.txt` and the docs `llms-small.txt` all return 200, and the Ask AI markup is present in the server-rendered HTML of the home page, `/agents` and a docs page.
- Docs gates: `lint:anchors`, `lint:links`, `lint:snippets`, `lint:boundaries` and `lint:navigation` pass, and `gen:llms` leaves no diff under `apps/docs`.

## Risk

Low for the product. Nothing in the binary changes. On the site: the marketing build now runs `bun scripts/gen-llms.ts` before `next build` and reads `web/apps/docs/content` from the workspace, so a stale docs nav entry fails the marketing build as well as the docs build (deliberate, and the failure names the entry). The theme preference now stores `"system"` in the shared `nl-theme` key, which next-themes on the docs site already understands. The PR-template check gains a required section, so open PRs will be asked to add an "AI assistance" line. The root llms files are 1.3 MB of committed text that the bot will regenerate after docs changes land.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2766)


modernmedici:
Screenshot 2026-09-16 at 1 20 56 AM
1. The toggle for dark mode at the bottom is redundant -- as its already at the top so please remove this. 

Screenshot 2026-09-16 at 1 23 06 AM
2. Copy should be consistent between docs and the website "ASK AI ABOUT THIS PAGE"
3.  Ask AI should be visually centered rather than clustered on the left corner. Which creates assymetry with the elements above. 
Screenshot 2026-09-16 at 1 24 50 AM

web/packages/ui/src/components/theme-switch.tsx
```diff
@@ -0,0 +1,88 @@
+"use client";
+
+import type * as React from "react";
+import { cn } from "../lib/cn";
+import { type ThemePreference, useTheme } from "./theme-provider";
+
+interface ThemeSwitchProps extends React.FieldsetHTMLAttributes<HTMLFieldSetElement> {}
```
modernmedici: This is redundant because this switch is on top of the page

modernmedici (changes requested):
Looks good, just remove the redundant switch, make the copy consistent and center the Ask AI buttons on the web page.

web/packages/ui/src/components/theme-switch.tsx
MarcusSorealheis: Removed. The header toggle is the only one now.

MarcusSorealheis:
> Screenshot 2026-09-16 at 1 20 56 AM 1. The toggle for dark mode at the bottom is redundant -- as its already at the top so please remove this. Screenshot 2026-09-16 at 1 23 06 AM 2. Copy should be consistent between docs and the website "ASK AI ABOUT THIS PAGE" 3. Ask AI should be visually centered rather than clustered on the left corner. Which creates assymetry with the elements above. Screenshot 2026-09-16 at 1 24 50 AM

I don't think this should be centered.

MarcusSorealheis:
but I did it because I'm not an agent, but I respect them. I do believe they still process left-to-right when english, but things are changing rapidly, so maybe not.

modernmedici (approved):
LGTM

Issue #2767, opened 2026-09-16
[Bug]: ONTAP S3 store sends an unpadded x-amz-checksum-sha256, which S3 rejects


harshavardhana:
### Version

`main` at 5d6e287d, and every released version back to v1.3.1. The store was added in #1630.

### Client

Other (not client-specific; the failure is between NativeLink and the object store)

### OS / Architecture

Linux x86_64

### Deployment method

Other (any deployment using the ONTAP store)

### Storage backend

S3 (`provider: "ontap"`)

### Configuration

```json5
{
  stores: [
    {
      name: "CAS",
      experimental_cloud_object_store: {
        provider: "ontap",
        endpoint: "https://s3.example.com",
        vserver_name: "us-east-1",
        bucket: "some-bucket",
      },
    },
  ],
}
```

### Expected behavior

A `PutObject` under 5MB stores the object.

The S3 API specifies `x-amz-checksum-sha256` as the base64 encoding of the 32-byte
SHA-256 digest. Base64 of 32 bytes is 44 characters and ends in a single `=` pad.
Amazon S3 validates this and rejects anything else, and S3-compatible
implementations follow suit.

### Actual behavior

The request carries a 43-character unpadded value, the server rejects it with
HTTP 400, and the object is never stored.

Objects at or above 5MB are fine, because the multipart path does not set the header.
So a deployment can look healthy while silently failing every small blob, which for a
CAS is most of them.

The cause is a single call on the small-object upload path:

https://github.com/TraceMachina/nativelink/blob/5d6e287dea9953b98791ccc0daface150167b4ab/nativelink-store/src/ontap_s3_store.rs#L400-L401

`BASE64_STANDARD_NO_PAD` strips the pad that the specification requires.

The test suite asserts the malformed value, so CI holds the bug in place:

https://github.com/TraceMachina/nativelink/blob/5d6e287dea9953b98791ccc0daface150167b4ab/nativelink-store/tests/ontap_s3_store_test.rs#L309-L312

### Logs / error output

Here is a reproducer against Amazon S3 itself, to show this is the protocol rejecting
the value rather than one vendor being strict. It sends the same object twice, once
with the padded encoding and once with the padding stripped.

```python
import base64
import hashlib

import boto3
from botocore.exceptions import ClientError

BUCKET = "your-bucket"
CONTENT = b"Demonstrating base64 padding enforcement in the S3 checksum header."

s3 = boto3.client("s3")

[...]

PR #2768, opened 2026-09-16
Let the S3 store address S3-compatible endpoints

harshavardhana:
## What and why

Towards #2767.

The `aws` provider can only reach AWS today. It has no way to set an endpoint, so
anyone pointing NativeLink at a self-hosted or third-party S3 service has to write a
whole new store. That is how `ontap_s3_store.rs` came about, and #2767 covers the bug
that grew inside it: a hand-rolled `x-amz-checksum-sha256` built with unpadded base64,
which S3 rejects with HTTP 400.

This adds four optional fields to `ExperimentalAwsSpec` so one store can serve any
S3-compatible endpoint:

| Field | Meaning |
| --- | --- |
| `endpoint` | Service URL to use instead of the AWS regional endpoint |
| `force_path_style` | Address objects as `{endpoint}/{bucket}/{key}` rather than `{bucket}.{endpoint}/{key}` |
| `access_key_id` | Explicit credentials, matching what the `r2` and `oci` providers already accept |
| `secret_access_key` | |

`S3Store::new` layers these onto the config it already loads from
`aws_config::defaults`, so environment and profile resolution still works the same way.
Explicit credentials short-circuit the default credential chain rather than being
applied on top of it.

Checksums are left entirely to the SDK. An earlier revision set
`request_checksum_calculation` and `response_checksum_validation` explicitly; that was
wrong, because `Builder::from(&config)` carries the values
`aws_config::defaults().load()` resolved from `AWS_REQUEST_CHECKSUM_CALCULATION` and
`AWS_RESPONSE_CHECKSUM_VALIDATION`, and setting them again overwrote an opt-out the
old `Client::new(&config)` path honoured. That opt-out matters for exactly the
S3-compatible services this PR is for.

This deliberately does not touch `ontap_s3_store.rs`. #2767 proposes rewriting that
store as an adapter over `S3Store`, the way `R2Store` and `OciStore` already are, and
then removing the duplicated `StoreDriver`. This only adds the capability that makes
the rewrite possible; the rewrite is a separate change and a separate review.
`root_certificates` is left out for the same reason, since #2747 already proposes
moving it onto `CommonObjectSpec` for every S3-compatible store.

## How was this verified?

`S3Store::new_with_http_clients` mirrors the existing `OciStore::new_with_http_client`,
so a test can build the SDK client from a spec and drive it with a
`StaticReplayClient`. Without it the endpoint and addressing behavior is unreachable
from a test, because the other `s3_store_test.rs` cases hand in a ready-made client and
so never exercise the spec-to-client mapping at all.

Three tests use it, each asserting on the URL or headers the SDK actually put on the
wire:

- `force_path_style_puts_the_bucket_in_the_path` expects
  `https://s3.example.com/dummy-bucket-name/{key}`. Before this change the field does
  not exist; with `force_path_style` left false it sees the virtual-hosted URL instead,
  which is what the second test pins.
- `endpoint_defaults_to_virtual_hosted_addressing` expects
  `https://dummy-bucket-name.s3.example.com/{key}`, so the new field cannot silently
  change addressing for anyone not asking for it.
- `single_put_lets_the_sdk_own_the_checksum` asserts the request carries
  `x-amz-sdk-checksum-algorithm` and an `x-amz-checksum-*` trailer, and that the store
  writes no `x-amz-checksum-*` header of its own. That last assertion is the one that
  fails against the `ontap_s3_store.rs` approach, and is the regression guard for
  #2767.

`nativelink-config/examples/s3_compatible_backend.json5` shows the fields in a working
config. The existing example-parsing test walks that directory, so it is loaded and
deserialized without any wiring.

`how-to/stores/s3-and-compatible.mdx` gains a section covering the four fields under
"Amazon S3". `lint:snippets` and `vale` pass on it. The fields are described in a table
rather than a JSON5 snippet because `lint:snippets` validates keys against
`reference/nativelink-config/index.mdx`, which tracks the latest *release* (v1.7.0);
`force_path_style` cannot appear in a fenced snippet until that reference regenerates.
The generated config reference is deliberately not touched here, matching how other
field-adding PRs leave it to the periodic `docs(config-reference)` regeneration.

Not verified: no live S3-compatible endpoint was used. The path-style and endpoint
behavior is checked against the request the SDK produces, not against a real server
accepting it.

## Risk

Low for existing deployments. All four fields default to today's behavior, and
`ExperimentalAwsSpec` uses `deny_unknown_fields`, so adding optional fields cannot
break a config that already parses. The checksum settings are set to the SDK default,
so no request changes shape.

The parts worth a reviewer's attention:

- `S3Store::new` used to pass the loaded `aws_config` straight to `Client::new`. It now
  layers an `aws_sdk_s3::config::Builder` on top of that same loaded config. Env and
  profile resolution is preserved, including the checksum settings, but this is the code
  path every existing AWS user takes, so it is the main thing that could regress.
- Explicit credentials now bypass `DefaultCredentialsChain` entirely. Anyone setting
  `access_key_id` gets only those credentials, with no fallback.
- `R2Store::build_aws_spec` and `OciStore::build_aws_spec` gained `..Default::default()`.
  Both already set path-style and credentials directly on their own SDK client, so the
  new spec fields stay unset and neither store changes behavior.

One thing that matters for the follow-up rather than for this PR: by default the SDK
uploads a single-shot `PutObject` as `Content-Encoding: aws-chunked` with the checksum
in a trailer. `oci_store.rs` sets `WhenRequired` because OCI rejects that form, and the
`x-amz-content-sha256: UNSIGNED-PAYLOAD` override in `ontap_s3_store.rs` looks like it
is there for the same reason. Operators of such a service can now reach for
`AWS_REQUEST_CHECKSUM_CALCULATION=when_required`, which this PR no longer overrides.
Whether ONTAP accepts the chunked form should still be confirmed against a real
endpoint before the two stores are merged.

## AI assistance

Claude Code (Opus 5) wrote the patch, the tests and the docs section under my
direction, and drafted this description. I reviewed every line, verified the review
feedback against the repo before acting on it, and ran the tests and docs lints
locally.

https://claude.ai/code/session_013cazn9TnJvMV3Bw3MxDzcH

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2768)


nativelink-store/src/s3_store.rs
```diff
@@ -107,34 +110,79 @@ where
     NowFn: Fn() -> I + Send + Sync + Unpin + 'static,
 {
     pub async fn new(spec: &ExperimentalAwsSpec, now_fn: NowFn) -> Result<Arc<Self>, Error> {
+        Self::new_with_http_clients(
+            spec,
+            TlsClient::new(&spec.common.clone()),
+            TlsClient::new_for_credentials(&spec.common),
```
MarcusSorealheis: Both of these return `Result<TlsClient, Error>` on main since #2748, so after the rebase they need `?`.

nativelink-store/src/s3_store.rs
```diff
@@ -107,34 +110,79 @@ where
...
+            };
+
+            let config = loader.load().await;
+
+            let mut config_builder = aws_sdk_s3::config::Builder::from(&config)
+                .force_path_style(spec.force_path_style)
+                // Let the SDK compute and validate the `x-amz-checksum-*`
+                // values. Writing them by hand is how they end up malformed.
+                .request_checksum_calculation(RequestChecksumCalculation::WhenSupported)
+                .response_checksum_validation(ResponseChecksumValidation::WhenSupported);
```
MarcusSorealheis: These four lines are a regression for exactly the S3-compatible case this PR is for, and they add nothing for AWS.

`WhenSupported` is already the SDK default. What the setters change is that `aws_config::defaults().load()` resolves `AWS_REQUEST_CHECKSUM_CALCULATION` / `AWS_RESPONSE_CHECKSUM_VALIDATION` (and the matching profile keys) into the `SdkConfig`, `Builder::from(&config)` copies them across, and these calls then overwrite them. The old `Client::new(&config)` path honoured them. `WHEN_REQUIRED` is the documented way to stop the SDK sending `aws-chunked` bodies with `x-amz-trailer` checksums, which some S3-compatible services reject; `oci_store.rs` hard-codes `WhenRequired` for that reason.

Removing them keeps the SDK computing the checksum, keeps `single_put_lets_the_sdk_own_the_checksum` passing, and keeps the opt-out working.

MarcusSorealheis (changes requested):
Thanks, this is the right shape: the fields default to today's behaviour and the tests pin the wire output. Three things before it can merge, two of them one-liners.

1. Rebase. `TlsClient::new` and `new_for_credentials` return `Result` on main since #2748, so the two calls in `S3Store::new` need `?`. That is why the branch shows as conflicting.
2. Drop the two explicit checksum setters. They override an opt-out that the old path honoured; details inline.
3. Docs. `web/apps/docs/content/docs/how-to/stores/s3-and-compatible.mdx` is the page AGENTS.md pairs with a store behaviour change. A short paragraph under "Amazon S3" covering `endpoint`, `force_path_style` and the credential pair is enough, plus a note that a plain `http://` endpoint also needs `insecure_allow_http`.

nativelink-store/src/s3_store.rs
harshavardhana: Done. Rebased onto main; both calls now use `?`.

nativelink-store/src/s3_store.rs
harshavardhana: You are right, and the reasoning is worth restating so it does not get re-added: `Builder::from(&config)` carries across whatever `aws_config::defaults().load()` resolved from `AWS_REQUEST_CHECKSUM_CALCULATION` / `AWS_RESPONSE_CHECKSUM_VALIDATION`, so setting them again silently overrode an opt-out the old `Client::new(&config)` path honoured. Since `WhenSupported` was already the default, the lines bought nothing for AWS and removed a working escape hatch for exactly the S3-compatible services this PR is meant to serve.

Both setters and their imports are gone. `single_put_lets_the_sdk_own_the_checksum` still passes, which is the point: the SDK keeps computing the checksum, it just no longer has its configuration overwritten.

harshavardhana:
Thanks, all three addressed and pushed.

1. **Rebase.** Done, onto `84a0e4ea`. Both `TlsClient` calls take `?` now.
2. **Checksum setters.** Removed, along with their imports. Replied on the thread with the reasoning so it does not get re-added by accident.
3. **Docs.** `how-to/stores/s3-and-compatible.mdx` gains an "Other services that speak S3" section under "Amazon S3", covering `endpoint`, `force_path_style` and the credential pair, plus the `insecure_allow_http` note for a plain `http://` endpoint. `vale` and `lint:snippets` pass locally.

One judgement call on the docs worth flagging. I wrote the four fields as a table rather than a JSON5 snippet, because `lint:snippets` validates snippet keys against `reference/nativelink-config/index.mdx`, which tracks the latest **release** (v1.7.0). `force_path_style` does not exist there yet, so any fenced snippet using it fails the lint. The options were a table, a snippet that omits the most important field, or the `lint-snippets: ignore` escape, which is meant for snippets that are deliberately not NativeLink config. The table seemed the honest one. Happy to switch it to a snippet in a follow-up once the config reference regenerates, or now if you would rather I use the escape.

I also left the generated config reference alone, since other field-adding PRs (the OCI store, for one) leave it to the periodic `docs(config-reference)` regeneration. Say the word if you want it in this PR instead.

Not verified: still no live S3-compatible endpoint. The path-style and endpoint behaviour is asserted against the request the SDK produces, not against a real server accepting it.

nativelink-store/src/s3_store.rs
MarcusSorealheis: ahh, makes sense.

MarcusSorealheis:
@harshavardhana thank you for addressing the comments. The docs issue is not a problem, nor is a test against a live server. Our integration tests do that, and our internal release checks evaluate against a number of live servers across an array of providers. 

Again, I cannot thank you enough. I am very grateful and hope to learn more about your Nativelink usage as you progress in your journey. To me, personally, you're an absolute legend!

MarcusSorealheis (approved):
LGTM.

PR #2769, opened 2026-09-16
Steer the Ask AI prompt towards the reader's use case

MarcusSorealheis:
## What and why

The Ask AI prompt asked assistants to compare NativeLink with other systems. It now asks for the features, then asks about the reader's build tool, team size and hosting, and works towards a configuration for that use case. Shared deployments are pointed to contact@nativelink.com for Enterprise pricing on the Business Source modules.

## How was this verified?

Biome on the three changed files, `tsc --noEmit` in `packages/ui` and `apps/web`, and the UI package's tests. Rendered the prompt with `bun` to read it back: 693 characters, 935 encoded, so it fits every assistant's query parameter.

## Risk

Low. Copy only, no binary change. The prompt now comes from one exported `askAiPrompt()` used by the footer and `/agents`, so both sites change together. The email differs from the site's contact@tracemachina.com; it matches the BSL file headers.

## AI assistance

An agent drafted the prompt and the refactor to one builder function. I set the direction, reviewed every line and the rendered prompt.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2769)


modernmedici (approved):
LGTM

PR #2772, opened 2026-09-16
Emit cache size and entries, and keep active action counts for Redis schedulers

amankrx:
## What and why

Three of the 30 documented instruments were never emitted by a running server.

`cache.size` and `cache.entries` are declared but nothing ever recorded them. `EvictingMap` now tracks both, and the `cache_metrics` wrapper turns them on for the tier it wraps, so they follow inserts, replacements, evictions and removals.

`execution.active.count` was only maintained by the in-memory scheduler, so it stayed empty on any install keeping scheduler state in Redis, which is every multi-replica install. A periodic pass now reports the per-stage counts from the store.

Both are opt-in, because both cost something.

The active action count queries the scheduler store every 15 seconds from every replica, against the same backend that serves scheduling, so it is off unless asked for:

```json5
{ "scheduler": { "simple": { "enable_active_action_count_metric": true } } }
```

The count is also cheap now. The first version routed through `search_by_index_prefix`, which hardcodes `LOAD data version`, so Redis read and shipped the full body of every matching action just to count rows client-side, and the cost grew with queue depth and with the whole completed-retention window. `SchedulerStore::count_by_index_prefix` issues `FT.SEARCH ... LIMIT 0 0` instead, which returns the total and no documents.

The cache size metrics stay off until a `cache_metrics` wrapper enables them, and they no longer report from inside the `EvictingMap` state lock. The mutation paths do integer arithmetic under the lock and the totals are flushed after it is released, so enabling them does not lengthen the critical section on the cache's hot path. A missed flush only delays the next update, it never loses a change.

Every replica reports the same store-wide totals for the active count, so aggregate that series with `max`, not `sum`.

## How was this verified?

- 685 tests pass across nativelink-util, nativelink-store, nativelink-scheduler and nativelink-config.
- A new test drives an `EvictingMap` through insert, replace, evict and remove against an in-process OTLP collector and asserts the exported totals.
- A new test asserts the store is never queried while the active count is disabled, and counted at least once per stage per minute when enabled.
- `count_by_index_prefix` is covered by tests that pin the exact command, so re-adding a `LOAD` fails them, plus RESP2 and RESP3 reply parsing.
- Run against a live AKS cluster: 30 of 30 instruments produced series, against 26 on v1.7.0.

## Risk

The active count is off by default, and when enabled adds one `FT.SEARCH` per stage every 15 seconds per replica. `EvictingMap` gains two integer adds on the insert and remove paths whether or not the metrics are enabled. After a scheduler starts there is a 15 second window before the first count is reported.

## AI assistance

An Agent helped with the implementation, while tests and decisions were mine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2772)


nativelink-store/src/cache_metrics_store.rs
```diff
@@ -44,10 +44,16 @@ pub struct CacheMetricsStore {
 
 impl CacheMetricsStore {
     pub fn new(spec: &CacheMetricsSpec, backend: Store) -> Arc<Self> {
+        let base_attrs = [KeyValue::new(CACHE_TYPE, spec.cache_type.clone())];
+        // A store that tracks its own size reports it under the same cache
+        // type as the operation metrics. Other stores report no size.
+        backend
+            .inner_store(None::<StoreKey<'_>>)
+            .enable_cache_size_metrics(&base_attrs);
```
MarcusSorealheis, line 52: [P2] Enable size reporting after backend references resolve

With a supported `cache_metrics` backend of `ref_store` pointing to a memory/filesystem store, this constructor runs before `StoreManager::run_post_init()`. At this point `RefStore::inner_store()` cannot resolve its target and returns the `RefStore` itself, whose default `enable_cache_size_metrics()` is a no-op. Nothing retries after initialization, so `cache.size` and `cache.entries` remain absent for this configuration. I reproduced this through `store_factory`: after `run_post_init()`, a successful 5-byte write/read emitted no size measurement, while an inline memory backend emitted 5. Please enable reporting after reference resolution and add coverage for a reference-backed wrapper.

nativelink-store/src/redis_utils/ft_search_count.rs
```diff
@@ -0,0 +1,150 @@
...
+                )));
+            }
+        },
+        Value::Map(entries) => {
+            let total = entries.iter().find_map(|(key, value)| match (key, value) {
+                (Value::SimpleString(key), Value::Int(total)) if key == "total_results" => {
+                    Some(*total)
+                }
+                _ => None,
+            });
```
MarcusSorealheis, line 80: [P2] Reject incomplete search replies before publishing the count

This RESP3 branch ignores `warning` and accepts any `total_results`. Redis can return partial results on timeout under its default [`ON_TIMEOUT RETURN` policy](https://redis.io/docs/latest/develop/ai/search-and-query/administration/configuration/#search-on-timeout), with warnings in the [RESP3 reply](https://redis.io/docs/latest/commands/ft.search/#return-information). A mocked reply with `total_results: 7` and `warning: ["Timeout limit was reached"]` currently returns `Ok(7)`. `report_active_counts()` then replaces the previous valid total with that incomplete count, potentially making an overloaded queue look smaller. Please reject nonempty warning replies, as the existing `ft_aggregate.rs` parser does, so the reporter retains its last valid sample; add a warning-bearing reply test.

nativelink-scheduler/src/store_awaited_action_db.rs
```diff
@@ -542,6 +579,42 @@ const fn get_state_prefix(state: SortedAwaitedActionState) -> &'static str {
...
+            .await
+            .map(|count| i64::try_from(count).unwrap_or(i64::MAX));
+        match count {
+            // The first pass records even a zero, so every stage has a series
+            // and an empty queue reads as 0 rather than no data.
+            Ok(count) if *last != Some(count) => {
+                EXECUTION_METRICS.execution_active_count.add(
+                    count - last.unwrap_or(0),
+                    &[KeyValue::new(EXECUTION_STAGE, *stage)],
+                );
```
MarcusSorealheis, line 605: [P2] Update the shipped consumers for replicated queue totals

When this option is enabled on multiple schedulers sharing Redis, each replica publishes the entire queue under the existing metric name. The shipped [`execution_active_by_stage` recording rule](https://github.com/TraceMachina/nativelink/blob/85915c9b49ea8a7bbf9837945633d4cc3a8b202f/deployment-examples/metrics/prometheus-config.yaml#L106-L109) and [`NativeLinkQueueBacklog` alert](https://github.com/TraceMachina/nativelink/blob/85915c9b49ea8a7bbf9837945633d4cc3a8b202f/deployment-examples/metrics/kubernetes/prometheus.yaml#L146-L149) still use `sum`. Two replicas with 60 queued actions therefore report 120 and trigger the >100 alert even though the real backlog is below its threshold. The new instruction to use `max` does not update these consumers. Please preserve additive semantics or distinguish shared-store totals and update the bundled aggregation rules; account for the in-memory schedulers, whose independent counts still need summing.

MarcusSorealheis (commented):
Reviewed 85915c9b. CI is green, but I found three correctness issues below. Two added local regression tests reproduce the reference-backed cache initialization issue and acceptance of an incomplete Redis reply; the other 62 tests in those two binaries passed. Please address these before merging.

MarcusSorealheis (commented):
LGTM

MarcusSorealheis (approved):
LGTM

PR #2780, opened 2026-09-19
Fix signal-terminated worker property commands

immanuwell:
## What and why

A `query_cmd` killed by a Unix signal has no numeric exit code, so worker registration panics on `unwrap()`. 
Return an `Internal` error with the process status instead, so the normal retry path handles it

## How was this verified?

Added a regression using `bash -c "kill -TERM $$"`. 
It panics on `main`, while all eight `worker_utils` tests pass with this fix using `bazel test //nativelink-worker:integration_tests/worker_utils_test_test`

## Risk

Low. Numeric exit codes keep their existing mapping and message. 
Only signal-terminated query commands take the new error path

## AI assistance

Codex helped me a bit

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2780)


palfrey (approved):
Nice catch, thanks!

PR #2784, opened 2026-09-22
scheduler: fail queued actions with criteria no worker can meet

corcillo:
## What and why

`SimpleScheduler` cannot tell a queued action that is waiting on a busy worker from one that will never be satisfied based on the existing worker pool configuration. Both actions will get stuck `Queued` forever and never fail, thus preventing us from even a local fallback option.

`find_worker_for_action` now returns a `MatchOutcome`. An action is `Unsatisfiable` when no connected worker could take it even when idle, judged against the properties each worker is *registered* with (new `Worker.total_platform_properties`), rather than what it has free right now. The verdict is cached per property shape and dropped whenever the fleet changes.

A new `unsatisfiable_action_timeout_s` (default 0, never fail) fails such an action with `FAILED_PRECONDITION` and a message naming every property that could not be satisfied and what the fleet offers, e.g.

```
Action cannot be scheduled, no worker can satisfy: 'gpu_count' requested 1, largest worker total 0; 'memory_kb' requested 37748736, largest worker total 31457280. No capable worker connected within 60s.
```

Commit 2: Every 5 seconds each scheduler publishes what its own workers can run to an expiring Redis record and reads every peer's, so a scheduler only fails an action when no worker on any replica could run it.

## How was this verified?

`nativelink-scheduler/tests/unsatisfiable_action_test.rs` (28 tests), `fleet_capabilities_test.rs` (3), and 4 in `simple_scheduler_state_manager_test.rs`. The customer's exact shape is the main fixture. They cover:

- the four-way classification, including a saturated fleet hiding an impossible action, `Minimum` above the largest total, and both `gpu_count` and `memory_kb` being reported (sorted, stable);
- the hard contracts: `Priority` value mismatch is never a reason; an undeclared (`Unknown`) key never restricts a worker lacking it; empty fleet never failed; `fail_queued_operation` is terminal, leaves `attempts` alone, and is a no-op on an executing, completed or unknown operation;
- the resulting `ExecuteResponse.status` is `FAILED_PRECONDITION` with zero details;
- the timer: due after the timeout, cleared when a capable worker joins, the second action of a due shape fails on its first pass, an idle gap does not count as waiting, a stale verdict is recomputed after a capable worker leaves, backoff once a due shape stays queued;
- the loop reaches the deadline with `fallback_match_interval_s = 0` and before a long one;
- the waiting client's `changed()` stream is woken by the failure;
- the Redis exchange: two `StoreAwaitedActionDb`s over one fake Redis see each other's shapes and not their own, republishing replaces, every record carries the 15 s expiry;
- end to end: two real `SimpleScheduler`s on one fake Redis with the exchange interval set to 50 ms. Only B has a GPU worker and it is busy. A keeps the GPU action queued 10× past the timeout, then fails it within one timeout of B's worker leaving. The interval is a constructor parameter so this runs in under a second; `new_with_callback` is unchanged.

Line coverage of the added code, measured with `-C instrument-coverage`, is 95%; what remains is error plumbing and the version-conflict retry, which cannot be forced deterministically with the in-memory db.

The one pre-existing test that changed, `logs_when_no_workers_match`, asserted a "No workers matched" log for an action that is in fact unsatisfiable (`prop` min 1 against a worker with 0); it now asserts the new, more specific log line. The fake Redis needed a one-line change to return non-versioned records from a tag query, which is what the real `RedisStore` already does.

## Risk

Default is 0, so existing deployments only gain the classification, a rate-limited `warn!` and two metrics; nothing is failed and the fleet exchange does not run. The matcher's hot path gains a `HashMap` lookup per action and, on a miss, one pass over worker totals; for an unsatisfiable action this is cheaper than today's full index walk on every pass.

With the timeout set: one small `HSET` + `EXPIRE` and one `FT.AGGREGATE` per scheduler every 5 seconds against the scheduler's Redis, on the index-search path the matcher already uses; a new `fc_*` record type that expires on its own. A worker joining a peer is seen within ~5 s and one leaving stops counting within 15 s, so the timeout must sit well above that, and above a scale-from-zero pool's provisioning time. The config doc says both.

`find_worker_for_action` changes return type (one external caller, in `nativelink-service` tests). `MatchingEngineStateManager` gains `fail_queued_operation` and `AwaitedActionDb` gains `exchange_fleet_capabilities` with a no-op default; each has one implementor.

## AI assistance
I wrote clear cut requirements and iterated on a prompt, asking several edge case questions to further hone the prompt. I then iterated a few times with an agent until the code had all the functionality I asked for, was fully tested, and safe to roll out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2784)


MarcusSorealheis (commented):
Thanks for this. It solves a real problem (Chinchilla's GPU actions sitting `Queued` forever after the GPU pool was removed), and the core is sound:
- the four-way `MatchOutcome`;
- judging against `total_platform_properties` rather than current free capacity;
- never failing when the fleet is empty;
- correct handling of `ignore`, `priority` and undeclared keys;
- `fail_queued_operation` only touching actions that are still `Queued`.

The tests are thorough too.

I have concerns in three areas before this merges. Inline comments give the details.

**1. Multi-scheduler safety: the design fails closed when it should fail open.**
- A peer this scheduler can't see is treated as a peer with no workers. That covers a peer still on the old binary during a rolling upgrade, a peer with the timeout set to 0, and a failed or undecodable exchange.
- A GPU action can then be failed even though another replica holds the only GPU workers.
- Separately, peers list their workers in LRU order, which changes on every dispatch. So `set_peer_fleet` sees a "changed" fleet on most exchanges, and the per-shape cache and tracker linger get reset constantly on busy fleets.

**2. Hot-path cost.**
- `find_worker_for_action` now builds a `PropertyShape` (clone, sort, SipHash) for every queued action on every pass, including ones that match. It does this while holding the worker-scheduler mutex.
- The cache lookup happens after the index walk, so it doesn't save the walk.
- The description says an unsatisfiable action is cheaper than before. In practice the cost is strictly added.
- Classifying only after a failed match makes the added cost zero for actions that match.

**3. Client-facing error and operability.**
- The `FAILED_PRECONDITION` message lists the `exact` values workers registered with, such as container image URIs and pool names. Any client can list them by submitting a mismatched action.
- `fail_queued_operation` has no backoff, doesn't record the execution metrics, and doesn't log which operation it failed.

Since maintainer edits are enabled, I have fixes for all of these ready. I'll push them as separate commits on top of your branch so you can review each one. Nothing will be force-pushed.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -590,6 +743,42 @@ impl SimpleScheduler {
...
+        // can fail an action. The db decides whether it has peers to talk
+        // to; the in-memory one returns nothing.
+        let task_fleet_exchange_spawn = {
+            let scheduler_id = Uuid::new_v4().hyphenated().to_string();
+            let state_manager = Arc::downgrade(&state_manager);
+            let worker_scheduler = Arc::downgrade(&worker_scheduler);
+            let enabled = unsatisfiable_action_timeout.is_some();
+            let record_ttl = fleet_exchange_interval * FLEET_RECORD_TTL_INTERVALS;
+            spawn!("simple_scheduler_task_fleet_exchange", async move {
+                if !enabled {
```
MarcusSorealheis: **Fail-closed when a peer doesn't publish.** The exchange only runs when *this* scheduler's timeout is set.

- **Failure case:**
  - During a rolling upgrade (B still on the old binary), or with mixed configs (B has the timeout at 0), replica B can hold every GPU worker and never publish.
  - A sees an empty peer fleet and fails every GPU action one timeout later, even though B would run them.
- **Suggestion:**
  - Always publish when the db is shared, and gate only the *failing* on the timeout. The cost is one HSET, one EXPIRE and one FT.AGGREGATE per 5s.
  - Don't spawn the task at all for the in-memory db. Right now, with the timeout set, it wakes every 5s and takes the scheduler lock to clone the worker totals for an exchange that returns nothing.
  - Document the rollout order: ship the binary everywhere with the timeout at 0, then turn it on.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -590,6 +743,42 @@ impl SimpleScheduler {
...
+                        return;
+                    };
+                    let local = worker_scheduler.fleet_shapes().await;
+                    match state_manager
+                        .exchange_fleet_capabilities(&scheduler_id, local, record_ttl)
+                        .await
+                    {
+                        Ok(peer_fleet) => worker_scheduler.set_peer_fleet(peer_fleet).await,
+                        Err(err) => {
+                            error!(?err, "Error while exchanging fleet capabilities");
```
MarcusSorealheis: **A failed exchange leaves a stale peer view in use.** If the exchange keeps erroring, `peer_fleet` keeps its last value indefinitely. Causes include a Redis outage, or one peer record that won't decode, which fails the whole `try_fold`.

- If that snapshot had no capable peer, A keeps failing actions a peer could run.
- It's safer to treat "peers unknown" as "can't prove unsatisfiable". Set the peer fleet to unknown on error, and have `static_verdict` return `None` (satisfiable) until an exchange succeeds again.
- Start in that unknown state when the db is shared, so no failure happens before the first exchange.
- This `error!` also fires every 5s for the whole outage. Logging once when it starts failing and once when it recovers would be quieter.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -590,6 +743,42 @@ impl SimpleScheduler {
...
+                        }
+                    }
+                    drop((state_manager, worker_scheduler));
+                    tokio::time::sleep(fleet_exchange_interval).await;
```
MarcusSorealheis: **A new peer worker can be seen late, while due shapes fail at once.**

- **The mismatch:** once a shape is due, every new action with that shape fails on its first pass. But a GPU worker joining peer B reaches A only after B's next publish *and* A's next read, which is up to about 10s.
- **Suggestion:** publish as soon as the local fleet changes. Wake this loop from `add_worker` and `remove_worker` with a `Notify`, for example `tokio::select!` on the sleep or the notify. A peer's join then reaches A within one of A's read intervals.
- **Related:** re-check `fleet_generation()` just before `fail_queued_operation` (see the comment on line 378).

nativelink-config/src/schedulers.rs
```diff
@@ -177,6 +177,42 @@ pub struct SimpleSpec {
...
+    /// Set this higher than the time a worker pool needs to scale up from
+    /// zero and register, or actions that the pool would have served are
+    /// failed. Only properties listed in `supported_platform_properties`
+    /// are enforced strictly; a property that is not listed does not
+    /// restrict workers that do not declare it.
+    ///
+    /// When several schedulers share one Redis backend, each publishes what
+    /// its workers can run every 5 seconds, and an action is only failed
+    /// when no scheduler has a worker that could run it. A worker that
+    /// joins a peer is seen here within about 5 seconds, and one that leaves
```
MarcusSorealheis: The numbers here don't match the implementation:
- A worker that joins a peer is visible after the peer's next publish *plus* this scheduler's next read, so up to about 2 intervals (10s), not about 5s. That drops to about 5s if the peer publishes on change.
- A scheduler that dies stops counting after the record TTL (15s) plus up to one read interval, so up to about 20s, not 15s.

Could you also document the rollout order: ship the new binary everywhere with the timeout at 0 before enabling it?

nativelink-scheduler/src/api_worker_scheduler.rs
```diff
@@ -241,12 +262,37 @@ impl ApiWorkerSchedulerImpl {
...
+        self.workers
+            .iter()
+            .map(|(_, w)| &w.total_platform_properties)
+            .filter(|totals| seen.insert(PropertyShape::from(*totals)))
+            .cloned()
+            .collect()
+    }
+
+    fn set_peer_fleet(&mut self, peer_fleet: Vec<PlatformProperties>) {
+        if self.peer_fleet == peer_fleet {
```
MarcusSorealheis, line 389: **This comparison depends on order.**
- `fleet_shapes()` walks `self.workers` in LRU order, and `worker_notify_run_action` reorders it on every dispatch through `get_mut` (line 547).
- `FT.AGGREGATE` doesn't guarantee any order either.
- **Effect:** on a busy, heterogeneous fleet, most exchanges look like a change. They bump `fleet_generation`, empty `static_verdicts`, and drop the tracker's linger in `end_pass`, which checks `fleet_generation`. That defeats the cache and makes "later actions with the same properties fail straight away" unpredictable.
- **Suggestion:** derive `PartialOrd`/`Ord` on `PropertyShape`, then sort and dedup the peer fleet by shape here before comparing.

nativelink-scheduler/src/api_worker_scheduler.rs
```diff
@@ -281,11 +363,38 @@ impl ApiWorkerSchedulerImpl {
             .capability_index
             .find_matching_workers(platform_properties, full_worker_logging);
 
-        if candidates.is_empty() {
+        // A cached verdict skips the per-property logging, so recompute when
+        // that logging was asked for.
+        let shape = PropertyShape::from(platform_properties);
```
MarcusSorealheis: **This adds cost to every match, not just unsatisfiable ones.**

- **The added work:** for every queued action on every pass, while holding the worker-scheduler mutex (which keepalives and updates also need), this builds a `PropertyShape`: it clones every key and value, sorts, and hashes with SipHash. `handle_unsatisfiable` then builds it a second time.
- **The cache doesn't save the walk:** the lookup happens after `find_matching_workers`. So the description's claim that an unsatisfiable action is "cheaper than today's full index walk" doesn't hold.
- **Suggestion:** try to match first, and classify only when nothing matched, including in the "all workers fully allocated" case.
  - `Matched` then pays nothing.
  - The shape is only built on the unsatisfiable-or-waiting path.
  - The index walk can be skipped when every worker is busy and the verdict is cached.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -295,13 +331,67 @@ impl SimpleScheduler {
...
+            let operation_id = {
+                let (action_state, _origin_metadata) = action_state_result
+                    .as_state()
+                    .await
+                    .err_tip(|| "Failed to get state of an unsatisfiable action")?;
+                action_state.client_operation_id.clone()
+            };
+            let err = make_err!(
+                Code::FailedPrecondition,
+                "Action cannot be scheduled, {reason}. No capable worker connected within {}s.",
```
MarcusSorealheis: **This leaks fleet configuration to clients.**
- `{reason}` includes `workers offer [...]` (up to 8 values per property) for `exact` and undeclared keys, from every replica: container-image URIs, internal pool names and so on.
- Any client can list them by submitting an action with a mismatched property. That matters for multi-tenant deployments.
- **Suggestion:** give the client the property and what it asked for, and keep the offered values in the server `warn!`. The largest total for a `minimum` property is only a number and is useful ("gpu_count requested 1, largest worker total 0"), so it seems fine to keep.

**Nit:** `waited` is the *shape's* age, so an action submitted a second ago can read "No capable worker connected within 3600s". Wording like "No capable worker has been connected for {waited}s" describes that accurately.

nativelink-scheduler/src/simple_scheduler.rs
```diff
@@ -295,13 +331,67 @@ impl SimpleScheduler {
...
+                observation.waited.as_secs()
+            );
+            let failed = matching_engine_state_manager
+                .fail_queued_operation(&operation_id, err)
```
MarcusSorealheis, line 458: **A worker can join between the verdict and the failure.**

- **The window:** the verdict is made under the worker-scheduler lock. This failure happens after `as_state()` and the db update, which on Redis is one or more round trips later.
- **Failure case:** a capable worker (or a peer's fresh exchange) can land in between, and we fail an action that now has a worker.
- **Suggestion:** a cheap guard. Skip failing when `workers.fleet_generation().await != pass.fleet_generation`, and let the next pass decide.

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -433,6 +434,80 @@ where
...
+            .get_by_operation_id(operation_id)
+            .await
+            .err_tip(|| "In SimpleSchedulerStateManager::fail_queued_operation")?
+        else {
+            return Ok(false);
+        };
+        // The action may be racing an assignment or a client update, so on
+        // a version conflict reload it and, if it is still queued, try again.
+        let mut last_err = None;
+        for _ in 0..MAX_UPDATE_RETRIES {
```
MarcusSorealheis: This retry loop has no backoff. `inner_update_operation`, a little further down, backs off with jitter between version-conflict retries. It would be good to share that, perhaps by pulling it out as `version_conflict_backoff(attempt)`, so both loops behave the same under contention.

nativelink-scheduler/src/simple_scheduler_state_manager.rs
```diff
@@ -433,6 +434,80 @@ where
...
+            state.last_transition_timestamp = now;
+
+            let mut new_awaited_action = awaited_action;
+            new_awaited_action.worker_set_state(Arc::new(state), now);
+            match self
+                .action_db
+                .update_awaited_action(new_awaited_action)
+                .await
+            {
+                Ok(()) => return Ok(true),
```
MarcusSorealheis: **This skips the execution metrics and logs nothing.**

- **Missing metrics:** on success this doesn't record the stage transition or the completed count that `inner_update_operation` records. So `execution.completed.count` undercounts these failures, and dashboards show actions disappearing from the queue without completing.
- **Missing log:** no line names the operation that was failed. There's only the per-shape warning, rate-limited to once a minute, plus a counter.
- **Suggestion:**
  - Move the metrics block from `inner_update_operation` into a helper and call it from both places.
  - Add `info!(%operation_id, ?err, "Failed a queued action no worker can run")`.

nativelink-scheduler/src/unsatisfiable_tracker.rs
```diff
@@ -0,0 +1,168 @@
...
+        if should_warn {
+            state.last_warned = Some(now);
+        }
+
+        let waited = now
+            .duration_since(state.first_unsatisfiable)
+            .unwrap_or_default();
+        let is_due = self.timeout.is_some_and(|timeout| waited >= timeout);
+        if is_due {
+            state.due_passes = state.due_passes.saturating_add(1);
```
MarcusSorealheis, line 124: `due_passes` is incremented once per *action*, not once per pass. Five due actions with the same shape in one pass push the recheck straight to the 32s cap, and it never resets. The field's doc says "consecutive passes". Counting only on the first observation of the shape in a pass (`!state.seen_this_pass` before it's set) would match the doc.

nativelink-scheduler/src/match_outcome.rs
```diff
@@ -0,0 +1,221 @@
...
+                    property.name,
+                    property.requested.as_str()
+                )?;
+            }
+            if property.offered.is_empty() {
+                f.write_str(", no worker declares this property")?;
+            } else if matches!(property.requested, PlatformPropertyValue::Minimum(_)) {
+                write!(f, ", largest worker total {}", property.offered[0])?;
+            } else {
+                write!(f, ", workers offer [{}]", property.offered.join(", "))?;
```
MarcusSorealheis, line 111: See the comment on `simple_scheduler.rs:374`: this list is fine in the server log but shouldn't reach the client. A second `Display` adapter (for example `reason.for_client()`) that leaves out the offered values would keep a single formatting path.

nativelink-util/src/metrics.rs
```diff
@@ -43,6 +43,8 @@ pub const RPC_STATUS_CODE: &str = "rpc.grpc.status_code";
 
 // Metric attribute keys for the scheduler.
 pub const SCHEDULER_MATCH_RESULT: &str = "scheduler.match.result";
+/// The platform properties no worker could satisfy, comma separated.
+pub const SCHEDULER_UNSATISFIED_PROPERTIES: &str = "scheduler.unsatisfied.properties";
```
MarcusSorealheis: **Nit:** the attribute is `scheduler.unsatisfied.properties`, but the metrics are `scheduler.unsatisfiable.*`. Since nothing has shipped yet, it might be worth making them consistent with `scheduler.unsatisfiable.properties`.

nativelink-scheduler/src/api_worker_scheduler.rs
```diff
@@ -48,6 +49,9 @@ use tracing::{debug, error, info, trace, warn};
 /// How many state-manager lookups `kill_revoked_operations` has in flight
 /// at once while checking which running operations were revoked.
 const MAX_CONCURRENT_REVOKED_CHECKS: usize = 32;
+
+/// How many property shapes `static_verdicts` holds before it is emptied.
+const MAX_STATIC_VERDICTS: usize = 4096;
```
MarcusSorealheis, line 54: **Nit:** this constant sits between `MAX_CONCURRENT_REVOKED_CHECKS` and `use uuid::Uuid;`. A blank line before the `use` (or moving the `use` up with the others) keeps rustfmt's grouping readable.

MarcusSorealheis:
**The main risk: failing actions that would have run.**

This changes an action that hangs into an action that fails, so every wrong call shows up as a failed build.

**What's already safe:**
- An action that is only waiting for a busy worker is never failed.
- The check runs against what workers *registered* with, so busy, paused and draining workers all count as capable. `busy_fleet_is_not_unsatisfiable` and `saturated_fleet_does_not_hide_an_unsatisfiable_action` cover this.
- The default is 0, so nothing changes unless it's turned on.

**Where it can still fail an action that a worker would have picked up:**
1. **A capable pool is gone for longer than the timeout.** Examples: scale from zero, a rolling restart of the whole GPU pool, or spot preemption. Only the configured timeout guards against this.
2. **Replicas that can't see each other.** Each replica on shared Redis only sees the others' workers through the exchange, so an invisible replica with a GPU pool looked like it had no GPU workers. The commits below fix this: while peers are unknown, nothing is judged unsatisfiable.
3. **Later actions fail immediately.** Once one action has waited out the timeout, later actions with the same properties fail on their first pass. During an outage like (1), that turns one timeout into a burst of failures across every build that needs the pool.

**Suggestions before enabling this anywhere:**
- Roll out with the timeout at 0 on every scheduler first, and watch `scheduler.unsatisfiable.queued` for a while. Anything that shows up there during normal operation is an action this would have failed.
- Set the timeout above the longest pool provisioning or restart time, not above typical queue wait.
- @corcillo, would you consider giving each action a minimum wait of its own, even when its property shape is already due? Aardvarks's builds would still fail within about one timeout, because their GPU actions queue at the same time. A pool outage would then stop turning into an immediate burst of failures.

**Pushed as three commits on top of your branch, with no force-push:**
- `9a29a047`: `fail_queued_operation` now backs off like `inner_update_operation`, records the execution metrics, and logs the operation it failed.
- `b99d21e7`: the client error names only what the action asked for; the values workers offer stay in the server log. It also:
  - re-checks the fleet generation before failing;
  - counts a due shape once per pass;
  - logs at `info` while the timeout is off;
  - renames the attribute to `scheduler.unsatisfiable.properties`.
- `e0480d8b`: fails open when peers can't be seen. Specifically:
  - peers are unknown until the first exchange and again after any failed one;
  - every scheduler on shared state publishes whatever its timeout, and publishes as soon as a worker joins or leaves;
  - the peer fleet is compared independent of order;
  - an action is classified only after no idle worker matched it;
  - the config doc gives the corrected timings and the rollout order.

**Tests:**
- New: `unknown_peers_never_fail_an_action` and `peer_fleet_in_another_order_is_not_a_fleet_change`.
- `exact_mismatch_names_what_workers_offer_only_in_the_log` checks the new client contract.
- The scheduler and service suites pass (254 tests), and every commit builds on its own.

MarcusSorealheis (changes requested):
#2728 and #2782 conflict with each other

@corcillo asked about feature flags. We don't have a standard way to do that today. Instead, features are generally enabled through a matrix of config options. It's very expressive for advanced use cases and users, but it can be complex to manage.

I'd merge #2782's logic but express it through #2728's flag. `trust_fast_store_for_has` defaults to `true` for `read_only` and get slow stores, and to `false` otherwise. That gives one code path and one config knob. #2782's other parts have nothing to do with #2728 and can merge as they are.  The gRPC read timeout, the read-only availability fix, and its docs are all ready to go from my perspective.

corcillo:
default behavior will not change until you set unsatisfiable_action_timeout_s, so i think that should be safe to merge

MarcusSorealheis (approved):
LGTM

MarcusSorealheis:
nice work. This was a massive one.

PR #2785, opened 2026-09-22
scheduler: share worker capabilities between schedulers on one store

corcillo:
**Stacked on #2784.** This branch carries that PR's commit as its base; only the second commit, `029e5606`, is this change. Compare against `corcillo-unsatisfiable-actions` for the real diff. It will be rebased to a single commit on `main` once #2784 merges.

## What and why

#2784 lets a scheduler fail a queued action that no connected worker could ever run. On a shared Redis backend that is not safe on its own: every scheduler sees every queued action, but a worker only exists on the one scheduler holding its `connect_worker` stream. With three replicas and one GPU worker, the two replicas that cannot see it would judge every GPU action unsatisfiable and, with the timeout set, fail it while that worker was merely busy. #2784 therefore refuses to start with a non-zero timeout on a Redis backend.

This lifts that. Every 5 seconds each scheduler writes the distinct properties its workers registered with to an expiring Redis record, `fc_<scheduler uuid>` (15 s TTL), and reads every peer's record back through the store's existing index search. A peer's worker keeps an action `WaitingForCapacity` exactly as a local one would, and its properties appear in the failure message when nothing anywhere could run the action. The exchange only runs when `unsatisfiable_action_timeout_s` is set, since that is the only decision it informs; the in-memory backend shares nothing and sees nothing.

The stopwatches from #2784 stay per scheduler. They are cheap to rebuild, each replica counts independently against the same `aa_<op>` record, and whichever one reaches the timeout first writes the failure; the others see the version move and back off. A replica that just restarted should not be counting down while its own workers reconnect, so not sharing them is the point.

## How was this verified?

`nativelink-scheduler/tests/fleet_capabilities_test.rs` (3 tests) and one added to `unsatisfiable_action_test.rs`:

- two `StoreAwaitedActionDb`s over one fake Redis: each sees the other's shapes and not its own, republishing replaces the previous record, and every record carries the 15 s expiry;
- `set_peer_fleet` on the worker scheduler keeps an otherwise-unsatisfiable action queued past the timeout and starts the clock from zero once the peer shape is gone;
- `fleet_shapes` reports registered totals, not what is currently free, and de-duplicates identical workers;
- end to end: two real `SimpleScheduler`s on one fake Redis with the exchange interval set to 50 ms. Only B has a GPU worker and it is busy. A keeps the GPU action queued 10× past the timeout, then fails it within one timeout of B's worker leaving, with the message naming `gpu_count`. The exchange interval is a constructor parameter (`new_with_fleet_exchange_interval`) so this runs in under a second; `new_with_callback` is unchanged.

The fake Redis needed a one-line change to return non-versioned records from a tag query, which is what the real `RedisStore` already does (`version.unwrap_or(0)`).

## Risk

Off unless `unsatisfiable_action_timeout_s` is non-zero, so nothing changes for existing deployments. When on: one small `HSET` + `EXPIRE` and one `FT.AGGREGATE` per scheduler every 5 seconds against the scheduler's Redis, on the same index-search path the matcher already uses. A new record type, `fc_*`, in the scheduler keyspace; it expires on its own.

Timing: a worker joining a peer is seen within ~5 s; one leaving stops counting within 15 s. The config doc tells operators to keep the timeout well above that. `AwaitedActionDb` gains `exchange_fleet_capabilities` with a no-op default, so other implementors are unaffected.

## AI assistance

- - -
This change is [Reviewable](https://reviewable.io/reviews/TraceMachina/nativelink/2785)


corcillo:
Folded into #2784 as its second commit.

